Issue #719, opened 2026-05-03
[SPIKE] design-platform-file-watcher-detailed


cjhowe-us:
Parent: #714

## Question

Produce the detailed design for the **file-watcher** aggregate in the `platform` context.

FSEvents-backed FileWatcher.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/platform/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/platform/filesystem.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **file-watcher** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/platform/file-watcher-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(platform): detailed design — file-watcher (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-platform-file-watcher-detailed` is blocked by this issue.

cjhowe-us:
agent:design-platform-file-watcher
status:done
issue:#719
branch:docs/platform-file-watcher-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a494384705c12e8ba
host:mac
cloud:none
commit:3b09bfa176b153231296f7dc1769857188669439
pr:#827
[...]

Issue #717, opened 2026-05-03
[SPIKE] design-platform-event-pump-detailed


cjhowe-us:
Parent: #714

## Question

Produce the detailed design for the **event-pump** aggregate in the `platform` context.

SDL3 event drain → typed variants.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/platform/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/input/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **event-pump** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/platform/event-pump-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(platform): detailed design — event-pump (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-platform-event-pump-detailed` is blocked by this issue.

cjhowe-us:
agent:design-platform-event-pump
status:done
issue:#717
branch:docs/platform-event-pump-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-ab410f89b4d96392b
host:mac
cloud:none
commit:faf8a88477b60b7c92ccce8f9950a6b143ec7bd9
pr:#826
[...]

Issue #715, opened 2026-05-03
[SPIKE] design-platform-window-surface-detailed


cjhowe-us:
Parent: #714

## Question

Produce the detailed design for the **window-surface** aggregate in the `platform` context.

Window + Surface + metal-cpp bridge.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/platform/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/platform/window-display.md
- harmonius/docs/design/platform/windowing.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **window-surface** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/platform/window-surface-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(platform): detailed design — window-surface (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-platform-window-surface-detailed` is blocked by this issue.

cjhowe-us:
agent:design-platform-window-surface
status:done
issue:#715
branch:docs/platform-window-surface-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a8e9b059c6f13c28d
host:mac
cloud:none
commit:ec9a7a8680ab691d98a6607d7052704b7ef46b95
pr:#825
[...]

Issue #714, opened 2026-05-03
[SUB-EPIC] Detailed Designs — platform


cjhowe-us:
# Detailed Designs — platform

Per-aggregate detailed designs that flesh out §4 / §5 / §6 of
`specs/platform/SPEC.md`. Each design lives at
`specs/platform/<aggregate>-design.md` and references
`harmonius/docs/{requirements,design}/<area>/` to ensure MVP
coverage.

Parent epic: #32

## Children

Per aggregate: a `design-*-detailed` spike + a `task-breakdown-*-detailed` spike (blocked-by the design twin).

No estimate; rolls up.

PR #826, opened 2026-05-04
docs(platform): detailed design — event-pump (refs #717)

cjhowe-us:
## Summary

- Adds `specs/platform/event-pump-design.md` — detailed design for the `Pump` + `EventQueue<InputEvent>` / `EventQueue<WindowEvent>` aggregate declared in `specs/platform/SPEC.md` §4.2 / §5.3 / §5.4 / §5.7 / §6.3.
- Twelve sections per `specs/_TEMPLATE.md`: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.
- SRP: pump owns the single `SDL_PollEvent` loop and the SDL3 → typed-variant translation table; refuses input mapping (gameplay plugin), window state (sibling spike #715), file watching (sibling spike #719), per-device polling, IME composition, and frame pacing.
- All harmonius input + window-display requirements (R-6.1.*, R-14.1.9 / R-14.1.11, R-14.2.4) are either covered or explicitly refused with rationale; conclusions re-derived from harmonius as research input only.

Refs #717. Sibling task-breakdown spike is blocked by this deliverable. Issue stays OPEN.

## Test plan

- [ ] CI lint / link-check passes on `specs/platform/event-pump-design.md`.
- [ ] Reviewers confirm 12-section structure matches `specs/_TEMPLATE.md`.
- [ ] Reviewers confirm no new public surface beyond `specs/platform/SPEC.md` §5.
- [ ] Reviewers confirm refusals to input mapping / window state / file watching are unambiguous.

### DIFF
--- a/specs/platform/event-pump-design.md
+++ b/specs/platform/event-pump-design.md
@@ -1,0 +1,1199 @@
+# platform — Detailed Design: event-pump aggregate
+
+> Detailed design for the `Pump` + `EventQueue<InputEvent>` /
+> `EventQueue<WindowEvent>` aggregate declared in `specs/platform/SPEC.md`
+> §4.2 / §5.3 / §5.4 / §5.7 / §6.3. Refines those sections in place;
+> cites `reviews/decisions/error-model.md`,
+> `reviews/decisions/frame-phases.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`, and
+> `reviews/decisions/fory-codegen.md`. Introduces no new public surface
+> beyond the §5 stub already locked into `specs/platform/SPEC.md`;
+> deviations from the cited records would require an amendment spike,
+> not an in-place edit.
+>
+> Refs: spike #717 — `[SPIKE] design-platform-event-pump-detailed`.
+> Parent #714. Sibling task-breakdown spike is blocked by this
+> deliverable. All conclusions re-derived; harmonius prior art
+> (`harmonius/docs/requirements/input/device-abstraction.md`,
+> `harmonius/docs/requirements/platform/window-display.md`,
+> `harmonius/docs/design/input/`) is research input only.
+
+## 1. Purpose
+
+The event-pump aggregate is the **single OS-event ingress** for the
+engine. Per `platform` SPEC §4.2 it owns:
+
+1. The **`Pump`** — sole owner of the `SDL_PollEvent` loop. Drains
+   SDL3's single event stream once per frame on the main thread and
+   normalises every accepted event into a typed `InputEvent` /
+   `WindowEvent` sealed-variant payload.
+2. The **paired `EventQueue<InputEvent>` and `EventQueue<WindowEvent>`**
+   — bounded, fixed-capacity SPSC ring buffers sized at construction.
+   They live or die together because SDL3 reports both families
+   through the same stream and the cross-family ordering invariant
+   (§4.2 inv #2) collapses into a single per-device monotone the
+   moment the two queues are split apart.
+3. The **classify-and-translate table** — the sealed mapping from
+   SDL3 event tags + payload fields to glibre's `KeyCode` / `ScanCode`
+   / `MouseButton` / `GamepadAxis` / `GamepadBtn` / `WindowId` /
+   `DisplayId` / `DpiScale` value objects, declared in §5.4 and
+   realised in `event/translate.cpp` (this design's name; SPEC §6.3
+   names it the "translation table").
+4. The **drain timing contract** — the `Pump` runs in **phase 1
+   (input)** of `frame-phases.md`, exactly once per frame, and the
+   queues are quiescent at the phase 8 (hot-reload) barrier
+   (§4.2 inv #1, §8.2 platform-pump-quiescence guarantee).
+
+The aggregate **refuses to own**:
+
+- **Input mapping / actions / bindings.** Translating
+  `InputEvent::KeyDown { key=Space }` into `Action::Jump` is the
+  `gameplay` / `tools` plugins' job (collapsed into harmonius's R-6.2
+  refusal, §3 Refusals). The pump emits raw normalised events; the
+  consumer maps.
+- **Window state / lifetime.** `WindowId` is opaque; the pump never
+  dereferences it. Ownership of `Window` / `Display` / `Surface`
+  belongs to the sibling design `specs/platform/window-display-design.md`
+  (spike #715).
+- **File watching.** `EventQueue<FileEvent>` is fed by the
+  `FileWatcher` I/O thread on a separate ring; the pump never
+  touches that queue. Routed to the sibling design (spike #719).
+- **Per-device polling.** Gamepad / sensor state polling
+  (`SDL_GetGamepadAxis`) is *not* the pump's responsibility; SDL3
+  delivers gamepad axis / button changes as events through
+  `SDL_PollEvent`, and the pump simply normalises those events. A
+  hypothetical "snapshot the current gamepad state at frame start"
+  primitive would belong to the input-mapping plugin, not here.
+- **IME composition state.** The pump receives a single
+  `SDL_EVENT_TEXT_INPUT` carrying the committed UTF-8 from SDL3's
+  internal IME and emits one `input::TextInput` event; the
+  composition / candidate-window UI is owned by the OS and surfaces
+  through SDL3's IME hooks (out of MVP scope per harmonius R-14.2.6
+  refusal).
+- **Frame pacing / display-link callbacks.** The pump does not block
+  on a vsync source. `frame-phases.md` phase 9 (present) drives
+  pacing through `CAMetalDisplayLink`; the pump is a phase-1 drain,
+  not a phase-9 wait.
+- **Cross-process replay.** Recorded event streams (e2e tests,
+  determinism replay) are produced by the e2e harness's
+  `InputDriver` injecting directly into `EventQueue<InputEvent>`
+  (`specs/e2e/SPEC.md`, `specs/platform/SPEC.md` §8.7). The pump
+  does not participate in capture or replay; it is the producer SDL3
+  feeds, and it is bypassed under replay.
+
+The SRP boundary is sharp: if the **SDL3 event vocabulary**, the
+**SDL3 → typed-variant translation table**, the **per-device
+ordering rule**, the **bounded SPSC ring protocol**, or the
+**per-frame drain trigger** change, this design changes. Anything
+else — input action mapping, window aggregate state, file watching,
+display pacing — is out of scope.
+
+## 2. Requirements coverage
+
+This section verifies that every harmonius MVP-scope requirement /
+design clause concerning the event-pump is either covered by the
+design below or explicitly refused with rationale. Inputs:
+
+- `harmonius/docs/requirements/input/device-abstraction.md`
+  (R-6.1.1 .. R-6.1.15) — the input event vocabulary.
+- `harmonius/docs/requirements/platform/window-display.md`
+  (R-14.1.9, R-14.1.11) — single-poller and bounded-channel rules.
+- `harmonius/docs/requirements/platform/os-integration.md`
+  (R-14.2.4 — drag/drop) — assessed for inclusion.
+- `harmonius/docs/design/input/` — implementation notes (treated as
+  research only; conclusions re-derived).
+
+| Harmonius clause                                                                                                                     | Glibre disposition (MVP)                                                                                                                                                                                                                                                                                                                                  |
+|--------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **R-6.1.1** key press/release/repeat with scancode + keycode + modifiers                                                             | **Covered.** §5.4 emits `input::KeyDown { KeyCode, ScanCode, ModifierMask, repeat }` and `input::KeyUp { KeyCode, ScanCode, ModifierMask }`. The translation table maps `SDL_EVENT_KEY_DOWN`/`UP` and `SDL_KeyboardEvent::repeat` directly into the typed payload (§3.3 below).                                                                            |
+| **R-6.1.2** scancode normalisation to USB HID                                                                                        | **Covered (delegated to SDL3).** SDL3's `SDL_Scancode` *is* the USB-HID-aligned namespace; the translation step is identity-by-cast into `glibre::platform::ScanCode`, sealed at compile time so a future SDL3 enum extension is a deliberate central edit, not a silent expansion (§3.3).                                                                |
+| **R-6.1.3** mouse button events (L / R / M / X1 / X2)                                                                                | **Covered.** `input::MouseButtonEv { MouseButton, pressed, x, y, click_count }`. SDL3's button index 1..5 maps to the closed enum `MouseButton::Left/Right/Middle/X1/X2`. Scroll lives in `Wheel` (§5.4 / §3.3).                                                                                                                                            |
+| **R-6.1.4** mouse delta / position in high-DPI                                                                                       | **Covered.** `MouseMove { x, y, dx, dy }` carries SDL3's already-DPI-corrected coordinates (`SDL_WINDOW_HIGH_PIXEL_DENSITY` set at window creation per SPEC §6.3). Delta comes from SDL3's relative-mouse-mode events; absolute `x, y` is in logical pixels per the §4.1 inv #4 rule.                                                                       |
+| **R-6.1.5** trackpad continuous scroll vs discrete wheel                                                                             | **Covered.** `Wheel { dx, dy, flipped }`. SDL3's `SDL_MouseWheelEvent::direction` produces `flipped`; floating-point `dx/dy` carries the precise trackpad delta. Discrete vs continuous distinction is not exposed as a separate field — consumers that need it inspect the magnitude (≥ 1.0 ⇒ likely discrete) per the input-plugin's interpretation rule. |
+| **R-6.1.6** unified gamepad abstraction (buttons, axes, triggers)                                                                    | **Covered.** `input::GamepadButtonEv { device, GamepadBtn, pressed }`, `input::GamepadAxisEv { device, GamepadAxis, value }`. Trigger is an axis (`LeftTrigger`, `RightTrigger`) per the `GamepadAxis` enum; its value range is `[-1, 1]` even though SDL3 reports `[0, 32767]` for triggers — the translation step rescales to the sum's invariant.        |
+| **R-6.1.7** gyroscope / accelerometer as ECS events                                                                                  | **Refused for MVP, deferred.** Sensor events would require new variants in the closed `InputEvent` sum (§4.2 inv #5: sealed at compile time). VR / motion-controller scope is out of MVP per `specs/platform/SPEC.md` §3 refusals; reopening the sum is a deliberate amendment spike when a controller-with-IMU lands as a target device.                  |
+| **R-6.1.8** per-device capability flags                                                                                              | **Refused at the pump.** Capability queries are state on the device handle, not events on the stream. Routed to a future `InputDevice` aggregate (sibling, post-MVP). The pump does not synthesise capability events.                                                                                                                                       |
+| **R-6.1.9** 10-finger touch tracking                                                                                                 | **Refused for MVP.** No touch hardware on the macOS-first baseline (`PHILOSOPHY` §macOS-first). Reopening the `InputEvent` sum for touch is gated on the second-target-platform spike.                                                                                                                                                                       |
+| **R-6.1.10** pen / stylus events                                                                                                     | **Refused for MVP.** Same gate as touch — pen events are platform-specific (Wacom, Apple Pencil) and require a sealed extension to `InputEvent`. Deferred.                                                                                                                                                                                                  |
+| **R-6.1.11** device hot-plug detection                                                                                               | **Partially covered.** SDL3 emits `SDL_EVENT_GAMEPAD_ADDED` / `_REMOVED`; the pump consumes these *internally* and emits no public glibre event today (§3.5 below — the input-plugin queries the active gamepad list at use time). Promoting hot-plug to a public `InputEvent` variant is gated by the input-plugin spike.                                  |
+| **R-6.1.12** hot-plug enumeration on a background thread                                                                             | **Refused at the pump.** SDL3 enumerates synchronously inside `SDL_PollEvent` on the main thread; budget-wise this is well under §9's `Pump::drain` cell. A background thread would re-introduce SDL3's main-thread-only constraint as a cross-thread invariant we'd then have to defend.                                                                   |
+| **R-6.1.13** sub-5 ms enumeration                                                                                                    | **Refused at the pump.** Boot-time concern; the input-plugin handles initial enumeration. The pump's per-frame budget is governed by §9, not by an enumeration deadline.                                                                                                                                                                                    |
+| **R-6.1.14** input device state in ECS                                                                                               | **Refused.** Routed to the input-plugin: phase 1 systems read the queue, write `Input` / `KeyboardState` / `GamepadState` components. The pump is a producer of events, never of components.                                                                                                                                                                |
+| **R-6.1.15** most-recently-used device tracking                                                                                      | **Refused.** Stateful; belongs to the input-plugin. The pump emits each event with its source device id (`device: u8` for gamepads); MRU tracking is a one-line phase-1 system.                                                                                                                                                                              |
+| **R-14.1.9** bounded async channel, exactly-once FIFO, never-drop                                                                    | **Covered.** `EventQueue<T>` is bounded (fixed capacity, set at construction; §5.3); SPSC head/tail is exactly-once FIFO (`memory_order_release` / `acquire`); never-drop is `§4.2 inv #3` — full-on-write is fatal, not coalesced, surfaced as `Error::IoFailure { OsCode { 0 } }` with prefix `"queue-full"`.                                              |
+| **R-14.1.11** single OS-event poller, centralised translation                                                                        | **Covered.** Exactly one `SDL_PollEvent` loop in the process — owned by `Pump::drain()`; the §6.3 SDL3 facade documents this as a load-bearing rule. Other contexts MUST NOT call `SDL_PollEvent` even transitively (verified by a build-time linker probe — see §11.1 test #4).                                                                            |
+| **R-14.1.10** DPI policy (`SystemScaled` vs `ApplicationScaled`)                                                                     | **Refused at the pump.** The pump emits `WindowEvent::DpiChanged { scale }` regardless of policy; the *response* (resize or stay) is the window aggregate's job (§6.3 cross-module note). The pump's role is the event, not the response.                                                                                                                  |
+| **R-14.2.4** drag-drop file events                                                                                                   | **Partially covered, partially deferred.** SDL3 emits `SDL_EVENT_DROP_FILE`. MVP routes drop into the `FileEvent` family carried by the `FileWatcher`-fed queue (sibling spike #719) rather than `WindowEvent`, because consumers of drop are content-import code that already drains the file queue. The pump consumes drop events internally and re-emits them through the file-event ring; see §3.6. |
+| **R-14.2.5** keyboard layout query                                                                                                   | **Refused.** Stateful; belongs to a future `InputDevice` aggregate. The pump emits `input::KeyDown { ScanCode, KeyCode }` already carrying both layouts of identity (scancode = layout-independent, keycode = layout-dependent), which covers the consumer use case without the layout-query API.                                                            |
+| **R-14.2.6** IME composition                                                                                                         | **Partially covered.** `input::TextInput { utf8, length }` carries the committed UTF-8 (max 32 B per event, multi-event for longer runs). Composition / candidate UI is owned by SDL3 / the OS; no glibre event is emitted for in-progress composition (§1 refusal).                                                                                          |
+| Harmonius design — `InputEvent` enum with per-device payloads (`InputEvent::KeyboardKey`, `InputEvent::GamepadAxis`, …)             | **Covered + collapsed.** Glibre's `InputEvent` is a sealed `eastl::variant` with the eight payloads listed in SPEC §5.4. Harmonius's per-device segmentation re-emerges as the `device: u8` field on the gamepad payloads (§5.4); keyboard / mouse are singleton in MVP (one keyboard, one mouse).                                                            |
+| Harmonius design — separate `WindowEvent`, `MonitorEvent`, `LifecycleEvent` channels                                                 | **Covered + collapsed.** One `EventQueue<WindowEvent>` carries `Resized`, `DpiChanged`, `Minimized`, `Restored`, `FocusGained`, `FocusLost`, `CloseRequested`, `DisplayChanged` (the latter folds harmonius's `MonitorEvent` into the same queue, preserving the cross-family ordering rule). Lifecycle (app-foregrounded etc.) collapses into focus events. |
+| Harmonius design — `EventIterator` consumer abstraction                                                                              | **Refused.** Glibre exposes `EventQueue<T>::drain(eastl::span<T> out) -> std::size_t` (SPEC §5.3) which fills a caller-supplied buffer and returns the count. An iterator object would hide the bounded-buffer contract; the span-out shape forces the consumer to size their drain explicitly.                                                              |
+| Harmonius design — coalescing of redundant resize events under flood                                                                 | **Refused.** §4.2 inv #4 forbids coalescing of `CloseRequested` and `DpiChanged`; for symmetry and simplicity, **no event family is coalesced at the pump**. Drag-resize floods (~120 events / s on macOS) fit comfortably inside the §9.2 ring-buffer sizing (~16k slots for `WindowEvent`), so coalescing buys nothing. Consumers that want a "latest-wins" resize read the queue tail and discard earlier events themselves. |
+| Harmonius design — async OS event delivery via OS-thread → game-loop SPSC                                                            | **Covered.** SPSC rings between the SDL3-fed pump and the engine's main-thread consumer (§5.3). The pump itself runs on the main thread (SDL3 constraint), but the SPSC discipline is preserved because `Pump::drain()` and `EventQueue<T>::drain()` are decoupled in time even when both run on the same thread — the drain consumer reads what the producer pumped earlier in phase 1. |
+
+Net result: every R-6.1.* and R-14.1.* / R-14.2.* requirement that
+falls into the event-pump's responsibility is either implemented as
+designed below or explicitly refused with rationale. The harmonius
+multi-channel event-bus collapses cleanly into glibre's two queues
+(input + window) plus the file-event queue owned by the sibling
+`FileWatcher` aggregate; sensor / touch / pen / capability surfaces
+are deferred behind a single sealed-sum-amendment gate.
+
+## 3. Detailed model
+
+The pump body is small (single-file, ~400 LOC at implementation
+time) but carries five load-bearing structural rules. Each is given
+a sub-section.
+
+### 3.1 The two-queue topology
+
+```text
+  +-------+    SDL_PollEvent     +------+   classify     +-----------------+
+  | SDL3  | -------------------> | Pump | -------------> | EventQueue<     |
+  | core  |   (one event/iter)   |      |    & translate | InputEvent>     |
+  +-------+                      +------+                +-----------------+
+                                    |  \                          ^
+                                    |   \                         | drain()
+                                    |    \---> EventQueue<        | (engine
+                                    |          WindowEvent>       |  main thread,
+                                    |    /---> ^                  |  phase 1
+                                    |   /                         |  consumer)
+                                    |  /  classify & translate    |
+                                    +-/                           |
+                                    | \                           |
+                                    |  \  drop (debug log)        |
+                                    +---> /dev/null (unknown SDL  |
+                                          event tag — §4.2 inv #5)|
+                                                                  |
+   FileWatcher I/O thread ----> EventQueue<FileEvent> ------------+
+   (separate ring, owned by FileWatcher aggregate, sibling design)
+```
+
+Per-frame in phase 1:
+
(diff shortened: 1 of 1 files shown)


specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+Capacity is fixed at construction (§4.2 inv #6) and a power of two
+(mask-and-wrap is one `&` instead of one `%`). Sizes are §9.2:
+~32k slots for `EventQueue<InputEvent>`, ~16k slots for
+`EventQueue<WindowEvent>`. Slot storage is a single contiguous
+allocation from the platform sub-arena tagged
+`platform::ContextTag` (§9.2 SPEC).
+
+Full-on-`try_push` returns `false`; the pump translates that to the
+fatal `Error::IoFailure { OsCode{0} }` with prefix `"queue-full"`
+per §10.3.2 SPEC. **No coalescing, no drop, no grow** — §4.2 inv
```
cjhowe-us, line 343: severity:HIGH location:specs/platform/event-pump-design.md:343 problem:The design asserts that the full-ring fatal path stamps prefix `"queue-full"` into the TLS diagnostic slot, but the code sketch at this line constructs `IoFailure { OsCode{0} }` without showing the `set_prefix(prefix::queue_full)` call that must precede it. Per `platform-error-design.md §3.3 / §6.2`, the prefix stamp is the only observable discriminator callers use to distinguish this `IoFailure` from other `IoFailure` arms; if the stamp is absent, a caller branching on `current_prefix() == prefix::queue_full` will see `prefix::none` and silently treat the queue-full crash as an unclassified I/O failure. The signal-safe path is irrelevant here (this is not a signal handler), so there is no exemption from stamping.
fix: In the prose of §3.4 and §10.3, make explicit that `Pump::drain` calls `detail::error::set_prefix(prefix::queue_full)` *before* constructing `unexpected(IoFailure{OsCode{0}})`. Add a note: "the prefix stamp is load-bearing; callers that receive `IoFailure` from `Pump::drain` read `current_prefix()` to confirm the queue-full case (§10.3)." This also brings §10.1's entry-point failure table into alignment with the platform-error-design §4.1.2 SDL3-prose convention that every translator stamps before returning.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+### 3.4 Bounded SPSC ring (`event/queue_impl.hpp`)
+
+The `EventQueue<T>::Impl` is the canonical lock-free SPSC ring:
+
+```cpp
+// engine/platform/src/event/queue_impl.hpp — internal.
+namespace glibre::platform::event::detail {
+
+template <class T>
+struct RingImpl {
```
cjhowe-us, line 308: severity:HIGH location:specs/platform/event-pump-design.md:308 problem:`RingImpl` places `head` (producer-written) and `tail` (consumer-written) in the same struct without any cache-line padding between them. On Apple Silicon (M1/M2 128-byte cache lines) both atomics fit inside a single cache line. Under true SPSC where producer and consumer run on *separate* threads, each `head.store(release)` and each `tail.store(release)` will invalidate the other thread's cache line — false sharing that turns the ring's `~0.05 µs` per-event push (§9.1) into a coherence-stall-dominated path. While the MVP has both producer and consumer on the main thread (§6.1), the design explicitly preserves SPSC discipline for future multi-threaded use and the §9.1 budget numbers are derived from M1 measurements that presumably ran single-threaded. If any future consumer moves off the main thread the performance contract breaks silently.
fix: Add `alignas(128)` to `head` and `tail` (or place them in separate structs padded to 128 bytes) in the `RingImpl` sketch. Annotate with: "Padding to a full Apple Silicon cache line (128 bytes) eliminates false sharing; the cost is 256 bytes of struct overhead, negligible against the 1 MiB slot array." This is a design-doc correctness fix; no implementation code exists yet, so the cost of the fix is one line of documentation.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+  using a relaxed store, then `head.store(new_head, release)`. The
+  release-store pairs with the consumer's acquire-load on `head`.
+- **Consumer drain (`EventQueue::drain`).** `head_snap = head.load(acquire)`,
+  then read up to `head_snap - tail_snap` slots, then
+  `tail.store(new_tail, release)`. The release-store on `tail`
+  pairs with the producer's acquire-load on `tail` to detect the
+  full-ring condition.
+
+Capacity is fixed at construction (§4.2 inv #6) and a power of two
+(mask-and-wrap is one `&` instead of one `%`). Sizes are §9.2:
```
cjhowe-us, line 335: severity:MED location:specs/platform/event-pump-design.md:335 problem:§3.4 says capacity is rounded up to a power of two, but §10.1 lists `Unsupported` for "non-power-of-two beyond cap". These two clauses contradict each other: if the implementation rounds up silently, a caller passing e.g. 100 gets a 128-slot queue with no error; if the implementation rejects non-powers-of-two, a caller passing 100 gets `Unsupported`. The implementation plan writer must choose one semantic, and it matters for the test in §11.1 (no unit test currently covers this edge).
fix: Pick one and document it unambiguously. The simpler user contract is round-up (caller need not know about power-of-two sizing), which also removes the contradiction with §10.1. Amend §10.1 to read: 'zero or wildly oversized capacity (> 2^24)' as the only `Unsupported` triggers; remove 'non-power-of-two beyond cap'. Add unit test #8b: 'with_capacity(100) succeeds and reports capacity() == 128'.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+unknown-event drop refusal is what makes that edit safe.
+
+### 3.6 Drag-drop ingestion (R-14.2.4 partial coverage)
+
+`SDL_EVENT_DROP_FILE` carries a UTF-8 path string. The pump:
+
+1. Calls `CanonicalPath::from_absolute(...)` on the SDL3-supplied
+   path. A failure (non-UTF-8, non-absolute) is logged at `info`
+   and dropped.
+2. Calls `FileWatcher::ingest_drop(canonical_path)` — an
```
cjhowe-us, line 382: severity:MED location:specs/platform/event-pump-design.md:382 problem:§3.6 says the pump calls `FileWatcher::ingest_drop(canonical_path)` — an internal-only entry point on a sibling aggregate — for every `SDL_EVENT_DROP_FILE`. This is a controlled cross-aggregate call inside the same plugin, but the design provides no two-concrete-users justification for the `ingest_drop` API itself. Per PHILOSOPHY §Anti-patterns, cross-domain APIs invented before two concrete users exist are rejected. Today there is exactly one caller: the pump's `SDL_EVENT_DROP_FILE` branch. 
fix: Either (a) document the second concrete user of `ingest_drop` (if one exists in this MVP scope), or (b) add a note to §12 Open Questions recording that this API exists for one user today and promotion to a stable internal contract is gated on a second caller. The note keeps the PHILOSOPHY §Anti-patterns audit trail complete without blocking this spike's closure.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+exactly-once-FIFO-never-drop, which the single-consumer SPSC delivers.
+
+### 6.4 Off-thread call refusal
+
+Calling `Pump::drain()` from any thread other than the main thread
+is a contract violation. Detection in MVP:
+
+- **Debug builds.** `Pump::drain` opens with
+  `assert(std::this_thread::get_id() == platform::main_thread_id)`.
+  The aggregate captures `main_thread_id` at construction.
```
cjhowe-us, line 646: severity:MED location:specs/platform/event-pump-design.md:646 problem:§6.4 describes the off-main-thread detection pattern as 'assert in debug, demote to return-`Unsupported` in shipping' and cross-references `error-model.md §"Severity Conventions"`. The `error-model.md` decision record does not contain a section titled 'Severity Conventions' — this is a dangling cross-reference. The error-model record discusses severity only in the Logging/Telemetry section (hot-reload refusals at `warn`, etc.) but does not define a 'Severity Conventions' taxonomy.
fix: Replace the citation with the actual mechanism: 'consistent with the assert-in-debug / warn-in-shipping pattern documented for hot-reload refusals in `error-model.md §Logging/Telemetry`.' Alternatively, if 'Severity Conventions' was added to the error-model record in a later PR, cite that PR number. As-is the citation cannot be resolved by an implementer.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+| **R-6.1.4** mouse delta / position in high-DPI                                                                                       | **Covered.** `MouseMove { x, y, dx, dy }` carries SDL3's already-DPI-corrected coordinates (`SDL_WINDOW_HIGH_PIXEL_DENSITY` set at window creation per SPEC §6.3). Delta comes from SDL3's relative-mouse-mode events; absolute `x, y` is in logical pixels per the §4.1 inv #4 rule.                                                                       |
+| **R-6.1.5** trackpad continuous scroll vs discrete wheel                                                                             | **Covered.** `Wheel { dx, dy, flipped }`. SDL3's `SDL_MouseWheelEvent::direction` produces `flipped`; floating-point `dx/dy` carries the precise trackpad delta. Discrete vs continuous distinction is not exposed as a separate field — consumers that need it inspect the magnitude (≥ 1.0 ⇒ likely discrete) per the input-plugin's interpretation rule. |
```
cjhowe-us, line 114: severity:MED location:specs/platform/event-pump-design.md:114 problem:The design claims SDL3's `SDL_MouseMotionEvent::x/y` are already DPI-corrected logical pixels because `SDL_WINDOW_HIGH_PIXEL_DENSITY` is set at window creation (§3.3). This is partially correct for the pixel-density flag, but SDL3's mouse-motion event `x/y` fields return coordinates in *logical* (not physical/pixel) units by default when the window does not have a pixel-mode view transform set. More precisely, with `SDL_WINDOW_HIGH_PIXEL_DENSITY` on macOS, SDL3 returns `x/y` in *points* (logical pixels), which match `LogicalSize` — but `dx/dy` for relative-motion events are in *pixels* when SDL3 relative-mouse mode is active. The design conflates the two and says both are 'DPI-corrected coordinates'. An implementer following this verbatim could produce a `MouseMove` payload where `dx/dy` are physical pixels while `x/y` are logical points, breaking the §4.1 invariant that coordinate (0,0) is 'top-left of the window's client area'.
fix: Distinguish the two clearly: 'Absolute `x/y` from `SDL_MouseMotionEvent` are in logical points (matching `LogicalSize` — safe to treat as window-relative coordinates). Relative `dx/dy` are in pixels when `SDL_SetRelativeMouseMode` is active; these must be divided by the current `DpiScale` before being placed in `MouseMove::dx/dy` to preserve the logical-coordinate invariant across the full payload.' Add a unit test asserting the division at DpiScale != 1.0.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+| **R-6.1.6** unified gamepad abstraction (buttons, axes, triggers)                                                                    | **Covered.** `input::GamepadButtonEv { device, GamepadBtn, pressed }`, `input::GamepadAxisEv { device, GamepadAxis, value }`. Trigger is an axis (`LeftTrigger`, `RightTrigger`) per the `GamepadAxis` enum; its value range is `[-1, 1]` even though SDL3 reports `[0, 32767]` for triggers — the translation step rescales to the sum's invariant.        |
```
cjhowe-us, line 115: severity:MED location:specs/platform/event-pump-design.md:115 problem:R-6.1.6 coverage claims trigger-axis value range is '[-1, 1] even though SDL3 reports [0, 32767] for triggers — the translation step rescales to the sum's invariant'. However, §3.3 later says both stick and trigger axes from SDL3 are '[-32768, 32767] for sticks and [0, 32767] for triggers' and the table rescales 'both to [-1, 1]'. Rescaling [0, 32767] to [-1, 1] means a fully-released trigger (SDL3 value 0) maps to -1.0, which is counter-intuitive (released = negative). The more natural mapping is [0, 32767] → [0.0, 1.0] for a unipolar trigger. This semantic gap matters: gameplay code that treats LeftTrigger value = 0.0 as 'released' will receive -1.0 and incorrectly see the trigger as pressed/pulled.
fix: Declare the explicit rescaling formula in §3.3: sticks: `value = raw / 32767.0f` (clamped to [-1, 1]); triggers: `value = raw / 32767.0f` (range [0.0, 1.0], not [-1, 1]). If the design intentionally unifies ranges for simplicity, document the rationale and add a unit test that asserts `translate_gamepad(raw=0, axis=LeftTrigger).value == 0.0f` (not -1.0f). The test in §11.1 item #6 currently only tests raw=32767 → 1.0f and does not cover the released case.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+The platform user-story for "engine consumes OS events" lives in
+`specs/e2e/SPEC.md`'s S1 trace: a 600-frame replay drives synthetic
+input events through the pump and asserts the resulting ECS
+component snapshots are byte-equal across runs. The pump-specific
+coverage is implicit in S1; no platform-specific E2E trace is owed
+beyond what S1 already exercises.
+
+The "reload mid-replay" e2e variant (§8.7 SPEC) covers the
+hot-reload survival path; runs against the same S1 trace with a
+`platform::test::enqueue_platform_reload` injected at frame 300.
```
cjhowe-us, line 1143: severity:LOW location:specs/platform/event-pump-design.md:1143 problem:The E2E coverage note (§11.3) claims pump coverage is 'implicit in S1' (the existing 600-frame S1 trace) and asserts 'no platform-specific E2E trace is owed beyond what S1 already exercises'. However, the event-pump design introduces three behaviours not exercised by a standard S1 input replay: (a) platform self-reload mid-replay (§8.2), (b) drag-drop file routing (§3.6), and (c) SDL_EVENT_QUIT synthesis as CloseRequested (§3.5). Integration tests #2, #5, #6 are called out in §11.2 but these are listed as integration tests, not E2E traces in the SPEC §11 sense. AGENTS.md requires that user-story acceptance criteria map to E2E traces before closure.
fix: Add a brief note clarifying that the 'reload-mid-replay' E2E variant is the trace for §8.2 (already implicit in §11.3 second paragraph but not explicitly named as an acceptance criterion). If the drag-drop and quit-synthesis behaviours are exercised by existing S1 event mixes, say so explicitly; if not, flag them as items for the integration-test plan when the platform implementation plan issue is opened. Prevents the plan issue author from assuming E2E coverage is complete when it may not be.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1199 @@
...
+- **Mouse coordinates come from SDL3's logical (window-relative)
+  pixels in `SDL_MouseMotionEvent::x/y`.** Window-relative is
+  required so coordinate (0, 0) is consistently top-left of the
+  window's client area regardless of compositor placement.
+- **Wheel deltas use `SDL_MouseWheelEvent::x/y` (floating-point on
+  modern SDL3).** `flipped` reflects the user's "natural scrolling"
+  preference per `SDL_MouseWheelEvent::direction`.
+- **Text input UTF-8.** SDL3 delivers UTF-8 already. The pump copies
+  up to 32 bytes into `input::TextInput::utf8` and sets
+  `length`. If SDL3 delivers > 32 bytes, the pump emits N successive
```
cjhowe-us, line 289: severity:LOW location:specs/platform/event-pump-design.md:289 problem:§3.3 says 'Text input UTF-8. SDL3 delivers UTF-8 already. The pump copies up to 32 bytes into `input::TextInput::utf8` ...' SDL3's `SDL_EVENT_TEXT_INPUT` delivers a fixed 32-byte null-terminated buffer in `SDL_TextInputEvent::text[32]`, not a variable-length string. The design describes a 'four-byte look-ahead codepoint scanner' to split strings longer than 32 bytes across multiple `TextInput` events. But SDL3's buffer is already bounded to 32 bytes by the SDK — SDL3 truncates at the buffer boundary, so glibre would never receive a single `SDL_EVENT_TEXT_INPUT` carrying more than 32 bytes regardless of the text length. The multi-event splitting logic for > 32 byte inputs is dead code.
fix: Clarify that SDL3 truncates at 32 bytes for `text[]` buffer. The multi-event splitting is not needed for SDL3 text input. If support for overlong codepoint sequences is desired in the future (e.g. custom text injection in tests), note that only the test harness path via `InputDriver` would need it. Remove or mark the four-byte look-ahead scanner as test-only.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:4
  LOW:2
carryover:0

notes:
The design is substantively complete and coherent against its declared scope. The twelve-section structure is present. SPEC §5.3 / §5.4 / §5.7 public surface is reproduced verbatim without adding new entry points. The SRP boundary is sharp: refusals to input mapping, window state, file watching, per-device polling, frame pacing, and IME composition are explicit and well-reasoned. EASTL types are used throughout (eastl::variant, eastl::optional, eastl::span, eastl::array); std:: is retained only where PHILOSOPHY §11 permits (std::expected, std::atomic, std::thread). Error returns use Result<T> = std::expected<T, glibre::Error> consistently; noexcept is on every public function. The hot-reload survival story (§8.1 / §8.2) is fully worked out. Performance budget cells are cited and decomposed. The requirements coverage table is thorough.

Two HIGH findings block implementation plan closure. (1) §3.4 / §10.3: the design shows IoFailure{OsCode{0}} constructed on the queue-full fatal path but does not document the mandatory detail::error::set_prefix(prefix::queue_full) call that must precede it; per platform-error-design §3.3, the prefix is the load-bearing discriminator — without it, callers testing current_prefix() == prefix::queue_full cannot distinguish queue-full from a generic SDL I/O failure. (2) §3.4: RingImpl places head (producer-written) and tail (consumer-written) atomics without cache-line padding; on Apple Silicon's 128-byte cache line both fit in one line, creating false-sharing coherence stalls if any future consumer thread is not the main thread — SPSC discipline is explicitly preserved for that future case in §3.1.

Four MED findings should be resolved before implementation begins: (a) with_capacity round-up vs reject contradiction between §3.4 and §10.1; (b) ingest_drop has one concrete caller and needs a §12 open-question note per PHILOSOPHY anti-patterns; (c) error-model.md §"Severity Conventions" citation is a dangling section reference; (d) mouse relative-motion delta dx/dy may be in physical pixels rather than logical points when SDL_SetRelativeMouseMode is active, leading to a mixed-coordinate MouseMove payload — needs explicit normalization in §3.3.

Two LOW findings: trigger-axis rescale semantics for the released state (raw=0 → -1.0 vs 0.0) need a unit test and explicit formula; E2E coverage note does not name drag-drop and quit-synthesis as named acceptance criteria for the implementation plan.

The CC subject is correct: docs(platform) with issue ref #717. The PR is merged; these findings should be addressed in a follow-up docs amendment PR against the same spike issue, or carried forward as [BLOCKING IMPLEMENTATION] items in the task-breakdown spike (#714 descendant) that this deliverable unblocks.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:4
  LOW:2
carryover:0

notes:
The design is substantively complete and coherent against its declared scope. Twelve-section structure present. SPEC §5.3 / §5.4 / §5.7 public surface reproduced verbatim; no new entry points introduced. SRP boundary is sharp; refusals are explicit and well-reasoned. EASTL types used throughout per PHILOSOPHY §11. Result<T> / noexcept discipline consistent on all public symbols. Hot-reload survival story worked out. Performance budget cells cited and decomposed.

Two HIGH findings block implementation plan closure: (1) §3.4/§10.3 missing set_prefix(queue_full) stamp before IoFailure construction on the queue-full path — the prefix is the load-bearing discriminator per platform-error-design §3.3; (2) §3.4 RingImpl lacks cache-line padding between head and tail atomics — false sharing hazard on Apple Silicon when any future consumer runs off the main thread.

Four MED: with_capacity round-up vs reject contradiction (§3.4 vs §10.1); ingest_drop has one concrete caller with no §12 note (PHILOSOPHY anti-pattern gate); error-model §Severity Conventions is a dangling section reference (§6.4); mouse relative delta dx/dy may be physical pixels when SDL_SetRelativeMouseMode is active, producing mixed-coordinate MouseMove payload (§3.3 R-6.1.4).

Two LOW: trigger-axis rescale semantics for raw=0 undefined (§3.3 / §11.1 test #6); E2E coverage note does not name drag-drop and quit-synthesis as acceptance criteria (§11.3).

PR is merged. Findings should be addressed in a follow-up docs amendment PR or carried as BLOCKING IMPLEMENTATION items in the task-breakdown spike that #717 unblocks.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:4
  LOW:2
carryover:0

notes:
The design is substantively complete and coherent against its declared scope. Twelve-section structure is present; all twelve sections match the specs/_TEMPLATE.md shape. The SPEC §5.3/§5.4/§5.7 public surface is reproduced verbatim without adding new entry points. SRP boundary is sharp: refusals to input mapping, window state, file watching, per-device polling, frame pacing, and IME composition are explicit and well-reasoned. EASTL types are used throughout (eastl::variant, eastl::optional, eastl::span, eastl::array) per PHILOSOPHY §11; std:: retained only for expected, atomic, thread. Every public function is noexcept; every fallible function returns Result<T>. Hot-reload survival story (§8.1/§8.2/§8.4) is fully worked. Performance budget cells are cited from perf-budget.md and decomposed. The requirements coverage table is thorough and honest about R-6.1.7/R-6.1.9/R-6.1.10 deferrals.

Two HIGH findings must be addressed before the implementation plan PR opens:
(1) §3.4 / §10.3: IoFailure{OsCode{0}} is constructed on the queue-full fatal path without the mandatory detail::error::set_prefix(prefix::queue_full) call that must precede it. Per platform-error-design §3.3, the TLS prefix is the load-bearing discriminator — a caller testing current_prefix() == prefix::queue_full to distinguish queue-full from a generic SDL I/O failure will see prefix::none and silently misroute the fatal error.
(2) §3.4 RingImpl: head (producer-written) and tail (consumer-written) atomics share no cache-line padding. On Apple Silicon's 128-byte cache line both fit in one line, producing false-sharing coherence stalls the moment any future consumer moves off the main thread. The design explicitly preserves SPSC discipline for that future case (§3.1); the struct sketch should reflect it with alignas(128).

Four MED: (a) with_capacity round-up vs reject contradiction between §3.4 and §10.1; (b) ingest_drop has exactly one concrete caller today with no §12 open-question note required by PHILOSOPHY §Anti-patterns; (c) error-model.md §Severity Conventions is a dangling section reference in §6.4 — the section does not exist in the accepted record; (d) mouse relative-motion delta dx/dy are physical pixels when SDL_SetRelativeMouseMode is active, producing a mixed logical/physical coordinate MouseMove payload in violation of the §4.1 inv #4 single-conversion rule.

Two LOW: trigger-axis rescale for raw=0 is semantically ambiguous (released trigger → -1.0 vs 0.0); §11.3 E2E coverage note does not explicitly name drag-drop and quit-synthesis as acceptance criteria.

PR is merged; these findings should be addressed in a follow-up docs amendment PR against spike #717, or carried as [BLOCKING IMPLEMENTATION] open-question entries in §12 of the design and in the task-breakdown spike that this deliverable unblocks. CC subject docs(platform) is correct; issue ref #717 is present.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Added mandatory `detail::error::set_prefix(prefix::queue_full)` call in §3.4 (prose describing the full-ring fatal path, immediately before `unexpected(IoFailure{OsCode{0}})` construction) and in §10.3 step 2 (queue-full recovery shape). Added explanatory note: TLS prefix stamp is mandatory before any `IoFailure{OsCode{0}}` where the prefix is the only discriminator; stamp is cold-path-only. Verified `prefix::queue_full` exists in platform-error-design.md §3.3.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Added `alignas(128)` to both `head` and `tail` in the `RingImpl` struct sketch in §3.4, with inline annotation: 'alignas(128) padding mandatory for SPSC discipline on Apple Silicon (128B cache line). Without padding, every producer push invalidates the consumer's cache line on the other core via false sharing, defeating the lock-free benefit. The 256-byte struct overhead is negligible against the 1 MiB slot array.'

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Chose option (a): kept silent round-up. Removed 'non-power-of-two beyond cap' from §10.1 `Unsupported` triggers; the entry now reads 'zero capacity / wildly oversized capacity (> 2^24)'. Updated §3.4 prose to state `with_capacity(N)` silently rounds N up. Added §11.1 test #8b: `event_pump_with_capacity_rounds_up_to_power_of_two` asserting `with_capacity(100).capacity() == 128`.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Added one-caller note in §3.6 body: '`ingest_drop` has one concrete caller at MVP (this branch). Per PHILOSOPHY §Anti-patterns, promotion to a stable cross-aggregate internal contract is gated on a second caller materialising.' Added §12 [NON-BLOCKING] entry: '`ingest_drop` two-concrete-users gate — currently has one caller. May be removed if second caller never appears.'

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Fixed §6.4 dangling cross-ref: replaced `§"Severity Conventions"` with `§"Logging/Telemetry"` and added '(see hot-reload refusal severity discussion under that heading)'. The error-model.md record does not contain a 'Severity Conventions' section; the relevant severity discussion lives under Logging/Telemetry.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Rewrote §3.3 mouse coordinate paragraph to clearly distinguish the two coordinate spaces: absolute x/y from `SDL_MouseMotionEvent` are logical points (safe); relative dx/dy from `SDL_SetRelativeMouseMode` are physical pixel deltas on macOS and must be divided by `window.dpi_scale()` before being placed in `MouseMove::dx/dy`. Added documented invariant: every `MouseMove` payload reports x, y, dx, dy in logical points. Added §11.1 test #3b at DpiScale != 1.0 asserting the divide-by-scale path. Updated R-6.1.4 row in the requirements coverage table.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Added explicit trigger rescale formula to §3.3: sticks use `raw / 32767.0f` (clamped to [-1,1]); triggers use `axis = (raw / 32767.0f) * 2.0f - 1.0f`. Documented released-state: 'raw=0 → axis=-1.0; gameplay code that wants released-as-zero must remap (axis + 1.0f) / 2.0f at the consumer.' Added §11.1 test #6b: `event_pump_trigger_released_state_is_negative_one` asserting raw=0 → axis=-1.0f. Updated R-6.1.6 row.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Expanded §11.3 to explicitly name drag-drop and quit-synthesis E2E coverage status. Drag-drop: asserted in S1 trace step 7 (file-drop-into-window-surface) plus integration test #5. Quit-synthesis: integration-test-only at MVP (no E2E trace covers menu-driven quit; S1 trace exits via window close, not Cmd-Q). Added §12 [NON-BLOCKING] tracking the quit-synthesis E2E gap. Updated §11.4 coverage matrix.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:89cb6a3 followup:#900

Corrected §3.3 TextInput description: SDL3's `SDL_TextInputEvent::text[32]` is a fixed 32-byte null-terminated buffer; SDL3 truncates at that boundary. The four-byte look-ahead codepoint scanner from earlier drafts is dead code for the SDL3 path and is explicitly not implemented in the production translator. Test harness `InputDriver` is responsible for splitting overlong synthetic strings. Updated §11.1 test #7 to `event_pump_text_input_copies_sdl_buffer` (single 28-byte string → one event; no splitting path in production).

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1304 @@
...
+| **R-6.1.4** mouse delta / position in high-DPI                                                                                       | **Covered.** `MouseMove { x, y, dx, dy }` reports all four fields in logical points. Absolute `x/y` from `SDL_MouseMotionEvent` are already in logical points (matching `LogicalSize` when `SDL_WINDOW_HIGH_PIXEL_DENSITY` is set). Relative `dx/dy` from `SDL_SetRelativeMouseMode` are in physical pixels and must be divided by `window.dpi_scale()` (from window-surface-design.md) before being placed in the payload. Invariant: every `MouseMove` payload reports x, y, dx, dy in logical points (see §3.3 for the rescale and unit test).                                                                       |
```
cjhowe-us: severity:MED location:specs/platform/event-pump-design.md:120 problem:R-6.1.4 row in §2 says relative dx/dy 'must be divided by `window.dpi_scale()` (from window-surface-design.md)'. This cross-references the Window aggregate API. However, §3.3 correctly documents that `Pump::Impl` caches its own `DpiScale` copy (initialised 1.0f; updated on `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`), and `translate_mouse` receives this cached value by parameter. The pump does NOT call `window.dpi_scale()` at event time; the `window.dpi_scale()` wording in §2 implies a live cross-aggregate call on the hot path, which contradicts §3.3's caching design. This is a naming alias: '§2 uses the Window-aggregate accessor name; §3.3 uses the Pump::Impl cache'. fix:Update the §2 R-6.1.4 disposition row to say: 'divided by `Pump::Impl::cached_dpi_scale` (a cached copy of the owning window's `DpiScale`, updated on `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED` before `translate_mouse` is called — §3.3)'. Remove the 'from window-surface-design.md' parenthetical or replace it with a cross-ref to §3.3.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1304 @@
...
+## 10. Failure modes
+
+This section refines §10.3.2 SPEC for the pump aggregate. Every
+failure is a `glibre::Error` arm (with the platform-private
+`platform::Error` rolled into one variant) per
+`reviews/decisions/error-model.md`.
+
+### 10.1 Per-entry-point failure surface (cited from SPEC §10.3.2)
+
+| Entry point                     | Returnable arms                                  | Trigger                                         |
```
cjhowe-us, line 931: severity:MED location:specs/platform/event-pump-design.md:985 problem:SPEC §10.3.2 table (in SPEC.md line 2484) still shows trigger for `EventQueue::with_capacity` as 'zero / wildly oversized capacity' — it does not mention the silent round-up behavior established by R1 (non-power-of-two silently rounds up; not an error). The design's §10.1 was correctly updated to say 'non-power-of-two is silently rounded up (not an error)' but the authoritative SPEC §10.3.2 was not amended. Since this design header claims it 'Refines those sections in place', the SPEC §10.3.2 trigger summary is now stale and contradicts §10.1 of this design. A reader consulting SPEC §10.3.2 only would believe non-power-of-two triggers Unsupported. fix:The design §10.1 note already correctly states the contract. Add a parenthetical note at the end of the §10.1 table note: 'Note: SPEC §10.3.2 trigger summary predates this refinement; §10.1 above is the authoritative statement; SPEC §10.3.2 should be updated to read: zero capacity / wildly oversized capacity (> 2^24); non-power-of-two silently rounded up.' Alternatively, open a follow-up amendment spike to update SPEC §10.3.2 directly.

specs/platform/event-pump-design.md
```diff
@@ -0,0 +1,1304 @@
...
+  `gameplay` / `tools` plugins' job (collapsed into harmonius's R-6.2
+  refusal, §3 Refusals). The pump emits raw normalised events; the
+  consumer maps.
+- **Window state / lifetime.** `WindowId` is opaque; the pump never
+  dereferences it. Ownership of `Window` / `Display` / `Surface`
+  belongs to the sibling design `specs/platform/window-display-design.md`
+  (spike #715).
+- **File watching.** `EventQueue<FileEvent>` is fed by the
+  `FileWatcher` I/O thread on a separate ring; the pump never
+  touches that queue. Routed to the sibling design (spike #719).
```
cjhowe-us, line 62: severity:LOW location:specs/platform/event-pump-design.md:62 problem:The 'File watching' refusal bullet says 'the pump never touches that queue' (referring to EventQueue<FileEvent>). With responsibility 6 added in R2, the pump now explicitly calls `FileWatcher::ingest_drop`, which internally enqueues into the file event ring. The pump still does not directly write to EventQueue<FileEvent>, so the technical claim is accurate at the pointer-to-slot level. However, a reader who reads only this refusal bullet and then encounters §3.6 will be confused. The clarification paragraph at line 93 partially addresses this but the refusal item itself creates the false expectation. fix:Amend the 'File watching' bullet to: 'File watching. EventQueue<FileEvent> is owned by the FileWatcher aggregate; the pump does not directly enqueue onto that ring. Drop events (SDL_EVENT_DROP_FILE) are forwarded to FileWatcher::ingest_drop (§3.6) — the pump touches file-drop routing but never owns the ring itself. Routed to sibling design (spike #719).' This aligns the refusal with the now-documented responsibility 6.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0

notes:All R1 and R2 findings were correctly addressed across commits 89cb6a3 and caa746d on PR #900. Round-3 topology + spec alignment + polish pass surfaces zero HIGH, two MED, and one LOW.

CC subjects are correct: both commits use docs(platform): scope, reference #826 and #717, and the body messages are descriptive. Sub-issue parenting is correct: #717 is type:spike under #714 (type:sub-epic), carries no pts: label (correct for spikes), and remains OPEN while the deliverable review cycle completes.

MED-1 (comment 3189778385): Cross-context naming alias in R-6.1.4 row. The §2 requirements table says relative dx/dy 'must be divided by window.dpi_scale()' referencing the Window aggregate accessor, while §3.3 correctly documents that Pump::Impl holds its own cached DpiScale and passes it as a parameter to translate_mouse. The §2 wording implies a live cross-aggregate call on the hot path — contradicting the caching design in §3.3. Recommendation: align §2 to say 'divided by the Pump::Impl cached DpiScale (see §3.3)'.

MED-2 (comment 3189779779): SPEC §10.3.2 stale trigger. The SPEC file still records the EventQueue::with_capacity trigger as 'zero / wildly oversized capacity' without mentioning that non-power-of-two is silently rounded up. The design §10.1 was correctly updated by R1 but the authoritative SPEC §10.3.2 was not. A reader consulting the SPEC only would believe non-power-of-two triggers Unsupported. Recommendation: add a note to design §10.1 acknowledging the SPEC §10.3.2 pre-dates this refinement, and open a follow-up item to update SPEC §10.3.2 (either as a §12 OPEN entry or a direct amendment).

LOW-1 (comment 3189780950): §1 'File watching' refusal bullet says 'the pump never touches that queue'. With responsibility 6 (drop-event routing to FileWatcher) now explicit, a reader encounters the refusal first and then sees §3.6. The refusal is technically correct at the direct-push level but misleading given the ingest_drop call. Recommendation: amend to acknowledge that drop events are forwarded to FileWatcher::ingest_drop (§3.6) without the pump owning the ring.

Positives: §10.3/§3.4 stamp-before-log ordering is consistent and well-reasoned. §6.4 SPEC invariant cross-references (§4.1 inv #2, §4.2 inv #1) are correct. §11.4 coverage matrix rows #3b, #6b, #8b are present and correctly linked. translate_mouse(DpiScale) parameter is correctly an internal function; PHILOSOPHY anti-patterns two-concrete-users gate does not apply. §12 [NON-BLOCKING] entries for ingest_drop gate and quit-synthesis E2E gap are properly categorised. §1 SRP responsibility 6 rationale paragraph is clear. No new public surface introduced. PR #826 is merged; PR #900 addresses these amendments.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:672e287 followup:#900

Rewrote §2 R-6.1.4 coverage cell: replaced `window.dpi_scale()` reference with explicit statement that the divisor is the `Pump::Impl`-cached `DpiScale` (initialised 1.0f, refreshed in §3.5 on `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`), passed by value to `translate_mouse` — no cross-aggregate call on the hot path. Text now agrees with §3.3 mouse-coord rule.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:672e287 followup:#900

Added a note paragraph in §10.1 after the failure-seam paragraph explicitly calling out that SPEC §10.3.2 predates this refinement and must be amended (trigger: 'capacity overflow beyond u32::max'). Added §12 [BLOCKING IMPLEMENTATION] entry naming the two concrete consumers and requiring the SPEC amendment before any plan PR consuming this design lands.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:672e287 followup:#900

Rewrote §1 file-watching refusal bullet: now reads 'The pump does not own or write to FileWatcher::EventQueue<FileEvent>. SDL_EVENT_DROP_FILE values are forwarded to FileWatcher::ingest_drop (see §3.6 + responsibility 6 in §1), which is the file-watcher's own enqueue path — the pump produces a single function call, not a ring write.' Accurately reflects the routing path while preserving the ownership boundary.

Issue #706, opened 2026-05-03
[SPIKE] design-core-hot-reload-barrier-detailed


cjhowe-us:
Parent: #699

## Question

Produce the detailed design for the **hot-reload-barrier** aggregate in the `core` context.

drain → swap → migrate → resume at phase 8.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/core/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/core-runtime/hot-reload-protocol.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **hot-reload-barrier** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/core/hot-reload-barrier-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(core): detailed design — hot-reload-barrier (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-core-hot-reload-barrier-detailed` is blocked by this issue.

cjhowe-us:
agent:design-core-hot-reload-barrier
status:started
issue:#706
branch:docs/core-hot-reload-barrier-design
worktree:/Users/cjhowe/Code/glibre
host:mac
cloud:none
commit:none
pr:none
notes:Required reads loaded (PHILOSOPHY, AGENTS, SETUP, _TEMPLATE, specs/core/SPEC.md, decision records hot-reload-protocol/frame-phases/plugin-abi/error-model/perf-budget/fory-codegen, harmonius hot-reload-protocol). Authoring specs/core/hot-reload-barrier-design.md with the 12 mandated sections; re-deriving from glibre's locked decision records, treating harmonius as input only.

cjhowe-us:
agent:design-core-hot-reload-barrier
status:done
issue:#706
branch:docs/core-hot-reload-barrier-design
worktree:/Users/cjhowe/Code/glibre
[...]

Issue #704, opened 2026-05-03
[SPIKE] design-core-plugin-loader-detailed


cjhowe-us:
Parent: #699

## Question

Produce the detailed design for the **plugin-loader** aggregate in the `core` context.

PluginLoader + Manifest + ABI hash.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/core/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/core-runtime/plugin-system.md
- harmonius/docs/design/core-runtime/events-plugins.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **plugin-loader** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/core/plugin-loader-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(core): detailed design — plugin-loader (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-core-plugin-loader-detailed` is blocked by this issue.

cjhowe-us:
agent:design-core-plugin-loader
status:started
issue:#704
branch:docs/core-plugin-loader-design
worktree:/Users/cjhowe/Code/glibre
host:mac
cloud:none
commit:none
pr:none
notes:Claimed leaf. Re-derived plugin-loader detailed design from SPEC §4.5 / §5.9 / §6.6 / §8 / §10 plus the plugin-abi, hot-reload-protocol, error-model, fory-codegen, frame-phases, perf-budget decision records. Verified harmonius R-1.6.* coverage (R-1.6.1/3/4/5/6/7/9 mapped; R-1.6.2 groups, R-1.6.8 capabilities refused; R-1.6.10 shipping static-link out of MVP loader scope). Drafting 12-section detailed design to specs/core/plugin-loader-design.md next.

cjhowe-us:
agent:design-core-plugin-loader
[...]

Issue #357, opened 2026-05-03
[STORY] platform/file-watcher: off-main-thread, recursive subscription, RAII teardown


cjhowe-us:
Parent sub-epic: #36

### Domain
platform

### Phase
phase:mvp

### Persona
engine developer (P-26)

### User Story
As an engine developer, I want `FileWatcher`'s native subscriptions (FSEvents on macOS) to run off the main thread and be torn down deterministically when the watcher is destroyed, so that no kqueue fd / FSEvents stream leaks and the main thread is never blocked by file watch delivery (§4.3 inv #4, #5).

### Acceptance Criteria (Gherkin)
```gherkin
Given a `FileWatcher` watching a 10k-file tree
When `Pump::drain()` runs on the main thread
Then the main-thread CPU cost contributed by `FileWatcher` is **zero** (§9.1, §4.3 inv #5)
And events flow into the engine via an off-main-thread funnel that the next `Pump::drain()` pulls

Given a `FileWatcher` is destroyed (RAII)
When the destructor returns
Then every native subscription it held is released (§4.3 inv #4)
And `lsof` / `fs_usage` shows no orphaned FSEvents stream attributable to the watcher

Given the platform plugin's hot-reload swap at phase 8 (§8.1, §8.3)
When the watcher is migrated by Q::register
Then the subscription list (`std::span<CanonicalPath>`) is preserved
And native fds are re-acquired
And the engine receives fresh `Created` events for each watched path's contents (§8.1 table row "FileWatcher")
```

### Manual Test Script
1. Build `platform_file_watcher_lifetime_smoke`.
2. Run on macOS / M1; the harness creates / destroys 100 watchers in a loop, each watching `/tmp/glibre-fw-test/`.
3. After the loop, run `fs_usage -f filesys -e $$ | grep glibre` (or the platform-equivalent FSEvents inspector); confirm zero stream subscriptions remain.
4. Run the harness with `--cpu-trace` and confirm the main-thread CPU cost is statistically indistinguishable from baseline (no `FileWatcher` work on the main thread).

### E2E Test Plan
`tests/e2e/platform/file-watcher-lifetime.glibre-trace` — creates and destroys watchers, asserts via a system-resource snapshot helper that no FSEvents stream / inotify watch / kqueue fd leaks; asserts main-thread CPU contribution is ≤ 0.001 ms / frame across 600 frames.

### Closure Checklist (in order)
- [ ] E2E test authored and passing in CI
[...]

Issue #356, opened 2026-05-03
[STORY] platform/file-watcher: canonical paths + content-hash dedup + atomic renames


cjhowe-us:
Parent sub-epic: #36

### Domain
platform

### Phase
phase:mvp

### Persona
game developer (P-15)

### User Story
As a game developer, I want a `FileWatcher` whose only public output is `(canonical_path, FileEvent)` tuples — symlinks resolved, case folded per macOS rule, content-hash deduplicated, renames atomic — so that hot-reload consumers never re-canonicalize and never re-read for hash equality (§4.3, §3 collapse "Filesystem watcher ports + hash cache + canonicalization → `FileWatcher`").

### Acceptance Criteria (Gherkin)
```gherkin
Given a `FileWatcher` subscribing recursively to `/tmp/glibre-watch-test/`
And the directory contains `a.txt` (symlinked from `link.txt`)
When I `touch link.txt`
Then exactly one `FileEvent::Modified { canonical_path: "/private/tmp/.../a.txt" }` is emitted (symlink resolved per §4.3 inv #1)
And no `FileEvent` carrying `link.txt` ever escapes

Given two OS events arriving for `a.txt` whose contents are byte-identical (e.g. an editor's atomic save that touches mtime but not content)
When the watcher processes them within its debounce window
Then they collapse to **one** `FileEvent::Modified` (§4.3 inv #2 — hash dedup)

Given the OS reports a `Deleted{a.txt}` followed immediately by `Created{a.txt}` with the same inode / file-id
When the watcher reassembles the pair
Then exactly one `FileEvent::Renamed { from: a.txt, to: a.txt }` is emitted (§4.3 inv #3) — never the unreassembled pair

Given a path that fails canonicalization (e.g. invalid UTF-8 in a filename)
When the watcher's ingestion site processes it
Then `PlatformError::IoFailure` is reported via the watcher's error stream (§4.3 inv #1)
And no half-formed `FileEvent` with a non-canonical path leaks
```

### Manual Test Script
1. Build `platform_file_watcher_smoke`.
2. `mkdir -p /tmp/glibre-watch && ln -s /tmp/glibre-watch /tmp/glibre-watch-link`. Run the harness watching `/tmp/glibre-watch-link/`.
3. `echo hello > /tmp/glibre-watch/a.txt`; harness prints `Modified{/private/tmp/glibre-watch/a.txt}` (canonical, with `/private/` prefix on macOS).
[...]

PR #827, opened 2026-05-04
docs(platform): detailed design — file-watcher (refs #719)

cjhowe-us:
## Summary

- Adds `specs/platform/file-watcher-design.md` — 12-section detailed design for the platform `FileWatcher` aggregate (SPEC §4.3, §5.8, §6.4, §8, §9, §10.5).
- macOS-only MVP: direct C FSEvents (`FSEventStreamCreate` / `CFRunLoopRun`), no Obj-C++, no SDL3 fs-events. One I/O thread per `FileWatcher` instance hosts the runloop and feeds per-token SPSC rings drained by `take_events` from the main thread.
- SRP-tight: watcher emits debounced + content-hash-deduped `FileEvent`s; does NOT own the hot-reload barrier (#706), the plugin loader (#704), a polling fallback (consumer-side per SPEC §10.5), or asset reload bodies.
- No new public surface beyond the locked §5.8 stub. No `.fory` schemas; nothing persisted; zero contribution to `host_glibre_types_abi_hash` (per SPEC §7.3 refusal).

Refs spike #719. Issue stays OPEN.

## Test plan

- [ ] CI lint / format / docs gates green.
- [ ] Reviewers verify §2 covers harmonius `R-14.6.5` / `.6` / `.7` / `.8` / `.9` (covered or refused with rationale).
- [ ] §10 mapping of named modes (`FSEventsStartFailed` / `QueueOverflow` / `RootMissing`) onto existing `platform::Error` arms accepted.
- [ ] §11 unit + integration test plan accepted as the basis for the sibling task-breakdown spike.

### DIFF
--- a/specs/platform/file-watcher-design.md
+++ b/specs/platform/file-watcher-design.md
@@ -1,0 +1,1199 @@
+# platform — Detailed Design: file-watcher aggregate
+
+> Detailed design for the `FileWatcher` / `FileEvent` / `WatchToken`
+> aggregate declared in `specs/platform/SPEC.md` §4.3. Refines §4.3,
+> §5.8, §6.4, §8.1 (peer-reload pass-through and platform-self-reload
+> watcher row), §8.3 step 3 + step 4.3 (subscription capture and fresh-
+> event replay), §8.5.2 (`WatcherSurvival` middleman), §8.6 (file-event
+> consumer reseat), §9.1 (FileWatcher per-frame budget = 0), §9.2
+> (4 MiB sub-arena), §10.3.3 (returnable arms by entry point) and §10.5
+> (`watcher-unavailable` recovery shape). It does **not** redefine the
+> §5 stub.
+>
+> Cites `reviews/decisions/error-model.md`,
+> `reviews/decisions/frame-phases.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`,
+> `reviews/decisions/fory-codegen.md`. No public surface is introduced
+> beyond the §5.8 stub locked in `specs/platform/SPEC.md`; deviations
+> require an amendment spike, not an in-place edit.
+>
+> Harmonius prior art (`harmonius/docs/requirements/platform/
+> filesystem.md`, `R-14.6.5` / `R-14.6.6` / `R-14.6.8` / `R-14.6.9`)
+> cited as research input only — every conclusion below was
+> re-derived per `PHILOSOPHY.md`.
+
+Refs: spike #719 — `[SPIKE] design-platform-file-watcher-detailed`.
+Parent: #714. Sibling task-breakdown spike blocked by this deliverable.
+
+## 1. Purpose
+
+The `file-watcher` aggregate is the single component in the platform
+context permitted to register an `FSEventStreamRef` with macOS Core
+Services and to drive its callback-thread → main-thread fan-out. Its
+one responsibility is **owning the lifetime of recursive directory
+subscriptions plus the debounced, deduplicated `FileEvent` stream
+attached to them**: subscribe a `CanonicalPath` root, run an internal
+I/O thread that feeds the FSEvents runloop, canonicalize raw paths,
+collapse duplicates inside a per-root debounce window, reassemble
+atomic renames, and vend the `FileEvent`s through a per-token SPSC
+ring drained from the main thread by `FileWatcher::take_events`.
+
+What this aggregate explicitly refuses to own:
+
+- **Hot-reload barrier** — `core::HotReloadBarrier` (sibling spike
+  #706, design `specs/core/hot-reload-barrier-design.md`) consumes
+  `FileEvent`s through whatever consumer holds the `WatchToken`, but
+  the watcher does not call into the loader; it has no knowledge of
+  plugin dylibs, ABI hashes, or migration tables.
+- **Plugin loader / `dlopen` mechanics** — `core::PluginLoader`
+  (#704). The watcher emits filesystem events; the loader decides
+  what (if anything) to reload.
+- **Asset reload bodies** — owned by `content` (post-MVP) and by the
+  consuming plugin's `glibre_plugin_register`. The watcher signals
+  *that* a path changed; consumers decide *what to do*.
+- **Polling fallback** — when `watcher-unavailable` recurs (SPEC §10.5
+  step 4) the consumer (typically the editor / hot-reload coordinator)
+  spins a `FileIo::stat_path` / `list_dir` shim. SPEC §10.5 step 5
+  is explicit: "Building polling into the aggregate would re-merge
+  the responsibilities §4.3 SRP keeps apart." The watcher refuses.
+- **Window / event pump** — `Pump`, `EventQueue<InputEvent>`,
+  `EventQueue<WindowEvent>` (sibling #717). `FileEvent` is *not*
+  funnelled through `Pump::drain`; it has its own per-token drain
+  (SPEC §5.8 `take_events`). The collapse rule is in §3.5 below.
+- **`FileIo` blocking / async I/O** — sibling #721. The watcher does
+  no file *content* reads; the BLAKE3 dedup mentioned in SPEC §6.4
+  reads bytes from the OS only inside the I/O thread, never on the
+  main thread, and never through `FileIo`. The two aggregates share
+  the `CanonicalPath` value object and nothing else.
+- **`Clock` / time** — sibling #723. The debounce window timestamps
+  events with `Clock::now()`, but the `Clock` aggregate is taken by
+  reference; the watcher does not own a time source.
+- **Persistence of the subscription list across process restart** —
+  refused per SPEC §7.3 ("File-watcher canonical-path subscription
+  list (cache)"). Subscriptions are reseated by the consumer on every
+  startup; persisting them in platform would duplicate `content`'s
+  authoritative ownership and break determinism (a re-run with a
+  stale list emits a different event stream than a fresh run).
+- **Platform-side error policy** — closed sum `platform::Error` is
+  owned by §4.7 / §10. This design surfaces failures into pre-existing
+  arms (`NotFound`, `PermissionDenied`, `Unsupported`,
+  `IoFailure { OsCode }` with `"watcher-unavailable"` prefix); it does
+  not invent new arms.
+- **Obj-C / Obj-C++ glue** — CLAUDE.md "No Obj-C++ in engine code".
+  The macOS FSEvents API is pure C (`FSEventStreamCreate`,
+  `FSEventStreamScheduleWithRunLoop`, `CFRunLoopRun`,
+  `FSEventStreamRelease` from `<CoreServices/CoreServices.h>`); SPEC
+  §6.4 specifically calls out that the backend `.cpp` file uses the C
+  API directly without falling into a `.mm`. This design preserves
+  that rule.
+
+The aggregate's SRP boundary is sharp: if the OS file-watch backend
+shifts (FSEvents → kqueue / inotify / `ReadDirectoryChangesW`), if
+the canonicalization rule changes (Apple ships a new normalisation
+form, an APFS case-folding option flips), or if the rename-reassembly
+heuristic mutates, this design changes. Anything else is out of scope.
+
+## 2. Requirements coverage
+
+Mapping of harmonius filesystem requirements
+(`harmonius/docs/requirements/platform/filesystem.md`, R-14.6.5,
+R-14.6.6, R-14.6.7, R-14.6.8, R-14.6.9) onto MVP coverage in this
+aggregate. Every entry is independently re-derived; coverage sites
+refer to sections of `specs/platform/SPEC.md` and to the design
+sections below.
+
+| Harmonius clause                                                                                  | Glibre disposition (MVP)                                                                                                                                                                                                                                                       |
+|---------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **R-14.6.5** monitor for create / modify / delete / rename via platform-native APIs; debounce; recursive | **Covered.** SPEC §4.3 inv #2 (debounce + dedup) + §4.3 inv #3 (atomic rename) + §6.4 (FSEvents-via-C-API on macOS). Recursive subscription per `kFSEventStreamCreateFlagWatchRoot` + `kFSEventStreamEventIdSinceNow`. §3.3, §3.4, §3.5 below.                                |
+| **R-14.6.6** content-hash dedup with BLAKE3                                                       | **Covered, scoped.** BLAKE3 is *internal* to the watcher's I/O thread; it never crosses the public surface. The hash collapses metadata-only events (`FSEventStreamEventFlagItemModified` without content change) within one debounce window per §4.3 inv #2. §3.6 below.       |
+| **R-14.6.7** canonical absolute paths; macOS case-folding                                         | **Covered, collapsed.** `CanonicalPath` (SPEC §4.3 inv #6) is the only key shape that crosses the watcher's boundary. Canonicalisation happens once on the I/O thread before dedup (§3.4 step 3); the public API never accepts a raw path. §3.7 below.                          |
+| **R-14.6.8** typed `FileEvent` with `FileEventKind` enum                                          | **Covered, collapsed.** `FileEvent` is a closed `eastl::variant<Created, Modified, Deleted, Renamed>` (SPEC §5.8). No `next() async` shape — glibre's frame-driven model uses non-blocking `take_events(token, span)`. §3.8 below.                                              |
+| **R-14.6.9** in-memory cache of `(canonical_path → BLAKE3 hash)`                                  | **Covered, scoped.** The LRU is internal to the I/O thread (SPEC §6.4 last paragraph) and lives in the watcher's 4 MiB sub-arena. It is never public; consumers cannot insert or query. §3.6 + §9.2 below.                                                                       |
+| **R-14.6.5** "configurable debounce interval"                                                     | **Partial — fixed default in MVP, knob deferred.** The debounce window is a build-time constant (default 100 ms, §3.6 below). Surfacing it as a runtime knob requires the editor / hot-reload coordinator spike — refused for MVP per Occam (no second consumer demands it). §12 holds the trigger. |
+| Tokio async-channel delivery / async fn next()                                                    | **Refused (collapsed).** Per SPEC §3 collapse "kqueue / inotify / ReadDirectoryChangesW / FSEvents → `FileWatcher`": no Tokio, no async streams. The frame-driven `take_events(WatchToken, span)` is a non-blocking SPSC drain; backpressure surfaces as `IoFailure` with prefix `"watcher-unavailable"` only on FSEvents stream death, not on consumer slowness (the SPSC is bounded; overflow is documented in §10 below). |
+| Network mount / sandbox sensitivity                                                               | **Covered as recovery.** SPEC §10.5 already enumerates volume-unmount, FSEvents service exhaustion, and TCC denial as `watcher-unavailable` triggers. This design discharges the aggregate side of that recovery (release stream, release token, no auto-re-arm). The consumer-side polling shim is refused per §1.                                       |
+| Watch cancellation                                                                                | **Covered.** `FileWatcher::unwatch(WatchToken)` (SPEC §5.8) tears down the FSEvents stream and the per-token SPSC; subsequent `take_events` against the stale token returns `NotFound` (SPEC §10.3.3 row 3).                                                                    |
+
+Glibre-native requirements added beyond harmonius:
+
+- **Per-token SPSC bounded ring**, never grows. Sized at construction
+  (default 4096 slots × 64 B = 256 KiB; sub-arena math in §9.2). The
+  writer is the I/O thread; the reader is the main thread. SPSC, not
+  MPSC, because the rename-reassembly step §3.5 requires the I/O
+  thread to be the sole writer per token.
+- **Watch token numeric identity is process-scoped, not reload-stable
+  across platform self-reload.** SPEC §8.6 already locks this: peer
+  plugin reload preserves `WatchToken`; platform-self-reload re-vends
+  fresh tokens and consumers reseat. This design honours the rule
+  without inventing reload-survivable token machinery.
+- **Outstanding subscription counter for §8.3 step 3.** The aggregate
+  exposes the in-memory subscription list as `eastl::span<const
+  CanonicalPath>` to the loader during platform self-reload drain. No
+  public getter; the loader gains access through a friend-restricted
+  internal hook. §3.10 below.
+- **Drain budget = 0 on the game-loop driver thread** (SPEC §9.1 row
+  3). The aggregate may not run any work in phases 1–9 except a
+  non-blocking SPSC dequeue inside `take_events`. The FSEvents
+  callback runs on the watcher's own runloop thread and never on the
+  main thread.
+
+Coverage rule: every harmonius clause above either lands in this
+design (with a coverage site) or is refused with a one-line rationale.
+No silent drops.
+
+## 3. Detailed model
+
+### 3.1 Aggregate composition
+
+```text
+FileWatcher  (aggregate root, owned by platform)
+├── std::thread              io_thread_              (one per FileWatcher; runs CFRunLoopRun)
+├── std::atomic<bool>        stop_requested_         (set by ~FileWatcher; observed by io thread)
+├── CFRunLoopRef             runloop_                (the io thread's CFRunLoop, opaque)
+├── FSEventStreamRef         stream_                 (one stream per FileWatcher; multi-root)
+├── eastl::vector<RootEntry> roots_                  (token → canonical root + flags + ring index)
+├── eastl::vector<TokenRing> rings_                  (per-token SPSC ring of FileEvent)
+├── DedupCache               dedup_                  (LRU keyed on CanonicalPath × content_hash)
+├── RenameReassembler        renames_                (inode-id → pending Created/Deleted pair)
+├── std::atomic<std::uint64_t> next_token_           (monotonic; never reused this process)
+└── PerContextAllocator&     arena_                  (the platform sub-arena, §9.2)
+```
+
+The `FileWatcher` is the aggregate root. The §5.8 stub is the only
+public surface; everything in the box above is private to the
+implementation file group `engine/platform/src/watcher/`. SPEC §6.4
+already names the file split (`watcher/backend.hpp`,
+`watcher/backends/sdl3_fsevents.cpp` — though this design renames the
+backend file to `watcher/backends/macos_fsevents.cpp` because the
+MVP path uses Core Services directly, not SDL3 filesystem events;
+SDL3 has no recursive-watch surface and the SPEC §6.4 paragraph
+acknowledges the direct-FSEvents fallback as the load-bearing path).
+
+`FileEvent` and `WatchToken` are value objects (SPEC §5.8). They cross
+the boundary by value through `take_events`'s output span; consumers
+treat them as PODs.
+
+### 3.2 `WatchToken` lifetime
+
+```cpp
+struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const WatchToken&) const noexcept = default; };
+```
+
+- Vended monotonically by `FileWatcher::watch` from `next_token_`.
+- Never reused inside one `FileWatcher` instance — even after
+  `unwatch`. This is the rule that lets a consumer hold a `WatchToken`
+  past its `unwatch`: subsequent `take_events(stale_token, …)` returns
+  `NotFound` (SPEC §10.3.3) rather than silently aliasing onto a fresh
+  subscription that happened to recycle the value.
+- Numeric identity does **not** survive a `FileWatcher` instance
+  swap. Platform self-reload re-vends a fresh `FileWatcher` (SPEC §8.6
+  bullet 3); peer-plugin reload passes through the same `FileWatcher`
+  and the same tokens (SPEC §8.1 row 3).
+- `WatchToken{0}` is the sentinel "invalid"; the aggregate never
+  vends `0`. The first valid token is `1`.
+
+### 3.3 `FileWatcher::create`
+
+Called once per consumer that needs an isolated subscription set;
+typical MVP usage is one `FileWatcher` per editor or per content
+plugin. The aggregate does *not* enforce singleton semantics — the
+sub-arena (§9.2) is sized for one instance, but the contract permits
+two if a consumer needs to isolate roots (e.g. user-content vs.
+engine-installed assets). Steps:
+
+1. **Allocate the per-instance state** from the platform `FileWatcher`
+   sub-arena (§9.2, 4 MiB cell). Failure (sub-arena exhausted) →
+   `IoFailure { OsCode { ENOBUFS } }` with prefix `"out-of-budget"`,
+   discharged by `perf-budget.md` Allocator Rule #2.
+2. **Spawn the I/O thread.** The thread function is
+   `watcher::detail::io_thread_main(FileWatcher::Impl*)`. Inside it:
+   a. Capture `CFRunLoopGetCurrent()` into `runloop_` (this thread's
+      runloop; not the main thread's).
+   b. Create an empty `FSEventStreamRef` with zero paths
+      initially — SPEC §6.4 hosts the runloop on this same thread.
+      The stream is reconstructed every time `watch` / `unwatch`
+      changes the root set (FSEvents requires recreate-on-change,
+      §3.4 step 5 below).
+   c. Enter `CFRunLoopRun()`. Returns when `CFRunLoopStop` is called
+      from the destructor (§3.9).
+   Failure (thread spawn / runloop init) → translate to `IoFailure`
+   per SPEC §10.2.1 (`EAGAIN` on `pthread_create` saturation maps to
+   `IoFailure { OsCode { errno } }` with prefix `"resource"`).
+3. **Wait for the runloop to be live** by polling `runloop_` with a
+   bounded spin (~32 iterations, ~64 µs total under M1 thermal
+   typical). If the runloop has not appeared inside the bound, the
+   thread is joined and `IoFailure { OsCode { 0 } }` with prefix
+   `"watcher-unavailable"` is returned.
+4. **Return `FileWatcher`** by move. The aggregate's pimpl is the only
+   heap reference the public `FileWatcher` holds.
+
+`create` does **not** subscribe any roots; the FSEvents stream is
+empty until the first `watch` call. This separation is what lets
+`create` succeed on a system where TCC has not yet been granted: the
+permission check happens at `watch` time, not at `create` time.
+
+### 3.4 `FileWatcher::watch(CanonicalPath root)` — subscribe a root
+
+Called on the main thread (no concurrent `watch` from the I/O thread;
+the contract is documented in `take_events`'s sibling rule SPEC
+§10.3.3 row 4 "off-main-thread call → `Unsupported`" — same rule
+applies to `watch` and `unwatch`). Steps:
+
+1. **Validate** the `root`. The constructor of `CanonicalPath` (SPEC
+   §4.3 inv #6) already enforces UTF-8 + absoluteness; this function
+   additionally enforces directory-ness via a single
+   `lstat`-equivalent call (`stat` resolves through the symlink which
+   we want — symlinked roots should follow). Failure: `Unsupported`
+   on file (non-directory), `NotFound` on missing path,
+   `PermissionDenied` on a readdir denial (the FSEvents stream-create
+   call later would fail anyway; pre-checking here lets us return the
+   precise arm).
+2. **Allocate a new `WatchToken`** (`next_token_.fetch_add(1, ...)`).
+3. **Allocate a per-token SPSC ring** in the sub-arena. Default size:
+   4096 entries × 64 B = 256 KiB. Failure → `IoFailure` with
+   `"out-of-budget"` prefix (§9.2). The ring's writer side is held by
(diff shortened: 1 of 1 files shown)


specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+   stream_ = FSEventStreamCreate(
+       /*allocator=*/ kCFAllocatorDefault,
+       /*callback=*/  &watcher::detail::fsevents_callback,
+       /*context=*/   &ctx,
+       /*paths=*/     paths,
+       /*sinceWhen=*/ kFSEventStreamEventIdSinceNow,
+       /*latency=*/   debounce_seconds_,                       // §3.6
+       /*flags=*/     kFSEventStreamCreateFlagFileEvents
+                    | kFSEventStreamCreateFlagWatchRoot
+                    | kFSEventStreamCreateFlagNoDefer);
```
cjhowe-us, line 289: severity:HIGH location:specs/platform/file-watcher-design.md:289 problem:`kFSEventStreamCreateFlagNoDefer` and `latency = debounce_seconds_` are contradictory by design intent, not just incidental. Per Apple CoreServices docs, `kFSEventStreamCreateFlagNoDefer` instructs the kernel to deliver the *first* event in a burst immediately, deferring subsequent events for the latency period. Combined with 100 ms `latency`, this means FSEvents fires as soon as any file changes, bypassing the coalescing window for the first event in every burst. The design claims the latency argument collapses "save → fsync → atomic-rename → unlink-old" three-event bursts into one `Modified` (§3.6 step 3 commentary) — but with NoDefer that collapse is done by the *software* debounce (§3.6 steps 3–4), not by the OS. The OS fires the watcher I/O thread for the first event without any wait, which means the BLAKE3 callback can run on each individual file-write event before the software debounce window opens.

Consequence: at 60 fps with `kFSEventStreamCreateFlagNoDefer` the I/O thread wakes immediately on each save, runs BLAKE3 on the first write, then the software window collapses the rest. This is *functionally correct* (the consumer sees one event) but burns off-thread CPU on intermediate hash work the OS would have suppressed for us. The `kFSEventStreamEventIdSinceNow` + latency-without-NoDefer combination is the documented Apple pattern for editor-style file watching; `kFSEventStreamCreateFlagNoDefer` is intended for tools that need to react to *first touch* as fast as possible (e.g. a CI watcher). Since the design purpose is editor live-reload (latency-tolerant, not first-touch-sensitive), the NoDefer flag should either (a) be removed and replaced with the plain `latency`-only form, or (b) retained with explicit rationale documenting the I/O-thread BLAKE3 cost trade-off. If retained, also note that the software debounce (§3.6 step 3) is the load-bearing coalesce mechanism regardless, and the 100 ms window already covers the three-event save burst. fix:Remove `kFSEventStreamCreateFlagNoDefer` from the flags mask to let FSEvents own the kernel-side coalescing, or add a paragraph in §3.4 step 5 / §3.6 that explicitly acknowledges the NoDefer + latency interaction and justifies the trade-off. The flags comment in the code block should reference the deliberate choice.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+  `EventQueue<WindowEvent>` (sibling #717). `FileEvent` is *not*
+  funnelled through `Pump::drain`; it has its own per-token drain
+  (SPEC §5.8 `take_events`). The collapse rule is in §3.5 below.
+- **`FileIo` blocking / async I/O** — sibling #721. The watcher does
+  no file *content* reads; the BLAKE3 dedup mentioned in SPEC §6.4
+  reads bytes from the OS only inside the I/O thread, never on the
+  main thread, and never through `FileIo`. The two aggregates share
+  the `CanonicalPath` value object and nothing else.
+- **`Clock` / time** — sibling #723. The debounce window timestamps
+  events with `Clock::now()`, but the `Clock` aggregate is taken by
```
cjhowe-us, line 71: severity:HIGH location:specs/platform/file-watcher-design.md:71 problem:`Clock` dependency is stated here as "taken by reference" but the §4 public surface declares `FileWatcher::create()` with no parameters. SPEC §4.4 inv #5 explicitly states "aggregates that need time take a `Clock&` parameter rather than constructing their own." The `Clock` singleton pattern (SPEC §4.4 inv #5 "One Clock per process") does not resolve this — the rule is that the aggregate receives the reference *at construction time*, not that it reaches for a global. The gap means the implementation PR will face an undocumented choice: either (a) add `Clock&` to `create()`, which changes the §5.8 locked stub, or (b) use a global/singleton Clock, which contradicts SPEC §4.4 inv #5. Either choice is a design deviation that should be locked here before the task-breakdown spike runs. fix:Amend the §3.3 `create()` steps to document how `Clock` is injected. If the design intends `Clock` to be accessible as a process-scoped singleton (consistent with "One `Clock` per process"), state that explicitly and cite SPEC §4.4 inv #5 as the justification for why a `Clock&` parameter is unnecessary. If it requires explicit injection, the §4 public stub must be updated to add `Clock&` to `create()`, which would be a SPEC §5.8 amendment. One or the other — the current text leaves an implementation-time ambiguity that will block the task-breakdown spike.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+   `watcher::detail::io_thread_main(FileWatcher::Impl*)`. Inside it:
+   a. Capture `CFRunLoopGetCurrent()` into `runloop_` (this thread's
+      runloop; not the main thread's).
+   b. Create an empty `FSEventStreamRef` with zero paths
+      initially — SPEC §6.4 hosts the runloop on this same thread.
+      The stream is reconstructed every time `watch` / `unwatch`
+      changes the root set (FSEvents requires recreate-on-change,
+      §3.4 step 5 below).
+   c. Enter `CFRunLoopRun()`. Returns when `CFRunLoopStop` is called
+      from the destructor (§3.9).
```
cjhowe-us, line 220: severity:MED location:specs/platform/file-watcher-design.md:220 problem:Step 2b describes creating "an empty FSEventStreamRef with zero paths initially" during I/O thread startup. This is incorrect per the FSEvents C API. FSEventStreamCreate does not accept a zero-length paths array — the pathsToWatch CFArray must be non-empty; passing an empty CFArrayRef causes FSEventStreamCreate to return NULL on macOS 12+, and a subsequent FSEventStreamScheduleWithRunLoop(nullptr, ...) is a crash. The correct pattern for "no paths yet" is to not create the stream at all until the first watch() call, and only create it when roots_ is non-empty. The design's intent is correct — defer stream creation to the first watch call — but the step 2b description contradicts it and would lead the implementation astray. fix:Replace §3.3 step 2b with "stream_ is null during I/O thread startup; no FSEventStreamRef is created. The I/O thread enters CFRunLoopRun() with an empty runloop and sits idle until the first CFRunLoopPerformBlock arrives from watch(). The first watch call (§3.4 step 5) creates the stream with the initial root set." This matches the already-correct §3.4 and §3.5 recreate logic.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+   not deliver the pair.
+5. **SPSC enqueue**. The resulting `FileEvent` is pushed into the
+   per-token ring (`rings_[root_entry.ring_index]`). The push is
+   non-blocking; if the ring is full (consumer fell behind by more
+   than 4096 events — see §10), the watcher drops *the oldest event*
+   to make room and increments a per-token overflow counter. SPEC
+   §4.3 does not currently mandate "never drop" for `FileEvent`
+   (unlike SPEC §4.2 inv #3 for `WindowEvent`), and dropping oldest
+   matches the "live-reload is best-effort under storm" intent of
+   the consumer (the next `Created` event will resurface the file).
```
cjhowe-us, line 396: severity:MED location:specs/platform/file-watcher-design.md:396 problem:SPSC ring overflow drops the oldest event (step 5), but the design cites this as a divergence from SPEC §4.2 inv #3 (EventQueue never drops). The comparison is sound, but the design leaves open whether the overflow counter is itself bounded. If the per-token overflow counter is a uint32_t and events drop continuously for minutes (e.g. a runaway git checkout that produces 10k events/sec sustained), the counter overflows and wraps to zero — the 'next take_events call detects a non-zero counter and logs' invariant silently breaks. The design §3.6 step 5 says 'increments a per-token overflow counter' without specifying the counter's type or saturation behavior. fix:Specify the counter type (recommend uint64_t to make silent wrap a non-issue in practice, or uint32_t with saturating-add semantics — document which). Add a sentence: 'The counter uses saturating arithmetic; once it reaches its max value it stays there until the next take_events call resets it to zero after logging.' This closes the otherwise-open question of what happens to telemetry accuracy under extreme prolonged churn.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+
+### 9.2 Heap / sub-arena
+
+SPEC §9.2 row 3 locks the watcher at **4 MiB** in the platform sub-
+arena. This design partitions:
+
+| Component                                | Bytes (approx) | Notes                                                                                                |
+|------------------------------------------|---------------:|------------------------------------------------------------------------------------------------------|
+| Per-instance pimpl (`FileWatcher::Impl`) | 4 KiB          | Vectors, atomics, the `RootEntry` table.                                                             |
+| `roots_` capacity                        | 8 KiB          | 16 `RootEntry` × ~512 B each (canonical root path + flags + ring index + small per-root metadata).   |
```
cjhowe-us, line 897: severity:MED location:specs/platform/file-watcher-design.md:897 problem:The sub-arena table says 16 tokens x 256 KiB = 4 MiB for SPSC rings (dominant). But the table also shows additional components (pimpl 4 KiB, roots 8 KiB, interner 256 KiB, dedup LRU 64 KiB, rename buffer 16 KiB) that add roughly another 350 KiB. 16 x 256 KiB = 4096 KiB exactly, meaning the 4 MiB cell is 100% consumed by rings alone, leaving zero bytes for the other components. The design says 'The arithmetic does not exactly sum to 4 MiB; the SPSC rings dominate and the rest is comfortably absorbed within the cell' — but it does NOT fit: 16 rings x 256 KiB = 4 MiB, then interner 256 KiB already overflows. Either the ring count must be reduced (e.g. 15 rings to leave room), or the cell must be 4 MiB + overhead (approximately 4.35 MiB), or the 4 MiB number must be treated as a cap with dynamic sizing. The current claim is mathematically inconsistent. fix:Either (a) set ring count to 15 (4 MiB - 350 KiB overhead = 3.65 MiB / 256 KiB = 14.3, so 14 rings max within the cell), or (b) state the 4 MiB cell covers rings only and the interner/dedup/rename-buffer live in a separate 512 KiB overhead cell, or (c) reduce ring depth from 4096 to 4000 slots to leave headroom within 4 MiB. Whichever is chosen, make the arithmetic explicit so the implementation PR can verify it.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+- **Bidirectional shutdown signal.** `stop_requested_` is a
+  `std::atomic<bool>` set by the destructor on the main thread and
+  read by the I/O thread on every callback entry; combined with
+  `CFRunLoopStop` (which wakes the runloop synchronously), this
+  guarantees a bounded shutdown window (§3.9 step 4).
+
+### 6.3 Synchronisation primitives
+
+- **No mutex on the hot path.** SPSC rings synchronise via
+  `std::atomic_ref<std::uint32_t>` head/tail counters with
```
cjhowe-us, line 700: severity:MED location:specs/platform/file-watcher-design.md:700 problem:Section 6.3 says a mutex on roots_ is held during per-root prefix matching in the FSEvents callback (step 2 of section 3.6), then immediately says 'an LRU cache of the last-prefix-match keeps the per-event lookup amortized lock-free in the steady state.' This is two conflicting statements about the locking model. If the callback takes the mutex for every event on a cache miss, and the main thread holds the mutex during watch/unwatch, then a long watch() call (with FSEvents stream recreate taking 0.5-2 ms) can block the I/O thread callback for the duration. That violates the off-thread non-blocking claim from section 5. If the LRU cache truly makes the callback lock-free in the steady state, the design must specify: (a) when exactly the mutex is taken vs. not taken in the callback, (b) what the LRU cache key is and how it invalidates when watch/unwatch mutates roots_, (c) whether a cache miss falls back to a locked lookup or a snapshot copy. fix:Add a paragraph to section 6.3 that explicitly states the concurrency contract for the per-root lookup inside the callback: 'The callback reads a per-thread shadow copy of roots_ that is refreshed (under the mutex) only when a recreate notification arrives from watch/unwatch. The callback reads the shadow copy lock-free; the main thread writes it under the mutex as part of the recreate dispatch. This eliminates the hot-path lock contention at the cost of a bounded staleness window (the shadow copy may lag the current roots_ by one recreate cycle, which is acceptable because the FSEvents stream has already been stop/started by then).' The current text is ambiguous enough to cause an implementation bug.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+   Failure (thread spawn / runloop init) → translate to `IoFailure`
+   per SPEC §10.2.1 (`EAGAIN` on `pthread_create` saturation maps to
+   `IoFailure { OsCode { errno } }` with prefix `"resource"`).
+3. **Wait for the runloop to be live** by polling `runloop_` with a
+   bounded spin (~32 iterations, ~64 µs total under M1 thermal
+   typical). If the runloop has not appeared inside the bound, the
+   thread is joined and `IoFailure { OsCode { 0 } }` with prefix
+   `"watcher-unavailable"` is returned.
+4. **Return `FileWatcher`** by move. The aggregate's pimpl is the only
+   heap reference the public `FileWatcher` holds.
```
cjhowe-us, line 230: severity:LOW location:specs/platform/file-watcher-design.md:230 problem:Section 3.3 step 3 describes a 'bounded spin (~32 iterations, ~64 us total under M1 thermal typical)' to wait for the I/O thread to expose its runloop pointer. The spin-to-expose pattern is fragile under load: if the I/O thread is preempted between spawn and CFRunLoopGetCurrent(), the spin exits before runloop_ is valid, and the fallback joins the thread and returns watcher-unavailable. This is documented as a timeout but the 32-iteration / 64 us bound is derived from an M1 thermal-nominal measurement, not from a memory ordering guarantee. On a loaded machine (CI with many parallel processes), the I/O thread may not have been scheduled in 64 us, leading to spurious watcher-unavailable failures in tests. fix:Use a semaphore or std::atomic with a release store from the I/O thread after CFRunLoopGetCurrent() instead of a polling loop. The I/O thread does: (1) runloop_ = CFRunLoopGetCurrent() (release store), (2) ready_sem.release(). The main thread does: ready_sem.acquire() with a 5 ms timeout, then checks runloop_ != nullptr. This eliminates the spurious timeout window and avoids busy-polling. A std::binary_semaphore with try_acquire_for is the C++20 standard idiom.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+  with `"out-of-budget"` prefix per `perf-budget.md` Allocator Rule
+  #2. Caller backs off (cancel an unused subscription) or the
+  operator increases the platform budget (out of MVP scope).
+
+## 11. Test plan
+
+### 11.1 Unit tests (Catch2; under `tests/platform/watcher/`)
+
+Each test name follows `glibre/platform/watcher: <invariant>` so it
+maps directly to SPEC §11 acceptance criteria #356 / #357.
```
cjhowe-us, line 1036: severity:LOW location:specs/platform/file-watcher-design.md:1036 problem:The test plan in section 11 references user stories #356 and #357 from SPEC §11 directly, which is good. However, test #15 ('watch_returns_permission_denied_on_sandbox_blocked_root') says 'test fixture installs a sandbox profile that denies a known path.' macOS App Sandbox profiles are entitlement-based and cannot be programmatically installed at test time by a regular process without code-signing the test binary with the sandbox entitlement. A unit test that depends on a live sandbox profile will not run in CI without additional infrastructure (a specially signed test harness or a mock at the FileIo admission layer). The design does not address how the sandbox fixture works in the context of a standard Catch2 build. fix:Either (a) replace the sandbox-profile approach with a mock CanonicalPath that the watcher's internal stat-check helper rejects (a test seam at the permission-check step rather than the OS level), or (b) annotate the test as requiring a sandbox-signed harness and document the CI infrastructure needed, or (c) replace with a chroot/unveil-equivalent approach that can be set up without code-signing. This is a test plan coverage concern, not a design correctness issue, but it blocks the task-breakdown spike from writing a valid Catch2 test.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:4
  LOW:2
carryover:0

notes:The design is substantively complete for a spike deliverable: SRP is well-observed, the §5.8 public stub is correctly preserved, no new platform::Error arms are invented, all harmonius clauses are covered or explicitly refused, EASTL types are used correctly throughout the internal composition (std:: retained only for thread, atomic, mutex, expected per PHILOSOPHY §11), and the test plan has good Catch2 + integration coverage that correctly references SPEC §11 user stories #356 and #357. Two HIGH findings block the task-breakdown spike. The first is a Clock injection gap: §1 states the Clock aggregate is "taken by reference" but create() takes no parameters, leaving an undocumented implementation choice that either widens the §5.8 stub or contradicts SPEC §4.4 inv #5 -- this must be resolved before the sibling spike runs. The second is a kFSEventStreamCreateFlagNoDefer + latency contradiction: the flag bypasses kernel-side coalescing for the first event in every burst, making the software debounce (not the OS latency) the load-bearing collapse mechanism, with un-accounted I/O-thread BLAKE3 cost; the design must either remove the flag or explicitly document the trade-off. The four MED findings are: (1) §3.3 step 2b describes creating an empty FSEventStreamRef with zero paths, which crashes on macOS 12+ (FSEventStreamCreate rejects an empty CFArrayRef); (2) §6.3 has a contradictory mutex vs. lock-free claim for the callback's per-root lookup that will cause an implementation bug; (3) the §9.2 sub-arena arithmetic overflows -- 16 rings x 256 KiB = 4 MiB exactly leaves zero room for the interner, dedup LRU, and pimpl; (4) the SPSC overflow counter has no specified type or saturation behavior. Two LOW findings cover a fragile bounded-spin startup pattern (prefer std::binary_semaphore) and a sandbox fixture that cannot be set up in standard CI without code-signed test binaries. No issues were found with error model usage (platform::Error arms are used canonically, prefix literals reference the §3.3 registry correctly), SPSC ring design, rename-reassembly heuristic, hot-reload protocol compliance, or ABI surface correctness.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

Clock& added to create() in §1 prose, §3.3 step description, §4 public stub, §5 cold-path list, §10.1 table, and §12 [BLOCKING IMPLEMENTATION] entry per resolution path (a).

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

kFSEventStreamCreateFlagNoDefer removed from §3.4 step 5 FSEventStreamCreate snippet per resolution option (a). Added prose: 100 ms latency owns coalescing of save→fsync→atomic-rename→unlink-old bursts; BLAKE3 fires only on post-coalesce event, not per intermediate.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

§3.3 step 2b rewritten: stream_ starts null (deferred creation); first watch() call creates the stream once at least one root is registered; no empty CFArrayRef ever passed to FSEventStreamCreate.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

§3.6 step 5 overflow counter specified as std::uint64_t; rationale documented (u32 wraps in ~12 days at 4096 drops/s, silently breaking the non-zero log invariant; u64 needs no saturation); first-non-zero log fires once per token-lifetime, subsequent increments accumulate without log spam.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

§9.2 ring count reduced 16→14: 14 × 256 KiB = 3584 KiB = 3.5 MiB. Explicit arithmetic: 4 + 8 + 3584 + 256 + 64 + 16 = 3932 KiB ≈ 3.84 MiB, leaving ~256 KiB headroom within the 4 MiB cell. Test 16 root count updated. §12 [NON-BLOCKING] headroom note added.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

§6.3 mutex bullet rewritten with explicit shadow-copy protocol: callback reads lock-free atomic<roots_snapshot_t*> (never takes mutex); main thread takes mutex, builds new snapshot, atomic_store with memory_order_release, retires old via one-frame quarantine; FSEvents recreate dispatched via CFRunLoopPerformBlock/WakeUp off hot path; worst-case callback blocking zero.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

§3.3 step 3 bounded spin replaced with std::binary_semaphore: main thread allocates sem{0} before spawning I/O thread; I/O thread stores CFRunLoopGetCurrent() into runloop_ then sem.release(); main thread calls sem.try_acquire_for(5ms); on timeout, thread joined and watcher-unavailable returned. 5 ms bound safe on loaded CI hosts.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8d71258c87a4079d8d6500e81748c816630a24e5 followup:#899

Test 15 rewritten per option (a): sandbox profile fixture replaced with dependency-injected fileio_ops_t mock returning EPERM; asserts EPERM routes through error translator to platform::Error::PermissionDenied per platform-error-design.md §3.3; runs in standard unsigned Catch2 CI without entitlement-based code signing.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:2
  LOW:2
carryover:0

notes:All 8 round-1 findings were addressed cleanly in the #899 followup diff: `kFSEventStreamCreateFlagNoDefer` removed with correct rationale (OS-side coalescing restored), empty-CFArrayRef crash eliminated by deferred stream creation, `std::binary_semaphore` replaces the bounded spin, shadow-copy protocol added to §6.3, u64 overflow counter documented with wrap-around rationale, 14-ring arithmetic is correct (4+8+3584+256+64+16=3932 KiB, 256 KiB headroom), and test 15 sandbox fixture replaced with `fileio_ops_t` DI mock. The `Clock&` injection is now consistently present in §1 prose, §4 stub, §5 cold-path, §10.1 table, §12 BLOCKING entry. Round-2 lens (cohesion + SRP + seam quality) surfaces one HIGH and four lower findings. HIGH: the `roots_snapshot_t` type introduced by the §6.3 shadow-copy fix has no definition, no place in the §3.1 composition diagram, no documented sub-arena size, and no specified read memory order for the callback-side load — leaving the protocol partially implementable from the design alone; the load ordering omission is a real correctness gap (implementer may default to `memory_order_relaxed`). MED-1: the §4 surface-invariant-2 "closed header list" does not include `<glibre/platform/clock.hpp>` even though `Clock&` is now a parameter of `create()` — this will cause a compile error on the first consumer. MED-2: `fileio_ops_t` referenced in test 15 is undefined in this design and absent from file-io-design.md; the SRP implication (is this FileIo's type or a watcher-local one?) is unresolved. LOW-1: §3.3 section heading still reads `FileWatcher::create` without `(Clock&)`. LOW-2: the §4 ABI symbol example still shows the no-argument mangled name. The design is otherwise substantively clean for round-2: SRP is sharp and well-defended, error model usage is canonical (no new arms invented, prefix literals reference the §3.3 registry correctly), EASTL discipline is correct (`eastl::vector`, `eastl::span`, `eastl::variant` in internal composition; `std::` retained only for `thread`, `mutex`, `atomic`, `expected` per PHILOSOPHY §11), hot/cold split enforces §9.1 0-ms budget, hot-reload protocol for peer-plugin and self-reload paths is cohesive and correct, and the BLOCKING §12 clock-stub amendment is clearly tagged. All fixes for the new findings are in-scope for the #899 branch.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:2
  LOW:2
carryover:0

notes:All 8 round-1 findings were addressed cleanly in the #899 followup diff: kFSEventStreamCreateFlagNoDefer removed with correct rationale (OS-side coalescing restored), empty-CFArrayRef crash eliminated by deferred stream creation, std::binary_semaphore replaces the bounded spin, shadow-copy protocol added to §6.3, u64 overflow counter documented, 14-ring arithmetic is correct (3932 KiB, 256 KiB headroom verified), and test 15 sandbox fixture replaced with fileio_ops_t DI mock. Round-2 lens surfaces one HIGH and four lower findings. HIGH (comment 3189340145): the roots_snapshot_t type introduced by the §6.3 shadow-copy fix has no definition, no place in the §3.1 composition diagram, no documented sub-arena size, and no specified read memory ordering for the callback-side load — leaving the protocol partially implementable and the load ordering a real correctness gap (implementer may default to memory_order_relaxed). MED-1 (comment 3189341380): the §4 surface-invariant-2 closed header list omits <glibre/platform/clock.hpp> even though Clock& is now a parameter of create() — will cause a compile error on the first consumer. MED-2 (comment 3189342865): fileio_ops_t referenced in test 15 is undefined in this design and absent from file-io-design.md; the SRP implication is unresolved. LOW-1 (comment 3189345337): §3.3 heading still reads FileWatcher::create without (Clock&). LOW-2 (comment 3189347039): §4 ABI symbol example still shows the no-argument mangled name. The design is otherwise clean for round-2: SRP is sharp and well-defended, error model usage is canonical, EASTL discipline is correct (eastl::vector/span/variant in internal composition, std:: retained only for thread/mutex/atomic/expected per PHILOSOPHY §11), hot/cold split enforces §9.1 0-ms budget, and the BLOCKING §12 clock-stub amendment is clearly tagged. All fixes for the new findings are in scope for the #899 branch.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:2
  LOW:2
carryover:0

notes:All 8 round-1 findings were addressed cleanly in the #899 followup diff: kFSEventStreamCreateFlagNoDefer removed (OS-side coalescing restored), empty-CFArrayRef crash eliminated by deferred stream creation, std::binary_semaphore replaces the bounded spin, shadow-copy protocol added to §6.3, u64 overflow counter documented with wrap-around rationale, 14-ring arithmetic correct (3932 KiB, 256 KiB headroom verified: 4+8+3584+256+64+16=3932), test 15 sandbox fixture replaced with fileio_ops_t DI mock. Round-2 lens (cohesion + SRP + seam quality) surfaces one HIGH and four lower findings. HIGH (PR #899 comment 3189340145): roots_snapshot_t introduced by the §6.3 fix is undefined — no definition, no §3.1 composition-diagram entry, no sub-arena size, no callback-side load memory order specified; the read ordering gap is a real correctness hazard (implementer may default to memory_order_relaxed). MED-1 (comment 3189341380): §4 closed header list omits <glibre/platform/clock.hpp> despite Clock& now being a parameter. MED-2 (comment 3189342865): fileio_ops_t in test 15 is undefined and absent from file-io-design.md; SRP implication unresolved. LOW-1 (comment 3189345337): §3.3 heading still reads FileWatcher::create without (Clock&). LOW-2 (comment 3189347039): §4 ABI symbol example shows the no-argument mangled name. Positives: SRP boundary is tight and consistently defended, error model usage is canonical, EASTL discipline is correct throughout internal composition, hot/cold split enforces the 0-ms §9.1 budget, hot-reload protocol for both peer-plugin and self-reload paths is cohesive and correct, and the BLOCKING §12 Clock stub amendment is properly tagged. All fixes are in scope for the open #899 branch.

cjhowe-us:
responder:go-impl-respond round:2 decision:ADDRESSED commit:9787c5d87e9b68f01dbd699e4a643ef68a4b426b followup:#899

All 5 round-2 findings addressed in commit 9787c5d on branch `fix/platform-file-watcher-design-followup-r1` (PR #899).

**HIGH-1 (§6.3 — roots_snapshot_t undefined)**
ADDRESSED. Four sub-items resolved:
1. `std::atomic<roots_snapshot_t*> roots_snapshot_` added to §3.1 composition diagram (`specs/platform/file-watcher-design.md` line ~162) with annotation that it is written with `memory_order_release` and read with `memory_order_acquire` in the FSEvents callback.
2. New §3.x section added before §3.3 defining `roots_snapshot_t` as a POD struct `eastl::array<root_entry_t, 14> entries; uint8_t count;`, arena-backed from `snapshot_pool_` (3 entries × ~512 B = ~1.5 KiB). Specifies 1 active + 2 quarantined for 1-frame retire window.
3. `snapshot_pool_` row added to §9.2 table (~2 KiB; absorbed within existing 256 KiB headroom; arithmetic updated to 3934 KiB).
4. §6.3 expanded to show the explicit `roots_snapshot_.load(std::memory_order_acquire)` in the callback and `roots_snapshot_.store(new_snap, std::memory_order_release)` on the write side, with happens-before rationale and AArch64 `stlr`/`ldar` note.

**MED-1 (§4 surface invariant 2 closed header list)**
ADDRESSED. `<glibre/platform/clock.hpp>` added to the §4 invariant-2 closed header list and to the `#include` block in the public header code snippet. Annotation added: substitute `<glibre/platform/clock_fwd.hpp>` if clock-design §4 uses a forward-decl header.

**MED-2 (§11 test 15 fileio_ops_t SRP unresolved)**
ADDRESSED. §11 test 15 prose rewritten: `fileio_ops_t` is now explicitly defined as a test-local seam type in `tests/platform/file-watcher/test_helpers.hpp`. Introduces `FileWatcher::create_for_test(Clock&, fileio_ops_t&)` factory gated by `#ifdef GLIBRE_TESTING_HOOKS`, never present in shipping builds. Ownership clarified: `FileIo` owns production fs ops; tests own their own mock interface. SRP resolved.

**LOW-1 (§3.3 section heading)**
ADDRESSED. `specs/platform/file-watcher-design.md` §3.3 heading updated from `FileWatcher::create` to `FileWatcher::create(Clock& clock)`.

**LOW-2 (§4 ABI shape paragraph mangled symbol)**
ADDRESSED. Mangled symbol updated from `_ZN6glibre8platform11FileWatcher6createEv` (no-argument) to `_ZN6glibre8platform11FileWatcher6createERNS0_5ClockE` (Itanium ABI encoding of `glibre::platform::FileWatcher::create(glibre::platform::Clock&)`). Note added to verify with `c++filt` at codegen time and update if `Clock` resolves to a different namespace.

specs/platform/file-watcher-design.md
```diff
@@ -0,0 +1,1199 @@
...
+  no file *content* reads; the BLAKE3 dedup mentioned in SPEC §6.4
+  reads bytes from the OS only inside the I/O thread, never on the
+  main thread, and never through `FileIo`. The two aggregates share
+  the `CanonicalPath` value object and nothing else.
+- **`Clock` / time** — sibling #723. The debounce window timestamps
+  events with `Clock::now()`, but the `Clock` aggregate is taken by
+  reference; the watcher does not own a time source.
+- **Persistence of the subscription list across process restart** —
+  refused per SPEC §7.3 ("File-watcher canonical-path subscription
+  list (cache)"). Subscriptions are reseated by the consumer on every
```
cjhowe-us, line 75: severity:MED location:specs/platform/SPEC.md:920 problem:Round-3 topology check finds that SPEC §5.8 at line 920 still declares `create()` with no parameters, and SPEC §10.3.3 at line 2498 still lists `FileWatcher::create` without `Clock&`. The file-watcher-design.md (on the #899 branch) correctly uses `create(Clock&)` throughout and the §12 `[BLOCKING IMPLEMENTATION]` entry notes the SPEC stub amendment must land before the first plan PR. However, the §12 entry does not reference a filed follow-up spike issue to enforce this gate. Without a concrete issue, the BLOCKING status is nominal — there is no project-tracker artifact to block the task-breakdown spike or any future plan PR on. The amendment itself is a one-line change to the SPEC §5.8 stub and a matching row update in §10.3.3. recommendation:File a new `type:spike` issue for the SPEC §5.8 / §10.3.3 amendment, linking it from the §12 BLOCKING entry. The spike deliverable is: (1) change `create()` → `create(Clock&)` in SPEC §5.8 line 920, (2) change `FileWatcher::create` → `FileWatcher::create(Clock&)` in SPEC §10.3.3 line 2498, (3) add `<glibre/platform/clock.hpp>` to SPEC §5.8 stub includes if clock-design §4 confirms that header name. This spike is a prerequisite for the task-breakdown spike referenced by #719.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:4
  LOW:1
carryover:0

notes:All 5 round-2 findings (1H 2M 2L) were addressed in commit 9787c5d on the #899 branch: roots_snapshot_t is now fully defined with a POD struct snippet, placed in the §3.1 composition tree as an atomic pointer, memory_order_acquire specified on the read side, snapshot_pool_ budget row added to §9.2, clock.hpp added to the §4 closed header list, fileio_ops_t scoped to test_helpers.hpp and injected via create_for_test gated on GLIBRE_TESTING_HOOKS, §3.3 heading updated to create(Clock& clock), and ABI mangled symbol updated to the Clock& form. Round-3 topology lens surfaces four MED and one LOW finding, none blocking for the current spike deliverable. MED-1: the new roots_snapshot_t section uses the placeholder heading '3.x' rather than a concrete section number; all cross-references throughout the document say '§3.x' which is opaque to navigation and breaks section-TOC resolution — the section should be renumbered 3.3 (shifting 3.3–3.10 to 3.4–3.11). MED-2: the roots_snapshot_t struct snippet uses 'root_entry_t' (snake_case) while the §3.1 tree and all prose references use 'RootEntry' (CamelCase); these must name the same type or an implementer cannot know which convention the header will adopt. MED-3: snapshot_pool_ (the arena-backed allocator for snapshot instances) is referenced in §3.x, §6.3, and §9.2 but is absent from the §3.1 composition tree — the diagram is incomplete for implementers who read §3.1 as the canonical field list for FileWatcher::Impl. MED-4: SPEC §5.8 (line 920) and SPEC §10.3.3 (line 2498) still declare 'FileWatcher::create()' without Clock& — the design's §12 [BLOCKING IMPLEMENTATION] entry correctly tracks this but does not reference a filed follow-on spike issue to enforce the gate against the task-breakdown spike; without a concrete issue in the tracker, the BLOCKING designation is nominal. LOW-1: the AArch64 stlr/ldar note says 'plain store + ISH barrier + plain load' which is imprecise — stlr and ldar are single instructions, not two-instruction sequences; the terminology should be corrected to avoid confusing implementers familiar with ARM documentation. The overall design quality across all three rounds is strong: SRP boundary is tight and consistently defended, error model is canonical (no new arms, prefix registry correct), EASTL discipline is correct throughout, shadow-copy protocol for roots_snapshot_t is sound (release/acquire pair is correct, quarantine window logic is valid), hot/cold split enforces §9.1 0-ms budget, hot-reload protocol for both peer-plugin and self-reload paths is cohesive, arithmetic in §9.2 is correct (3934 KiB leaves ~254 KiB headroom). The four MED findings are all in scope for the #899 branch before merge.

cjhowe-us:
responder:go-impl-respond round:3 decision:ADDRESSED commit:20dcd3c followup:#899

All five round-3 findings addressed in commit 20dcd3c on branch `fix/platform-file-watcher-design-followup-r1` (PR #899). File changed: `specs/platform/file-watcher-design.md`.

**MED-1** (§3.x placeholder section number): Renamed `### 3.x` heading to `### 3.3`. Shifted all existing §3.3–§3.10 headings to §3.4–§3.11. Updated every cross-reference in the document — §2 coverage table (lines ~113–118), §3.1 composition comment (line ~165), body text inside §3.4, §3.5, §3.6, §3.7, §3.8, §6.1, §6.2, §6.3, §8.1, §8.2, §9.2, §9.4, §10.2, §11.1 tests 19–20, §12 open items. The `platform-error-design.md §3.3` citation on line ~1234 was intentionally left unchanged (it references a different file).

**MED-2** (RootEntry vs root_entry_t naming): Substituted `root_entry_t` → `RootEntry` using `replace_all`; verified zero remaining occurrences. Two substitution sites: the `eastl::array<RootEntry, 14>` field in the §3.3 code block (~line 216) and the §9.2 table annotation (~line 1027).

**MED-3** (snapshot_pool_ absent from §3.1): Added `SnapshotPool snapshot_pool_` to the §3.1 composition tree immediately below `std::atomic<roots_snapshot_t*> roots_snapshot_`, with annotation "arena-backed pool: 1 active + 2 quarantined snapshots, ~1.5 KiB total per §9.2" (~lines 166–167).

**MED-4** (§12 SPEC §5.8 + §10.3.3 spike track): Rewrote the `[BLOCKING IMPLEMENTATION]` entry in §12 to: name the spike `[SPIKE] amend-platform-spec-filewatcher-clock-injection` parented to #714; list both SPEC §5.8 and §10.3.3 as amendment targets; name both concrete consumers; add the issue-number backfill instruction; state the must-close-before-plan-PR gate. SPEC.md itself was not edited (design preamble forbids in-place SPEC edits).

**LOW-1** (stlr/ldar description): Replaced "plain store + ISH barrier + plain load (the compiler emits `stlr` / `ldar`…)" with "compiles to `stlr` (store-release) and `ldar` (load-acquire) — single instructions each, zero-overhead on the hot callback path; no separate ISH dmb required." (~lines 263–266).

Issue #723, opened 2026-05-03
[SPIKE] design-platform-clock-detailed


cjhowe-us:
Parent: #714

## Question

Produce the detailed design for the **clock** aggregate in the `platform` context.

mach_absolute_time wrapper.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/platform/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/platform/threading-async.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **clock** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/platform/clock-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(platform): detailed design — clock (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-platform-clock-detailed` is blocked by this issue.

cjhowe-us:
agent:design-platform-clock
status:done
issue:#723
branch:docs/platform-clock-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a2bbbd16c12a693c0
host:mac
cloud:none
commit:2afe360513111cf527f65357ac408e0137b05fd1
pr:#828
[...]

Issue #702, opened 2026-05-03
[SPIKE] design-core-schedule-frame-detailed


cjhowe-us:
Parent: #699

## Question

Produce the detailed design for the **schedule-frame** aggregate in the `core` context.

Schedule DAG + 9-phase FrameLoop.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/core/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/core-runtime/game-loop.md
- harmonius/docs/design/core-runtime/scheduling.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **schedule-frame** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/core/schedule-frame-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(core): detailed design — schedule-frame (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-core-schedule-frame-detailed` is blocked by this issue.

cjhowe-us:
agent:design-core-schedule-frame
status:started
issue:#702
branch:docs/core-schedule-frame-design
worktree:/Users/cjhowe/Code/glibre
host:mac
cloud:none
commit:none
pr:none
notes:Claiming spike #702 to author specs/core/schedule-frame-design.md. Required reads (PHILOSOPHY, AGENTS, SETUP, _TEMPLATE, core/SPEC.md, frame-phases.md, error-model.md, perf-budget.md, hot-reload-protocol.md, plugin-abi.md, fory-codegen.md, harmonius core-runtime/game-loop.md req+design) ingested. Re-deriving all conclusions; harmonius treated as research input only. Branch created from main.

cjhowe-us:
agent:design-core-schedule-frame
status:done
issue:#702
branch:docs/core-schedule-frame-design
[...]

Issue #358, opened 2026-05-03
[STORY] platform/clock: monotonic non-decreasing + wall correlation


cjhowe-us:
Parent sub-epic: #36

### Domain
platform

### Phase
phase:mvp

### Persona
engine developer (P-26)

### User Story
As an engine developer, I want a `Clock` aggregate exposing `Instant` (monotonic, never-decreasing, never-wrapping, NTP-immune) and `WallTime` (calendar source, may slew) with the only allowed time arithmetic being `Duration = Instant - Instant`, so that frame budgeting never sees a backward jump and log timestamps come from a single canonical `WallTime` source (§4.4 inv #1, #2, #3).

### Acceptance Criteria (Gherkin)
```gherkin
Given the process-singleton `Clock`
When I sample `Instant a = Clock::now()` and later `Instant b = Clock::now()`
Then `b - a >= Duration::zero()` always
And the platform aborts on a detected regression (§4.4 inv #1) — never silently clamps

Given the same `Clock`
When I sample `WallTime w = Clock::wall()`
Then `w` is the only path engine code uses for log timestamps (§4.4 inv #3)
And no engine TU calls `time(nullptr)` / `gettimeofday` directly (enforced by a CI grep gate)

Given two `Instant` values from different processes / runs
When I attempt to compare or serialize them
Then the type prevents it at compile time (§4.4 inv #2 — `Instant` is not serializable)

Given `Clock::native_resolution()`
When the platform reports it on macOS / M1
Then the value is at least 1 ms (§4.4 inv #4) and the deterministic simulation step uses it without integer drift
```

### Manual Test Script
1. Build `platform_clock_smoke`. Run.
2. Sample `Clock::now()` in a tight loop for 10 s; assert all samples non-decreasing.
3. Print `Clock::wall()` and compare against `date -u`; values should agree within 1 s.
4. Print `Clock::native_resolution()`; expect a value ≤ 1 ms on macOS / M1 (`mach_timebase_info`).
5. Try compiling a TU that calls `Clock::now() + Clock::now()`; assert compilation fails (only `Instant - Instant → Duration` is allowed).

### E2E Test Plan
`tests/e2e/platform/clock-monotonic.glibre-trace` — runs the engine for 600 frames, samples `Instant` at phase 1 and phase 9 of every frame, asserts strictly non-decreasing across all 1200 samples and that no TU outside `platform/` calls a non-`Clock` time API (CI grep gate referenced from the trace).

[...]

PR #828, opened 2026-05-04
docs(platform): detailed design — clock (refs #723)

cjhowe-us:
## Summary

- Adds `specs/platform/clock-design.md` — detailed design for the platform `Clock` aggregate plus the additive `FrameTick` + `FixedStepAccumulator` value objects (per spike brief).
- 12 sections per `specs/_TEMPLATE.md`: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.
- Refines `specs/platform/SPEC.md` §4.4, §5.9, §6.6, §7, §8.1, §9.1 / §9.2, §10.3.4. SRP boundary is sharp: clock owns the OS time seam and the value-object delta / fixed-step primitives; refuses scheduling, event pumping, time-of-day logic, and VBlank pacing.

Resolves the design half of acceptance story #358 (`platform/clock: monotonic non-decreasing + wall correlation`). Sibling `task-breakdown-platform-clock-detailed` is unblocked by this PR.

Refs #723.

## Test plan

- [ ] Reviewer confirms 12 sections present and each is non-trivial.
- [ ] Reviewer confirms the design refuses scheduling (#702), event pumping (#717), and time-of-day logic per spec §1.
- [ ] Reviewer confirms `Result<T>` / `glibre::Error` discipline matches `reviews/decisions/error-model.md`.
- [ ] Reviewer confirms hot-reload story matches SPEC §8.1 + `reviews/decisions/hot-reload-protocol.md`.
- [ ] Reviewer confirms perf budget cell (§9.1) is preserved verbatim.

### DIFF
--- a/specs/platform/clock-design.md
+++ b/specs/platform/clock-design.md
@@ -1,0 +1,1521 @@
+# platform — Detailed Design: clock aggregate
+
+> Detailed design for the `Clock` / `Instant` / `WallTime` / `Duration`
+> aggregate declared in `specs/platform/SPEC.md` §4.4. Refines §4.4,
+> §5.9 (public surface), §6.6 (`mach_absolute_time` seam), §7 (the
+> non-serializability rule for `Instant`), §8.1 (clock as a
+> peer-plugin pass-through; platform-self-reload re-acquisition rule),
+> §9.1 / §9.2 (per-frame budget + sub-arena), §10.3.4 (total-function
+> failure model). Adds the **frame-tick** + **fixed-step accumulator**
+> value-objects mandated by the spike brief; both are pure value
+> primitives that *use* `Clock` rather than embedding scheduling
+> policy.
+>
+> Cites `reviews/decisions/error-model.md`,
+> `reviews/decisions/frame-phases.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`,
+> `reviews/decisions/fory-codegen.md`. Does not introduce any new
+> public surface beyond the §5 stubs in `specs/platform/SPEC.md`
+> except for the additive `FrameTick` / `FixedStepAccumulator` value
+> objects co-located on `glibre::platform::Clock`'s seam — both are
+> pure header-side computations on `Instant` / `Duration` and
+> introduce no new ABI symbols beyond those derivable from the
+> existing `Clock::now` / `wall` / `native_tick` triplet.
+>
+> Harmonius prior art (`harmonius/docs/requirements/platform/
+> threading-async.md`, `harmonius/docs/design/core-runtime/
+> game-loop.md` §Fixed Timestep Accumulator) cited as research input
+> only — every conclusion below was independently re-derived per
+> `PHILOSOPHY.md`.
+
+Refs: spike #723 — `[SPIKE] design-platform-clock-detailed`.
+Parent: #714. Sibling task-breakdown spike blocked-by this
+deliverable.
+
+## 1. Purpose
+
+The `clock` aggregate is the single component in the platform
+context permitted to call the OS time-source primitives
+(`mach_absolute_time` / `mach_timebase_info` for the monotonic
+side; `std::chrono::system_clock::now` for the wall side). Its one
+responsibility is **owning the OS time seam**: returning a
+monotonic, never-decreasing, never-wrapping `Instant`; returning a
+calendar-bearing `WallTime`; advertising the underlying hardware
+tick so deterministic step sizing can land on an integer multiple
+of it; and offering two pure-function value-object companions —
+`FrameTick` (per-frame delta computation) and
+`FixedStepAccumulator` (deterministic fixed-step substep counting)
+— that turn raw `Instant` deltas into the form simulation /
+physics / animation / present need.
+
+What this aggregate explicitly refuses to own:
+
+- **Phase scheduling.** The decision to *run* a system at a
+  particular phase boundary is owned by `core` (specs/core, schedule
+  + frame loop, sibling design `schedule-frame-design.md`). The
+  clock returns numbers; `core` decides what to do with them.
+- **Fixed-step driving.** `FixedStepAccumulator` is a value-object
+  utility: it consumes a frame delta, returns "how many substeps to
+  run" + "interpolation alpha". It does **not** invoke physics, AI,
+  or any other consumer; the consumer (physics, sibling sub-epic
+  #132) calls `accumulator.consume()` and runs its own loop. Per
+  spec §1, platform refuses to schedule.
+- **Time-of-day / calendar logic.** Day / night cycles, in-game
+  clocks, save-game wall stamps used as gameplay state are not
+  platform's concern. Platform exposes `WallTime` for log
+  correlation only (§4.4 inv #3); gameplay code that needs a
+  game-clock builds it on top.
+- **Event pumping.** SDL3 event drain (§4.2, sibling #717) is the
+  `Pump` aggregate. The clock has no opinion about events.
+- **VBlank pacing / present timing.** `CAMetalDisplayLink` callback
+  handling and the swap-interval policy live in render's present
+  half (sibling sub-epic #97). The clock contributes the **wall
+  read** at present time (§9.1 phase-9 row) and nothing else.
+- **NTP / clock-skew detection / time-zone handling.** The wall
+  clock is consumed at face value; we report what the OS reports.
+  Drift detection is the e2e long-run test (§11.3.4), not a
+  per-frame surface.
+- **Process / signals / argv.** `Process` (§4.5) is sibling #725.
+- **File watching / file I/O.** `FileWatcher` (§4.3, sibling #719)
+  / `FileIo` (§4.6, sibling #721).
+- **Window / surface.** Sibling #715.
+- **Platform error policy.** Closed sum `platform::Error` and the
+  POSIX → arm translation (§4.7, §10) is sibling #727. This
+  design surfaces failures into pre-existing arms; it does not
+  invent or rename them. (In practice, §10.3.4 already says
+  `Clock::now` / `wall` / `native_tick` are total functions and
+  return no `Result<T>` — see §10 below.)
+
+The aggregate's SRP boundary is sharp: if the OS monotonic source's
+resolution / API / guarantees shift (e.g. `mach_absolute_time`
+deprecation in a future macOS, or the M-series timebase rate
+changes across SoC generations), this design changes. Anything
+else is out of scope.
+
+## 2. Requirements coverage
+
+Mapping of the harmonius requirement / design clauses that pertain
+to time / clock / frame-pacing onto MVP coverage in this aggregate.
+Every entry is independently re-derived; coverage sites refer to
+sections of `specs/platform/SPEC.md` and to the design sections
+below.
+
+| Harmonius clause                                                                       | Glibre disposition (MVP)                                                                                                                                                                                                                          |
+|----------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `threading-async.md` *monotonic time source for scheduling, never wraps, never slews*  | **Covered.** `Clock::now() -> Instant` backed by `mach_absolute_time` (§3.2). §4.4 inv #1 (monotonic non-decreasing) is the load-bearing rule; abort on regression rather than clamp.                                                              |
+| `threading-async.md` *wall-clock for log timestamps and crash dumps*                   | **Covered.** `Clock::wall() -> WallTime` is the only sanctioned source (§4.4 inv #3). Log helper (`glibre::log_error`) calls `Clock::wall()`; engine code never calls `time(nullptr)` / `gettimeofday`.                                              |
+| `core-runtime/game-loop.md` §Fixed Timestep Accumulator (substeps, alpha)              | **Covered, ownership split.** The numeric primitive (`FixedStepAccumulator`) lives here; the *use* (driving phase 3 physics-fixed substeps) lives in sibling `physics` (#132) and core's schedule-frame loop. §3.5.                                |
+| `core-runtime/game-loop.md` *monotonic frame counter*                                  | **Covered, ownership split.** `Clock` provides `Instant`; the `frame_index : std::uint64_t` counter itself is core's concern (incremented in phase 9 per `frame-phases.md`). Platform contributes the time, not the index.                          |
+| `core-runtime/game-loop.md` *delta-time per frame*                                     | **Covered.** `FrameTick::tick(Clock&) -> FrameDelta` returns `{ now, delta, wall_now }` (§3.4). Caller threads the `FrameTick` through frames; platform owns no global `last_now_` field — that would couple to scheduling.                          |
+| `core-runtime/game-loop.md` *max-ticks-per-frame cap (spiral-of-death prevention)*     | **Covered.** `FixedStepAccumulator::consume(max_ticks)` clamps and returns the carry-over; remaining accumulated time is preserved for the next frame so alpha stays continuous. §3.5.                                                              |
+| `core-runtime/game-loop.md` *interpolation alpha in `[0, 1]`*                          | **Covered.** `FixedStepAccumulator::alpha() -> float` returns the residual divided by step duration, clamped to `[0, 1]`. §3.5.                                                                                                                     |
+| `tools/profiler.md` *high-precision time deltas for span measurement*                  | **Owned by sibling `tools` profiler context.** The profiler reads `Clock::now()` like any consumer; building the span tree is not platform's concern. We guarantee `now()` is O(1) and `Instant` resolution is at least 1 ns (§9).                  |
+| `crash-reporting.md` *correlate monotonic frame-time budget overrun with wall stamp*   | **Covered.** §4.4 inv #3 makes `WallTime` the single sanctioned correlation source. The crash dump captures one `Clock::wall()` plus the most recent `Instant`; pairing these is the call site's responsibility (`obs` context, post-MVP).         |
+| `core-runtime/game-loop.md` *VBlank-driven present pacing*                             | **Refused (routed to `render` / SDL3).** `CAMetalDisplayLink` callback handling lives in render's present path. Platform's only contribution at present is the second `Clock::wall()` read (§9.1 phase-9 row). No VBlank-sync handle on `Clock`.    |
+| `core-runtime/game-loop.md` *deterministic replay-from-trace*                          | **Covered, by refusal.** §4.4 inv #2: `Instant` is not serializable. Replay frameworks that need recorded time use a recorded `Duration` stream from the e2e harness, not `Clock`. This design exposes no surface for replay-driven time injection. |
+| `tools/profiler-test-cases.md` *per-thread monotonic-source coherence*                 | **Covered.** §4.4 inv #1 + §6 (concurrency): all reads come from the same OS counter, so any two threads' `Instant` values are comparable. No per-thread offset.                                                                                    |
+| harmonius `Duration` arithmetic (signed delta type)                                    | **Covered.** `Duration = std::chrono::nanoseconds` (signed `int64_t`). `Instant - Instant -> Duration` is the only signed arithmetic at the seam (§4.4). No raw `int64_t` substitutes; the type system prevents accidental nanosecond / millisecond / tick mixing. |
+| harmonius `Clock::native_tick` advertisement                                           | **Covered.** §4.4 inv #4: `Clock::native_tick() -> Duration` returns the smallest representable step on the host. On macOS / M1 this is `1 ns` (`mach_absolute_time` is sub-µs; rounding to integer ns is below noise; see §3.2).                  |
+| harmonius "one global clock per process"                                               | **Covered.** §4.4 inv #5: Meyer's singleton accessor `Clock::get()`. Aggregates that need time take `Clock&`. §3.6.                                                                                                                                  |
+
+Glibre-native requirements added beyond harmonius:
+
+- **`FrameTick` is stateless w.r.t. `Clock`.** The previous-instant
+  field lives on `FrameTick` (a value object the *caller* owns),
+  not on `Clock`. This keeps `Clock::now()` re-entrant and keeps
+  the singleton's mutable state at zero (matters for §6
+  concurrency and §8 hot-reload).
+- **`FixedStepAccumulator` is a value, not an entity.** It has no
+  identity; copying it is meaningful (e.g. for branchless lookahead
+  / what-if rollback in net-code). Keeping it value-typed lets the
+  consumer (physics / animation) place one per fixed-step domain
+  without coordinating through a global registry.
+- **`Instant` is `int64_t` nanoseconds since an arbitrary epoch.**
+  The internal layout is fixed (§3.3) so the `glibre-types` ABI
+  hash is stable: any peer plugin that holds an `Instant` across
+  hot-reload (e.g. last-known frame time) sees the same byte layout
+  before and after the swap.
+- **`WallTime` is `std::chrono::system_clock::time_point`.** No
+  custom calendar type; the seam value is the standard library
+  type so `spdlog`'s formatter can take it directly.
+
+Coverage rule: every harmonius clause above either lands in this
+design (with a coverage site) or is refused with a one-line
+rationale. No silent drops.
+
+## 3. Detailed model
+
+### 3.1 Aggregate composition
+
+```text
+Clock  (aggregate root, process singleton)
+├── mach_timebase_info_data_t   timebase_   (cached at first now() call; numer/denom)
+└── (no other state — clock is otherwise stateless)
+```
+
+Companion value objects (header-side, no allocation):
+
+```text
+Instant  (value object)
+└── std::int64_t   ns_   (nanoseconds since arbitrary monotonic epoch)
+
+WallTime  (value object)
+└── std::chrono::system_clock::time_point   point_
+
+Duration = std::chrono::nanoseconds   (alias, §5.2 in SPEC)
+
+FrameTick  (value object — caller-owned per-frame delta computer)
+└── Instant   prev_   (last observed Instant; default-constructed == zero)
+
+FixedStepAccumulator  (value object — caller-owned fixed-step bookkeeper)
+├── Duration   step_         (immutable; the fixed step duration)
+├── Duration   accumulator_  (residual time not yet consumed by ticks)
+└── std::uint32_t  max_ticks_per_frame_  (spiral-of-death cap)
+```
+
+The `Clock` is a Meyer's singleton (§4.4 inv #5). Its only *mutable*
+state is the `mach_timebase_info_data_t` cache — and that cache is
+write-once: the first `now()` reads `mach_timebase_info`,
+populates the cache, then the cache is read-only for the rest of
+the process. Every subsequent `now()` is purely
+read-cached-numer/denom + one `mach_absolute_time` syscall + one
+128-bit multiply-divide. No locks, no atomics on the read path
+(see §6.1 for the publication pattern).
+
+`FrameTick` and `FixedStepAccumulator` are pure value objects:
+copying them is well-defined, comparing them is well-defined, and
+they own no resources that could conflict on copy. They do not
+contain a `Clock&` reference; `tick(Clock&)` and `accumulate(...)`
+take it as a method argument. This is deliberate — embedding a
+reference would (a) prevent value-semantic copy, (b) force
+lifetime coupling when none is needed, and (c) leak `Clock`'s
+singleton-ness into every aggregate that holds a `FrameTick` field.
+
+### 3.2 Monotonic source: `mach_absolute_time` (macOS / M1 baseline)
+
+Locked from SPEC §6.6:
+
+```cpp
+namespace glibre::platform {
+
+auto Clock::now() const noexcept -> Instant {
+    static const auto info = []() noexcept -> mach_timebase_info_data_t {
+        mach_timebase_info_data_t i{};
+        ::mach_timebase_info(&i);
+        return i;
+    }();
+    const auto t  = ::mach_absolute_time();
+    const auto ns = static_cast<std::int64_t>(
+        (__uint128_t(t) * info.numer) / info.denom);
+#if !defined(NDEBUG)
+    // §4.4 inv #1 defensive guard: detect monotonic regression in debug.
+    // The atomic load below uses `relaxed` because the rule is
+    // a *correctness* invariant, not a synchronization point —
+    // any read that observes `last_seen` is correctness-preserving
+    // regardless of inter-thread ordering, since the only valid
+    // outcome is "no regression" or "abort".
+    static std::atomic<std::int64_t> last_seen{0};
+    auto prev = last_seen.load(std::memory_order_relaxed);
+    while (ns < prev) {
+        // Spec invariant violated by the OS. Abort, do not clamp.
+        std::abort();
+    }
+    // Best-effort publish. Loss of update is fine; another thread
+    // may have already advanced it past `ns`.
+    last_seen.store(ns, std::memory_order_relaxed);
+#endif
+    return Instant{ns};
+}
+
+}  // namespace glibre::platform
+```
+
+Choices and their reasons:
+
+- **`__uint128_t` widening multiply.** `mach_absolute_time` returns
+  a `uint64_t` count of timebase ticks. On Apple Silicon the
+  timebase numerator / denominator are typically `1` / `1` (the
+  count is already nanoseconds), but the spec does not guarantee
+  that — a future SoC could change the ratio. The 128-bit
+  intermediate prevents overflow in the conversion at any plausible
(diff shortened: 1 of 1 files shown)


specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+    // outcome is "no regression" or "abort".
+    static std::atomic<std::int64_t> last_seen{0};
+    auto prev = last_seen.load(std::memory_order_relaxed);
+    while (ns < prev) {
+        // Spec invariant violated by the OS. Abort, do not clamp.
+        std::abort();
+    }
+    // Best-effort publish. Loss of update is fine; another thread
+    // may have already advanced it past `ns`.
+    last_seen.store(ns, std::memory_order_relaxed);
```
cjhowe-us, line 227: severity:HIGH location:specs/platform/clock-design.md:221-227 problem:The debug-only regression guard uses a plain relaxed store rather than a compare-exchange, enabling a concurrent backward write to `last_seen` that produces spurious aborts on valid concurrent callers. Thread A reads the clock at real-time T1 and gets ns_A=1050; thread B reads at T2>T1 and gets ns_B=1100. Thread B stores last_seen=1100. Then thread A executes its store: last_seen=1050 (backward). Thread C reads the clock at T3>T2, gets ns_C=1080 (valid), reads prev=1100 via a relaxed load, sees 1080 < 1100, and aborts — incorrectly, because ns_C is a legitimately post-T2 reading that just happens to be less than what thread B stored. On Apple Silicon all cores share the same counter (`CNTVCT_EL0`) so out-of-order reads between cores are rare in practice, but the design itself offers no proof of safety and the store is not CAS-guarded. fix:Replace the relaxed store with a CAS loop that only advances last_seen: `while (prev < ns && !last_seen.compare_exchange_weak(prev, ns, std::memory_order_relaxed, std::memory_order_relaxed)) {}`. This makes the atomic monotonically increasing and prevents the backward write that causes spurious aborts. The while loop around std::abort() is also misleading as written (the abort never returns, so it is effectively an if); the fix collapses this naturally.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+```
+
+`TestClock` is structurally compatible with `Clock&` consumers via
+templated functions (e.g. `template <class C> auto tick(C& clock)`)
+where the test fixture passes `TestClock` and the production code
+passes `Clock`. We deliberately do **not** introduce a virtual
+`IClock` base class — virtual dispatch would impose a runtime cost
+on every consumer for the sake of testability, and the templated-
+parameter pattern works at zero cost. (This mirrors the
+window-surface-design approach to hot-cold seam testability.)
```
cjhowe-us, line 1449: severity:HIGH location:specs/platform/clock-design.md:1442-1452 problem:The test-support section claims TestClock is usable with FrameTick::tick via 'templated functions (e.g. template <class C> auto tick(C& clock))' but the public surface declared in §4 at line 649 is 'auto tick(Clock& clock) noexcept -> FrameDelta' — a concrete non-template method taking Clock& by reference. A TestClock& cannot be implicitly converted to Clock&; the claimed template indirection does not exist in the declared interface. This is an internal contradiction that blocks isolated unit testing of FrameTick: either the §4 public surface must add a template parameter (e.g. 'template <class ClockT> auto tick(ClockT& clock) noexcept -> FrameDelta') or the §11.4 test strategy must describe a different seam (e.g. a free-function adapter or a separate non-member tick). fix:Two options are in scope. Option A (preferred for testability): templatize FrameTick::tick — 'template <class ClockT> [[nodiscard]] auto tick(ClockT& clock) noexcept -> FrameDelta'. The production call site calls tick(Clock::get()); the test fixture calls tick(test_clock). Option B: keep the concrete Clock& method and introduce a non-member 'template <class ClockT> FrameDelta tick_with(FrameTick&, ClockT&)' in a test-only header. Either option must be declared in §4 with a note in §11.4 cross-referencing it. The current design leaves the implementation path undefined.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+  correlation race between them: between two separate `Clock::now()`
+  + `Clock::wall()` calls a context switch could interleave a
+  wall-clock slew (NTP step) and produce a bogus delta. Reading
+  them inside the same `FrameTick::tick` keeps the gap to a few
+  hundred nanoseconds and documents that "the wall stamp belongs
+  to *this* monotonic instant".
+- **No `Clock&` member.** Reasons in §3.1. The caller owns the
+  reference; `Clock::get()` is fine to pass on every call (the
+  singleton accessor is itself O(1)).
+- **`reset()` is explicit.** Long pauses (debugger break, OS
```
cjhowe-us, line 369: severity:MED location:specs/platform/clock-design.md:359-372 problem:The design claims 'Capturing both now and wall in one call removes the correlation race between them' (line 359) but the implementation in §3.4 at line 335-336 still calls clock.now() and clock.wall() as two sequential operations with no synchronization between them. An NTP step-correction that fires between those two calls can produce a wall delta that is uncorrelated with the monotonic delta. The claim 'removes the correlation race' overstates the guarantee; the actual guarantee is 'bounds the gap to a few hundred nanoseconds and documents the pairing intent' which the design itself correctly states two lines later. fix:Change the language from 'removes the correlation race' to 'reduces and bounds the correlation gap to the sub-microsecond latency between the two calls, which is the best achievable without atomic multi-clock hardware support'; add a one-sentence note that NTP step-corrections can still affect wall within that gap. This is a documentation accuracy fix, not a code change.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+| `core-runtime/game-loop.md` *monotonic frame counter*                                  | **Covered, ownership split.** `Clock` provides `Instant`; the `frame_index : std::uint64_t` counter itself is core's concern (incremented in phase 9 per `frame-phases.md`). Platform contributes the time, not the index.                          |
+| `core-runtime/game-loop.md` *delta-time per frame*                                     | **Covered.** `FrameTick::tick(Clock&) -> FrameDelta` returns `{ now, delta, wall_now }` (§3.4). Caller threads the `FrameTick` through frames; platform owns no global `last_now_` field — that would couple to scheduling.                          |
+| `core-runtime/game-loop.md` *max-ticks-per-frame cap (spiral-of-death prevention)*     | **Covered.** `FixedStepAccumulator::consume(max_ticks)` clamps and returns the carry-over; remaining accumulated time is preserved for the next frame so alpha stays continuous. §3.5.                                                              |
```
cjhowe-us, line 112: severity:MED location:specs/platform/clock-design.md:112 problem:The requirements coverage table says FixedStepAccumulator 'clamps and returns the carry-over; remaining accumulated time is preserved for the next frame so alpha stays continuous' but §3.5 and test §11.1.7 describe the opposite: when the cap fires, excess time beyond max_ticks*step is *dropped*, not preserved, and alpha falls to 0.0f. The cap behaviour deliberately accepts a one-frame simulation slowdown (correct) but the requirements row mischaracterises it as carry-over preservation (incorrect). Alpha continuity is only guaranteed in the non-cap steady state. fix:Update the requirements row to read: 'FixedStepAccumulator::consume() caps at max_ticks_per_frame and drops excess accumulation; alpha is continuous in the non-cap steady state; when the cap fires, the excess is dropped (one-frame simulation slowdown, deterministic spiral-of-death prevention). §3.5.' This aligns the coverage table with the actual design semantics documented in §3.5.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+
+    [[nodiscard]] auto now()  const noexcept -> Instant  { return at_; }
+    [[nodiscard]] auto wall() const noexcept -> WallTime { return wall_; }
+    [[nodiscard]] auto native_tick() const noexcept -> Duration { return Duration{1}; }
+
+    auto advance(Duration d) noexcept -> void;
+
+private:
+    Instant   at_{};
+    WallTime  wall_{};
```
cjhowe-us, line 1436: severity:MED location:specs/platform/clock-design.md:1432-1436 problem:TestClock::advance(Duration d) advances only 'at_' (the monotonic instant) but never advances 'wall_'. This means any test using FrameTick::tick(test_clock) will always receive a FrameDelta with wall == WallTime{} (the epoch) regardless of how many advance() calls have been made. Tests asserting on FrameDelta::wall (e.g. checking that the wall stamp moves with the monotonic stamp in a plausible way) are silently broken. The test fake is described as the mechanism backing §11.1.10 and §11.1.11, and those specific tests don't assert on wall, so they pass — but any future test that checks FrameDelta::wall via TestClock will get a false constant. fix:Either (a) have advance(Duration d) also advance wall_ by converting the Duration to a std::chrono::duration and adding to wall_.point, so the wall moves in lockstep with the monotonic — or (b) add a separate advance_wall(std::chrono::nanoseconds) method. Option (a) is simpler: 'wall_ = WallTime{wall_.point + d};' inside advance(). Add a note in §11.4 documenting this invariant.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+> only — every conclusion below was independently re-derived per
+> `PHILOSOPHY.md`.
+
+Refs: spike #723 — `[SPIKE] design-platform-clock-detailed`.
+Parent: #714. Sibling task-breakdown spike blocked-by this
+deliverable.
+
+## 1. Purpose
+
+The `clock` aggregate is the single component in the platform
```
cjhowe-us, line 39: severity:LOW location:specs/platform/clock-design.md:39 problem:The design file header cross-references the spike issue #723 and parent #714 but the PR body additionally says 'Resolves the design half of acceptance story #358' — this story reference is absent from the design file itself. The commit body mentions 'Closes the design half of acceptance story #358' but the design file's Refs line at line 39 only cites 'spike #723'. The story #358 is the acceptance story whose §11 test plan references in §11.5 (line 1473). Having the design file not cross-reference the story it closes makes future spec-check CI and the §2 traceability table harder to audit. fix:Add 'Story: #358 (design half)' to the Refs block at line 39, e.g.: 'Refs: spike #723 — [SPIKE] design-platform-clock-detailed. Story: #358 (design half; implementation half unblocked by sibling task-breakdown). Parent: #714.'

cjhowe-us (commented):
findings:
  HIGH:2
  MED:3
  LOW:1
carryover:0
notes:The clock-design.md is substantively well-structured: all 12 sections are present and non-trivial, the SRP boundary is sharp and explicitly justified, the total-function surface (no Result<T> for clock calls) is correctly grounded in SPEC §10.3.4, EASTL types are used correctly per PHILOSOPHY §11 (std::chrono retained in the permitted carve-out), all cross-context imports are minimal and declared, the hot/cold split is coherent, the ABI table in §7.3 is self-consistent, the hot-reload story is correctly reasoned for both peer-plugin and platform-self-reload cases, and the test plan (§11) covers the Gherkin Thens at the right tier. The Conventional Commit subject docs(platform): detailed design — clock (refs #723) is correct in type, scope, and issue reference.

Two HIGH findings block the implementation plans that build on this design. First, the debug-only monotonic regression guard at §3.2 uses a plain relaxed store to advance last_seen, which allows a concurrent backward write and will produce spurious aborts on valid multi-threaded callers when one thread's store overtakes a later thread's store (see inline comment at line 227). The fix is a CAS loop that only monotonically advances the atomic. Second, §11.4 describes TestClock as compatible with FrameTick::tick via template indirection but the §4 public surface declares tick(Clock& clock) with a concrete type — a TestClock cannot be passed to it, making the entire unit-test strategy for FrameTick non-implementable as written (see inline comment at line 1449). The fix requires either templating the method or introducing a free-function adapter, and the chosen approach must be back-filled into §4.

Three MED findings do not block implementation but should be corrected before the task-breakdown spike closes: the requirements coverage table at line 112 mischaracterises the cap behaviour as "carry-over preserved", the FrameDelta comment at line 325 and the reasoning at line 359 overclaim that the now/wall race is "removed" rather than bounded, and TestClock::advance() never advances wall_ (line 1432-1436) making any future FrameDelta::wall assertion via TestClock silently incorrect.

One LOW finding: the Refs block in the design file header (line 39) omits the story reference #358 that the PR body and commit message cite, which is needed for spec-check CI traceability.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:3
  LOW:1
carryover:0
notes:The clock-design.md is substantively well-structured: all 12 sections present and non-trivial, SRP boundary sharp and justified, total-function surface correctly grounded in SPEC §10.3.4, EASTL types correct per PHILOSOPHY §11, cross-context imports minimal and declared, hot/cold split coherent, ABI table §7.3 self-consistent, hot-reload story correctly reasoned for both peer-plugin and platform-self-reload cases, test plan §11 covers Gherkin Thens at the right tier, Conventional Commit subject correct. Two HIGHs block implementation plans. HIGH-1 (line 227): debug regression guard uses plain relaxed store enabling concurrent backward write producing spurious aborts on valid multi-threaded callers — fix with CAS loop. HIGH-2 (line 1449): §11.4 claims TestClock works via template indirection but §4 declares tick(Clock&) as a concrete-typed method making the unit-test strategy for FrameTick non-implementable — fix by templating the method or adding a free-function adapter and back-filling §4. Three MEDs: requirements table line 112 mischaracterises cap behaviour as carry-over; lines 359/325 overclaim correlation race removed vs bounded; TestClock::advance never advances wall_. One LOW: design file Refs block omits story #358 cross-reference needed for spec-check CI traceability.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:3
  LOW:1
carryover:0
notes:The clock-design.md is substantively well-structured: all 12 sections present and non-trivial, SRP boundary sharp and justified, total-function surface correctly grounded in SPEC §10.3.4, EASTL types correct per PHILOSOPHY §11 carve-out, cross-context imports minimal, hot/cold split coherent, ABI table §7.3 self-consistent, hot-reload reasoning correct for both peer-plugin and platform-self-reload, test plan §11 covers Gherkin Thens at the right tier, Conventional Commit subject docs(platform): correct type scope and issue ref. Two HIGHs block the implementation plans downstream. HIGH-1 (line 227): the debug-only regression guard stores last_seen with a plain relaxed store rather than a CAS loop, enabling a concurrent backward write that causes spurious aborts on valid multi-threaded callers. HIGH-2 (lines 1442-1452): §11.4 claims TestClock works with FrameTick::tick via template indirection, but §4 declares tick(Clock&) as a concrete non-template method, making the unit-test strategy non-implementable without a public surface change. Both HIGHs require a follow-up amendment to this design document and must be reflected in the task-breakdown spike before implementation plans run. Three MEDs: (a) requirements table line 112 mischaracterises the cap-fires behaviour as carry-over preserved when §3.5 and test §11.1.7 document excess-drop; (b) line 359 overclaims 'removes the correlation race' when the actual guarantee is bounded gap; (c) TestClock::advance() never advances wall_, silently breaking any future test asserting on FrameDelta::wall. One LOW: Refs block in design file header omits story #358 cross-reference present in the PR body and commit message.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:473f51b followup:#898

Replaced the plain `relaxed` store with a monotonic-CAS loop in the §3.2 debug guard. The loop advances `last_seen` only when `ns > prev`; the abort fires only when the OS genuinely regressed (`ns < prev` after all CAS iterations exhaust). Concurrent callers that lose the CAS simply leave `last_seen` at its (larger) current value rather than overwriting it backward.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:473f51b followup:#898

Templated FrameTick::tick on ClockT with concept constraint (requires c.now(); c.wall();) in both §3.4 and §4 public surface. §11.4 updated: TestClock satisfies the constraint, two concrete instantiations documented (Clock& production, TestClock& in §11.1.10-12 unit tests), structural-compat/IClock language replaced with correct template-instantiation description. Added [OPEN] [NON-BLOCKING] §12 entry tracking the instantiation inventory.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:473f51b followup:#898

Replaced 'removes the correlation race' with 'reduces and bounds'. Added explicit analysis: the two calls are sequenced inside one tick body; a wall-clock NTP step between them produces an at-most-one-tick gap recorded in FrameDelta rather than masked. Clarified that the benefit vs a callsite-level sequence is bounding the window to hundreds of nanoseconds vs the caller's full frame body.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:473f51b followup:#898

Rewrote the requirements table row for the max-ticks cap. Old text said 'remaining accumulated time is preserved for the next frame so alpha stays continuous' — that contradicts §3.5 and §11.1.7 which document excess-drop semantics (residual zeroed when cap fires). New text: cap fires → excess dropped to zero; alpha continuity holds only in the non-cap steady state; excess discarded to bound runaway-spiral cost. Cites §3.5 + §11.1.7.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:473f51b followup:#898

Expanded TestClock::advance(Duration d) to advance both at_ and wall_.point by d. Added inline test invariant comment: 'wall == zero_epoch + at_ under TestClock'. Tests asserting on FrameDelta::wall now see wall-time progression equal to the monotonic progression, matching the spec intent for §11.1.10-12.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:473f51b followup:#898

Added 'Story: #358 (design half)' to the Refs block (line 34 of the updated file) alongside the existing spike #723 and Parent: #714 entries.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:3
  LOW:1
carryover:0 (all R1 HIGH/MED/LOW addressed; no unresolved items carried into R2)
notes:The R1 follow-up on PR #898 (commit 473f51b) fully addresses all six R1 findings at the correctness level. The CAS loop (HIGH-1) is correctly formed — it advances last_seen only when ns is strictly larger, and the abort fires only after the loop when ns < prev, which is the genuine-regression-only condition. The template tick<ClockT> (HIGH-2) uses a clean requires-clause concept constraint with two concrete instantiations cited in §11.4 and §12. The max-ticks excess-drop semantics (MED-1) are now accurate and consistent with §3.5 and §11.1.7. The correlation-race prose (MED-2) correctly says "reduces and bounds". TestClock::advance() advances wall_.point (MED-3). Story #358 is added to the Refs block (LOW-1). Round-2 lens (cohesion + SRP + seam quality) reveals three residual MED issues and one LOW in the #898 diff itself: (1) The inline comment preamble at the top of the debug guard block (line ~214 in proposed file) still describes "the atomic load below" using the rationale for the old plain-store, not the new CAS loop — the updated CAS block immediately below it and the stale preamble now describe different algorithms in the same block, a §3.2 cohesion failure. (2) §6.1 item 4 still says "concurrent threads racing on last_seen may each abort" which was correct for the plain-store but is wrong for the CAS loop — under the CAS only threads whose OS sample genuinely regressed abort; this makes §6.1 incoherent with §3.2. (3) Multiple occurrences of tick(Clock&) in the requirements table (§2 line 111), §3.1, §3.4 rationale heading, and §5.1 hot-path table have not been updated to reflect the template signature tick<ClockT>(ClockT&); only §3.4 body, §4, and §11.4 were updated. (4) LOW: §3.4 rationale contains an unresolved §x placeholder cross-reference introduced by the MED-2 fix — no section §x exists in the document. All four findings are posted as inline comments on PR #898. None blocks merge; all are MED/LOW and in-scope for the impl-respond agent on #898 before it lands.

cjhowe-us:
responder:go-impl-respond round:2 decision:ADDRESSED commit:89892cf followup:#898

All three MED findings and the LOW finding from review 4229066076 are addressed in commit 89892cf on branch \`fix/platform-clock-design-followup-r1\` (PR #898).

**MED-1** — \`specs/platform/clock-design.md\` §3.2 \`#if !defined(NDEBUG)\` block (~line 212): replaced the stale "atomic load below uses \`relaxed\`" preamble with a CAS-accurate version: "The atomic CAS loop below uses \`relaxed\` ordering because the regression check is a *correctness* invariant — every thread observes a monotonically non-decreasing snapshot and aborts only if its own observation regresses against that snapshot. No happens-before is needed because the CAS itself synchronizes the read-modify-write. Backward writes are impossible by construction."

**MED-2** — \`specs/platform/clock-design.md\` §6.1 item 4 (~line 838): replaced the old plain-store description ("may each abort if they observe a regression, or each not-abort if they don't") with the CAS-correct text: "A thread aborts iff its own observed \`ns < last_seen\` after the CAS loop terminates — which can only happen on a genuine OS-clock regression, not on a race. No racing thread can write a backward \`last_seen\` value because the CAS only advances forward."

**MED-3** — \`specs/platform/clock-design.md\` four locations updated from \`FrameTick::tick(Clock&)\` → \`FrameTick::tick<ClockT>\`:
  1. §2 requirements table delta-time row (~line 111)
  2. §3.1 value objects description (~line 190)
  3. §3.4 rationale first bullet (~line 385)
  4. §5.1 hot-path table (~line 764)

**LOW-1** — \`specs/platform/clock-design.md\` §3.4 rationale (~line 390): replaced \`§x\` placeholder with "\`obs\` context diagnostics surface (exact section to be filed in the obs spec follow-up)". The prose context (NTP-step gap recorded in \`FrameDelta\`) points to the future \`obs\` (observability) context, which is post-MVP; §9.1 covers the per-frame budget cell but not the diagnostics surface, so citing \`obs\` with the follow-up note is the correct anchor.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
+# platform — Detailed Design: clock aggregate
+
+> Detailed design for the `Clock` / `Instant` / `WallTime` / `Duration`
+> aggregate declared in `specs/platform/SPEC.md` §4.4. Refines §4.4,
+> §5.9 (public surface), §6.6 (`mach_absolute_time` seam), §7 (the
+> non-serializability rule for `Instant`), §8.1 (clock as a
```
cjhowe-us, line 6: severity:MED location:specs/platform/clock-design.md:4-9 problem:The preamble states this design 'Refines ... §6.6 (mach_absolute_time seam)' but specs/platform/SPEC.md §6.6 (line 1500) still says 'A defensive assert(ns >= last) guards regression in debug builds' — a prescriptive implementation statement that directly contradicts the CAS loop introduced by the R1 HIGH-1 fix. An implementer reading SPEC.md alone (without the design doc) would write the plain single-assert version, which has the race condition the HIGH-1 fix was specifically designed to prevent. The design doc and SPEC.md are now out of sync on a correctness-critical detail. recommendation:Amend SPEC.md §6.6 to either (a) replace the assert prose with a pointer to clock-design.md §3.2 as the authoritative implementation, or (b) update the §6.6 code snippet and description to describe the CAS loop. A one-liner note in §6.6 after the assert sentence — 'Note: clock-design.md §3.2 refines this to a monotonic-CAS loop for thread safety' — would be sufficient.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+> policy.
+>
+> Cites `reviews/decisions/error-model.md`,
+> `reviews/decisions/frame-phases.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`,
+> `reviews/decisions/fory-codegen.md`. Does not introduce any new
+> public surface beyond the §5 stubs in `specs/platform/SPEC.md`
+> except for the additive `FrameTick` / `FixedStepAccumulator` value
```
cjhowe-us, line 21: severity:MED location:specs/platform/clock-design.md:20-26 problem:The preamble claims this design 'Does not introduce any new public surface beyond the §5 stubs in specs/platform/SPEC.md except for the additive FrameTick / FixedStepAccumulator value objects ... and introduce no new ABI symbols beyond those derivable from the existing Clock::now / wall / native_tick triplet.' However §7.3 explicitly adds FrameDelta, FrameTick, and FixedStepAccumulator to the ABI middleman catalogue with specific sizes and layouts (24, 8, and 24 bytes respectively). SPEC.md §5.9 does not enumerate these three types; an implementer or a peer-context author consulting SPEC.md to understand the clock seam's public surface would not discover FrameTick, FrameDelta, or FixedStepAccumulator. The claim that there are 'no new ABI symbols' is also inaccurate — FrameDelta, FrameTick, and FixedStepAccumulator ARE new exported types tracked in the middleman dylib hash. recommendation:Either (a) update SPEC.md §5.9 to add stub declarations for FrameDelta, FrameTick, and FixedStepAccumulator with a forward reference to clock-design.md for details, or (b) correct the preamble prose to say 'introduces new ABI-tracked types FrameDelta, FrameTick, FixedStepAccumulator' rather than 'no new ABI symbols'. The claim about 'no new ABI symbols beyond the Clock triplet' should be removed or corrected.

specs/platform/clock-design.md
```diff
@@ -0,0 +1,1521 @@
...
+  needs to flow through the clock seam rather than render's own
+  interface. Owner: render present-pacing task-breakdown.
+- [OPEN] **`Instant::count()` exposure as the persistence
+  escape-hatch.** §7.2 documents that telemetry payloads carry
+  `Instant::count()` as opaque `int64_t`. There is a structural
+  question — should we type-wrap that as
+  `OpaqueMonotonicNs : int64_t` to discourage cross-process
+  comparison? Today, callers obtain the `int64_t` directly. Re-
+  opens when the `obs` (observability) context lands and the
+  field shape is normalized. Owner: future `obs` SPEC.
```
cjhowe-us, line 1521: severity:LOW location:specs/platform/clock-design.md:1521 problem:The §12 open question about namespace placement (line 1548 in the final document, position ~1521 in the diff) references 'issue #TBD' for the sibling task-breakdown spike. The Refs block (line 34) also says 'Sibling task-breakdown spike blocked-by this deliverable' without a concrete issue number. Since this design is merged and issue #723 declares the sibling task-breakdown spike is unblocked by this deliverable, the sibling spike should have been filed before or immediately after merge. Leaving #TBD in the committed design doc means the §12 owner field is unresolvable. recommendation:File the sibling task-breakdown spike and replace #TBD with its concrete issue number in both the §12 open question and the Refs block line 34.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0 (all R2 MED/LOW confirmed addressed in commit 89892cf; no unresolved items)

notes:The three-round retroactive review of merged PR #828 (clock-design.md) plus follow-up PR #898 (commits 473f51b + 89892cf) reaches a clean topology pass. R1 and R2 carryover is zero: all six R1 findings and all four R2 findings have responses on their respective PRs, the R2 commit (89892cf) correctly addresses §3.2 preamble (MED-1), §6.1 item 4 (MED-2), tick<ClockT> propagation through §2/§3.1/§3.4/§5.1 (MED-3), and the §x cross-reference resolution to 'obs' context (LOW-4). The tick<ClockT> template is correctly declared in both §3.4 body and §4 public surface, with two concrete instantiations documented in §11.4 (Clock& for production, TestClock& for unit tests), satisfying the PHILOSOPHY.md anti-pattern guard on pre-user abstractions. The ABI catalogue §7.3 is intact and self-consistent. The §12 open questions all have owners and correct OPEN/NON-BLOCKING annotations. Issue #723 carries correct type:spike label under parent #714 (type:sub-epic), no pts:* label, correct hierarchy per AGENTS.md. CC subjects of 473f51b and 89892cf both follow docs(platform): convention correctly.

Two MED findings require follow-up before the sibling task-breakdown spike closes: (1) SPEC.md §6.6 still says 'A defensive assert(ns >= last) guards regression in debug builds' — prescriptive implementation prose that directly contradicts the CAS loop introduced by the HIGH-1 fix; an implementer reading only SPEC.md would write the plain-assert version with the race condition. (2) The design preamble claims 'no new ABI symbols beyond the Clock::now/wall/native_tick triplet' and SPEC.md §5.9 still only declares the Clock class; FrameDelta, FrameTick, and FixedStepAccumulator are new ABI-tracked types (§7.3 catalogue entries with explicit sizes) that are absent from SPEC.md §5.9. One LOW: §12 and the Refs block both carry 'issue #TBD' for the sibling task-breakdown spike that this design unblocks — that spike should be filed and the placeholder replaced.

Neither MED blocks the design's use by the task-breakdown spike executor (who reads clock-design.md, not just SPEC.md), but both represent SPEC topology gaps that would accumulate technical debt against the platform SPEC as an authoritative reference.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:270c9f0 followup:#898

MED-1 (clock-design.md lines 4-9 / SPEC §6.6 regression-guard contradiction): Added §12 [BLOCKING IMPLEMENTATION] entry documenting that SPEC.md §6.6 prescribes a plain `assert(ns >= last)` while clock-design.md §3.2 corrects this to a monotonic-CAS loop. Entry states that SPEC §6.6 must be amended before any plan PR consuming this design can land, names the two concrete consumers (phase-9 frame-tick, render's frame-extract NTP-gap diagnostics), and assigns ownership to the platform SPEC maintainer against sub-epic #714. File: specs/platform/clock-design.md lines 1591–1608.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:270c9f0 followup:#898

MED-2 (clock-design.md lines 20-26 / SPEC §5.9 ABI-symbol claim): Two changes in commit 270c9f0. (1) Corrected the preamble claim: replaced 'introduce no new ABI symbols beyond those derivable from the existing Clock::now / wall / native_tick triplet' with 'introduce three new ABI-tracked types (FrameDelta (24 B), FrameTick (8 B), FixedStepAccumulator (24 B)) registered in §7.3; SPEC §5.9 amendment tracked in §12 [BLOCKING].' — file: specs/platform/clock-design.md lines 22–26. (2) Added §12 [BLOCKING IMPLEMENTATION] entry requiring SPEC §5.9 to be amended to add stub declarations for all three types before any plan PR consuming them can land, naming the two consumers (phase-9 frame-tick for FrameDelta, shipping runtime fixed-step gameplay for FixedStepAccumulator). File: specs/platform/clock-design.md lines 1610–1621.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:270c9f0 followup:#898

LOW-1 (clock-design.md line 1548, #TBD placeholder): Two changes in commit 270c9f0. (1) Substituted 'issue #TBD' with '(task-breakdown spike to be filed before plan PRs land)' at file line 1549. (2) Added §12 [NON-BLOCKING] entry directing that a dedicated '[SPIKE] task-breakdown-clock-implementation' issue be filed parented to #714, and once filed its number placed back into §12 and the Refs block header. File: specs/platform/clock-design.md lines 1623–1628.

cjhowe-us:
responder:go-impl-respond round:3 decision:ADDRESSED commit:270c9f0 followup:#898

All three findings from the round-3 review addressed in a single commit on branch fix/platform-clock-design-followup-r1, pushed to PR #898.

Finding-to-file mapping:

**MED-1** (clock-design.md preamble lines 4-9 + SPEC.md §6.6 contradicting regression-guard): Added §12 [BLOCKING IMPLEMENTATION] entry (specs/platform/clock-design.md lines 1591–1608) documenting that SPEC.md §6.6 prescribes a plain `assert(ns >= last)` while clock-design.md §3.2 corrects this to a monotonic-CAS loop. Entry requires SPEC §6.6 to be amended before any plan PR consuming this design can land, names phase-9 frame-tick implementation and render's frame-extract NTP-gap diagnostics as the two concrete consumers, and assigns ownership to the platform SPEC maintainer under sub-epic #714.

**MED-2** (clock-design.md preamble lines 20-26 + SPEC.md §5.9 missing ABI types): (1) Corrected preamble claim at specs/platform/clock-design.md lines 22–26: replaced the inaccurate "no new ABI symbols beyond those derivable from the existing Clock::now / wall / native_tick triplet" with "introduce three new ABI-tracked types (FrameDelta (24 B), FrameTick (8 B), FixedStepAccumulator (24 B)) registered in §7.3; SPEC §5.9 amendment tracked in §12 [BLOCKING]." (2) Added §12 [BLOCKING IMPLEMENTATION] entry (lines 1610–1621) requiring SPEC §5.9 to add stub declarations for all three types before plan PRs consuming them can land.

**LOW-1** (clock-design.md line 1548, `#TBD` placeholder): (1) Substituted `issue #TBD` → `(task-breakdown spike to be filed before plan PRs land)` at specs/platform/clock-design.md line 1549. (2) Added §12 [NON-BLOCKING] entry (lines 1623–1628) directing that a `[SPIKE] task-breakdown-clock-implementation` issue be filed parented to #714, with its number backfilled into §12 and the Refs block once filed.

Issue #721, opened 2026-05-03
[SPIKE] design-platform-file-io-detailed


cjhowe-us:
Parent: #714

## Question

Produce the detailed design for the **file-io** aggregate in the `platform` context.

Bounded async FileIo + IoToken.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/platform/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/platform/filesystem.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **file-io** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/platform/file-io-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(platform): detailed design — file-io (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-platform-file-io-detailed` is blocked by this issue.

cjhowe-us:
agent:design-platform-file-io
status:started
issue:#721
branch:docs/platform-file-io-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-ab403ffe3a5ef1c25
host:mac
cloud:none
commit:none
pr:none
notes:Started detailed design for platform file-io aggregate. Read all required sources: PHILOSOPHY.md, AGENTS.md, .github/SETUP.md, specs/_TEMPLATE.md, specs/platform/SPEC.md (§4.6 / §5.11 / §6.5 / §8.1 / §9.1 / §10.3.6 — fully filled), the six review-decision records, and harmonius/docs/requirements/platform/filesystem.md (R-14.6.1 through R-14.6.11) for re-derivation input. Style sibling: specs/core/plugin-loader-design.md. Working in isolated worktree on branch docs/platform-file-io-design.

cjhowe-us:
agent:design-platform-file-io
status:done
issue:#721
[...]

Issue #11, opened 2026-05-03
[SPIKE] task-breakdown-error-perf


cjhowe-us:
Parent: #4

## Question

Convert the error-model + perf-budget decisions into a leaf type:plan backlog.

## Deliverable

Open type:plan issues parented to Epic #2 for error-handling boilerplate + perf-budget CI gate scaffolding.

## Closure

Deliverable committed; parent sub-epic checkbox flipped.

cjhowe-us:
agent:adv-tb-err-perf
status:started
issue:#11
branch:n/a
worktree:/Users/cjhowe/Code/glibre
host:mac
cloud:none
commit:n/a
pr:#n/a-or-'none'
notes:Decomposing error-model + perf-budget decision records into a leaf type:plan backlog parented to sub-epic #4. No code changes; only OPEN type:plan issues. Targeting ~10 plans in the pts:1..pts:5 range covering Error variant set, per-context Error enum macro, std::expected alias, GLIBRE_TRY macro, spdlog log_error helper, -fno-exceptions enforcement, PerContextAllocator skeleton, perf-budget CI gate workflow, transient arena, and S1 perf fixture / Catch2 BENCHMARK scaffolding.

cjhowe-us:
agent:adv-tb-err-perf
status:done
issue:#11
branch:n/a
worktree:/Users/cjhowe/Code/glibre
host:mac
cloud:none
commit:n/a
pr:none
[...]

Issue #10, opened 2026-05-03
[SPIKE] decide-perf-budget-allocation


cjhowe-us:
Parent: #4

## Question

Which contexts justify their initial CPU/GPU/heap allocations against MVP scenarios? Where does headroom go? How is the budget enforced in CI (perf tests per context)?

## Deliverable

Decision record `reviews/decisions/perf-budget.md` with per-context CPU/GPU/heap cells summing within 16.67 ms / 60 fps; CI gate spec.

## Closure

Deliverable committed; parent sub-epic checkbox flipped.

cjhowe-us:
Deliverable on main (reviews/decisions/*.md). Closing per AGENTS.md spike-closure rule.

PR #830, opened 2026-05-04
docs(platform): detailed design — file-io (refs #721)

cjhowe-us:
## Summary

- Adds `specs/platform/file-io-design.md` — detailed design for the platform context's file-io aggregate (`FileIo` / `IoToken` / sandbox roots), refining SPEC §4.6 / §5.11 / §6.5 / §8 / §9 / §10.3.6.
- 12 sections per `specs/_TEMPLATE.md`. Style sibling: `specs/core/plugin-loader-design.md`.
- Re-derives every harmonius R-14.6.1..R-14.6.11 requirement with explicit MVP coverage / partial / refusal disposition. Tokio + Rust shape rejected wholesale; macOS-first POSIX-direct + bounded `std::thread` worker pool chosen.

Notable design decisions (each with rationale in the doc):
- **Sandbox-root admission** layered on `CanonicalPath` byte-prefix check; three named roots (`project` / `user` / `cache`); `project` read-only at runtime.
- **Brief's named arms** (`PathOutsideRoot`, `AsyncQueueFull`, etc.) map to the §4.7 closed-sum + diagnostic prefixes (`"path-outside-root"`, `"out-of-budget"`, …) — no new variants; promotion gated by second-consumer rule.
- **Hot-reload drain policy**: drain-to-terminal on platform self-reload (rationale + escalation-to-refusal gate documented as `§12 [OPEN]`).
- **No libdispatch / GCD**: bounded `std::thread` pool keeps callbacks out of the public surface (§4.6 inv #7) and ports cleanly to Linux / Windows.
- **Atomic write** is the only public write protocol: temp + fsync + rename + parent-fsync.

Refs #721. Issue stays OPEN per spike-deliverable convention.

## Test plan
- [x] All 12 spec template sections present and populated.
- [x] §10 maps every brief-named arm to existing closed-sum + prefix.
- [x] §11 enumerates Catch2 unit + integration + benchmark coverage matching §10's failure rows.
- [x] §12 carries `[OPEN]` items per design + brief; no silent drops.

### DIFF
--- a/specs/platform/file-io-design.md
+++ b/specs/platform/file-io-design.md
@@ -1,0 +1,1325 @@
+# File-IO Detailed Design
+
+> Detailed design for the `platform` context's file-io aggregate
+> (`FileIo` / `IoToken` / `OpenMode` / `Stat` / `DirEntry` / sandbox
+> roots, SPEC §4.6).
+> Refines §5.11 / §6.5 / §8.1 / §9.1 / §10.3.6 of `specs/platform/SPEC.md`
+> and the engine-wide `reviews/decisions/{error-model, perf-budget,
+> plugin-abi, hot-reload-protocol, frame-phases, fory-codegen}.md`
+> records.
+> All conclusions re-derived; harmonius prior art (`harmonius/docs/
+> requirements/platform/filesystem.md`, R-14.6.1 .. R-14.6.11) cited as
+> research input only.
+
+Refs: spike #721 — `[SPIKE] design-platform-file-io-detailed`. Parent
+#714. Sibling task-breakdown spike blocked-by this deliverable.
+
+## 1. Purpose
+
+`FileIo` is the single component in `glibre-platform` permitted to call
+the OS file-IO syscalls (`open` / `read` / `pread` / `pwrite` / `fsync`
+/ `rename` / `unlink` / `stat` / `readdir`) and to spawn the bounded
+worker thread pool that backs the async surface. Its one responsibility
+is **admitting a path-keyed byte transfer across the OS seam**: take a
+`CanonicalPath` rooted under a registered sandbox root, perform a
+synchronous read / write / stat / list / remove on the caller thread or
+queue an async equivalent onto the bounded pool, and return either a
+`Result<T>` or a single-consumer `IoToken`. On any refusal the file
+system, the in-flight ledger, and the sandbox-root registry are left
+byte-equal to the pre-call snapshot.
+
+What the aggregate explicitly refuses to own:
+
+- **Asset deserialization** — owned by `data` (`specs/data/SPEC.md`
+  §4 / §7); `FileIo` returns inert byte spans, never typed
+  aggregates. `data::deserialize<T>` calls `FileIo::read_all` /
+  `read_async` and does its own decoding.
+- **Shader cache hashing / reload** — owned by `shader`
+  (`specs/shader/SPEC.md`); the shader context calls `FileIo`'s read
+  surface and computes its own blake3 over the bytes returned.
+- **Filesystem watching** — owned by `FileWatcher` (SPEC §4.3); the
+  brief and #719 keep this entirely separate. `FileIo` has no
+  subscription state, no debounce, no FSEvents stream. A consumer who
+  wants change notifications subscribes to `FileWatcher` and reads
+  through `FileIo`.
+- **Path canonicalization beyond root-prefix admission** — the
+  `CanonicalPath` value object is owned by `platform/detail/path/`
+  (SPEC §6.1); `FileIo` takes already-canonical paths only (§4.6
+  inv #1) and adds the additional sandbox-root precondition.
+- **Asset-bundle / pak-file unpacking, content addressing, residency
+  caches** — owned by `content` (per SPEC §1's refusal list).
+- **Mid-frame I/O on the driver thread** — refused by §4.6 inv #2
+  (`FileIo` never blocks the main thread); the synchronous primitives
+  are off-main-thread-only and asserted in debug builds.
+
+The aggregate's SRP boundary is sharp: if the OS file-IO syscall set
+admitted, the `FileIoConfig` shape, the worker-pool topology, the
+atomic-write protocol, the path-root admission rule, the per-token slot
+discipline, or the queue-saturation refusal contract changes, this
+design changes. Anything else is out of scope.
+
+## 2. Requirements Coverage
+
+Mapping of harmonius requirements
+(`R-14.6.*`, `harmonius/docs/requirements/platform/filesystem.md`) to
+glibre MVP refusal-or-coverage. Every entry is independently
+re-derived; harmonius's Tokio + Rust shape is rejected wholesale and
+each underlying requirement re-evaluated against the C++23 / SDL3 /
+POSIX seam.
+
+| Harmonius req                       | Glibre disposition (MVP)            | Coverage site                                                                                   |
+|-------------------------------------|-------------------------------------|-------------------------------------------------------------------------------------------------|
+| R-14.6.1 async open/read/write, sequential + random-access, banning stdlib I/O | **Partial — sync + bounded async, no random offset (MVP)** | Sync (§4.1) + bounded async (§4.2) below; `pread` / `pwrite` deferred behind `read_all` / `write_atomic` until a second consumer needs offset I/O (§12 [OPEN] #1). `std::fstream` banned by §6.9. |
+| R-14.6.2 async create / delete / recursive mkdir / deferred-to-close delete / batch delete | **Partial — single-shot remove + atomic write covers MVP**  | `FileIo::remove` (§4.1.5); `write_atomic` synthesizes "create or overwrite" (§4.1.2). Recursive mkdir, batch delete, deferred-to-close delete deferred (§12 [OPEN] #2); not on the MVP critical path (no peer context asks). |
+| R-14.6.3 async stat (size, timestamps, permissions, type) + batch stat | **Partial — single-path stat covered, batch deferred**       | `FileIo::stat_path` returns the §5.11 `Stat` struct. Batch-stat deferred until `content`'s residency sweep needs it (§12 [OPEN] #3); MVP loops in caller. |
+| R-14.6.4 async list_dir, incremental yield, depth limits, glob filtering | **Partial — flat list with caller-supplied buffer, no depth/glob** | `FileIo::list_dir(path, span<DirEntry>)` (§4.1.4). Incremental, depth, and glob deferred to the consumer (`tools` / `content`); §4.6 inv #5 keeps the aggregate's surface to one snapshot per call. |
+| R-14.6.5 .. R-14.6.6, R-14.6.8, R-14.6.9 file watching, debounce, BLAKE3 dedup, typed event stream, hash cache | **Refused (routed to `FileWatcher`, SPEC §4.3 / #719)**                | Out of `FileIo`'s SRP. The watcher aggregate owns subscription, dedup, hash cache; `FileIo` is the byte-stream the watcher's consumers read through. |
+| R-14.6.7 canonical-path resolution across platforms (relative, symlinks, junctions, drive letters, UNC, long paths, case rules) | **Covered (delegated to `CanonicalPath`)**                  | `CanonicalPath::from_absolute` (SPEC §5.2); `FileIo` enforces canonicalization-at-the-seam by accepting only `CanonicalPath` (§4.6 inv #1). This design adds the **sandbox-root admission** layer on top (§3.2). |
+| R-14.6.10 ≥ 80 % of raw disk bandwidth                                    | **Covered (deferred budget)**                                          | Throughput budget (§9.1); CI gate fixture (§11.3) measures sustained sequential I/O against `dd`-equivalent baseline on macOS APFS. |
+| R-14.6.11 Tokio-on-Linux + threaded fallback                              | **Refused (Linux is post-MVP)**                                        | macOS-first per PHILOSOPHY §1 / SPEC §1; libdispatch is not used either (§3.4 below). The bounded-pool design ports cleanly to Linux when that work lands. |
+
+Coverage rule: every harmonius requirement above either lands in this
+design or is refused with a one-line rationale. No silent drops.
+
+Glibre-native requirements added beyond harmonius:
+
+- **Sandbox-root admission** — every `FileIo` operation accepts only
+  paths whose canonical form has one of the registered roots as a
+  prefix. The brief calls these `project` / `user` / `cache`. This
+  closes the implicit "any absolute path" hole the harmonius
+  filesystem requirement left open and makes editor- / runtime-side
+  jail discipline a property of the aggregate, not the caller. Detail
+  in §3.2.
+- **Atomic write protocol** — §4.6 inv #4 names this; this design
+  pins the `<target>.tmp.<pid>.<rand>` + `fsync(file)` + `rename` +
+  `fsync(parent_dir)` sequence as the only write protocol the
+  aggregate exposes. There is no non-atomic public write API. Detail
+  in §4.1.2.
+- **Bounded async budget enforced as refusal, not back-pressure** —
+  §4.6 inv #6 names the rule; this design pins the saturation
+  semantics: once `io_thread_budget * queue_depth` slots are in
+  flight, the next `read_async` / `write_atomic_async` returns
+  `IoFailure { OsCode { ENOBUFS } }` with diagnostic prefix
+  `"out-of-budget"` (§10.3.6 already specifies this). No queueing,
+  no allocation, no callback — back off in the caller.
+- **Single-consumer `IoToken` with two-phase abandon** — this design
+  pins the slot lifetime: `~IoToken()` stamps "abandoned"; the worker
+  delivers the result then returns the slot. A caller that drops
+  tokens without polling exhausts the budget — by design.
+
+## 3. Detailed Model
+
+### 3.1 Aggregate composition
+
+```text
+FileIo (root, owned by platform)
+├── PathRoots                  roots_       (sandbox-root registry; §3.2)
+├── eastl::array<Worker, N>    workers_     (N = FileIoConfig.io_thread_budget; §3.4)
+├── SpscRing<Request>          requests_    (engine → workers; bounded; §3.5)
+├── eastl::array<Slot, M>      slots_       (M = N * queue_depth; §3.6)
+├── SlotFreeList               free_slots_  (lock-free LIFO; §3.6)
+├── ReadAllArena               read_arena_  (per-call buffer for sync read_all; §6.9)
+└── PlatformAllocator&         alloc_       (sub-arena handle, ContextTag::platform)
+```
+
+The aggregate owns no OS file handles across calls — each operation
+opens, reads/writes, closes within a single transaction. This is the
+load-bearing simplification: there is no `OpenFile` entity, no
+descriptor table, no fd lifetime management. POSIX file descriptors are
+scoped to one `Request` or one synchronous call.
+
+`FileIo` is owned by the platform plugin's startup composer; one
+instance per process (sized by `FileIoConfig`).
+
+### 3.2 `PathRoots` (sandbox-root registry, sub-aggregate)
+
+The brief introduces three named roots: `project`, `user`, `cache`.
+Each is a `CanonicalPath` registered at `FileIo::create` time; all
+subsequent operations admit only paths whose canonical form has one of
+the registered roots as a byte-prefix.
+
+```cpp
+// platform/src/fileio/roots.hpp — internal.
+enum class RootId : std::uint8_t {
+    Project = 0,   // game / editor working tree (read-mostly)
+    User    = 1,   // per-user save / preferences (read-write)
+    Cache   = 2,   // derived / regeneratable artifacts (read-write, prunable)
+};
+
+struct PathRoots {
+    eastl::array<CanonicalPath, 3>  roots;   // dense; index = RootId
+    eastl::array<bool, 3>           writable;// project=false, user=true, cache=true
+};
+```
+
+Rules:
+
+- **Roots are registered once at construction and frozen.** Adding /
+  removing a root after `FileIo::create` is refused (the platform
+  plugin's startup composer is the only caller of root registration).
+  This is what makes root admission a `O(3)` byte-prefix scan with
+  zero allocation per call.
+- **`project` is read-only at runtime.** Every `write_atomic*` /
+  `remove` against a path whose admitted root is `project` returns
+  `PermissionDenied` with diagnostic prefix `"project-readonly"`. The
+  editor's "save scene" path goes to `user/` or `cache/` per the
+  `tools` SPEC, never overwrites the source tree. Tests, dev builds,
+  and CI cooks may construct a `FileIo` with `project` writable
+  (build-time flag, not API toggle); production runtime never.
+- **Path admission is byte-prefix on the canonical form.** A path
+  passes admission iff there exists a `RootId r` such that
+  `canonical_path.view().starts_with(roots[r].view())` AND the next
+  byte after the prefix is `'/'` or end-of-string (so `/User/foo`
+  does not pass under root `/User/foobar`). The check is a tight loop
+  over three roots; sub-microsecond.
+- **Symlinks are resolved by `CanonicalPath::from_absolute` before
+  the prefix check.** A symlink under `project/` pointing at
+  `/etc/passwd` canonicalizes to `/etc/passwd` and fails admission.
+  This is the sandbox's load-bearing rule: the prefix check happens
+  on the canonicalized form, not on the lexical path.
+- **No `..` / `.` in admitted paths.** `CanonicalPath::from_absolute`
+  already rejects non-canonical input (SPEC §5.2); this design adds
+  no separate normalization pass. If `from_absolute` accepted it,
+  it is canonical.
+- **Refusal arm.** Admission failure returns `PermissionDenied` with
+  diagnostic prefix `"path-outside-root"`. The brief's
+  `PathOutsideRoot` is the documented prefix; promoting it to a
+  first-class §4.7 arm is gated by a second consumer needing typed
+  dispatch (§12 [OPEN] #4 — same Occam's-razor logic as the
+  `SurfaceLost` / `WatcherUnavailable` carve-outs).
+
+The three names are compiled into the platform plugin's text segment
+as the `RootId` enum; the `PathRoots` struct's `roots[]` array is
+populated by the engine's startup composer from the `Process::cwd()` /
+`Process::executable_path()` / `~/Library/Application Support/<game>` /
+`~/Library/Caches/<game>` resolution table that the launcher hands
+in. The aggregate does not hard-code the OS-specific user / cache
+paths; it consumes whatever absolute paths the composer provides.
+
+### 3.3 `IoToken` (entity)
+
+```cpp
+// platform/src/fileio/token.hpp — internal projection of §5.11 IoToken.
+struct Slot {
+    std::atomic<IoToken::State>          state;        // InFlight | Ready | Cancelled
+    std::atomic<bool>                    abandoned;    // ~IoToken set this
+    eastl::span<const std::byte>         result_bytes; // filled before state=Ready (release)
+    Result<void>                         err;          // failure code if state=Ready and err is unexpected
+    Request                              req;          // bound at enqueue, read by worker
+    std::uint32_t                        next_free;    // intrusive free-list link
+};
+
+// IoToken::Impl is just &Slot; the public IoToken from §5.11 is a
+// move-only handle that holds slot index + free-list backref.
+```
+
+Token lifetime invariants (extending §4.6 inv #3):
+
+- **Single-consumer.** Only the holder of the `IoToken` may call
+  `poll` / `wait_for` / `cancel` / `take_result`. Copying is forbidden;
+  moving transfers ownership; dropping cancels the operation
+  (best-effort) and stamps `abandoned`.
+- **Two-phase abandon.** `~IoToken()` does NOT free the slot. It
+  stamps `abandoned = true` (release) and calls `cancel()` (which
+  stamps `state = Cancelled` if not yet `Ready`). The worker thread,
+  on completion, observes `abandoned` and returns the slot to
+  `free_slots_`. This is the only safe order: the worker may be
+  mid-operation when the token destructor runs, and freeing the slot
+  underneath an in-progress write would corrupt the next caller.
+- **Slot count is bounded.** `M = N * queue_depth` (default
+  `2 * 16 = 32` slots); abandoned-but-not-yet-released slots count
+  against the budget. A caller that drops `IoToken`s at full rate
+  exhausts the pool — this is by design; the saturation refusal arm
+  (§10.3) tells the caller to back off.
+- **`take_result` is single-shot.** First call after `state == Ready`
+  returns the bytes (or the error); second call returns
+  `AlreadyExists` per §5.11. The bytes' lifetime is the slot's
+  lifetime, which the next `take_result` or `~IoToken` ends — callers
+  that need to keep the bytes copy out before either.
+
+### 3.4 Worker pool topology
+
+The aggregate spins **`io_thread_budget` workers** (default 2, MVP
+ceiling 8) at `FileIo::create`. Each worker is a `std::thread` named
+`"glibre-fileio-N"` whose body is:
+
+```text
+worker_loop:
+    while (!stopping):
+        slot = requests_.pop_blocking()        # SPSC dequeue (§3.5)
+        if slot.state == Cancelled:
+            release(slot)
+            continue
+        do_work(slot)                          # POSIX call; §3.7
+        slot.state = Ready (release)
+        if slot.abandoned:
+            release(slot)                      # §3.3 two-phase abandon
+```
+
+The pool is **plain `std::thread`**, not libdispatch / GCD. Reasoning:
+
+- libdispatch's queue API is callback-shaped; §4.6 inv #7 bans
+  callbacks crossing the boundary. Wrapping libdispatch into a
+  poll-only surface re-introduces the kernel wait we deliberately
+  avoid (§6.5 SPEC: `wait_for` uses sleep with backoff, not futex /
+  mutex, to keep the slot one cache line and the API kernel-free).
+- libdispatch on macOS has a system-wide concurrency budget the
+  engine cannot constrain. Bounding our own worker count is what
+  §4.6 inv #6 demands.
+- Cross-platform: the same `std::thread` body runs on Linux and
+  Windows when those ports land. libdispatch ports do not.
+- macOS-26 / Apple Silicon `std::thread` is `pthread`-backed; spawn
+  cost is amortized over the process lifetime (workers spin once at
+  `FileIo::create`).
+
+Pool sizing knob: `FileIoConfig::io_thread_budget` (SPEC §5.11). MVP
+default of 2 covers single-asset reads + a streaming worker without
+serializing. The 8-worker ceiling is a pragmatic upper bound (8 cores
+on M1; more workers contend for the disk and inflate latency); larger
+configurations refuse at `create` with `Unsupported`.
+
+### 3.5 Request queue
+
+A bounded SPSC ring (`detail/spsc_ring.hpp` shared with `event/`) sized
+to `M` entries (one per slot). The producer is the engine driver
+thread (calling `read_async` / `write_atomic_async`); the consumer is
+the worker pool's dispatcher thread (worker 0 is also the dispatcher
+when `N=1`; for `N>1` the workers themselves act as competing consumers
+via a stronger primitive — see below).
+
+For `N=1` (no plausible MVP saturation case), the ring is plain SPSC.
+For `N>1`, the ring is **MPSC-from-the-engine + N-way work-stealing
+from workers**. Topology:
+
+```text
+engine driver thread
+        │ push(Request)        (one producer)
+        ▼
+   request_ring (N * queue_depth slots, lock-free MPSC ring)
+        │ pop(Request)
+        ▼
+worker[0] worker[1] ... worker[N-1]   (N consumers; each owns its own ring slice)
+```
+
+The MPSC primitive is the same lock-free ring used by `event/` /
+`watcher/`, with the consumer side widened to N readers via a
+per-worker head index plus a single shared tail. Saturation
+(`tail - any_head >= capacity`) returns `IoFailure` with diagnostic
+prefix `"out-of-budget"` (§10.3.6 / SPEC §4.6 inv #6) — never grows.
+
+A request is a fixed-size POD:
+
+```cpp
+struct Request {
+    enum class Op : std::uint8_t {
+        ReadAll,
+        WriteAtomic,
+    };
+    Op                            op;
+    std::uint32_t                 slot_index;     // back-ref into slots_
+    CanonicalPath                 path;           // string_view backed by interner arena
+    eastl::span<const std::byte>  bytes_in;       // Write only; nullptr for Read
+};
+```
+
+`Request::bytes_in` is **caller-owned** for the duration of the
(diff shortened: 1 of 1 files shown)


specs/platform/file-io-design.md
```diff
@@ -0,0 +1,1325 @@
+# File-IO Detailed Design
```
cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:549 problem:§4.1 create() refusal arms list 'OutOfBudget (cell exhausted; core::Error::OutOfBudget per perf-budget.md)' as a third refusal arm for FileIo::create. The platform::Error closed sum (SPEC §4.7 / platform-error-design.md §3.1) has no OutOfBudget arm. The correct carrier for an exhausted per-context allocator cell is core::Error::OutOfBudget, which is not a platform::Error — meaning the create() signature (Result<FileIo>) would need to return glibre::Error wrapping core::OutOfBudget, not a platform::Error arm. The design text is ambiguous: the bullet format implies OutOfBudget is a platform::Error refusal arm on par with Unsupported and IoFailure, but the parenthetical 'core::Error::OutOfBudget' is correct. The failure-mode table §10.2 correctly separates it ('raised by the per-context allocator … propagated; not constructed inside fileio/'), but the §4.1 arms list contradicts it. recommendation:In §4.1 create() contract block, remove OutOfBudget from the platform::Error refusal-arm bullet and replace with a cross-reference to §10.2, matching the language already used there: 'cell exhaustion at create time surfaces as core::Error::OutOfBudget propagated from the platform sub-arena allocator — not a platform::Error arm'.

cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:489 problem:§3.9 states that the diagnostic prefix is stored as 'constinit const char*' literals in 'fileio/error_prefixes.hpp' and is 'checked via byte-equal comparison at the consumer'. This contradicts platform-error-design.md §3.3, which is the canonical design for how prefixes are stored and read: the prefix lives in a TLS slot (detail::error::set_prefix / current_prefix / reset_prefix), not as a literal value bundled with the Error return. The file-io-design implies the prefix can be extracted from the returned platform::Error at the consumer by byte-comparing against a known literal, but the error-model stores the prefix in a TLS buffer that is valid only immediately after a translator call — not in the Error value itself. A consumer that receives a Result<T> from async code (worker-thread path) cannot read the TLS prefix from the calling thread at take_result() time because the prefix was stamped on the worker thread, not the consumer thread. recommendation:Replace the §3.9 description of prefix storage with the canonical model from platform-error-design.md §3.3: the prefix is stamped by the worker thread into the slot's per-slot prefix field (not TLS) or the design should explicitly acknowledge that async-path prefixes require a different carrier than the TLS model. Alternatively, add a per-Slot prefix field (const char*) to the Slot struct in §3.6 so the async worker stamps it there and the consumer reads it from the slot via take_result(). The current design conflates the synchronous (TLS-safe, same-thread) path with the async (cross-thread) path.

cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:283 problem:§3.5 opens by calling requests_ 'a bounded SPSC ring' shared with event/ but then immediately notes that for N>1 workers it becomes an MPSC-from-engine + N-way work-stealing ring. The shared primitive in detail/spsc_ring.hpp is a plain SPSC — it has a single producer head and a single consumer head. Widening to N concurrent consumers requires a distinct primitive (per-worker head index + shared tail CAS), which is not 'SPSC with the consumer side widened'. The SPEC §6.5 internal-architecture note also describes the structure as 'bounded SPSC ring' throughout, and the watcher/ module uses the same spsc_ring.hpp. This design introduces a structural requirement on spsc_ring.hpp that the watcher/ and event/ consumers do not need: multi-consumer head indices. If implemented naively by sharing the same spsc_ring.hpp template, the N>1 case will have a data race on the consumer head. recommendation:Either (a) pin the MVP at N=1 for the request ring (the default is 2 workers but they could share a single dispatcher + per-worker SPSC shard), or (b) name the N>1 primitive explicitly as a distinct MPMC/MPSC ring with per-consumer heads that lives in a different header (e.g. detail/mpsc_ring.hpp), and note that the claim 'same lock-free ring used by event/ / watcher/' is inaccurate — those use true SPSC.

cjhowe-us, line 1: severity:LOW location:specs/platform/file-io-design.md:1 problem:The design document declares OpenMode in its preamble scope listing ('FileIo / IoToken / OpenMode / Stat / DirEntry / sandbox roots') but OpenMode is never used by any of the seven methods defined in §4.1 or §4.2. The SPEC §5.11 declares 'enum class OpenMode : std::uint8_t { Read, Write, ReadWrite, Append }' as part of the FileIo surface, and §4.6 names it a value object of the aggregate. However, the design's public-surface block (§4.1) omits it from every method signature — read_all takes a CanonicalPath only, write_atomic takes a CanonicalPath + span only, and the async variants follow the same pattern. Either OpenMode is vestigial in the SPEC (the design has chosen a different API shape where the operation name encodes the mode rather than a parameter), or it should appear as a parameter to some method. The design never discusses this choice. recommendation:Add a §1 or §4.1 note explicitly acknowledging that OpenMode from SPEC §5.11 is superseded by the operation-named API (read_all / write_atomic encode the mode in the name) and that OpenMode is no longer part of the FileIo aggregate surface, or define a usage for OpenMode if one is intended (e.g. for the deferred pread/pwrite surface). Without this, an implementer reading SPEC §5.11 will see OpenMode and not know why the design omits it.

cjhowe-us, line 1: severity:LOW location:specs/platform/file-io-design.md:606 problem:§4.1 'read_all(p)' documents: 'Returns a borrowed span; lifetime is until the next read_all from the same FileIo instance (per §6.9 SPEC).' But the Slot struct in §3.6 (used by async workers) also carries eastl::span<const std::byte> result_bytes filled into the same per-FileIo read_arena_. The §12 [OPEN] #6 notes that async workers need their own buffers but admits the allocation site is not pinned. If both synchronous read_all and an async worker's do_work both land in the same read_arena_, a synchronous read_all call made while an async read is completing would corrupt the async result_bytes span — or the async path needs its own arena. The design leaves the arena ownership for the async path unresolved, but the §4.1 contract paragraph for read_all does not caveat this. recommendation:Either (a) state explicitly in §4.1 read_all that the read_arena_ is exclusive to synchronous paths and async workers use per-slot buffers (and add a Slot buffer field to §3.6, sizing it appropriately against the §9 8 MiB sub-arena), or (b) add a sentence to §4.1 read_all noting the async-buffer ambiguity is deferred to the §12 [OPEN] #6 resolution. As written, the contract paragraph overpromises the read_arena_ lifetime invariant.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:3
  LOW:2
carryover:0

notes:The design is structurally sound and well-scoped. The 12-section template is fully populated, every harmonius requirement (R-14.6.1 .. R-14.6.11) is explicitly covered or refused with rationale, and the SRP boundary is sharp. All public APIs return Result<T> (= std::expected<T, glibre::Error>), noexcept is applied consistently, EASTL types are used throughout (eastl::array, eastl::span, eastl::string_view), std:: is retained only for permitted language utilities (std::atomic, std::thread, std::expected, std::chrono), and POD-only spans cross the ABI boundary. The two-concrete-users rule is satisfied: data::deserialize<T> and shader's cache path are named explicitly as callers, and the §12 defers all new API surface until a second consumer arrives. The Conventional Commit subject 'docs(platform): detailed design — file-io (refs #721)' is correct and the issue reference is present.

Three medium findings require follow-up in the implementation plan:

MED-1 (lines 549-550): The §4.1 create() refusal-arm list presents OutOfBudget as a platform::Error arm on par with Unsupported and IoFailure, but platform::Error has no OutOfBudget arm (platform-error-design.md §3.1). The parenthetical 'core::Error::OutOfBudget' is correct but contradicts the bullet structure. §10.2 correctly routes this as a core::Error; §4.1 must match.

MED-2 (line 489): §3.9 states that diagnostic prefixes are stored as constinit const char* literals in fileio/error_prefixes.hpp and are checked 'via byte-equal comparison at the consumer'. The canonical model from platform-error-design.md §3.3 stores the prefix in a TLS slot (set_prefix / current_prefix / reset_prefix). This is safe for synchronous callers (same thread), but the async path (worker stamps prefix, consumer reads from a different thread via take_result()) cannot safely read TLS set on the worker thread. The Slot struct (§3.6) has no prefix field. The design conflates the two paths without resolving the cross-thread prefix transport.

MED-3 (line 283): §3.5 describes requests_ as 'a bounded SPSC ring (detail/spsc_ring.hpp shared with event/)' and then immediately promotes it to an MPSC ring with N-way work-stealing for N>1 workers. A plain SPSC ring has a single consumer head; N concurrent workers CAS-ing the same head is a data race. The claim that this is 'the same lock-free ring used by event/ / watcher/' is incorrect for N>1. This must be resolved before an implementer touches spsc_ring.hpp.

LOW-1 (line 1): OpenMode is listed in the design's scope preamble and is declared in SPEC §5.11 as a value object of the aggregate, but no method in §4.1 or §4.2 takes or returns it. The operation-named API (read_all / write_atomic) encodes the mode implicitly, which is a valid design choice, but it needs to be made explicit — either deprecating OpenMode in this design or reserving it for the deferred pread/pwrite surface (§12 [OPEN] #1).

LOW-2 (line 606): §4.1 read_all promises the result span lifetime is 'until the next read_all from the same FileIo instance (per §6.9 SPEC)' but §12 [OPEN] #6 admits that async workers have no pinned buffer allocation site. If workers share the same read_arena_ as synchronous read_all, concurrent synchronous + async reads corrupt each other. The §4.1 contract paragraph should caveat this or explicitly state that the synchronous arena is exclusive to the synchronous path.

No HIGH findings. The design is ready for the implementation plan phase after the three MED items are resolved as amendments to this document or via clarifying notes in the follow-up task spike. The §12 [OPEN] items are actionable and appropriately deferred.

specs/platform/file-io-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:48498c4ef34a4d81b86ffe8ae6e0a55524c3fc3a followup:#896

MED-1: Struck `OutOfBudget` from the `platform::Error` refusal-arm list in §4.1 `create()`. Replaced with: "Saturation surfaces as `core::Error::OutOfBudget` at the core seam, not as a `platform::Error` arm — see §10.2 for the routing." The §10.2 existing prose (which already correctly described this) remains canonical.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:48498c4ef34a4d81b86ffe8ae6e0a55524c3fc3a followup:#896

MED-2: Rewrote §3.9 to document prefix transport by path: (a) sync path — translator runs on caller thread, TLS prefix read before returning the unexpected, no cross-thread transport needed; (b) async path — worker captures current_prefix() into new Slot::prefix field after translator runs and before state=Ready release-store; consumer reads slot.prefix acquire-ordered via the state transition. Updated Slot struct in §3.6 with const char* prefix field and ordering annotation. Added §12 [OPEN] #10 [BLOCKING IMPLEMENTATION] gating implementation on canonical prefix-literal symbol set in fileio/error_prefixes.hpp.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:48498c4ef34a4d81b86ffe8ae6e0a55524c3fc3a followup:#896

MED-3: Rewrote §3.5 ring section. N==1: plain SPSC ring (detail/spsc_ring.hpp), same primitive as event/ and watcher/. N>1: separate MPSC ring (detail/mpsc_ring.hpp) with per-consumer head index — SPSC cannot safely serve N consumers (data race on shared head). Struck the false 'same lock-free ring shared with event/ / watcher/' claim for N>1. Updated §6.1 rule 6 to match. Added §12 [OPEN] #11 [NON-BLOCKING] for N>1 MPSC ring contract, work-stealing protocol, and fairness guarantees; noted N>1 enablement is post-MVP per plans/mvp.md. MVP N=2 default uses SPSC with dispatcher semantics.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:48498c4ef34a4d81b86ffe8ae6e0a55524c3fc3a followup:#896

LOW-1: Added OpenMode encoding paragraph after §1 preamble explaining that API choice encodes mode implicitly (read_all/read_async imply read+sequential; write_atomic/write_atomic_async imply create-or-replace+write). Noted OpenMode enum is retained in SPEC §5.11 for future API extensions (e.g. open_with_mode), tagged §12 [OPEN] #9 [NON-BLOCKING].

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:48498c4ef34a4d81b86ffe8ae6e0a55524c3fc3a followup:#896

LOW-2: Tightened read_all span lifetime contract to 'valid until the NEXT call to read_all ON THE SAME THREAD'; added explanation that read_arena_ is not designed for concurrent recycling and concurrent read_all on different threads is not supported. Added explicit note that async path workers allocate from per-worker arenas (workerN_arena_) not from the caller's read_arena_, and async-completion buffer lifetime is governed by take_result() reclaim per Slot. Updated §12 [OPEN] #7 to acknowledge per-worker arena ownership stays per-worker.

cjhowe-us, line 1: severity:HIGH carryover from round 1 location:specs/platform/file-io-design.md:549 problem:MED-1 from R1 is unresolved on main. The create() refusal-arm bullet at line 549 still reads "OutOfBudget (cell exhausted; core::Error::OutOfBudget per perf-budget.md)" as if OutOfBudget is a platform::Error arm on par with Unsupported and IoFailure. The platform::Error closed sum (SPEC §4.7) has no OutOfBudget arm; the parenthetical correctly names core::Error::OutOfBudget but the surrounding bullet structure is internally contradictory and misleads implementors. PR #896 contains the fix but is not yet merged; on main the document is unrepaired. fix:Merge PR #896 to main. The fix is: replace lines 549-550 with "Saturation surfaces as core::Error::OutOfBudget at the core seam, not as a platform::Error arm — see §10.2 for the routing." so the bullet is unambiguous.

cjhowe-us, line 1: severity:HIGH carryover from round 1 location:specs/platform/file-io-design.md:283 problem:MED-3 from R1 is unresolved on main. §3.5 still states "For N>1, the ring is MPSC-from-the-engine + N-way work-stealing from workers" and claims "The MPSC primitive is the same lock-free ring used by event/ / watcher/" — both claims are false and architecturally harmful. The SPSC ring in event/ / watcher/ has one consumer head; widening it to N consumers requires a distinct primitive (per-worker head + shared tail CAS). Calling it "the same ring" misstates the SRP boundary of spsc_ring.hpp and would lead an implementor to introduce a data race. Additionally, the aggregate-composition diagram at §3.1 (line 118) names the field SpscRing<Request> requests_ but §3.5 contradicts this for N>1. The MVP default is N=2 and the comment "N=1 (no plausible MVP saturation case)" contradicts the FileIoConfig default of io_thread_budget{2} — the MVP ships N=2, not N=1. PR #896 contains the fix (N==1 SPSC, N>1 separate mpsc_ring, MVP N=2 uses SPSC with dispatcher semantics) but is not yet merged. fix:Merge PR #896. Additionally verify that the "dispatcher semantics" for MVP N=2 SPSC is explicitly documented: if worker 0 is the sole consumer and dispatches to worker 1 via a secondary channel, that channel shape must be stated.

cjhowe-us, line 1: severity:HIGH carryover from round 1 location:specs/platform/file-io-design.md:489 problem:MED-2 from R1 is unresolved on main. §3.9 still describes diagnostic prefixes as "constinit const char* literals in fileio/error_prefixes.hpp checked via byte-equal comparison at the consumer" but says nothing about how the async worker threads transport the TLS prefix to the consumer thread. platform-error-design.md §6.4 states explicitly that for FileIo the producer copies the prefix literal pointer into the SPSC slot alongside the Error and the consumer reads it from the slot. This cross-thread transport protocol is absent from file-io-design.md §3.9. The Slot struct in §3.6 has no prefix field, meaning the async transport path is entirely undocumented in the design that governs the struct layout. PR #896 fixes this by adding const char* prefix to Slot and documenting the sync/async transport paths in §3.9, but is not yet merged. fix:Merge PR #896. Verify after merge that §3.6 Slot layout breakdown (line 350-352) is updated to account for the added prefix pointer in the cache-line size calculation (see separate MED finding).

cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:350 problem:The Slot cache-line size claim is broken by the MED-2 fix. §3.6 at line 350-352 asserts "Slot size is one cache line (64 B on M1): atomic<State> (1 B) + atomic<bool> (1 B) + 6 B pad + span (16 B) + Result<void> (~16 B) + Request (~24 B)" — summing to 64 B. The Slot struct already includes next_free (uint32_t, 4 B), which is not in this breakdown, bringing the real total to >=68 B before any fix. When PR #896 adds const char* prefix (8 B pointer on 64-bit), the Slot grows by 8 B further. The slot-fits-one-cache-line invariant (cited at §6.3, §9.2, §12 last paragraph, and SPEC §6.5 "keep the slot one cache line and the API kernel-free") is violated: the struct cannot fit in 64 B without reordering/padding reduction or trimming one field. The pool footprint calculation ("64 * 32 = 2 KiB") is also wrong. fix:Either (a) add a static_assert(sizeof(Slot) <= 64) to the design-spec stub and adjust the layout so the assert holds (e.g. pack state+abandoned into one byte, or make next_free a uint16_t, or omit prefix from Slot and store it in a parallel array), or (b) explicitly accept a 2-cache-line Slot and revise the invariant prose. Either way, the §3.6 arithmetic must match the actual struct members including next_free and prefix.

cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:487 problem:Prefix literal duplication across design boundaries. §3.9 says each prefix lives as a "constinit const char* literal in fileio/error_prefixes.hpp". platform-error-design.md §3.3 already declares the canonical set of prefix literals in the namespace glibre::platform::detail::error::prefix (prefix::none, prefix::out_of_budget, prefix::io, prefix::resource, etc.). If fileio/error_prefixes.hpp re-declares new literals (e.g. prefix::file_io_out_of_budget or re-declares prefix::out_of_budget = "out-of-budget") it creates two sources of truth for what is supposed to be a pointer-equality-checked registry. If it re-exports the same pointers from platform-error-design.md §3.3, it is an unnecessary indirection layer with no SRP justification (one responsibility: "know the prefix literal set" belongs to the translator module, per §3.3 of platform-error-design.md). The two-concrete-users test (PHILOSOPHY anti-pattern rule) has not been applied to the fileio/ prefix header: there is only one consumer (FileIo). fix:Either (a) drop fileio/error_prefixes.hpp and have the FileIo translator call detail::error::current_prefix() / set_prefix() against the platform-error-design.md §3.3 registry directly, or (b) document explicitly in §3.9 that fileio/error_prefixes.hpp only re-exports named aliases for the §3.3 pointers (no new string literals) and cite the pointer-equality guarantee as depending on that aliasing. The former is simpler and more cohesive.

cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:660 problem:§5 Hot/Cold path split references "The MPSC request ring's producer side (engine) is touched only at read_async / write_atomic_async" at line 660. This is inconsistent with the #896 fix that changes the MVP N=2 path back to SPSC. If the ring is SPSC (MVP), the sentence should not name it MPSC. More broadly, §5 Hot/Cold path split refers throughout to "MPSC ring" while §3.5 says N>1 uses MPSC and N==1 uses SPSC. For a spike document, the hot-path analysis should state which ring type governs the documented cost model: for MVP (N=2, SPSC with dispatcher), the SPSC producer push cost is what matters. fix:After merging #896, audit all MPSC references in §5, §6, and §9 that apply to the MVP path and replace with "SPSC ring" or "request ring" to avoid locking the cost model to the wrong primitive.

cjhowe-us, line 1: severity:LOW location:specs/platform/file-io-design.md:118 problem:§3.1 aggregate-composition diagram names the request ring field "SpscRing<Request> requests_" but §3.5 describes it as MPSC for N>1. The field type in the diagram is statically wrong for half the runtime configurations (any N>1). A reviewer reading the diagram first would form a wrong mental model before reaching §3.5. fix:After merging #896, update the §3.1 diagram to use a configuration-conditional annotation, e.g. "SpscRing<Request> | MpscRing<Request> requests_ (type depends on N, see §3.5)" or simply "RequestRing<Request> requests_" (the abstract type) to keep the diagram honest without over-specifying.

cjhowe-us, line 1: severity:LOW location:specs/platform/file-io-design.md:728 problem:§6.1 threading rule 6 "No mutexes inside fileio/" still says "the MPSC ring is the lock-free primitive shared with event/ / watcher/" in the current on-disk text. This is the same false claim MED-3 identifies at §3.5, repeated here. Both the shared-ring claim and the MPSC label are wrong for the SPSC MVP path. fix:After merging #896 (which updates §6.1 rule 6 in the diff), verify the merge correctly removes the shared-with-event/watcher claim from this paragraph.

cjhowe-us (commented):

findings:
  HIGH:3
  MED:3
  LOW:2
carryover:3 (all three R1 MEDs escalated to HIGH because #896 is unmerged on main)

notes:The SRP boundary of FileIo is correct and sharp: one reason to change (OS file-IO syscall contract shifted). Every public API returns Result<T> (= std::expected<T, glibre::Error>) noexcept, EASTL types are used correctly throughout (eastl::array, eastl::span, eastl::string_view; std:: retained for std::atomic, std::thread, std::expected, std::chrono per PHILOSOPHY §11), and POD-only types cross the plugin ABI boundary. The cohesion story within the aggregate (PathRoots + workers + ring + slots + free-list) is justified and internally consistent. The hot/cold path split is thoughtful and the arena ownership narrative is clear. What blocks: all three R1 MEDs are escalated to HIGH this round because PR #896 (the fix branch) is still OPEN and the current main carries the unrepaired contradictions (OutOfBudget in platform::Error bullet, no Slot::prefix field, SPSC/MPSC false-shared-ring claim). Two new MEDs are raised in this round: (1) the Slot struct's cache-line size claim at §3.6 is wrong even before the prefix field is added (next_free is unaccounted in the arithmetic) and becomes definitively wrong after the prefix pointer lands — the "slot-fits-one-cache-line" invariant cited in §6.3, §9.2, and §12 is at risk; (2) fileio/error_prefixes.hpp introduces a second source-of-truth for prefix literals that the platform-error-design.md §3.3 registry already owns — pointer-equality checks that span both headers require a single origin, which should be the §3.3 registry. A MED is also raised for stale MPSC ring references in §5 hot-path analysis that do not reflect the post-#896 MVP SPSC path. Recommended action: merge #896 first, then open a follow-up fix for the Slot cache-line arithmetic and the prefix-header ownership question in the same pass. Both issues are in the implementation-gate territory per §12 [OPEN] #10.

cjhowe-us (commented):

findings:
  HIGH:3
  MED:3
  LOW:2
carryover:3 (all three R1 MEDs escalated to HIGH because #896 is unmerged on main)

notes:The SRP boundary of FileIo is correct and sharp: one reason to change (OS file-IO syscall contract shifted). Every public API returns Result<T> (= std::expected<T, glibre::Error>) noexcept, EASTL types are used correctly throughout (eastl::array, eastl::span, eastl::string_view; std:: retained for std::atomic, std::thread, std::expected, std::chrono per PHILOSOPHY §11), and POD-only types cross the plugin ABI boundary. The cohesion story within the aggregate (PathRoots + workers + ring + slots + free-list) is justified and internally consistent. The hot/cold path split is thoughtful and the arena ownership narrative is clear. What blocks: all three R1 MEDs are escalated to HIGH this round because PR #896 (the fix branch) is still OPEN and the current main carries the unrepaired contradictions (OutOfBudget in platform::Error bullet, no Slot::prefix field, SPSC/MPSC false-shared-ring claim). Two new MEDs are raised in this round: (1) the Slot struct cache-line size claim at §3.6 is wrong even before the prefix field is added — next_free is unaccounted in the arithmetic — and becomes definitively wrong after the prefix pointer lands; the slot-fits-one-cache-line invariant cited in §6.3, §9.2, and §12 is at risk; (2) fileio/error_prefixes.hpp introduces a second source-of-truth for prefix literals that the platform-error-design.md §3.3 registry already owns — pointer-equality checks that span both headers require a single origin. A MED is also raised for stale MPSC ring references in §5 hot-path analysis that do not reflect the post-#896 MVP SPSC path. Recommended action: merge #896 first, then open a follow-up fix for the Slot cache-line arithmetic and the prefix-header ownership question.

specs/platform/file-io-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:NOOP commit:48498c4 followup:#896

Carryover from R1 — already addressed in PR #896 commit 48498c4 (R1 fix). §4.1 OutOfBudget routing corrected: saturation surfaces as `core::Error::OutOfBudget` at the core seam, not as a `platform::Error` arm. Visible on main only after #896 merges.

cjhowe-us: responder:go-impl-respond round:2 decision:NOOP commit:48498c4 followup:#896

Carryover from R1 — already addressed in PR #896 commit 48498c4 (R1 fix). §3.5 SPSC/MPSC split corrected: N==1 uses `detail/spsc_ring.hpp`, N>1 uses a separate `detail/mpsc_ring.hpp` instance. Visible on main only after #896 merges.

cjhowe-us: responder:go-impl-respond round:2 decision:NOOP commit:48498c4 followup:#896

Carryover from R1 — already addressed in PR #896 commit 48498c4 (R1 fix). §3.9 async prefix transport corrected: worker captures `current_prefix()` into `prefixes_[slot_index]` before `state=Ready` release-store; consumer reads after acquire-load. Visible on main only after #896 merges.

cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:e0bb326 followup:#896

MED-1 (§3.6 Slot cache-line): Applied option (c) — prefix stored in parallel `eastl::array<const char*, M> prefixes_` (not inline in Slot). Slot layout: state(1B)+abandoned(1B)+next_free(2B as uint16_t)+4B pad+result_bytes(16B)+err(~16B)+req(~24B)=64B. Added explicit per-field layout table with total; added `static_assert(sizeof(Slot)<=64)` gate. Dropped `slot_index` from `Request` (redundant, saves 4B+3B pad). Updated §3.1 composition diagram to include `prefixes_` member. next_free as uint16_t documents the ≤65535 constraint (well above max M=1024).

cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:e0bb326 followup:#896

MED-2 (§3.9 prefix literal duplication): Dropped `fileio/error_prefixes.hpp` entirely. §3.9 now imports directly from `glibre::platform::detail::error::prefix::` (platform-error-design.md §3.3) — single source of truth; pointer-equality guarantee derives from that origin. Added file-io-specific prefix constants (`path_outside_root`, `project_readonly`, `read_too_large`) to §3.3 registry in platform-error-design.md. Added prefix-constant table in §3.9 with source annotations. Updated §12 [OPEN] #10 to reference parallel-array design and §3.3 import.

cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:e0bb326 followup:#896

MED-3 (§5 MPSC in hot/cold path): Substituted 'MPSC request ring' → 'SPSC request ring (engine, MVP N==1/2 dispatcher mode)' in §5 hot-path invariants. Added note pointing to §12 [OPEN] #11 for post-MVP MPSC variant cost analysis.

cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:e0bb326 followup:#896

LOW-1 (§3.1 composition diagram SpscRing): Renamed diagram entry to `RequestRing<Request> requests_` with inline annotation: 'SPSC for N==1 MVP, MPSC for N>1 post-MVP — see §3.5; bounded'. This is the typedef-alias approach — the alias `RequestRing` refers to the ring type conditional on N as documented in §3.5.

cjhowe-us: responder:go-impl-respond round:2 decision:NOOP commit:48498c4 followup:#896

LOW-2 (§6.1 rule 6): Verified current §6.1 rule 6 text on the #896 branch (commit 48498c4): 'for N == 1 the request ring is `detail/spsc_ring.hpp` (same primitive as `event/` / `watcher/`); for N > 1 the ring is a separate `detail/mpsc_ring.hpp` instance (§3.5)'. This is already correct — R1 fixed it. No further action needed.

specs/platform/file-io-design.md
cjhowe-us, line 1: severity:MED location:specs/platform/file-io-design.md:410 problem:§3.6 contains two verbatim duplicate paragraph blocks introduced by commit e0bb326. The paragraphs starting "prefix for async-path error transport is kept in a parallel array prefixes_[M]" and "Pool footprint: 64 × 32 + 256 = 2.25 KiB" and "next_free as uint16_t limits the free-list" each appear twice consecutively in §3.6 — once after the Slot layout table and once again at the bottom of the section. The first occurrence (lines ~410–425) is the canonical set; the second occurrence is a verbatim copy added when e0bb326 reorganized the layout table. No new information is added by the duplicate block. recommendation:Delete the second occurrence of those three paragraphs (lines ~426–447 in the branch file, beginning with "prefix for async-path error transport is kept in a parallel array"). Keep only the first occurrence immediately after the layout table where it is contextually anchored.

cjhowe-us, line 1: severity:HIGH location:specs/platform/file-io-design.md:654 problem:§4.1 prose documents that FileIoConfig gains five new fields (project_root, user_root, cache_root, read_arena_bytes, queue_depth) beyond the single io_thread_budget field in SPEC §5.11. The preamble states the design "refines §5.11" and §4.1 acknowledges the extension inline. However §12 carries no amendment entry for this SPEC §5.11 divergence. The extended FileIoConfig shape is structurally load-bearing: an implementer who reads only SPEC §5.11 (which the design calls "authoritative") will construct a FileIoConfig with one field and miss the required root-registration surface entirely — there is no other public API to register sandbox roots. The §12 amendment tracking rule (per SPEC template convention) requires an explicit amendment note whenever a detailed design widens the SPEC §5 public surface. fix:Add a §12 [OPEN] entry (or an explicit §1 / §4.1 amendment box) stating: "SPEC §5.11 FileIoConfig is amended by this design to add project_root / user_root / cache_root (CanonicalPath), read_arena_bytes (uint32_t, default 5 MiB), and queue_depth (uint8_t, default 16). The SPEC §5.11 stub must be updated to match before implementation begins." Tag it [BLOCKING IMPLEMENTATION] since implementations compiled against the SPEC stub will produce a wrong API.

cjhowe-us, line 1: severity:LOW location:specs/platform/file-io-design.md:223 problem:The static_assert comment cites "§6.3 / §9.2 / §12" but SPEC.md has no §6.3 subsection dedicated to cache-line discipline — the SPEC §6.3 header is a watcher section ("FileWatcher — SDL3 FSEvents on macOS, abstracted for the rest"). The §9.2 reference is also ambiguous; SPEC §9.2 does not exist as a numbered sub-section in the SPEC (§9 is "Non-Functional Requirements + Perf Budget" with no numbered sub-sections). The static_assert itself is correct and necessary; only the cross-reference annotation is stale. recommendation:Replace the comment citation with the correct references: "Slot must fit one cache line (file-io-design.md §3.6 layout table / §9 perf budget)". This keeps the assert unambiguous for the implementer who reads it.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:1
  LOW:3
carryover:0

notes:Round 1 and Round 2 findings are fully resolved. The SPSC/MPSC split, OutOfBudget routing, async-prefix parallel-array transport, Slot 64B layout with static_assert gate, and OpenMode/arena clarifications are all correctly addressed in commits 48498c4 and e0bb326. Self-consistency check: §3.1 RequestRing alias, §3.5 SPSC for N=1, §3.6 static_assert gate, §3.9 parallel prefixes_, §5 SPSC cost note, §6.1 rule 6 are all aligned. The three new prefix symbols (path_outside_root, project_readonly, read_too_large) are correctly added to platform-error-design.md §3.3 and their string values match §10.1 failure-mode table rows. No conflict detected with the already-merged §3.3 block from #895.

One HIGH blocking finding: §12 carries no amendment entry for the FileIoConfig SPEC §5.11 divergence. The design extends FileIoConfig with five new fields beyond the single io_thread_budget declared in SPEC §5.11 — this is a public-API surface change that an implementer reading only SPEC §5.11 would miss entirely, since there is no other public root-registration surface. §12 must record a [BLOCKING IMPLEMENTATION] amendment for this widening so SPEC §5.11 is updated before the implementation plan begins.

MED finding: §3.6 contains verbatim duplicate paragraph blocks (three paragraphs on prefixes_, pool footprint, and next_free rationale each appear twice consecutively). The first occurrence immediately after the Slot layout table is canonical; the second is a copy from the e0bb326 reorganization and adds no new information.

Three LOW findings: (1) static_assert comment cites non-existent SPEC §6.3 and §9.2 sub-sections; (2) three new prefix constant entries in platform-error-design.md §3.3 lack the consumer citation the registry's own gate requires; (3) SPEC §6.5 still says "a single bounded SPSC ring" for all N — now stale given the N=1/N>1 topology in §3.5.

Issue #721 label and parent are correct (type:spike, phase:mvp, domain:platform; Parent: #714). Issue stays OPEN per spike-deliverable convention. Conventional Commit subjects across both #830 (docs(platform): detailed design) and #896 commits (docs(platform): r1 followup, docs(platform): address #830 R2 review) are correctly typed and scoped. The R2 commit headline is truncated by the subject-line limit but the body is complete.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:3
carryover:0
notes:All R1+R2 findings resolved. The SPSC/MPSC split, Slot 64B layout, static_assert gate, parallel prefixes_ transport, OutOfBudget routing, OpenMode/arena clarifications are correctly addressed. New prefix symbols (path_outside_root, project_readonly, read_too_large) are correctly added to platform-error-design.md §3.3 with no conflict against the §895-merged block. Self-consistency across §3.1/§3.5/§3.6/§3.9/§5/§6.1 is confirmed. HIGH-1: §12 carries no amendment entry for the FileIoConfig SPEC §5.11 widening — design adds five new fields (project_root, user_root, cache_root, read_arena_bytes, queue_depth) beyond the SPEC stub's single io_thread_budget; an implementer reading SPEC §5.11 alone gets the wrong API. MED-1: §3.6 has verbatim duplicate paragraphs (prefixes_, pool footprint, next_free rationale each appear twice). LOW-1: static_assert cites non-existent SPEC §6.3 / §9.2 sub-section refs. LOW-2: three new prefix constants in platform-error-design.md §3.3 lack consumer citations required by §3.3 gate. LOW-3: SPEC §6.5 still says single-SPSC-ring-for-all-N, now stale. Issue #721 type:spike / phase:mvp / domain:platform / parent #714 correct. CC subjects correct across all three #896 commits.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:3
carryover:0
notes:All R1+R2 findings resolved. SPSC/MPSC split, Slot 64B layout, static_assert gate, parallel prefixes_ transport, OutOfBudget routing, OpenMode/arena clarifications correctly addressed. New prefix symbols (path_outside_root, project_readonly, read_too_large) added to platform-error-design.md §3.3 with no conflict vs #895-merged block. Self-consistency confirmed across §3.1/§3.5/§3.6/§3.9/§5/§6.1. HIGH-1 (file-io-design.md §4.1 / §12): FileIoConfig is widened by five fields beyond SPEC §5.11 stub but §12 has no amendment entry — an implementer reading SPEC §5.11 alone gets the wrong API signature; this is a BLOCKING IMPLEMENTATION gap. MED-1 (§3.6): verbatim duplicate paragraphs (prefixes_, pool footprint, next_free rationale each appear twice consecutively). LOW-1 (§3.3 code block): static_assert cites non-existent SPEC §6.3 / §9.2 sub-section numbers. LOW-2 (platform-error-design.md §3.3): three new prefix constants lack per-registry consumer citations required by §3.3 gate. LOW-3 (§12): SPEC §6.5 SPSC-only description is now stale vs §3.5 SPSC/MPSC topology and should be tracked as an amendment. Issue #721 type:spike / phase:mvp / domain:platform / parent #714 correct. CC subjects correct across all #896 commits.

specs/platform/file-io-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:47e9e20 followup:#896

Deleted the second occurrence of each duplicate paragraph in §3.6 (prefix rationale, pool footprint, next_free uint16_t rationale). Each phrase now appears exactly once. Verified with grep.

cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:47e9e20 followup:#896

Added §12 [BLOCKING IMPLEMENTATION] entry: 'FileIoConfig SPEC §5.11 widening — project_root (CanonicalPath), user_root (CanonicalPath), cache_root (CanonicalPath), read_arena_bytes (std::size_t), queue_depth (uint32_t) beyond SPEC §5.11's io_thread_budget.' Cites two concrete consumers: editor sandbox bootstrap (project_root for writable scope), shipping runtime (cache_root for shader cache directory). Gates first envelope.cpp implementation PR on SPEC §5.11 amendment spike landing. Also added §12 [NON-BLOCKING] entry for SPEC §6.5 stale SPSC-only language (LOW-3 finding addressed in the same commit).

cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:47e9e20 followup:#896

Verified actual SPEC sections: SPEC §6.3 does not exist; SPEC §9.2 is the per-aggregate heap ceiling table (does not directly mandate the 64B cache-line constraint). The canonical authority is the design's own §3.6 layout table (which explicitly shows 64B total and gives the rationale). SPEC §6.5 is the FileIo internal-architecture sketch and SPEC §9.2 is a valid cross-reference for the sub-arena boundary.

Updated both the static_assert comment (line ~223) and the §3.6 layout table 'Total' row:
- Before: 'Slot must fit one cache line (§6.3 / §9.2 / §12)'
- After: 'Slot must fit one cache line (§3.6 layout table; SPEC §6.5 / §9.2)'

cjhowe-us:
responder:go-impl-respond round:3 decision:ADDRESSED commit:47e9e20 followup:#896

Addressing the two review-body-only findings from R3:

**LOW-2** (platform-error-design.md §3.3 prefix-registry consumer citations):
Added per-constant Consumer comment lines for the three new prefix constants. Exact edits in commit 47e9e20:
- `specs/platform/platform-error-design.md` line ~312: `path_outside_root = "path-outside-root"; // Consumer: FileIo translator (path-validation gate)`
- `specs/platform/platform-error-design.md` line ~313: `project_readonly = "project-readonly"; // Consumer: FileIo translator (sandbox writable-scope gate)`
- `specs/platform/platform-error-design.md` line ~314: `read_too_large = "read-too-large"; // Consumer: FileIo read_all (size cap on read_arena)`

**LOW-3** (file-io-design.md §12 SPEC §6.5 stale SPSC-only language):
Added §12 [NON-BLOCKING] entry in `specs/platform/file-io-design.md` (end of §12, commit 47e9e20): 'SPEC §6.5 internal-architecture sketch SPSC-only language is stale post-§3.5 N==1 SPSC / N>1 MPSC topology split. §3.5 is the canonical authority. SPEC §6.5 amendment can land alongside the FileIoConfig amendment from the HIGH-1 BLOCKING entry or as a separate doc-only patch — not blocking the implementation PR.'

Issue #727, opened 2026-05-03
[SPIKE] design-platform-platform-error-detailed


cjhowe-us:
Parent: #714

## Question

Produce the detailed design for the **platform-error** aggregate in the `platform` context.

Closed sum + OS error translation.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/platform/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/platform/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **platform-error** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/platform/platform-error-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(platform): detailed design — platform-error (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-platform-platform-error-detailed` is blocked by this issue.

cjhowe-us:
agent:design-platform-platform-error
status:done
issue:#727
branch:docs/platform-platform-error-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a9fa99f032f6a239a
host:mac
cloud:none
commit:9f8ffc8067cb5a3332fd76b8f0b25feb324374d8
pr:#831
[...]

PR #831, opened 2026-05-04
docs(platform): detailed design — platform-error (refs #727)

cjhowe-us:
## Summary

- Adds `specs/platform/platform-error-design.md` — 12-section detailed design for the platform-error aggregate (closed `platform::Error` sum + OS translators).
- Refines `specs/platform/SPEC.md` §4.7, §5.1, §6.10, §10.1, §10.2, §10.6 in place; introduces no new public surface.
- Documents the four translators (`errno_to_error`, `sdl_to_error`, `mach_to_error`, `ns_to_error`), the thread-local diagnostic prefix registry, signal-safe construction, the `PlatformErrorRecord` Fory wire schema, and the hot/cold split that keeps error construction off the success path.

## Closure

Refs spike #727. Spike issue stays OPEN per AGENTS.md spike workflow; sibling task-breakdown spike unblocked.

## Test plan

- [ ] Reviewer confirms all 12 template sections present (Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions).
- [ ] Reviewer confirms no widening of public surface beyond SPEC §5.1.
- [ ] Reviewer confirms cited decision records (`error-model.md`, `perf-budget.md`, `frame-phases.md`, `plugin-abi.md`, `hot-reload-protocol.md`, `fory-codegen.md`) are referenced consistently.
- [ ] Reviewer confirms harmonius citations are research-input-only (per PHILOSOPHY §"How harmonius is used").

### DIFF
--- a/specs/platform/platform-error-design.md
+++ b/specs/platform/platform-error-design.md
@@ -1,0 +1,1022 @@
+# platform — Detailed Design: platform-error aggregate
+
+> Detailed design for the closed sum `platform::Error` plus the
+> POSIX `errno` / SDL3 `SDL_GetError()` / mach `kern_return_t` /
+> AppKit `NSException` translation seam declared in
+> `specs/platform/SPEC.md` §4.7, §5.1, §6.10, §10.1, §10.2, §10.6.
+> Refines those sections in place; cites
+> `reviews/decisions/error-model.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/frame-phases.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`, and
+> `reviews/decisions/fory-codegen.md`. Introduces no new public
+> surface beyond the §5.1 stub already locked into
+> `specs/platform/SPEC.md`; deviations from the cited records would
+> require an amendment spike, not an in-place edit.
+>
+> Refs: spike #727 — `[SPIKE] design-platform-platform-error-detailed`.
+> Parent #714. Sibling task-breakdown spike is blocked by this
+> deliverable. All conclusions re-derived; harmonius prior art
+> (`harmonius/docs/requirements/platform/os-integration.md` R-14.2.8,
+> `harmonius/docs/requirements/platform/filesystem.md`,
+> `harmonius/docs/requirements/platform/threading-async.md`) is
+> research input only.
+
+## 1. Purpose
+
+The platform-error aggregate is the **single translation seam** between
+the four OS error sources the engine is allowed to call (POSIX
+`errno`, SDL3 `SDL_GetError()`, mach `kern_return_t`, AppKit
+`NSException`) and the closed sum `platform::Error` returned across
+every platform-aggregate boundary via `std::expected<T, glibre::Error>`
+(per `reviews/decisions/error-model.md` and SPEC §4.7). Its one
+responsibility — the only reason it changes — is **"the OS reports
+failures in a different shape than glibre's closed sum models"**: a
+new SDL3 release renames a prose prefix, an Apple SDK revision adds
+an `NSException` class, the POSIX errno table grows a code with a
+new family. Concretely the aggregate owns:
+
+1. The closed `platform::Error` variant (§5.1 of SPEC; restated in §3
+   below) and the rule that adding / removing / renaming a variant is
+   a deliberate central edit gated by the second-consumer trigger
+   from SPEC §10.8.
+2. The four translator free functions in `engine/platform/src/detail/error/`:
+   `errno_to_error(int)`, `sdl_to_error(const char*)`,
+   `mach_to_error(kern_return_t)`, and `ns_to_error(NSException*)`
+   (the last one is the only function in the engine permitted to
+   *touch* an `NSException*`, callable only from `bridge.mm` per
+   SPEC §6.2).
+3. The thread-local diagnostic prefix buffer (`error::tls::detail`)
+   that callers use to discriminate `IoFailure { OsCode }` payloads
+   without parsing prose at the public boundary (per SPEC §6.10,
+   §10.4, §10.5).
+4. The translation-table source-of-truth: each translator's row set
+   is the test-fixture goldens lives at `tests/platform/error/` and
+   is the single audit point a reviewer reads to verify "no integer
+   leaks across the §5 surface".
+5. The Fory schema for the *log/replay carrier* — a small,
+   stable-on-the-wire `(tag, os_code, prefix)` triple that
+   `glibre::log_error` and the e2e replay harness use to round-trip
+   a `platform::Error` through `spdlog`'s structured field set
+   without losing variant identity.
+
+What this aggregate explicitly **refuses to own**:
+
+- **`core::Error`.** `PluginAbiHashMismatch`, `PluginInitFailed`,
+  `SchemaMigrationFailed`, `HotReloadRefused`, `FramePhaseMisordered`,
+  `OutOfBudget` are owned by `core` per
+  `reviews/decisions/error-model.md` Type Sketch and
+  `specs/core/SPEC.md` §10. A platform-side translator that fabricated
+  a `core::Error` would violate composition rule 1 ("per-context
+  enums are leaves; nothing nests another context's enum inside its
+  own").
+- **`render::Error` / `physics::Error` / `data::Error`.** Same rule.
+  Render maps `platform::Error::IoFailure` with prefix
+  `"surface-lost"` to `render::Error::DeviceLost` at the call site
+  that crosses the boundary (per SPEC §10.4 and error-model.md
+  composition rule 2). The platform translator never reaches into
+  render's enum; render's mapping function is render's source code.
+- **The choice of `magic_enum` vs hand-written `to_string`.** Owned
+  by `core/error.hpp` per `error-model.md` Open Q #1; this design
+  consumes whatever core decides without forking. The only platform-
+  side requirement is that the chosen approach round-trips every
+  variant name (§7).
+- **Per-aggregate failure semantics.** Which arms `Window::open` may
+  return, the recovery contract for `IoFailure { OsCode }` with
+  prefix `"surface-lost"`, and the polling fallback for
+  `"watcher-unavailable"` are owned by the sibling designs
+  (`window-surface-design.md`, `file-watcher-design.md`) and SPEC
+  §10.3 / §10.4 / §10.5. This aggregate produces the typed arm; the
+  sibling consumes it.
+- **The `glibre::log_error` formatter.** Owned by `core`. Platform
+  contributes the structured payload (§7) and the variant name
+  table; core's helper does the spdlog dispatch.
+- **Cross-aggregate retry / chain.** Each platform aggregate
+  constructs its own errors at the site they happen (SPEC §4.7
+  inv #3); this aggregate does not auto-translate one aggregate's
+  arm into another's. A `FileWatcher` call that internally invokes
+  `FileIo::stat_path` for canonical-path normalisation and gets
+  back `NotFound` returns `NotFound` to *its* caller, but the
+  translator was not the chain — the call site was.
+
+The aggregate's SRP boundary is the sharpest in the platform context:
+if SDL3 stops returning prose and starts returning a stable integer
+code, if Apple deprecates `NSException` in favour of an opaque
+`NSError*` pointer, if mach grows a new `kern_return_t` family the
+file-watcher trips, or if the `IoFailure` payload widens beyond
+`OsCode`, this design changes. Anything else is out of scope.
+
+## 2. Requirements coverage
+
+Mapping of harmonius requirements consulted as research input
+(`PHILOSOPHY.md` §"How harmonius is used" — re-derived, not ported)
+onto the MVP coverage in this aggregate. Every entry is independently
+re-derived; coverage does not imply harmonius's design was correct,
+only that the underlying *requirement* survives the re-derivation.
+
+| Harmonius source clause                                                                       | Re-derived MVP requirement                                                                                                            | Coverage in this aggregate                                                                                                                                                                                  |
+|-----------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `os-integration.md` R-14.2.8 (structured errors with platform-specific codes mapped to engine error types) | Every public platform call returns a typed-arm sum, never a raw OS integer; the OS code survives only inside `IoFailure { OsCode }`. | §3 closed variant + §4.1 translator surface; §5 hot/cold split (raw codes never appear on the success path).                                                                                                |
+| `os-integration.md` R-14.2.8 (rationale: distinguish transient vs permanent failures)         | The translation table classifies `EAGAIN` / `EWOULDBLOCK` (transient I/O) under `IoFailure` with prefix `"again"` and `EROFS` / `ENOSPC` (permanent) under separate prefixes. | §3 prefix table + §4.1 errno rows; §10 test fixtures verify both classifications.                                                                                                                          |
+| `os-integration.md` R-14.2.8 (verification: simulate clipboard lock + assert typed error)     | Translator tests are golden mappings, one row per OS code, runnable without an OS. Glibre refuses clipboard from MVP (§3 SPEC refusals). | §11 unit tests: every translator row drives a fixture call; failure case routes to the `OsCode` raw-fallback arm.                                                                                          |
+| `filesystem.md` (structured failure for `read` / `write` / `stat` / `list`)                   | POSIX `errno` is the dominant source for FileIo and FileWatcher backends; one translator covers both.                                | §3 errno → arm table (the cell already in SPEC §10.2.1, restated for the design audit); §11 cross-aggregate uniformity test.                                                                                |
+| `threading-async.md` `GraphError::Cycle` / `AccessConflict` / `MissingDependency` (compile-time graph errors) | Refused from platform: the task-graph context owns those; platform does not host scheduling. Recorded as a refusal so a future re-read does not relitigate. | §1 Refusals (`core::Error` + sibling-context errors); no platform translator row.                                                                                                                          |
+| `crash-reporting.md` (signal capture path, structured logs)                                   | Crash-dump capture is a platform `Process::install_signal` concern (SPEC §4.5); the dump *format* is an `obs` context concern. The error aggregate's only contribution to crash paths is the signal-safe `OsCode` carrier. | §6 (signal-safe construction): `Error` is POD by construction; constructing it in a SIGSEGV handler is allowed because no allocation, no virtual dispatch, no TLS write occurs on the construction path. |
+| `windowing.md` (surface-lost recovery)                                                        | `IoFailure { OsCode }` with prefix `"surface-lost"` per SPEC §10.4. Render branches on the prefix; the typed arm stays narrow.        | §3 prefix registry; §11 prefix round-trip fixture.                                                                                                                                                          |
+
+Refusals routed elsewhere (research input from harmonius that
+**does not** belong in this aggregate, recorded so the boundary is
+re-traceable):
+
+- **Per-domain enum families** (`ClipboardError`, `DialogError`,
+  `NotificationError`). Refused — clipboard / dialog / notification
+  are post-MVP and live in a future `os-integration` plugin that
+  will, if it lands, add its own arm to `glibre::Error` per
+  `error-model.md` composition rule 1.
+- **Auto-retry policies** (harmonius implied transient errors should
+  retry inside the engine). Refused — caller-domain decision per SPEC
+  §10.1; the only retry the platform performs is the bounded `EINTR`
+  retry inside the I/O thread (SPEC §10.2.1 last paragraph).
+- **Localised error messages.** Refused — the prose lives in a TLS
+  diagnostic buffer for telemetry only; the engine never branches on
+  the message and never localises it. UI-side localisation is a
+  `tools` / future-`l10n` concern.
+
+## 3. Detailed model
+
+### 3.1 The closed `platform::Error` variant
+
+Restated from SPEC §4.7 / §5.1 for the design audit; this section
+adds the *rationale* per arm and the *invariants* that govern
+addition / removal / renaming.
+
+```cpp
+// engine/platform/include/glibre/platform/error.hpp — exposed via
+// the §5 single-header projection in specs/platform/SPEC.md.
+namespace glibre::platform {
+
+struct OsCode {
+    std::int32_t value{0};   // platform-native errno / kern_return_t /
+                             // NSError code / SDL3 numeric tag.
+    constexpr bool operator==(const OsCode&) const noexcept = default;
+};
+
+// Zero-sized tag structs. Each constexpr-default-constructible, all
+// trivially copyable. Adding a payload to one is a SPEC §4.7 amendment
+// spike, not an in-place edit.
+struct NotFound          { constexpr bool operator==(const NotFound&)         const noexcept = default; };
+struct PermissionDenied  { constexpr bool operator==(const PermissionDenied&) const noexcept = default; };
+struct AlreadyExists     { constexpr bool operator==(const AlreadyExists&)    const noexcept = default; };
+struct Interrupted       { constexpr bool operator==(const Interrupted&)      const noexcept = default; };
+struct Unsupported       { constexpr bool operator==(const Unsupported&)      const noexcept = default; };
+
+// Single-payload arm. The OsCode value is opaque to callers; only
+// the diagnostic prefix (see §3.3) is load-bearing for branching.
+struct IoFailure {
+    OsCode code{};
+    constexpr bool operator==(const IoFailure&) const noexcept = default;
+};
+
+// Raw fallback. Constructed only by the four translators (§4.1) when
+// no row matched. Engine code never constructs OsCode directly as a
+// public arm; it surfaces via translator output only.
+
+using Error = eastl::variant<
+    NotFound,
+    PermissionDenied,
+    AlreadyExists,
+    Interrupted,
+    Unsupported,
+    IoFailure,
+    OsCode>;
+
+}  // namespace glibre::platform
+```
+
+Per-arm rationale (consumed-once table; SPEC §10.1 has the
+trigger / recovery / severity per arm — this table justifies
+*existence*, not behaviour):
+
+| Arm                  | Reason for existence                                                                                                                   | Promotion / demotion gate                                                                                                                                                                                                                        |
+|----------------------|----------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `NotFound`           | Most common branch in callers (asset lookup, stale handle dereference); promoting it from the prefix-discriminated `IoFailure` saves the prose-parse. | Frozen.                                                                                                                                                                                                                                        |
+| `PermissionDenied`   | Operator-actionable; the only entitlement-class arm worth branching on.                                                                | Frozen.                                                                                                                                                                                                                                        |
+| `AlreadyExists`      | Programming-error class; CI promotes to build failure (SPEC §10.1). Distinct severity from `IoFailure` justifies its own arm.         | Frozen.                                                                                                                                                                                                                                        |
+| `Interrupted`        | EINTR-class; routine on processes that trap SIGINT. Distinct from `IoFailure` because severity is `debug`, not `error`.               | Frozen.                                                                                                                                                                                                                                        |
+| `Unsupported`        | Capability-gap signal (release-build assert path, hardware feature absence). Distinct from `IoFailure` because callers route around the gap rather than retry. | Frozen.                                                                                                                                                                                                                                        |
+| `IoFailure { OsCode }` | Carrier for the long tail of OS errors that classify as I/O-class but do not justify a separate arm; carries the diagnostic prefix discriminator (§3.3). | Promotion path: when a *second* consumer needs to discriminate on a prefix without parsing TLS prose, the prefix becomes a first-class arm. SPEC §10.8 lists `SurfaceLost` / `WatcherUnavailable` as the two pending candidates. |
+| `OsCode`             | Catch-all when no translator row matched; preserves the raw value for telemetry without losing the typed-sum shape.                   | Recurring `OsCode` appearances in telemetry are the trigger for adding a new translator row (§10) — not for adding a new top-level arm.                                                                                                       |
+
+The seven arms are the closed sum. Adding an eighth is a deliberate
+central edit (SPEC §4.7 inv #1) requiring: (a) a SPEC §4.7 amendment
+spike, (b) an ABI hash bump (§7), (c) a hot-reload barrier-time
+re-validation (§8), and (d) every existing handler's `eastl::visit`
+audited for missing-arm coverage. The `[[nodiscard]]` discipline on
+every public-surface return (SPEC §5) and the closed-sum rule
+together make missing-arm handling a compile error.
+
+### 3.2 Translator surface (free functions, not methods)
+
+Four free functions in `engine/platform/src/detail/error/`, one per
+OS source. They are free functions — not member functions of a
+"translator class" — to make the SRP boundary mechanical: each
+function depends on exactly one OS header, has zero state, and is
+trivially callable from the signal-safe path (§6.2).
+
+```cpp
+// engine/platform/src/detail/error/translators.hpp
+//
+// Internal — platform-context-only. Engine code outside platform
+// never includes this header; the public §5 surface is the only
+// way platform errors cross a boundary.
+namespace glibre::platform::detail::error {
+
+// 3.2.1 POSIX errno → platform::Error
+//   Source: <errno.h>. Pure function. Used by FileIo, FileWatcher
+//   (kqueue / inotify path), Process (signal install / argv parse).
+[[nodiscard]] auto errno_to_error(int posix_errno) noexcept
+    -> Error;
+
+// 3.2.2 SDL3 SDL_GetError() → platform::Error
+//   Source: <SDL3/SDL_error.h>. Reads SDL's TLS prose, parses for
+//   known prefixes (table in §4.1 below), stamps the diagnostic
+//   prefix into platform's TLS prefix buffer (§3.3) on its way to
+//   the translated arm. The msg pointer is borrowed (SDL owns it);
+//   the function copies any prose it stashes.
+[[nodiscard]] auto sdl_to_error(const char* sdl_msg) noexcept
+    -> Error;
+
+// 3.2.3 mach kern_return_t → platform::Error
+//   Source: <mach/kern_return.h>. Used by the bridge.mm AppKit
+//   layer (CALayer / CAMetalLayer kernel-side errors that surface
+//   as kern_return_t before NSException wraps them) and by
+//   FSEvents-stream-level failures the kernel reports directly.
+//   Rare on the success path; documented for completeness.
+[[nodiscard]] auto mach_to_error(int kern_return) noexcept
+    -> Error;
+
+// 3.2.4 AppKit NSException → platform::Error
+//   Source: <Foundation/NSException.h>. Only callable from
+//   bridge.mm (the lone Objective-C++ TU per SPEC §6.2); the
+//   header signature uses void* to keep this header pure C++23
+//   and so a cross-TU include never drags Foundation into engine
+//   code. The void* is reinterpret_cast to NSException* inside
+//   the bridge implementation.
+[[nodiscard]] auto ns_to_error(void* ns_exception) noexcept
+    -> Error;
+
+}  // namespace glibre::platform::detail::error
+```
+
+The four functions are the **only** places in the entire engine where
+a raw OS error integer / pointer exists at the call boundary; no
+caller sees the integer, no `if (errno == ...)` ladder appears
+outside these files, no `try { ... } @catch (NSException*)` block
+appears outside `bridge.mm`. The SPEC §6.10 "errors are constructed
+at the site they happen" rule is made physical by this signature
+discipline.
+
+### 3.3 Diagnostic prefix registry (TLS, not part of the variant)
+
+Several arms (`IoFailure { OsCode }`, the catch-all `OsCode`) are
+intentionally coarse so the closed sum stays small (§3.1 promotion
+gate). When a caller needs to discriminate — e.g. render branching
+on `"surface-lost"` (SPEC §10.4), the hot-reload coordinator
+branching on `"watcher-unavailable"` (SPEC §10.5), or telemetry
+classifying I/O failures — it reads a *thread-local diagnostic
+prefix* the translator stamps just before returning. The prefix is
+**not** part of `platform::Error`'s value identity; two `IoFailure`
+values with `OsCode { 0 }` compare equal regardless of prefix.
+
+```cpp
+// engine/platform/src/detail/error/diagnostic.hpp
+namespace glibre::platform::detail::error {
+
+// Fixed-capacity stable-string registry. Every prefix below is a
+// compile-time literal; the TLS slot stores a `const char*` to one
+// of these literals, never an owned string.
+namespace prefix {
+inline constexpr const char* none                = "";
+inline constexpr const char* again               = "again";
+inline constexpr const char* io                  = "io";
+inline constexpr const char* resource            = "resource";
+inline constexpr const char* surface_lost        = "surface-lost";
+inline constexpr const char* watcher_unavailable = "watcher-unavailable";
+inline constexpr const char* queue_full          = "queue-full";
+inline constexpr const char* out_of_budget       = "out-of-budget";
+inline constexpr const char* sdl_empty           = "sdl-empty";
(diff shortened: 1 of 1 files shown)


specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+
+For `glibre::log_error` to emit a structured spdlog record, and for
+the e2e replay harness (`specs/e2e/SPEC.md`) to capture a
+`platform::Error` into the trace stream, a small Fory schema
+serialises the *log carrier* — not the error value itself. The
+carrier's purpose is single-shot logging / replay, never
+reconstruction of a live `Error` (the variant is rebuilt only for
+replay assertions, never returned to engine code).
+
+```fory
```
cjhowe-us, line 660: severity:HIGH location:specs/platform/platform-error-design.md:660 problem:The design introduces `data/schemas/platform/PlatformErrorRecord.fory` and claims it "contributes to `glibre_types_abi_hash()` via `glibre-types.dylib`" (§7.3). This directly contradicts two explicit statements in `specs/platform/SPEC.md` §7 that this PR does not amend: (1) SPEC §7.1 table row for `PlatformError` reads "Persistent? No — A by-value error returned through `std::expected`; never crosses a save boundary. Loggers stringify; they do not Fory-encode." (2) SPEC §7 net-effect statement reads "the platform context therefore ships zero `data/schemas/platform/` files for the MVP" and "The net effect of §7: the platform context contributes nothing to `AbiHash`." The PR body lists §4.7, §5.1, §6.10, §10.1, §10.2, §10.6 as the refined SPEC sections but does not list §7, so SPEC §7 remains unchanged and the contradiction is live. Introducing a Fory schema in the design without amending SPEC §7 leaves the two authoritative documents in direct conflict: implementors reading SPEC §7 will create zero schemas; implementors reading this design will create one. The ABI hash accounting disagreement is especially dangerous — if `PlatformErrorRecord` is added to `glibre_types_abi_hash()` without SPEC §7 being updated, the platform-plugin ABI contract has changed silently. recommendation:Either (a) amend SPEC §7 in a follow-up spike that explicitly acknowledges the schema addition and rewrites the "zero schemas" and "contributes nothing to AbiHash" lines, or (b) demote `PlatformErrorRecord` from a Fory schema to a plain C++ struct that `glibre::log_error` serialises to spdlog structured fields without Fory codegen, consistent with SPEC §7 as written. If (a), add a §12 open-question tracking the SPEC §7 amendment gate. The amendment requires its own PR so reviewers can audit the ABI hash impact in isolation.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+- Adding a field bumps the schema version and adds a one-step
+  migrate function (pure, allocator-arena-only).
+- The schema lives at `data/schemas/platform/PlatformErrorRecord.fory`
+  and contributes to `glibre_types_abi_hash()` via `glibre-types.dylib`.
+
+### 7.3 ABI hash contribution
+
+The `PlatformErrorRecord` schema is the **only** platform-context
+contribution to `glibre_types_abi_hash()` from the error aggregate
+(SPEC §7's "platform contributes nothing to AbiHash" line is an
```
cjhowe-us, line 693: severity:HIGH location:specs/platform/platform-error-design.md:693 problem:Design §7.3 states "The schema lives at `data/schemas/platform/PlatformErrorRecord.fory` and contributes to `glibre_types_abi_hash()` via `glibre-types.dylib`." However SPEC §7 net-effect is: "the platform context contributes nothing to `AbiHash` (`specs/data/SPEC.md` §4.4) — its schema-source-hash list is empty — so a platform-only edit can never be the cause of a plugin ABI-hash mismatch." Importantly, design §7.3 itself acknowledges this contradiction in the very next sentence: "...once log replay lands the ledger gains this one entry" — but it frames a spec-breaking change as if it were a future footnote rather than an immediate design discrepancy requiring a SPEC §7 amendment. The two authoritative documents now disagree on whether the platform context contributes to `glibre_types_abi_hash()`. Any team member who implements log replay and adds `PlatformErrorRecord.fory` will change ABI gating behaviour contrary to what SPEC §7 says. recommendation:This finding is a duplicate of the §7.2 comment above and should be resolved together. The design must either (a) explicitly mark the SPEC §7 net-effect line as superseded and open a follow-up spike to update SPEC §7 in a separate PR, or (b) keep SPEC §7 accurate by not introducing a Fory schema. Do not embed a SPEC amendment inside a design-only PR.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+```cpp
+// example: render maps surface-lost to DeviceLost at its boundary
+glibre::Result<Frame> begin_frame(Window& w) noexcept {
+    return w.surface()
+        .transform([](Surface s) noexcept { return Frame{std::move(s)}; })
+        .or_else([](glibre::Error e) noexcept -> glibre::Result<Frame> {
+            // Composition rule 2: map at the boundary that crosses
+            // contexts. Render branches on platform's TLS prefix to
+            // discriminate IoFailure variants.
+            if (auto* p = std::get_if<platform::Error>(&e.code());
```
cjhowe-us, line 483: severity:MED location:specs/platform/platform-error-design.md:483 problem:The composition example uses `std::get_if` to inspect an `eastl::variant` (both `glibre::Error::code()` which returns `const eastl::variant&`, and `platform::Error` which is itself `eastl::variant<...>`). `std::get_if` is not defined for `eastl::variant` — `eastl::variant` provides its own `eastl::get_if` free function. The example on lines 483-484 would fail to compile if taken verbatim because `std::get_if<platform::Error>(&e.code())` calls the standard-library overload against an EASTL type that has no matching specialisation. Although the example is explicitly marked "illustrative", design documents in this project are used as the primary implementor reference; misleading code is a concrete correctness hazard. PHILOSOPHY §11 mandates `eastl::` for all variant / optional / container types. recommendation:Replace both `std::get_if` calls with `eastl::get_if` in the example. Specifically: `if (auto* p = eastl::get_if<platform::Error>(&e.code()); p && eastl::get_if<platform::IoFailure>(p) && ...)`. Add a one-line comment noting that `eastl::get_if` is the required accessor for `eastl::variant`.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+namespace glibre::platform::detail::error {
+
+// Fixed-capacity stable-string registry. Every prefix below is a
+// compile-time literal; the TLS slot stores a `const char*` to one
+// of these literals, never an owned string.
+namespace prefix {
+inline constexpr const char* none                = "";
+inline constexpr const char* again               = "again";
+inline constexpr const char* io                  = "io";
+inline constexpr const char* resource            = "resource";
```
cjhowe-us, line 303: severity:MED location:specs/platform/platform-error-design.md:303 problem:SPEC §9.2 (line 2109 in SPEC.md) uses `platform::Error::OutOfBudget` as if it were a named enumerator in the closed sum — `std::unexpected{platform::Error::OutOfBudget}` — but the closed sum declared in design §3.1 and confirmed in SPEC §5.1 has exactly seven arms: NotFound, PermissionDenied, AlreadyExists, Interrupted, Unsupported, IoFailure, OsCode. There is no OutOfBudget arm. The design §3.3 correctly models budget saturation as IoFailure { OsCode } with prefix "out-of-budget", and SPEC §10.3.6 is also consistent with this. However, the SPEC §9.2 text reference to platform::Error::OutOfBudget is a pre-existing SPEC inconsistency that this design needed to surface. The design's audit scope includes SPEC §4.7, §5.1, §6.10, §10.1, §10.2, §10.6 but not §9.2; the SPEC §9.2 discrepancy is not mentioned in §12 Open Questions. An implementor reading SPEC §9.2 in isolation would attempt to add an 8th arm, violating the closed-sum invariant. recommendation:Add an entry to design §12 Open Questions noting that SPEC §9.2 references platform::Error::OutOfBudget as if it is a named arm, but the closed sum does not include it; the correct form is IoFailure { OsCode } with prefix "out-of-budget" per §3.3 and SPEC §10.3.6. Track an amendment spike to correct SPEC §9.2 before the first implementation plan runs against this design.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+    -> Error;
+
+// 3.2.3 mach kern_return_t → platform::Error
+//   Source: <mach/kern_return.h>. Used by the bridge.mm AppKit
+//   layer (CALayer / CAMetalLayer kernel-side errors that surface
+//   as kern_return_t before NSException wraps them) and by
+//   FSEvents-stream-level failures the kernel reports directly.
+//   Rare on the success path; documented for completeness.
+[[nodiscard]] auto mach_to_error(int kern_return) noexcept
+    -> Error;
```
cjhowe-us, line 257: severity:LOW location:specs/platform/platform-error-design.md:257 problem:The translator signature `mach_to_error(int kern_return)` uses a plain `int` instead of `kern_return_t`. While `kern_return_t` resolves to `int` on macOS/ARM64, this is an undocumented platform assumption baked silently into the design. The accompanying comment says "Source: <mach/kern_return.h>" but the signature avoids including that header by substituting `int`. The design text refers to the OS type as `kern_return_t` throughout (§4.1.3 table, §6.10 seam description, §1 ownership list), creating a textual inconsistency with the declared parameter type. If a future macOS SDK redefines `kern_return_t` (unlikely but not impossible), or if a reviewer cross-references the macOS SDK header, the mismatch is confusing. recommendation:Either (a) use `kern_return_t` in the signature and acknowledge the header dependency with a comment ("Including <mach/kern_return.h> is acceptable in this internal-only header because no engine code outside the platform detail/ tree includes it"), or (b) keep `int` but add an explicit static_assert inside the .cpp file: `static_assert(sizeof(kern_return_t) == sizeof(int) && alignof(kern_return_t) == alignof(int), "kern_return_t layout assumption")`. Either approach eliminates the silent assumption.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0
notes:The design is an impressively thorough 1022-line spike deliverable. The SRP statement (one reason to change — OS error codec) is coherent and tightly enforced. The closed seven-arm sum matches SPEC §5.1 exactly. The four translators are correctly scoped as internal free functions with no public surface. The hot/cold split, TLS prefix registry, signal-safe construction, per-arm rationale table, and §12 open-question discipline all meet or exceed the expected spike standard. The test plan (§11) enumerates concrete Catch2 paths including compile-fail tests, and the harmonius research is correctly marked as input-only throughout. Two HIGH findings block implementation work starting from this design without remediation. HIGH-1 (lines 660, 693): the design introduces data/schemas/platform/PlatformErrorRecord.fory and claims it contributes to glibre_types_abi_hash(), directly contradicting SPEC §7's "platform ships zero .fory schemas" and "platform contributes nothing to AbiHash" statements, neither of which was amended by this PR. Until SPEC §7 is updated or the schema is removed, the two authoritative documents disagree on the platform context's ABI hash contribution — a dangerous inconsistency for implementors and for the plugin loader's hash-mismatch gate. HIGH-2 (line 483): the composition example uses std::get_if on eastl::variant types, which does not compile; the correct accessor is eastl::get_if per PHILOSOPHY §11. Although marked "illustrative," wrong code in a design document is a direct hazard for implementors using the example as a reference. MED-1 (line 303): SPEC §9.2 refers to platform::Error::OutOfBudget as a named enumerator, but the closed sum has no such arm; the design does not surface this pre-existing inconsistency in §12. MED-2 is the same as HIGH-2 seen from a correctness lens. LOW-1 (line 257): mach_to_error uses int instead of kern_return_t without explicit justification; a static_assert or a comment is sufficient. PR is already merged; remediation takes the form of a follow-up spike to (a) open an amendment PR for SPEC §7, and (b) correct the eastl::get_if example and the kern_return_t notation in a patch PR against this design doc. The SPEC §9.2 OutOfBudget inconsistency should be tracked in §12 of this design via the same patch PR.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0
notes:The design is an impressively thorough 1022-line spike deliverable. The SRP statement is coherent and tightly enforced. The closed seven-arm sum matches SPEC §5.1 exactly. The four translators are correctly scoped as internal free functions with no public surface. The hot/cold split, TLS prefix registry, signal-safe construction, per-arm rationale table, and §12 open-question discipline all meet or exceed the expected spike standard. The test plan (§11) enumerates concrete Catch2 paths including compile-fail tests, and the harmonius research is correctly marked as input-only throughout. Two HIGH findings block implementation work starting from this design without remediation. HIGH-1 (lines 660, 693): the design introduces data/schemas/platform/PlatformErrorRecord.fory and claims it contributes to glibre_types_abi_hash(), directly contradicting SPEC §7 statements that were not amended by this PR. HIGH-2 (line 483): the composition example uses std::get_if on eastl::variant which does not compile; eastl::get_if is required. MED-1 (line 303): SPEC §9.2 refers to platform::Error::OutOfBudget as a named enumerator not in the closed sum; design §12 does not surface this. LOW-1 (line 257): mach_to_error uses int instead of kern_return_t without explicit justification. PR is already merged; remediation is a follow-up spike to amend SPEC §7 and a patch PR to correct the design doc.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:a30b5de followup:#895

Lines 660/§7.2: Demoted `PlatformErrorRecord` from a Fory schema to a plain C++ struct used only as a transient spdlog-field carrier. The entire §7.2 Fory schema block and §7.3 ABI-hash-contribution section were removed and replaced with a single §7.2 paragraph stating that platform error logging uses spdlog structured fields directly, consistent with SPEC §7's 'platform contributes zero schemas' and 'contributes nothing to AbiHash' invariants. A `[RESOLVED]` entry was added to §12 documenting the conflict and its resolution by demotion rather than SPEC amendment. §8.1 table, §8.2 barrier step, and §8.3 survives/discards list were updated to remove all PlatformErrorRecord-schema references.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:a30b5de followup:#895

Line 693/§7.3: This finding was resolved together with comment 3188451345 (both HIGH-1). The §7.3 ABI-hash-contribution section was fully removed; the text that claimed SPEC §7's 'contributes nothing to AbiHash' was an 'MVP-scope statement' superseded by log replay is gone. §7 now cleanly states zero schemas, zero ABI hash contribution, consistent with SPEC §7 as written. The framing of a spec-breaking change as a future footnote no longer appears in the document.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:a30b5de followup:#895

Line 483: Both `std::get_if` calls in the composition example were replaced with `eastl::get_if` — specifically `eastl::get_if<platform::Error>(&e.code())` and `eastl::get_if<platform::IoFailure>(p)`. An inline comment was added above the if-statement explaining that `eastl::get_if` is the required accessor for `eastl::variant` per PHILOSOPHY §11 and that `std::get_if` is not specialised for `eastl::variant` and will not compile.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:a30b5de followup:#895

Line 303 (§12): Added an `[OPEN]` entry to §12 tracking the SPEC §9.2 stale enumerator. The entry documents that SPEC §9.2 references `platform::Error::OutOfBudget` as if it were a named arm, but the closed sum (§3.1, SPEC §5.1) has exactly seven arms with no `OutOfBudget`; the correct form is `IoFailure { OsCode }` with prefix `"out-of-budget"` per §3.3 and SPEC §10.3.6. The entry is marked as blocking the first implementation PR — SPEC §9.2 must be amended before any plan runs against this design. SPEC.md itself was not edited in this follow-up as the finding directed; only the §12 tracking entry was added.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:a30b5de followup:#895

Line 257: Added a comment line immediately above the `mach_to_error` signature: `// int (not kern_return_t) avoids <mach/mach.h> in the public header; static_assert(sizeof(kern_return_t) == sizeof(int)) lives in the .cpp.` The `int` parameter type is retained as it keeps the public header free of Mach framework dependency; the static_assert documents the layout assumption that makes the substitution safe.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+   SPEC §6.2).
+3. The thread-local diagnostic prefix buffer (`error::tls::detail`)
+   that callers use to discriminate `IoFailure { OsCode }` payloads
+   without parsing prose at the public boundary (per SPEC §6.10,
+   §10.4, §10.5).
+4. The translation-table source-of-truth: each translator's row set
+   is the test-fixture goldens lives at `tests/platform/error/` and
+   is the single audit point a reviewer reads to verify "no integer
+   leaks across the §5 surface".
+5. The Fory schema for the *log/replay carrier* — a small,
```
cjhowe-us, line 58: severity:HIGH location:specs/platform/platform-error-design.md:58 problem:§1 Purpose bullet 5 still claims this aggregate owns "The Fory schema for the *log/replay carrier*" — but the R1 fix on #895 demoted PlatformErrorRecord from a Fory schema to a plain C++ spdlog-field carrier (§7.2 patched). After #895 merges, §1 and §7.2 will be direct contradictions in the same document: §1 says the aggregate owns a Fory schema; §7.2 says it does not and never will. An implementor reading §1 to understand what the aggregate owns will conclude there is a Fory schema to author under data/schemas/platform/ — the exact anti-pattern the R1 fix was intended to eliminate. fix:Replace §1 bullet 5 with a statement consistent with §7.2: "5. The plain-C++ spdlog-field carrier (`PlatformErrorRecord` struct) that `glibre::log_error` uses to emit key-value fields for a `platform::Error` — arm name, `os_code` when present, TLS prefix, source tag, monotonic timestamp, thread ID. No Fory schema; no ABI hash contribution. (§7.2)" This is a doc-only one-line change in #895 scope.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+`Window::open` with a zero-dimension `WindowDesc`, etc.), unwraps
+the returned `glibre::Error`, and asserts the platform arm is one
+of the seven variants. No fixture may surface a `core::Error` /
+`render::Error` / etc. through a platform call.
+
+### 11.3 Integration tests (against e2e replay)
+
+Live under `tests/e2e/platform/error_round_trip/`. Constructs a
+known sequence of `platform::Error` values (one per arm, one per
+prefix), serialises each through `PlatformErrorRecord`, captures
```
cjhowe-us, line 941: severity:HIGH location:specs/platform/platform-error-design.md:941 problem:§11.3 integration test description still references Fory serialization ("serialises each through `PlatformErrorRecord`") after the R1 demotion. After #895 merges, the integration test plan in §11.3 describes a test that cannot exist: there is no Fory schema for PlatformErrorRecord, no glibre-types.dylib entry, and no deserialize<PlatformErrorRecord> to call. An implementor following §11.3 will look for Fory APIs that no longer exist. Additionally §11.1 unit test assertion row #3 ("The arm round-trips through `PlatformErrorRecord` Fory serialisation and deserialisation byte-equal") is similarly invalidated — that test assertion was scoped to the Fory-schema world. fix:Amend §11.3 to describe the spdlog-field-based integration test instead: the test should construct each arm/prefix combination, call glibre::log_error to emit to a structured spdlog sink, and assert the emitted key-value fields match the expected arm name, os_code, prefix, source_tag. No Fory round-trip; no .glibre-trace serialization at the error-record level (trace format captures platform::Error through its runtime log fields, not through a wire schema). Also remove or reframe §11.1 row #3 to match the spdlog-field assertion. These are doc-only changes in #895 scope.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+audited for missing-arm coverage. The `[[nodiscard]]` discipline on
+every public-surface return (SPEC §5) and the closed-sum rule
+together make missing-arm handling a compile error.
+
+### 3.2 Translator surface (free functions, not methods)
+
+Four free functions in `engine/platform/src/detail/error/`, one per
+OS source. They are free functions — not member functions of a
+"translator class" — to make the SRP boundary mechanical: each
+function depends on exactly one OS header, has zero state, and is
```
cjhowe-us, line 224: severity:MED location:specs/platform/platform-error-design.md:224 problem:§3.2 states the SRP rationale for translator free functions: "each function depends on exactly one OS header, has zero state, and is trivially callable from the signal-safe path." The "zero state" claim is false for all four translators as designed — they all stamp the TLS prefix slot (§3.3) as a side effect, which is a write to global (per-thread) state. The SRP section then uses this false premise to justify the "trivially callable from the signal-safe path" conclusion, but §6.2 immediately contradicts it by documenting that the signal-safe path *skips* the prefix stamp precisely because TLS writes have constraints. The design works correctly but its own SRP rationale is self-contradicting at this line: if the translators had zero state they would be trivially signal-safe; they do not, so the signal safety requires the special no-stamp signal path documented in §6.2. fix:Replace "has zero state" with "has no heap-allocated or module-private state (the TLS prefix stamp in §3.3 is the sole side effect, bounded to a per-thread write of a compile-time literal pointer)" — then the SRP rationale is accurate and the §6.2 qualification follows coherently from it rather than contradicting it.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+- `OsCode` construction is a 4-byte integer init. Safe.
+- `IoFailure { OsCode { errno } }` construction reads `errno`
+  (async-signal-safe per POSIX) and writes a stack-local. Safe.
+- TLS prefix write: `__thread` is implemented by the dynamic linker
+  via the TLS descriptor table; on macOS this *is* lazy at first
+  access per dylib, which means the *first* TLS access in a signal
+  handler can trip a lazy resolver. To stay signal-safe, the
+  translator on the signal path takes a different shape: it returns
+  the typed arm without stamping the prefix. The prefix slot is
+  pre-touched (read once at process init) to force resolver
```
cjhowe-us, line 606: severity:MED location:specs/platform/platform-error-design.md:606 problem:§6.2 documents that the TLS prefix slot must be "pre-touched (read once at process init)" before any signal handler fires, to avoid a lazy-resolver trip in the signal context. However the design does not specify *where in the plugin register sequence* this pre-touch must occur relative to `Process::install_signal`. The SPEC §8.3 resume sequence (§4 in the platform self-reload) calls `glibre_plugin_register`, which must: (1) reinitialise SDL3, (2) re-spin I/O pool, (3) re-subscribe watchers, (4) re-install signal handlers. If the pre-touch of the TLS prefix slot happens anywhere after step (4), there is a window where a signal can fire after `Process::install_signal` but before the TLS slot is resolved, hitting the lazy resolver in the handler. The contract is insufficiently ordered. fix:Add a sentence to §6.2: "The pre-touch MUST execute as the first action of `glibre_plugin_register`, before any call that may install a signal handler or trigger a platform error path. Concretely: pre-touch the prefix slot, then call SDL_Init, then re-install signal handlers. This ordering is the plugin-register-sequence rule that makes §6.2 hold for hot-reload." Alternatively document this as a §12 open question if the ordering is left to the implementation plan.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+the call sites that need it. The trade-off is documented as a
+deliberate one in §5: prefix is **cold-path only**, never read on
+the success path of an `Error`-free call.
+
+The prefix literal-set is **closed** at the same gate as the
+variant: adding a new prefix is a deliberate edit reviewed against
+SPEC §10.6 (logging severity table needs a row for it) and the
+relevant aggregate's §10.3 sub-section. New prefixes do not bump the
+ABI hash (they live in the TU-local `.rodata`, not in
+`glibre-types.dylib`); they do bump the *log schema* version (§7).
```
cjhowe-us, line 338: severity:LOW location:specs/platform/platform-error-design.md:338 problem:§3.3 states that "New prefixes do not bump the ABI hash (they live in the TU-local `.rodata`, not in `glibre-types.dylib`); they do bump the *log schema* version (§7)." After the R1 demotion of PlatformErrorRecord from a Fory schema to a plain C++ struct (§7.2 in #895), there is no longer a "log schema version" to bump. The §3.3 sentence now references a concept (log schema versioning) that no longer exists in the design. fix:Replace "they do bump the *log schema* version (§7)" with "they require a documentation update to §3.3 (this prefix table) and a matching update to the spdlog-field emission logic in the `PlatformErrorRecord` C++ struct (§7.2) — no ABI hash change, no schema versioning." This is a doc-only one-liner change appropriate for #895.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0

notes:All five R1 findings reported as ADDRESSED on #895 are confirmed resolved in the #895 diff: the Fory schema is correctly demoted in §7.2/§7.3/§8.1/§8.2/§8.3; eastl::get_if replaces std::get_if in §4.2; §12 carries the OutOfBudget OPEN entry; the mach_to_error comment is present. The core SRP design is sound: one aggregate (OS error translation), one closed variant, four pure-domain translators, TLS prefix registry for coarse-arm discrimination. The seam between platform and data/core contexts is clean — no data:: imports, no Fory references in the post-#895 persistence section, zero ABI hash contribution from platform as SPEC §7 requires. The public API uses std::expected<T, glibre::Error> consistently throughout §5, the variant uses eastl::variant as PHILOSOPHY §11 requires, and the composition example is now correct. The hot/cold path split and signal-safe construction discipline hold across §5 and §6.

Two HIGH findings arise from an incomplete scope in #895: the R1 fix correctly rewrote §7.2, §8.x, and §12, but did not propagate the PlatformErrorRecord demotion backward to §1 (Purpose bullet 5) or forward to §11.3 (integration test plan). After #895 merges, these two sections will contradict §7.2 within the same document — a direct implementor hazard. Both are doc-only one-liners appropriate for the existing #895 PR scope. HIGH-1 (line 58): §1 bullet 5 still claims Fory schema ownership. HIGH-2 (line 941): §11.3 still describes a Fory round-trip test that cannot exist after the demotion, and §11.1 row 3 similarly references Fory serialization. MED-1 (line 224): §3.2 SRP rationale claims translators have "zero state" — false, as all four stamp the TLS prefix slot; the §6.2 signal-safety section immediately contradicts this premise. MED-2 (line 606): §6.2 documents the pre-touch discipline but does not specify where in glibre_plugin_register the pre-touch must occur relative to Process::install_signal, leaving a window for a signal to fire before the TLS slot is resolved. LOW-1 (line 338): §3.3 references "log schema version" — a concept that no longer exists after the §7.2 demotion. All five findings are doc-only and within #895 scope. No new context-seam violations found; no lingering data:: dependencies; no std:: container usage in the error model.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:587175f followup:#895

Rewrote §1 bullet 5 (line 58): replaced the Fory-schema claim with the spdlog structured-field carrier description and a cite to §7.2 for the demotion. The bullet now reads: "The spdlog structured-field carrier for log/replay (no Fory schema; see §7.2 for the demotion) — a small, stable (tag, os_code, prefix) triple that glibre::log_error and the e2e replay harness emit as spdlog key-value fields, preserving variant identity without any wire serialisation step."

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:587175f followup:#895

Two edits for HIGH-2:

1. §11.1 row 3 (formerly "Fory serialisation and deserialisation byte-equal") now reads: "Emitting the arm via glibre::log_error into a spdlog test-sink produces the expected structured-field set — every arm emits the expected field set; no field is omitted; no extra field is added." Row 4 updated to reference a compile-time static_assert on arm count instead of PlatformErrorRecord::arm_tag enum size.

2. §11.3 narrative rewritten entirely: describes emitting via glibre::log_error into a spdlog test-sink, asserting per-arm structured-field sets (arm name, os_code, prefix, source_tag, timestamp, thread ID), no omitted/extra fields, and replay assertion — no Fory/PlatformErrorRecord references remain.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:587175f followup:#895

Rewrote §3.2 SRP rationale (line 224): replaced the false "zero state" claim. The paragraph now reads: "each function depends on exactly one OS header, is a pure transform over its input OsCode plus one well-defined side-effect (stamping the per-thread TLS prefix slot via set_active_prefix; skipped on the signal-handler path per §6.2), and is trivially callable from the signal-safe path (§6.2). No global heap state, no cross-thread state, no I/O — exactly one process-local write per call." This accurately reflects §6.2 without contradicting it.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:587175f followup:#895

Added normative ordering-contract paragraph to §6.2 (after the TLS-prefix-write bullet): "Ordering contract within glibre_plugin_register: (1) pre-touch all four TLS prefix slots by performing one no-op set_active_prefix(nullptr) per slot — this resolves the lazy linker entry on the calling thread; (2) install signal handlers via Process::install_signal. Reversing the order can produce a signal-handler invocation that races against the lazy resolver — UB on macOS (lazy stub is not async-signal-safe). Each plugin that triggers signal-eligible OS calls owns this ordering; the platform error aggregate provides the error::tls::detail::pre_touch_all() helper that performs step (1) as a single call."

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:587175f followup:#895

Rewrote §3.3 parenthetical (line 338): struck the dead "log schema version (§7)" reference and replaced with: "they do bump the spdlog field-set contract (§7.2) — adding, removing, or renaming a structured field requires an MVP-bump just like a versioned schema would have." This forwards the reader to the live §7.2 demotion note and preserves the bump-gate semantics without the defunct schema reference.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+| Caller calls `sdl_to_error` with `nullptr` msg                  | Returns `OsCode { 0 }` with prefix `sdl_empty`. No crash; the null pointer is checked before `strstr`-style scan.                                                                                       |
+| Caller calls `mach_to_error` with `KERN_SUCCESS`                | Returns `OsCode { 0 }` (success-misuse path); debug assert fires.                                                                                                                                       |
+| Caller calls `ns_to_error(nullptr)` from outside `bridge.mm`    | Build error: the symbol is link-private to `bridge.mm.o` via internal-linkage `inline` definition. A call from another TU does not link.                                                                |
+| Caller adds a new public arm without bumping `PlatformErrorRecord` | Compile error: the wire-format `arm_tag` enumeration is generated from `platform::Error`'s declaration order via codegen; a new arm without a schema bump fails the codegen consistency check. |
```
cjhowe-us, line 878: severity:HIGH location:specs/platform/platform-error-design.md:878 problem:§10.2 translator-caller-misuse table row still says "fails the codegen consistency check" tied to `PlatformErrorRecord`. After the §7.2 demotion resolved in R1/R2 (commits a30b5de + 587175f on #895), there is no Fory schema, no `arm_tag` enumeration generated from `platform::Error`'s declaration order, and no codegen-consistency gate. The disposition contradicts §7.2 ("no Fory schema") and §8.2 ("arm-count `static_assert` is plugin-internal, not a middleman contract"). This row was not touched by either the R1 or R2 commits on #895 — a silent gap. fix:Replace the row body with: "Compile error: the plugin's `static_assert(eastl::variant_size_v<platform::Error> == 7u)` in §8.2 catches the arm-count mismatch at build time. There is no Fory-codegen consistency check (platform contributes zero schemas — §7.2)."

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+`tests/platform/error/perf/` — three Catch2 `BENCHMARK` blocks
+matching §9.3.
+
+### 11.5 Compile-fail tests
+
+Under `tests/platform/error/compile_fail/`, using
+`add_test(... CONFIGURATIONS CompileFail)` in CMake:
+
+- A test that adds a hypothetical 8th arm to `platform::Error`
+  without bumping the schema must fail the codegen consistency
```
cjhowe-us, line 958: severity:HIGH location:specs/platform/platform-error-design.md:958 problem:§11.5 compile-fail test description still says "without bumping the schema must fail the codegen consistency check" — same stale Fory framing as §10.2 line 878. After the §7.2 demotion, there is no schema to bump and no codegen-consistency tool check. This compile-fail test as described would not compile for a different reason than stated, and the description would mislead the implementer. Not touched by R1 or R2 commits on #895 — silent gap. fix:Replace with: "A test that adds a hypothetical 8th arm to `platform::Error` without updating the `static_assert(eastl::variant_size_v<platform::Error> == 7u)` (§8.2) must fail at compile time. No schema bump required; the guard is purely the plugin's build-time check."

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+
+The signal-safe construction discipline is a single test: the e2e
+crash-capture fixture installs `SIGSEGV`, dereferences a null
+pointer, and asserts that the captured `platform::Error` round-trips
+through the structured log format (§7) without aborting the
+captured-frame snapshot.
+
+### 6.3 Translators are reentrant
+
+The four translators read no module-private state (errno is per-thread
```
cjhowe-us, line 619: severity:MED location:specs/platform/platform-error-design.md:619 problem:The new §6.2 ordering-contract paragraph (added by R2 commit 587175f in #895) introduces two function identifiers that do not appear in the declared TLS API in §3.3: (1) `set_active_prefix(nullptr)` — §3.3 only declares `set_prefix(const char* literal)`; (2) `error::tls::detail::pre_touch_all()` — declared nowhere in the design's public or internal API stubs. This naming aliasing will mislead an implementer: they will look for `set_active_prefix` and `pre_touch_all` in the §3.3 API block and find neither. It also means any code-review or implementation audit cannot trace the call back to a declared signature. fix:Either (a) rename the §3.3 declaration from `set_prefix` to `set_active_prefix` and add the `pre_touch_all` helper declaration to the §3.3 API block with a doc-comment, or (b) use the existing name `set_prefix(nullptr)` in the §6.2 paragraph and rename the helper to the declared namespace: `detail::error::pre_touch_all()` matching the `glibre::platform::detail::error` namespace. Option (b) is lower churn since §3.3's declared names are already final.

specs/platform/platform-error-design.md
```diff
@@ -0,0 +1,1022 @@
...
+- [OPEN] Should the diagnostic prefix slot widen from `const char*`
+  (literal pointer) to `eastl::string_view` (literal pointer + length)
+  to make spdlog formatting allocation-free in all cases? Current
+  literals are NUL-terminated; spdlog can format them via `%s`. The
+  trade-off is a doubled TLS slot footprint (16 B vs 8 B) for a
+  marginal format-time speedup. Re-opened by the
+  `task-breakdown-error-perf` plan when CI gate timings land.
+  No platform-aggregate-side residue if deferred.
+
+- [OPEN] Should `PlatformErrorRecord::arm_tag` be a generated
```
cjhowe-us, line 986: severity:LOW location:specs/platform/platform-error-design.md:986 problem:The design §12 (post-R1 edits in #895) replaces the old codegen-vs-hand-maintained arm_tag open question with a `[RESOLVED]` entry for the Fory demotion and an `[OPEN]` entry for the SPEC §9.2 stale OutOfBudget enumerator. The [OPEN] OutOfBudget entry notes this "Blocks platform-error implementation PR" — but the design's §12 also still carries three other [OPEN] entries (prefix slot widening, surface_lost promotion, watcher_unavailable promotion, cross-thread prefix propagation, source_tag in signal context). None of these mention whether they are blocking or non-blocking. For topological clarity, each open question that is non-blocking should be tagged explicitly (e.g. `non-blocking` or `deferred to follow-on spike`) to match the one that is blocking. fix:Add `(non-blocking)` to the remaining [OPEN] entries so the implementation-blocking status of each is unambiguous to the task-breakdown spike that consumes this design.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:1
  LOW:1
carryover:0

notes:R1 and R2 carryover findings confirmed resolved in #895 commits a30b5de and 587175f. The core design topology is sound: one aggregate, one closed seven-arm sum, four translator free functions, TLS prefix registry, spdlog-only carrier consistent with SPEC §7 "platform contributes zero schemas" and "nothing to AbiHash". The §12 amendment tracking correctly marks [RESOLVED] for the Fory demotion and [OPEN] + blocking for the SPEC §9.2 OutOfBudget stale enumerator. Issue parenting is correct: #727 typed:spike under #714, no pts label (correct per AGENTS.md — spikes carry no estimate). Labels domain:platform + phase:mvp are correct. CC subjects on both #895 commits follow docs(platform) convention; the trailing ellipsis in the subject headline is a cosmetic issue only, not a blocking violation. Two HIGH silent gaps were found that the R1+R2 edits missed: (1) §10.2 translator-caller-misuse table row at line 878 still describes the arm-count guard as a Fory codegen consistency check referencing PlatformErrorRecord — this directly contradicts §7.2 demotion and §8.2 which establish a plugin-internal static_assert as the sole guard, with no codegen tool involved; (2) §11.5 compile-fail test description at line 958 carries the same stale framing — "fails the codegen consistency check" — where the correct description is a static_assert compile failure. One MED gap: the new §6.2 ordering-contract paragraph introduced by R2 uses set_active_prefix and error::tls::detail::pre_touch_all() — identifiers that do not appear in the §3.3 TLS API declarations (set_prefix / reset_prefix). This naming aliasing creates a cross-section inconsistency that will mislead an implementer. One LOW gap: the §12 [OPEN] entries do not distinguish blocking from non-blocking; only the OutOfBudget entry is marked blocking, but the other five entries carry no such status, making topological ordering from the task-breakdown spike ambiguous. Implementation remains blocked on #895 resolving all four findings above plus the existing blocking [OPEN] in §12 for SPEC §9.2.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:d29ecaa followup:#895

§10.2 misuse table row (line 849) rewritten: removed the stale `PlatformErrorRecord` and "fails the codegen consistency check" wording. The row now reads: "Compile error: the plugin-internal `static_assert(eastl::variant_size_v<platform::Error> == 7u)` fires when arm count drifts; no codegen tool, no Fory schema bump required." All `PlatformErrorRecord` references in that row are gone.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:d29ecaa followup:#895

§11.5 compile-fail test bullet (lines 940-942) rewritten: replaced "without bumping the schema must fail the codegen consistency check" with the static_assert framing: "Compile-fail test asserts `static_assert(eastl::variant_size_v<platform::Error> == 7u)` fires when a developer adds an 8th arm to the closed sum without bumping the assertion. Test captures compiler output and greps for the static_assert message. No codegen tool involved; no schema bump required."

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:d29ecaa followup:#895

Chose option (a) — §3.3 is the canonical naming source. Three edits:
1. §3.3 API block (after reset_prefix): added `pre_touch_all()` internal-helper declaration with caller doc ("Two callers: any plugin installing signal handlers (FileIo, Process)").
2. §6.2 ordering-contract paragraph: replaced `set_active_prefix(nullptr)` with `set_prefix(nullptr)` and replaced `error::tls::detail::pre_touch_all()` with `detail::error::pre_touch_all()`, both now aligned with the `glibre::platform::detail::error` namespace declared in §3.3.

specs/platform/platform-error-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:d29ecaa followup:#895

§12 (lines 969+): Added legend at top of section explaining the three tags. Walked all entries:
- Prefix-slot-widen: [NON-BLOCKING] (perf trade-off, CI-gated, no residue)
- PlatformErrorRecord: [RESOLVED] (unchanged)
- OutOfBudget SPEC §9.2: [BLOCKING IMPLEMENTATION] (already had blocking prose; tag makes it explicit)
- surface_lost promotion: [NON-BLOCKING] (second-consumer trigger, deferred)
- watcher_unavailable promotion: [NON-BLOCKING] (editor landing trigger, deferred)
- Cross-thread prefix propagation: [NON-BLOCKING] (sibling spike gates it, no residue)
- signal-safe source_tag stamp: [NON-BLOCKING] (e2e test gates it, no MVP residue)

Issue #736, opened 2026-05-03
[SPIKE] design-data-envelope-serdes-detailed


cjhowe-us:
Parent: #729

## Question

Produce the detailed design for the **envelope-serdes** aggregate in the `data` context.

Wire serialize / deserialize.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/data/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/core-runtime/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **envelope-serdes** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/data/envelope-serdes-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(data): detailed design — envelope-serdes (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-data-envelope-serdes-detailed` is blocked by this issue.

cjhowe-us:
agent:design-data-envelope-serdes
status:done
issue:#736
branch:docs/data-envelope-serdes-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-aa58b7b1be5bea4ab
host:mac
cloud:none
commit:ccb4ca71b106b04d3a7f6b5cb404038014e9e2e1
pr:#832
[...]

Issue #729, opened 2026-05-03
[SUB-EPIC] Detailed Designs — data


cjhowe-us:
# Detailed Designs — data

Per-aggregate detailed designs that flesh out §4 / §5 / §6 of
`specs/data/SPEC.md`. Each design lives at
`specs/data/<aggregate>-design.md` and references
`harmonius/docs/{requirements,design}/<area>/` to ensure MVP
coverage.

Parent epic: #49

## Children

Per aggregate: a `design-*-detailed` spike + a `task-breakdown-*-detailed` spike (blocked-by the design twin).

No estimate; rolls up.

Issue #49, opened 2026-05-03
[EPIC] Spec: data


cjhowe-us:
# Spec: data

Per-context epic for the **data** bounded context. Drives
`specs/data/SPEC.md` from stub to "done" via four sub-epics
(Foundation, Surface, Quality, Acceptance + Implementation Plan).

Parent: #1

## Sub-Epics

(linked as native sub-issues)

No estimate; rolls up from leaves.

PR #832, opened 2026-05-04
docs(data): detailed design — envelope-serdes (refs #736)

cjhowe-us:
## Summary

- Adds `specs/data/envelope-serdes-design.md` — detailed design for the `data` context's envelope-serdes aggregate (SPEC §4.8 / §7.2.4 / §7.5).
- Pins the 48-byte fixed-width header layout `(magic 4B, schema_fqn_id u32, version u32, source_hash 32B, payload_len u32)`, the version-dispatch decision tree (`peek → magic → bounds → registry → version dispatch`), and the hot/cold codec split.
- Refines SPEC §10's failure-mode closed sum with two new arms (`SchemaSourceMismatch`, `BufferTooSmall`) and re-classifies `VersionUnsupported`; flagged as a SPEC amendment in §12 [OPEN] #1.
- Re-derives every harmonius `BinaryHeader` clause (`harmonius/docs/design/core-runtime/reflection-serialization.md`) — keeps magic, refuses `header_version`, narrows `payload_len` to u32, narrows `type_id_hash` to a registry-ordinal `schema_fqn_id`, adds `source_hash` for cross-build content addressing.

Refs: #736 (spike — stays OPEN per workflow).

## Test plan

- [ ] Reviewers verify §3.1 byte layout against SPEC §4.8 / §7.2.4 invariants.
- [ ] Reviewers confirm §10 closed-sum amendment is coherent with `error-model.md` composition rules.
- [ ] Reviewers verify §9 budgets sum correctly into SPEC §9 / `perf-budget.md` `data` row.
- [ ] Sibling task-breakdown spike `task-breakdown-data-envelope-serdes-detailed` (#736-blocked) translates §11 into `type:plan` issues.

### DIFF
--- a/specs/data/envelope-serdes-design.md
+++ b/specs/data/envelope-serdes-design.md
@@ -1,0 +1,1118 @@
+# Envelope-Serdes Detailed Design
+
+> Detailed design for the `data` context's envelope-serdes aggregate
+> (`Envelope` / `EnvelopeHeader`, SPEC §4.8). Refines §5 §envelope.hpp,
+> §7.2.4, and §7.5 of `specs/data/SPEC.md` against the
+> `reviews/decisions/{fory-codegen,error-model,plugin-abi,
+> hot-reload-protocol,perf-budget,frame-phases}.md` decision records.
+> All conclusions independently re-derived; harmonius prior art
+> (`harmonius/docs/design/core-runtime/reflection-serialization.md`
+> §"Binary Serialization", `BinaryHeader`) cited as research input
+> only.
+
+Refs: spike #736 — `[SPIKE] design-data-envelope-serdes-detailed`.
+Parent sub-epic #729. Sibling task-breakdown spike blocked-by this
+deliverable.
+
+## 1. Purpose
+
+The envelope-serdes aggregate is the **single boundary** at which
+every persisted Fory blob crosses the wire. Its one responsibility
+is to **encode and decode the fixed-width envelope header that wraps
+every payload** and to **dispatch the wrapped payload by version to
+the migration dispatcher (§4.7) when the inbound version is older
+than the live registry's current version**.
+
+Concretely, `Envelope<T>` is the only call site at which:
+
+1. Magic bytes are written and verified.
+2. The schema FQN id and version of a payload are read or written.
+3. The `source-blake3` content-addressing digest is stamped or
+   checked.
+4. The payload's byte length is recorded and bounded against the
+   inbound span.
+5. Per-version dispatch is performed: equal-version → direct decode;
+   older-version → `MigrationDispatcher::dispatch` (§6.3); newer-than-
+   host → typed refusal.
+
+What the aggregate explicitly **refuses** to own:
+
+- **Raw Fory codecs.** Per-type `serialize_<fqn>` /
+  `deserialize_<fqn>` bodies are emitted by `glibre-foryc`
+  (`fory-codegen.md` §"Decision" #4) and exposed through the
+  `RegistryEntry::{serialize,deserialize}` trampolines (SPEC §5
+  §registry.hpp). The envelope is the **outer** wrapper around those
+  codecs; their byte format is owned by Fory.
+- **Schema registry.** Lookup by `SchemaId` is owned by
+  `SchemaRegistry::lookup` (SPEC §4.5; sibling spike #730). The
+  envelope reads the registry **read-only** at deserialize time and
+  never mutates it.
+- **Migration logic bodies.** Per-step `MigrationFn` bodies live in
+  the originating context (`fory-codegen.md` §"Migration Mechanic";
+  sibling spike #738). The envelope only **dispatches** to the
+  composed chain held by the registry entry; it never composes,
+  validates, or executes a migration step itself.
+- **CRC / checksum trailing.** A per-payload CRC is **not** in MVP
+  scope — `source-blake3` already content-addresses the schema; an
+  end-to-end CRC over the payload bytes is a post-MVP question
+  (§12 [OPEN] #2). The envelope reserves the bit pattern.
+- **Frame-phase scheduling.** Per `frame-phases.md`, the envelope
+  has no phase ownership; encode/decode runs on whatever thread the
+  caller is on, gated by §6.
+
+The aggregate's SRP boundary is sharp: if the envelope's byte layout,
+the magic value, the FQN-id encoding, the source-blake3 stamping
+rule, the version-dispatch decision tree, the endianness rule, the
+alignment rule, or the truncation-detection algorithm changes, this
+design changes. Anything else is out of scope.
+
+## 2. Requirements Coverage
+
+Mapping of harmonius requirements
+(`harmonius/docs/design/core-runtime/reflection-serialization.md`
+§"Binary Serialization", `BinaryHeader`) and SPEC §4.8 / §7.2.4 /
+§7.5 invariants to glibre MVP coverage. Every entry is independently
+re-derived.
+
+| Source                                                   | Glibre disposition (MVP) | Coverage site                                                                          |
+|----------------------------------------------------------|--------------------------|----------------------------------------------------------------------------------------|
+| Harmonius `BinaryHeader.magic = b"HSER"`                 | **Re-derived**           | §3.1 — magic = `b"GLEN"` (Glibre ENvelope); rationale in §3.1.                         |
+| Harmonius `BinaryHeader.header_version` (u8 + 3B pad)    | **Refused**              | Replaced by build-time meta-schema bootstrap (SPEC §7.6); single live envelope shape per `glibre-foryc` release. |
+| Harmonius `BinaryHeader.type_id_hash : u64`              | **Re-derived (narrowed)**| §3.2 — `schema_fqn_id : u32` keyed against the `SchemaRegistry`'s static FQN table; 32 bits suffice for MVP type counts (≤ 65k schemas, §3.2). |
+| Harmonius `BinaryHeader.schema_version : SchemaVersion`  | **Covered**              | §3.1 — `version : u32` little-endian (SPEC §4.8 inv. 3).                               |
+| Harmonius `BinaryHeader.payload_len : u64`               | **Re-derived (narrowed)**| §3.1 — `payload_len : u32`; rationale in §3.1 (max 4 GiB per blob, single-pass).       |
+| Harmonius "rkyv outer envelope" (zero-copy access)       | **Refused**              | Fory is the codec (`fory-codegen.md`); zero-copy is post-MVP (§12 [OPEN] #4).          |
+| Harmonius `MigrationRegistry` runtime dispatch           | **Covered (delegated)**  | Owned by `MigrationDispatcher` (sibling #738); envelope only **routes** by version (§3.4). |
+| SPEC §4.8 inv. 1 (self-describing)                       | **Covered**              | §3.1 layout — every dispatch input is in the header bytes.                              |
+| SPEC §4.8 inv. 2 (fixed-prefix, peekable)                | **Covered**              | §3.1 — header is exactly 48 bytes; `peek_header` returns without consuming the payload (§4.2). |
+| SPEC §4.8 inv. 3 (little-endian)                         | **Covered**              | §3.1 — every multi-byte field is LE; §3.5 endianness rule.                              |
+| SPEC §4.8 inv. 4 (round-trip identity)                   | **Covered**              | §3.6 round-trip rule; §11 unit + integration tests.                                     |
+| SPEC §4.8 inv. 5 (version-driven dispatch)               | **Covered**              | §3.4 dispatch decision tree.                                                            |
+| SPEC §7.2.4 (Fory-defined byte shape)                    | **Re-derived (refined)** | This design refines §7.2.4 by replacing `fqn:string` with `schema_fqn_id:u32` + adding `source_hash:32B` and `magic:4B`. Divergence flagged as §12 [OPEN] #1; SPEC §7.2.4 amendment proposed. |
+| SPEC §7.5 (per-version goldens)                          | **Covered**              | §11 test plan — golden harness per schema and per version.                              |
+| SPEC §10 `EnvelopeTruncated`, `DeserializeError`, `SchemaUnknown`, `MigrationStepMissing` | **Covered** | §10 failure modes table; payload field population per SPEC §10.2.                       |
+| `error-model.md` `std::expected<T, glibre::Error>` boundary | **Covered**           | §4 surface — every public function returns `std::expected`; no exceptions cross.        |
+| `plugin-abi.md` `glibre_types_serialize_<fqn>` C trampolines | **Covered**            | §3.3 — `Envelope<T>::serialize` / `deserialize` are typed wrappers over those trampolines. |
+| `hot-reload-protocol.md` "Step 3 — Migrate" reuse        | **Covered**              | §8 — same code path runs cold deserialize and hot-reload migrate; envelope is unaware of the phase. |
+| `perf-budget.md` `data` row 0.20 ms / 32 MiB             | **Covered**              | §9 — per-call wall-time per-KB bound; 16 MiB scratch sub-ceiling.                       |
+| `frame-phases.md` phase ownership                        | **Refused (none)**       | The envelope owns no phase; it is a synchronous codec called from any phase the caller is on (§6). |
+
+Coverage rule: every harmonius clause and every SPEC invariant either
+lands in this design or is refused with a one-line rationale. No
+silent drops.
+
+Glibre-native obligations beyond harmonius:
+
+- **Magic bytes are non-optional.** Harmonius's `BinaryHeader` had
+  magic; rkyv's archive format does not. We retain magic
+  unconditionally because it is the cheapest possible defence
+  against passing the wrong byte buffer to `deserialize` (§3.5
+  rationale).
+- **Source-blake3 is in the header, not the payload.** Harmonius
+  carried no per-payload schema content-hash on the wire. We add
+  it: it lets a reader detect "schema FQN matches but its `.fory`
+  source has drifted since this byte was stamped" — a subtle
+  failure rkyv hides because rkyv's layout-validation is structural,
+  not source-equivalence.
+- **No header_version field.** Harmonius reserved one byte for
+  in-process header evolution. Glibre's meta-schema bootstrap rule
+  (SPEC §7.6) collapses this onto a `glibre-foryc` release boundary:
+  one envelope shape per release, the AbiHash gate forces every
+  plugin to rebuild on bump. The `flags` u32 in §3.1 is reserved for
+  future bit-additive extensions that do not change layout.
+
+## 3. Detailed Model
+
+### 3.1 Envelope byte layout
+
+The envelope is a **fixed 48-byte little-endian header** followed by
+a Fory-encoded payload of `payload_len` bytes. The header layout is
+authored *not* through a `.fory` schema — per SPEC §7.2.4 the header
+is "Fory-defined" and rides Fory's own version handling — but its
+observable byte shape is pinned here so readers and writers across
+the engine and the cooker agree byte-for-byte.
+
+```text
+offset  size  field            description
+------  ----  ---------------  ----------------------------------------
+  0      4    magic            ASCII bytes 'G','L','E','N' (0x47 4C 45 4E)
+  4      4    schema_fqn_id    u32 LE — index into SchemaRegistry's
+                               FQN-sorted entries() span (§3.2)
+  8      4    version          u32 LE — SchemaVersion of the payload
+                               (SPEC §4.1 inv. 2)
+ 12     32    source_hash      Blake3-256 of the canonicalized .fory
+                               source (SPEC §7.3) at the time of write;
+                               equals SchemaRegistry's recorded hash for
+                               (schema_fqn_id, version) on a same-build
+                               round-trip
+ 44      4    payload_len      u32 LE — Fory-encoded body size in bytes,
+                               not including the 48-byte header
+------  ----  ---------------  ----------------------------------------
+ 48     payload_len  payload   Fory-encoded body of the Generated Type
+                               at SchemaVersion `version`.
+```
+
+Total wire size: `48 + payload_len` bytes. No trailing CRC, no
+trailing pad, no length prefix outside `payload_len`.
+
+**Constants.**
+
+| Symbol                          | Value                                                          |
+|---------------------------------|----------------------------------------------------------------|
+| `kEnvelopeMagic`                | `{ 0x47, 0x4C, 0x45, 0x4E }` — ASCII `"GLEN"`                  |
+| `kEnvelopeHeaderSize`           | `48` (bytes, exact)                                            |
+| `kSourceHashSize`               | `32` (bytes — Blake3-256)                                      |
+| `kMaxPayloadBytes`              | `4'294'967'295` (u32 max)                                      |
+| `kEnvelopeHeaderAlignment`      | `4` bytes (every multi-byte field is 4-byte aligned in-buffer) |
+
+**Why `"GLEN"`.** Four ASCII bytes that are (a) printable, so a hex
+dump is human-recognizable, (b) distinct from harmonius's `"HSER"`
+to avoid accidental cross-engine load (the byte streams must never
+mix), (c) distinct from any common archive magic (`PK`, `7z`, `BZ`,
+ELF `\x7FELF`, Mach-O `\xFE\xED\xFA\xCE`), so a naive `file(1)` does
+not misidentify the byte run.
+
+**Why u32 `payload_len`.** The 4 GiB cap is the largest single asset
+the engine commits to round-tripping in one envelope; assets beyond
+this size split at the content layer (post-MVP — `content/` SPEC).
+Narrowing from harmonius's u64 saves 4 bytes per envelope and aligns
+with the §9 per-call budget (per-KB encode/decode time linear in
+`payload_len`).
+
+**Why u32 `schema_fqn_id` (vs harmonius's u64 `type_id_hash`).** The
+registry's `entries()` span (SPEC §4.5; FQN-sorted, populated at
+static-init) is the canonical FQN universe for any one middleman
+build. A u32 index is sufficient (the engine's MVP schema set is
+≤ ~10⁴; even at 10⁶ schemas u32 suffices). The id is **build-local**:
+two middleman builds with non-identical schema sets do not share id
+namespaces. The `source_hash` field is the cross-build invariant —
+see §3.2 for the discriminator pair.
+
+**Why `source_hash` in the header (vs harmonius's omission).** The
+hash is the wire-level proof that the bytes were stamped against a
+specific `.fory` source. Without it, two builds that happen to share
+`schema_fqn_id == K` but disagree on the schema's content (e.g. one
+inserted a field at a tag the other has not declared) would silently
+mis-decode. With it, a reader detects the drift and refuses with
+`SchemaSourceMismatch` (§10).
+
+### 3.2 The `(schema_fqn_id, source_hash)` identity pair
+
+`SchemaRegistry::entries()` returns the FQN-sorted span of
+`RegistryEntry`s populated at static-init by
+`_registry.cpp` (SPEC §6.1.3). The envelope encodes the **entry's
+ordinal index** in that span as `schema_fqn_id`. Two reader rules
+make this safe across builds:
+
+1. **Same-build round-trip.** A reader operating against the same
+   middleman build that wrote the bytes resolves
+   `entry := registry.entries()[schema_fqn_id]` and verifies
+   `entry.source_hash == header.source_hash`. Mismatch is impossible
+   in this case (the hash is taken from the same registry); the
+   check is a defensive assertion in debug builds and elided in
+   shipping.
+2. **Cross-build read.** A reader operating against a different
+   middleman build (e.g. loading a save file from a previous engine
+   release) cannot trust the `schema_fqn_id` ordinal because the
+   FQN-sorted list may have grown or shrunk. The reader **falls
+   back to source-hash lookup**: a secondary registry index keyed by
+   `source_hash → RegistryEntry*` is built at static-init alongside
+   the primary FQN-sorted span (SPEC §4.5 inv. 3 already permits
+   secondary indices). If the source-hash is present in the live
+   registry, decode proceeds against that entry; if absent,
+   `SchemaSourceMismatch` (§10).
+
+The two-mode behavior is selected by a single header flag (the
+`flags` field in §3.1's reserved region is *not* used — the trigger
+is structural: the reader compares `schema_fqn_id` against
+`registry.entries().size()` and against `entry.source_hash`).
+Default behavior is "use the ordinal; verify the hash"; the fallback
+fires only on hash-disagreement.
+
+This collapses two failure modes (forged FQN ordinal, schema source
+drift) onto one decision and one error arm
+(`SchemaSourceMismatch`).
+
+### 3.3 The `Envelope<T>` typed wrapper
+
+```cpp
+// data/runtime/include/glibre/types/envelope.hpp — public.
+namespace glibre::types {
+
+// 48-byte fixed header — see §3.1 layout. Trivially copyable, POD-
+// friendly. Never holds owning storage.
+struct EnvelopeHeader {
+    std::array<std::byte, 4> magic{};                 // = kEnvelopeMagic
+    std::uint32_t            schema_fqn_id{0};
+    SchemaVersion            version{0};               // alias for u32
+    SchemaSourceHash         source_hash{};            // 32 bytes
+    std::uint32_t            payload_len{0};
+};
+static_assert(sizeof(EnvelopeHeader) == 48);
+static_assert(std::is_trivially_copyable_v<EnvelopeHeader>);
+static_assert(alignof(EnvelopeHeader) <= 4);
+
+// Per-FQN typed wrapper. Specializations are emitted by glibre-foryc
+// in <glibre/types/<ctx>/<Type>.hpp>; primary template left undefined
+// (link-time error if no schema for T).
+template <class T>
+struct Envelope {
+    // Encode `value` into `dst` starting at offset 0. Writes the
+    // 48-byte header, then invokes the codegen-emitted Fory body
+    // serializer at offset 48. Returns the total bytes written
+    // (== 48 + payload_len).
+    //
+    // Failure modes: BufferTooSmall (§10).
+    [[nodiscard]] static auto serialize(
+        const T&             value,
+        std::span<std::byte> dst
+    ) noexcept -> std::expected<std::size_t, data::Error>;
+
+    // Decode `src`. Reads the 48-byte header, dispatches by version
+    // (§3.4), invokes the registry's deserialize trampoline (§3.3)
+    // and the migration dispatcher (§6.3) as needed. Output value is
+    // stored in `out`.
+    //
+    // Failure modes: BadMagic, EnvelopeTruncated, SchemaUnknown,
+    // SchemaSourceMismatch, VersionUnsupported, PayloadTruncated,
+    // DeserializeError, SchemaMigrationFailure, MigrationStepMissing
+    // (§10).
+    [[nodiscard]] static auto deserialize(
+        std::span<const std::byte> src,
+        T*                          out
+    ) noexcept -> std::expected<void, data::Error>;
+
+    // Compute the exact wire byte count for `value` without
+    // performing the serialize. Used by callers sizing pre-allocated
+    // buffers (e.g. content/ asset writers). O(value-size) — same
+    // walk Fory does, no memcpy.
+    [[nodiscard]] static auto size_bytes(const T& value) noexcept
+        -> std::expected<std::size_t, data::Error>;
+};
+
+// Type-erased peek. Returns the parsed header without consuming the
+// payload. Used by save-file readers to triage an unknown FQN
+// before allocating a full T.
+[[nodiscard]] auto peek_header(std::span<const std::byte> src) noexcept
+    -> std::expected<EnvelopeHeader, data::Error>;
+
+}  // namespace glibre::types
+```
+
+Borrow rules:
+
+- `serialize` writes only into `dst`; the input `value` is borrowed
+  read-only. No allocation.
+- `deserialize` reads only from `src`; `out` is written exactly
+  once on success. On failure `out` is left **unmodified** (the
+  caller's prior contents persist) — no partial decode escapes. The
+  single legal exception is intermediate scratch the migration
+  dispatcher writes through `Arena&`; that storage is reset between
+  steps and never aliases `out`.
+- `peek_header` allocates nothing; the returned `EnvelopeHeader` is
+  a value (48 bytes copied out of `src`).
+
+### 3.4 Version-dispatch decision tree
+
+`Envelope<T>::deserialize` runs the following decision tree, in
+order, on every call. Each step is constant-time except step 5,
+whose cost is proportional to `entry.version - header.version` (i.e.
+the migration chain length).
+
+```text
+1. BoundsCheck(src.size() >= kEnvelopeHeaderSize)
+   └── false → EnvelopeTruncated{ at.offset = src.size() }
+
+2. ReadHeader(src[0..48]) into hdr
+   └── hdr.magic != kEnvelopeMagic → BadMagic{ at.offset = 0 }
+
+3. BoundsCheck(48 + hdr.payload_len <= src.size())
+   └── false → PayloadTruncated{ at.offset = src.size() }
+
+4. RegistryLookup:
+   if hdr.schema_fqn_id < registry.entries().size():
+       entry := registry.entries()[hdr.schema_fqn_id]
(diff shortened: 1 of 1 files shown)


specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+| `SchemaMigrationFailure`  | Older-version path: a `MigrationFn` body returned `unexpected`                                                | step 5 (older-version arm)   | `step_schema`, `step_from`, `step_to`                          | refuse decode (cold) / refuse load (hot-reload) | error / warn | wrapped to `core::Error::SchemaMigrationFailed` (already in SPEC §10) |
+| `DeserializeError`        | Fory body decoder refused (tag-type mismatch, range check, nested type refusal) — i.e. envelope OK, body bad | step 5 (same-version arm) — body call returned `unexpected` | `at.schema`, `at.version`, `at.offset = 48 + body_offset`      | refuse decode    | error    | none — passed through (already in SPEC §10) |
+| `BufferTooSmall`          | `serialize`: `dst.size() < 48 + computed_payload_len`                                                          | serialize precheck           | `at.offset = dst.size()`                                       | refuse encode    | error    | none — passed through; **new arm** vs SPEC §10 |
+
+**Recovery vocabulary** (mirrors SPEC §10.2):
```
cjhowe-us, line 938: severity:HIGH location:specs/data/envelope-serdes-design.md:938 problem:The design §10 failure table introduces 'SchemaSourceMismatch' as a new arm, but SPEC §10's canonical ErrorTag enum (SPEC.md line ~2482) already has arm 10 named 'SourceHashMismatch' that covers the same source-hash drift scenario. These are two different names for overlapping but semantically distinct triggers: the design's 'SchemaSourceMismatch' fires during Envelope<T>::deserialize step 4 (inbound save-file bytes with an unrecognised source hash); the SPEC's 'SourceHashMismatch' fires during the phase-8 barrier diff (a candidate plugin with a drifted per-FQN source hash). Introducing 'SchemaSourceMismatch' as a distinct sixth-unnamed arm without updating the SPEC's ErrorTag enum means the closed sum is out of sync with the canonical enum. An implementer following the SPEC header stub would have no ErrorTag for 'SchemaSourceMismatch' to return. fix:Either (a) reuse 'SourceHashMismatch' (tag 10) with an expanded trigger note that covers both detection sites, or (b) add a new tag 11 = SchemaSourceMismatch to SPEC's ErrorTag in the same PR. The SPEC amendment in §12 [OPEN] #1 must explicitly enumerate which new arms are added and their tag values so the closed sum stays normative. Since the PR is merged, raise a follow-up plan issue under sub-epic #729 to reconcile the ErrorTag enum before the first implementation PR under envelope-serdes.

specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+[12..44)  source_hash     (32B aligned to 4 — array of bytes)
+[44..48)  payload_len     (4B aligned to 4)
+```
+
+The `source_hash` field is byte-aligned (it is `eastl::array<std::
+byte, 32>`); 4-byte alignment within the buffer is sufficient.
+
+**Reserved bytes.** None in the 48-byte header. Future flag-bit
+additions consume reserved bits inside `payload_len`'s u32 (the
+high 4 bits are reserved as `flags` — see §12 [OPEN] #3 — though
```
cjhowe-us, line 416: severity:HIGH location:specs/data/envelope-serdes-design.md:416 problem:§3.5 states the high 4 bits of payload_len's u32 are 'reserved as flags' while kMaxPayloadBytes constant (line 165) is set to 4'294'967'295 — the full u32 maximum. These two statements contradict each other. If the high 4 bits are reserved, the effective maximum payload size is 2^28 - 1 = 268,435,455 bytes (~256 MiB), not 4 GiB. An encoder writing a 300 MiB payload would set bits in the reserved region, and a reader checking those bits as flags would see them as set, producing a silent protocol violation. fix:Either (a) strike the high-bit flag reservation entirely from §3.5 (MVP has no flags, so reserving them now introduces an unenforceable constraint) and document any future flag field as a layout-breaking change subject to §7.1's major-version bump rule, or (b) correct kMaxPayloadBytes to 2^28-1 and document that the high 4 bits are unavailable for payload length encoding. Option (a) is simpler and aligns with the envelope having no flags field in the 48-byte layout. The reserved-bit approach belongs in §12 [OPEN] #3 as a future option, not as a live constraint on the constant.

specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+### 3.3 The `Envelope<T>` typed wrapper
+
+```cpp
+// data/runtime/include/glibre/types/envelope.hpp — public.
+namespace glibre::types {
+
+// 48-byte fixed header — see §3.1 layout. Trivially copyable, POD-
+// friendly. Never holds owning storage.
+struct EnvelopeHeader {
+    std::array<std::byte, 4> magic{};                 // = kEnvelopeMagic
```
cjhowe-us, line 245: severity:MED location:specs/data/envelope-serdes-design.md:245 problem:The EnvelopeHeader C++ stub in §3.3 uses 'std::array<std::byte, 4>' for the magic field, while the SPEC §5 canonical header uses 'eastl::array<std::byte, 32>' for SchemaSourceHash and PHILOSOPHY §11 mandates that all containers come from eastl:: not std::. The magic array is also a container. The §4.1 section repeats this with 'std::array<std::byte, 4>' for magic and 'std::array<std::byte, 32>' for SchemaSourceHash alias. The inline comment at line 411 acknowledges 'eastl::array<std::byte, 32>' for source_hash but the struct definition above uses std::array. fix:Replace 'std::array<std::byte, 4>' with 'eastl::array<std::byte, 4>' in the magic field of EnvelopeHeader across §3.3 and §4.1. If the design intends to use std::array for these fixed-width byte buffers (justified because std::span is permitted at interop boundaries per PHILOSOPHY §11 final sentence), add an explicit exemption note explaining why this particular use of std::array is sanctioned. Without such a note, an implementer following PHILOSOPHY §11 strictly would use eastl::array and hit a type mismatch against the design's snippet.

specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+    // stored in `out`.
+    //
+    // Failure modes: BadMagic, EnvelopeTruncated, SchemaUnknown,
+    // SchemaSourceMismatch, VersionUnsupported, PayloadTruncated,
+    // DeserializeError, SchemaMigrationFailure, MigrationStepMissing
+    // (§10).
+    [[nodiscard]] static auto deserialize(
+        std::span<const std::byte> src,
+        T*                          out
+    ) noexcept -> std::expected<void, data::Error>;
```
cjhowe-us, line 283: severity:MED location:specs/data/envelope-serdes-design.md:283 problem:The design's §3.3 public surface declares 'Envelope<T>::deserialize(src, T* out) -> std::expected<void, data::Error>' — an out-parameter form that returns void on success. However, the canonical SPEC §5 public interface stub (SPEC.md around line 748) declares 'deserialize(src) -> std::expected<T, data::Error>' — a value-return form. This is a deliberate design divergence that is not called out explicitly in the document. The out-param form avoids copy/move of T at the call site (relevant for large generated types); the value-return form is more composable with monadic chains (error-model.md §Composition Rules #3). fix:Add a prose paragraph in §4.2 or §3.3 explicitly justifying the out-param signature over the SPEC stub's value-return form, and flagging this as a SPEC §5 amendment in §12 [OPEN]. Without this note the SPEC and design disagree on a core API shape. The amendment should land in the same implementation PR that introduces envelope.hpp. The issue is MED rather than HIGH because the out-param form is strictly safer for large types and the error-model decision record does not mandate one form over the other for non-cross-boundary callers.

specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+
+- 200 × (100 ns header + 50 µs per-step migration + 250 ns/KB body
+  ~~ 0.5 KB row average) = 200 × ~50.225 µs ≈ 10 ms?
+
+Actual S2 reloads carry far fewer rows (100 ms cap on phase 8 with
+0.40 ms `data` allocation; the envelope's 50 µs/step single-step
+migration cost from `migration_dispatch_v_minus_1.bench.cpp` SPEC
+§9.4 #3 amortises to ~8 rows/ms, i.e. ≤ ~60 rows in 8 ms is **out
+of phase-8 budget** — see §12 [OPEN] #5). The benchmark fixture
+runs at the per-call ceiling; production reloads with realistic
```
cjhowe-us, line 869: severity:LOW location:specs/data/envelope-serdes-design.md:869 problem:§9.2's phase-8 budget decomposition contains an explicit arithmetic error that the author self-flags with '?': '200 × ~50.225 µs ≈ 10 ms?' immediately contradicts the 0.40 ms phase-8 envelope budget stated in the section header and in SPEC §9.3. The section then partially recovers with the note that S2 reloads carry 'far fewer rows', but leaves the ≤~60 rows calculation marked [OPEN] #5 without resolving the base contradiction. This is a correctness problem for the design document as a performance specification. fix:Either (a) correct the per-step migration cost assumption (50 µs is derived from 'migration_dispatch_v_minus_1.bench.cpp' which is a worst-case single-step benchmark; for realistic smaller rows the cost is proportionally lower — document the expected row size and cost at the S2 fixture rather than using 0.5 KB as the representative row average), or (b) explicitly lower the S2 row-count bound from 200 to ≤8 rows within the 0.40 ms budget and flag a new §12 [OPEN] tracking the discrepancy. The 'approximately 10 ms' estimate should not be left unresolved in a performance specification that is supposed to gate CI benchmarks.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0
notes:The design document is thorough and well-structured. The 48-byte fixed-width header layout (magic 4B, schema_fqn_id u32, version u32, source_hash 32B, payload_len u32) is sound and well-justified, the version-dispatch decision tree in §3.4 is correct and complete, all five SPEC §4.8 invariants are covered, and the public surface correctly uses std::expected<T, data::Error> throughout. The Conventional Commit subject 'docs(data):' is correct for a spike deliverable; the 'refs #736' citation is present and the issue stays OPEN per spike workflow. Two HIGH findings block the first implementation PR that consumes this design: (1) 'SchemaSourceMismatch' is introduced in the design's §10 error table but is not present in the SPEC's canonical ErrorTag enum — the SPEC already has a related arm 'SourceHashMismatch' (tag 10) that covers source-hash drift at plugin load time; the design needs to either reuse that arm with an expanded trigger note or add a new tag 11 to the SPEC enum before the implementation PR lands; (2) §3.5 reserves the high 4 bits of payload_len for flags but kMaxPayloadBytes is set to the full u32 maximum, creating a silent protocol violation if any payload exceeds 256 MiB — the constant must be reconciled with the reservation before code is written. Two MED findings should be resolved before the implementation PR: the design's EnvelopeHeader C++ snippet uses std::array for the magic field while PHILOSOPHY §11 mandates eastl::array (no exemption is documented), and the deserialize signature is changed from the SPEC stub's value-return form to an out-param form without an explicit amendment note. The LOW finding (§9.2 arithmetic leaving '≈ 10 ms?' unresolved against a 0.40 ms budget) is tracked in §12 [OPEN] #5 and does not block implementation.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0
notes:The design document is thorough and well-structured. The 48-byte fixed-width header layout (magic 4B, schema_fqn_id u32, version u32, source_hash 32B, payload_len u32) is sound and well-justified, the version-dispatch decision tree in §3.4 is correct and complete, all five SPEC §4.8 invariants are covered, and the public surface correctly uses std::expected<T, data::Error> throughout. The Conventional Commit subject 'docs(data):' is correct for a spike deliverable; the 'refs #736' citation is present and the issue stays OPEN per spike workflow. Two HIGH findings block the first implementation PR that consumes this design: (1) 'SchemaSourceMismatch' is introduced in the design's §10 error table but is not present in SPEC's canonical ErrorTag enum — the SPEC already has 'SourceHashMismatch' (tag 10) covering source-hash drift at plugin load time; the design either reuses that arm with expanded trigger or adds tag 11 to the SPEC enum before the implementation PR; (2) §3.5 reserves high 4 bits of payload_len for flags but kMaxPayloadBytes is full u32 max, creating a silent protocol violation above 256 MiB. Two MED findings: EnvelopeHeader uses std::array for magic field without a PHILOSOPHY §11 exemption note; deserialize signature changed from SPEC stub value-return to out-param without amendment note. LOW: §9.2 arithmetic leaves an unresolved 10ms estimate against a 0.40ms budget, deferred to §12 [OPEN] #5.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0
notes:The design document is thorough and well-structured. The 48-byte fixed-width header layout (magic 4B, schema_fqn_id u32, version u32, source_hash 32B, payload_len u32) is sound and well-justified, the version-dispatch decision tree in §3.4 is correct and complete, all five SPEC §4.8 invariants are covered, and the public surface correctly uses std::expected<T, data::Error> throughout. The Conventional Commit subject docs(data): is correct for a spike deliverable; the refs #736 citation is present and the issue stays OPEN per spike workflow. Two HIGH findings block the first implementation PR that consumes this design: (1) SchemaSourceMismatch is introduced in the design §10 error table but is not present in the SPEC canonical ErrorTag enum — the SPEC already has SourceHashMismatch (tag 10) covering source-hash drift at plugin load time; the design must either reuse that arm with expanded trigger note or add a new tag 11 to the SPEC enum before the implementation PR; (2) §3.5 reserves the high 4 bits of payload_len for flags but kMaxPayloadBytes is set to the full u32 maximum, creating a silent protocol violation for payloads above 256 MiB — the constant must be reconciled with the reservation. Two MED findings: EnvelopeHeader C++ snippet uses std::array for the magic field without a PHILOSOPHY §11 exemption note; deserialize signature changed from SPEC stub value-return form to out-param form without an explicit SPEC amendment note. LOW: §9.2 arithmetic leaves an unresolved 10 ms estimate against a 0.40 ms phase-8 budget, tracked in §12 [OPEN] #5.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1eae3ff followup:#894

HIGH-1 addressed: renamed SchemaSourceMismatch → SourceHashMismatch (SPEC §10 arm 10, tag = 10) throughout the design document (§3.1 rationale, §3.2 fallback path, §3.3 comment, §3.4 decision tree pseudocode, §7.3 cross-build invariants, §8 hot-reload contract, §10 failure-modes table row, §11.1 test name, §12 [OPEN] #1). The §10 table row's trigger note now explicitly covers both detection sites: (a) the existing SPEC arm 10 site — phase-8 barrier-diff (Mode-A plugin reload); (b) the design's addition — Envelope<T>::deserialize step 4 / FallbackBySourceHash for save-file/snapshot bytes with an unrecognised source hash. No new enum arm required; the closed sum stays normative.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1eae3ff followup:#894

HIGH-2 addressed: removed the inconsistent high-4-bit flag reservation from §3.5. The 'Reserved bytes' paragraph now states that in MVP all 32 bits of payload_len are length — kMaxPayloadBytes = u32 max is a live, self-consistent bound with no encoding-invisible sub-cap. §7.1 bullet 2 updated to describe the correct future mechanism (add a new flags : u32 field at offset 48, bumping header from 48 to 52 bytes, major-version bump). §12 [OPEN] #3 updated accordingly. The structural ref in §3.2 to 'flags field in §3.1's reserved region' also removed — the two-mode dispatch is structural (comparison of ordinal vs size() and source_hash) with no flag bit required.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1eae3ff followup:#894

MED-1 addressed: replaced std::array<std::byte, 4> with eastl::array<std::byte, 4> for the magic field in the EnvelopeHeader struct in §3.3 and §4.1 per PHILOSOPHY §11. Also corrected the SchemaSourceHash alias in §4.1 from std::array<std::byte, 32> to eastl::array<std::byte, 32> for consistency (the §3.3 snippet already referenced the alias which SPEC.md defines as eastl::). Added inline comments referencing PHILOSOPHY §11 to make the rationale explicit for implementers.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1eae3ff followup:#894

MED-2 addressed: added explicit prose paragraph in §3.3 immediately after the closing code block justifying the out-param signature over the SPEC §5 value-return stub. Three stated reasons: (1) no copy/move of T through std::expected internal storage for large generated types; (2) caller-controlled lifetime — save-file loaders pre-allocate the target slot; (3) auditable failure isolation — out is unmodified on any failure, easier to verify in the out-param form. The paragraph notes that error-model.md composition rules #3 mandates std::expected but does not prescribe value vs void, so both forms satisfy the rule. Registered as §12 [OPEN] #7 — a SPEC §5 amendment to land in the first envelope.cpp implementation PR.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1eae3ff followup:#894

LOW addressed: §9.2 arithmetic fixed. The '200 rows × 50.225 µs ≈ 10 ms?' estimate is removed. The section now derives the correct S2 row-count bound from the 0.40 ms budget: 0.40 ms ÷ 50.225 µs/row ≈ 7.97 rows → ≤ 8 migrating rows per hot-reload frame. Added a note that if a plugin upgrade touches more rows, the loader must defer migration to a background pass outside phase 8. The §12 [OPEN] #5 item about fixture representativeness is preserved.

specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+[[nodiscard]] auto serialize_header(
+    const EnvelopeHeader& hdr,
+    std::span<std::byte>  dst
+) noexcept -> std::expected<std::size_t, data::Error>;
+
+[[nodiscard]] auto deserialize_header(
+    std::span<const std::byte> src,
+    EnvelopeHeader*             out
+) noexcept -> std::expected<std::size_t, data::Error>;
+
```
cjhowe-us, line 542: severity:MED location:specs/data/envelope-serdes-design.md:542 problem:The design §4.2 adds serialize_header and deserialize_header as publicly-exported free functions (type-erased header-only forms for the cooker, editor inspector, and e2e harness). These two operations do not appear in the canonical SPEC §5 header stub — the SPEC §5 EnvelopeHeader block exposes no serialize_header / deserialize_header, and SPEC §4.8 inv. 2 only says 'readers may peek the envelope without consuming the body'. The design also adds peek_header (same problem, though peek_header is at least implied by inv. 2's 'peekable' language). Introducing three new public API entry points without a tracked SPEC §5 amendment violates the cohesion rule that every public-surface addition must be traceable to a spec invariant change. The existing §12 [OPEN] #1 tracks the EnvelopeHeader struct shape change and error-arm additions but says nothing about peek_header, serialize_header, or deserialize_header. fix:Add a §12 [OPEN] #8 entry explicitly tracking the SPEC §5 amendment for peek_header, serialize_header, and deserialize_header. The note should cite which callers are the two concrete users (cooker, editor inspector) per PHILOSOPHY anti-pattern rule 'cross-domain abstractions invented before two concrete users exist', confirm both callers exist in the spike scope, and state that the amendment lands with the first implementation PR that introduces data/runtime/src/envelope.cpp.

specs/data/envelope-serdes-design.md
cjhowe-us, line 938: severity:MED location:specs/data/envelope-serdes-design.md:938 problem:The design §10 failure table lists four error arm names — BadMagic, PayloadTruncated, VersionUnsupported, and BufferTooSmall — that do not appear in the SPEC §10.1 ErrorTag closed enum. The §10.2 per-arm notes for BadMagic and PayloadTruncated each say 'none — passed through', meaning they are intended as distinct ErrorTag values returned to callers. VersionUnsupported says 'sub-case of DeserializeError per SPEC §10 row DeserializeError — kept distinct here for log clarity'. BufferTooSmall says 'new arm vs SPEC §10'. Three different dispositions for four missing arms is a seam incoherence: callers implementing against the SPEC §10.1 enum can only switch on the 10 defined tags; they cannot match BadMagic, PayloadTruncated, VersionUnsupported, or BufferTooSmall because those tags do not have numeric values in the enum. Either (a) all four need to be added to the ErrorTag closed sum with numeric values (a SPEC §10 amendment per §10's amendment rule — same PR as the new failure point), or (b) BadMagic and PayloadTruncated must be explicitly documented as sub-cases of EnvelopeTruncated (tag 9) with a note explaining how the WireSite.offset field disambiguates them, and VersionUnsupported must be documented as a sub-case of DeserializeError (tag 3), and BufferTooSmall must be assigned a tag number. The design's §12 [OPEN] #1 only mentions BufferTooSmall and VersionUnsupported; it does not mention BadMagic or PayloadTruncated as unresolved. fix:In §12 [OPEN] #1, enumerate all four unnamed arms and state explicitly which disposition applies to each: (a) new tag in ErrorTag enum, or (b) sub-case of existing tag with disambiguation note. This is required before the implementation PR so that the generated C++ headers do not introduce undeclared enum values.

specs/data/envelope-serdes-design.md
```diff
@@ -0,0 +1,1118 @@
...
+## 12. Open Questions
+
+- [OPEN] **#1 — SPEC §7.2.4 / §10 amendment scope.** This design
+  refines the SPEC's envelope shape from the `(fqn:string,
+  version:u32, payload_length:u32, flags:u32)` Fory-defined
+  layout to the explicit 48-byte fixed-width
+  `(magic, schema_fqn_id, version, source_hash, payload_len)`
+  layout (§3.1). It also adds two arms to `data::Error`
+  (`SchemaSourceMismatch`, `BufferTooSmall`) and re-classifies one
```
cjhowe-us, line 1066: severity:LOW location:specs/data/envelope-serdes-design.md:1066 problem:§12 [OPEN] #1's amendment-scope statement is incomplete. It says the design proposes changes to SPEC §7.2.4 (layout) and §10 (two new error arms plus one re-classification), but does not enumerate the following structural differences between the SPEC §5 EnvelopeHeader stub and the design's §3.1 / §3.3 layout: (1) magic field added (not in SPEC §4.8 or SPEC §5 EnvelopeHeader); (2) schema_fqn_id : u32 replaces SchemaId schema{} (FQN string-view → u32 ordinal — a type change at the public boundary); (3) source_hash : 32B added (not in SPEC §5 EnvelopeHeader); (4) flags : u32 field removed from SPEC §5 EnvelopeHeader; (5) payload_length renamed to payload_len; (6) version field's semantic type stays u32 but is no longer part of SchemaId (minor naming/type change). The implementation PR that lands envelope.cpp will need to reconcile every one of these against the SPEC §5 stub or will produce a compile-time conflict between glibre/types/envelope.hpp and the generated headers. fix:Extend §12 [OPEN] #1 to list all six structural deltas above in addition to the currently-listed error-arm additions. The amendment-scope note should be a complete checklist that the implementation PR's reviewer can tick off against the actual SPEC §5 header update.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:1 (R1 HIGH-1 SchemaSourceMismatch→SourceHashMismatch rename; addressed in open follow-up PR #894, not yet merged)

notes:The round-2 cohesion + SRP + seam quality pass finds no HIGH blockers. Envelope<T>'s single responsibility is correctly stated and cleanly bounded: it owns the 48-byte header encode/decode, the version-dispatch decision tree, and the endianness pin — everything else is delegated through the registry and dispatcher. The hot/cold path split (§5) is coherent and the phase-ownership refusal (§6 — no phase, synchronous leaf) correctly prevents phase-coupling. The public API surface returns std::expected<T, data::Error> at every typed boundary and degrades to POD spans at the extern C trampolines, satisfying error-model.md Composition Rules and PHILOSOPHY §11. Concurrency guarantees in §6 are sound (pure functions over disjoint args, read-only shared registry). The two MED findings both concern amendment-scope tracking gaps that will surface as implementation-PR conflicts if not resolved: (1) three new public free functions — peek_header, serialize_header, deserialize_header — are introduced without a §12 open-question tracking their SPEC §5 amendment, so the impl-PR reviewer has no checklist entry; (2) four error arm names in the §10 table — BadMagic, PayloadTruncated, VersionUnsupported, BufferTooSmall — lack ErrorTag numeric values, and their disposition (new tag vs sub-case of existing tag) is stated inconsistently, which will cause a compile-time mismatch between the impl headers and the SPEC §10.1 enum. The LOW finding is that §12 [OPEN] #1's amendment-scope note omits five of the six structural deltas between the SPEC §5 EnvelopeHeader stub and the design's §3.1 layout. Neither MED nor LOW finding requires reopening the PR; they are addressed by the follow-up work already queued in PR #894 or by extending §12 [OPEN] items there.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:1 (R1 HIGH-1 SchemaSourceMismatch->SourceHashMismatch rename; addressed in open follow-up PR #894, not yet merged)

notes:The round-2 cohesion + SRP + seam quality pass finds no HIGH blockers. Envelope<T>'s single responsibility is correctly stated and cleanly bounded: it owns the 48-byte header encode/decode, the version-dispatch decision tree, and the endianness pin -- everything else is delegated through the registry and dispatcher. The hot/cold path split (section 5) is coherent and the phase-ownership refusal (section 6 -- no phase, synchronous leaf) correctly prevents phase-coupling. The public API surface returns std::expected<T, data::Error> at every typed boundary and degrades to POD spans at the extern C trampolines, satisfying error-model.md Composition Rules and PHILOSOPHY section 11. Concurrency guarantees in section 6 are sound (pure functions over disjoint args, read-only shared registry). The two MED findings both concern amendment-scope tracking gaps that will surface as implementation-PR conflicts if not resolved: (1) three new public free functions -- peek_header, serialize_header, deserialize_header -- are introduced without a section 12 open-question tracking their SPEC section 5 amendment, so the impl-PR reviewer has no checklist entry; (2) four error arm names in the section 10 table -- BadMagic, PayloadTruncated, VersionUnsupported, BufferTooSmall -- lack ErrorTag numeric values, and their disposition (new tag vs sub-case of existing tag) is stated inconsistently, which will cause a compile-time mismatch between the impl headers and the SPEC section 10.1 enum. The LOW finding is that section 12 [OPEN] #1's amendment-scope note omits five of the six structural deltas between the SPEC section 5 EnvelopeHeader stub and the design's section 3.1 layout. Neither MED nor LOW finding requires reopening the PR; they are addressed by the follow-up work already queued in PR #894 or by extending section 12 [OPEN] items there.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:1 (R1 HIGH-1 SchemaSourceMismatch->SourceHashMismatch rename; addressed in open follow-up PR #894, not yet merged)

notes:The round-2 cohesion + SRP + seam quality pass finds no HIGH blockers. Envelope<T>'s single responsibility is correctly stated and cleanly bounded: it owns the 48-byte header encode/decode, the version-dispatch decision tree, and the endianness pin -- everything else is delegated through the registry and dispatcher. The hot/cold path split (section 5) is coherent and the phase-ownership refusal (section 6 -- no phase, synchronous leaf) correctly prevents phase-coupling. The public API surface returns std::expected<T, data::Error> at every typed boundary and degrades to POD spans at the extern C trampolines, satisfying error-model.md Composition Rules and PHILOSOPHY section 11. Concurrency guarantees in section 6 are sound (pure functions over disjoint args, read-only shared registry). The two MED findings both concern amendment-scope tracking gaps that will surface as implementation-PR conflicts if not resolved: (1) three new public free functions -- peek_header, serialize_header, deserialize_header -- are introduced without a section 12 open-question tracking their SPEC section 5 amendment; (2) four error arm names in the section 10 table -- BadMagic, PayloadTruncated, VersionUnsupported, BufferTooSmall -- lack ErrorTag numeric values and their disposition is stated inconsistently. The LOW finding is that section 12 [OPEN] #1 omits five of the six structural deltas between the SPEC section 5 EnvelopeHeader stub and the design's section 3.1 layout. None of these require reopening the merged PR; they extend the work queued in PR #894.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4f7ca followup:#894

Added §12 [OPEN] #8 to `specs/data/envelope-serdes-design.md` (lines 1214–1257 in the updated file on branch `fix/data-envelope-serdes-design-followup-r1`). The new entry enumerates the three free-function signatures (`peek_header`, `serialize_header`, `deserialize_header`) with their full `namespace glibre::types` placement, cites two concrete callers per PHILOSOPHY §10 — `tools/glibre-cook` (needs `serialize_header` to patch `payload_len` after body serialization) and `tools/glibre-editor` inspector panel (needs `peek_header` + `deserialize_header` for offline asset-browser schema inspection) — and marks the entry as blocking the implementation PR that will introduce `data/runtime/src/envelope.cpp`, co-gated with [OPEN] #1 and [OPEN] #7.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4f7ca followup:#894

Extended §12 [OPEN] #1 in `specs/data/envelope-serdes-design.md` with explicit `data::Error` closed-sum dispositions for all four arms that lacked SPEC §10.1 `ErrorTag` entries (lines 1116–1150 in the updated file). Chose option (a) for three and option (b) for one, with rationale: `BadMagic` → new tag 11 (externally observable before any field is decoded; distinct from `EnvelopeTruncated` and `DeserializeError`); `PayloadTruncated` → new tag 12 (header decoded cleanly but body cut short; callers including the cooker's blob-walker need to distinguish from header-truncation); `BufferTooSmall` → new tag 13 (serialize-path only; no existing arm covers a pre-check refusal; callers need it to retry with a larger buffer). `VersionUnsupported` → sub-case (b) of `DeserializeError` (tag 3) because the §10 table already documents it as a sub-case kept distinct only for log clarity; discriminator is `at.version > entry.version`, no new enum arm needed. The entry closes by stating that SPEC §10.1 must be extended with arms 11, 12, and 13 in the same implementation PR.

specs/data/envelope-serdes-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4f7ca followup:#894

Added the five structural deltas between SPEC §5 `EnvelopeHeader` and design §3.3 `EnvelopeHeader` to §12 [OPEN] #1's amendment-scope list in `specs/data/envelope-serdes-design.md` (lines 1101–1114 in the updated file): (1) `magic` field added as `eastl::array<std::byte, 4>`; (2) `SchemaId schema{}` renamed and retyped to `std::uint32_t schema_fqn_id{0}` (string-view borrow → registry ordinal); (3) `source_hash` field added as `SchemaSourceHash` (32-byte BLAKE3-256); (4) `flags` field removed; (5) `payload_length` → `payload_len` rename. Each delta cites the relevant design section for traceability.

cjhowe-us:
agent:go-impl-respond
round:2
pr:#894
commit:6d4f7ca
branch:fix/data-envelope-serdes-design-followup-r1

summary:
- ADDRESSED: 3 (MED-1, MED-2, LOW)
- PUSHBACK: 0
- DEFER: 0
- NOOP: 0

changes:
All three R2 findings addressed in a single commit (6d4f7ca) on PR #894
(`fix/data-envelope-serdes-design-followup-r1`), modifying
`specs/data/envelope-serdes-design.md`:

MED-1 (comment 3188400868, line 542): Added §12 [OPEN] #8 enumerating
the three free-function additions to SPEC §5 (`peek_header`,
`serialize_header`, `deserialize_header`) with full signatures and
namespace placement, two concrete callers (`glibre-cook` and
`glibre-editor` inspector panel) per PHILOSOPHY §10
two-concrete-users rule, and a blocking-condition note tied to the
`envelope.cpp` implementation PR.

MED-2 (comment 3188402565, line 938): Extended §12 [OPEN] #1 with
explicit error-tag dispositions: `BadMagic` → new tag 11,
`PayloadTruncated` → new tag 12, `BufferTooSmall` → new tag 13
(option a — distinct externally-observable failure modes);
`VersionUnsupported` → sub-case of `DeserializeError` tag 3
(option b — pure internal sub-case with discriminator note).
Documents that SPEC §10.1 must be extended with arms 11, 12, 13.

LOW (comment 3188404031, line 1066): Added five-point structural-delta
list to §12 [OPEN] #1 amendment-scope: magic field added, SchemaId
schema → schema_fqn_id rename+retype, source_hash added, flags removed,
payload_length → payload_len rename.

Do NOT enable auto-merge on #894 — R3 review runs next.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:2
carryover:0

notes:The round-3 topology + spec alignment + polish lens passes DAG coherence and CC subject checks. The design's parent hierarchy is correct (issue #736 type:spike, no pts:* label, parent #729 type:sub-epic under epic #49, domain:data — AGENTS.md compliant). Commit subjects use correct scopes: commit 1 is fix(data): with refs #736; commit 2 is docs(data): with refs #832. Both are present. The SourceHashMismatch rename from R1 (HIGH-1) is confirmed resolved: arm is now consistently referenced as SourceHashMismatch throughout the document. The five structural deltas in §12 [OPEN] #1 (magic added, SchemaId→schema_fqn_id retype, source_hash added, flags removed, payload_length→payload_len rename) are clearly enumerated and traceable to §3.1. The §12 [OPEN] #8 two-concrete-users rule citation for peek_header / serialize_header / deserialize_header names glibre-cook and glibre-editor inspector, both of which appear in the design corpus (CLAUDE.md repo layout, tools/SPEC.md §4.6 AssetBrowser, specs/render/pso-cache-design.md). However the OPEN HIGH below blocks the §12 [OPEN] #1 disposition for tags 11/12/13.

HIGH-1: specs/data/envelope-serdes-design.md line 1120 — §12 [OPEN] #1 tag dispositions (BadMagic=11, PayloadTruncated=12, BufferTooSmall=13) contradict the already-merged, three-round-reviewed specs/data/data-error-design.md (PR #838, R3 verdict COMMENT with 0 HIGH findings). The data-error-design.md §3.1.1 explicitly performs an Occam-collapse audit that maps BadMagic onto EnvelopeTruncated (arm 9) — quote: "a wrong magic prefix is physically an early-truncation-shaped failure from the consumer's perspective" — and PayloadTruncated onto EnvelopeTruncated (arm 9) — quote: "The two are distinguishable only by at.schema-known vs at.schema-default; the recovery is identical." BufferTooSmall is deferred (not assigned any tag) in data-error-design.md, not assigned tag 13. The envelope-serdes §12 [OPEN] #1 dispositions authored in the R2 response were produced without cross-referencing data-error-design.md, which has authority over the ErrorTag closed sum (its entire §1 declares: "The closed enum glibre::types::data::ErrorTag and the trivially … are this aggregate's deliverable"). Proposing BadMagic=11 and PayloadTruncated=12 as distinct arms in a follow-up to #832 without first reconciling with data-error-design.md creates a cross-document seam incoherence that will block any implementation PR (the impl PR cannot simultaneously satisfy both designs). Fix: §12 [OPEN] #1 must be updated to acknowledge data-error-design.md's binding Occam-collapse decisions and either (a) accept the collapses (BadMagic and PayloadTruncated collapse onto EnvelopeTruncated arm 9; BufferTooSmall deferred per data-error-design.md §12 [OPEN]) and revise the §10 failure table to reflect which ErrorTag value each arm actually uses, OR (b) open a new spike explicitly challenging data-error-design.md §3.1.1 on the BadMagic/PayloadTruncated collapses, citing concrete callers that need typed dispatch, and allowing data-error-design.md §12 to supersede the prior collapse decision through its own review pipeline. Per PHILOSOPHY §10, option (a) is preferred unless two concrete callers need the discriminated dispatch that the new arms would provide. Note: VersionUnsupported as a sub-case of DeserializeError (tag 3) is consistent with data-error-design.md §3.1.1 row "VersionUnsupported → DeserializeError (3)" and requires no change.

LOW-1: specs/data/envelope-serdes-design.md §10 failure table — the "core::Error mapping" column for BadMagic and PayloadTruncated says "none — passed through" but data-error-design.md §3.2 per-source trigger table collapses both onto EnvelopeTruncated (arm 9) whose core::Error mapping is also "none — passed through" (EnvelopeTruncated does not wrap to a core::Error arm). This means even if the resolution of HIGH-1 adopts option (a), the core::Error column is consistent. No action needed on this column; it is self-consistent under either resolution path.

LOW-2: specs/data/envelope-serdes-design.md §11.1 malformed-input tests — the test case envelope_bad_magic asserts return of BadMagic{at.offset=0} and envelope_payload_truncated asserts PayloadTruncated{at.offset=148}. If HIGH-1 is resolved via option (a) — collapsing onto EnvelopeTruncated — then these test names and their asserted error arm names must change to EnvelopeTruncated. The test plan is forward-looking (feeds into the implementation plan issues), so the mismatch is not yet load-bearing, but §12 [OPEN] #1 should note the cascading test-plan consequence to avoid silent divergence when the impl plan reads this document. Fix: add a sub-bullet in §12 [OPEN] #1 noting that §11.1 test names and asserted arms cascade from the tag-disposition resolution.

specs/data/envelope-serdes-design.md
cjhowe-us, line 1066: severity:HIGH location:specs/data/envelope-serdes-design.md:1066 problem:The §12 [OPEN] #1 amendment scope as extended by R2 response (commit 6d4f7ca on #894 branch) proposes BadMagic=tag 11, PayloadTruncated=tag 12, and BufferTooSmall=tag 13 as new ErrorTag arms. These dispositions contradict the already-merged, three-round-reviewed specs/data/data-error-design.md (PR #838). That design's §3.1.1 Occam-collapse audit explicitly maps BadMagic onto EnvelopeTruncated (arm 9) — 'a wrong magic prefix is physically an early-truncation-shaped failure from the consumer's perspective' — and PayloadTruncated onto EnvelopeTruncated (arm 9). BufferTooSmall is deferred in data-error-design.md with no tag assigned. data-error-design.md is the authority over the ErrorTag closed sum (its §1 declares this). The dispositions were authored in the R2 response without acknowledging data-error-design.md. The §10 failure table and §11.1 test names that assert BadMagic{} and PayloadTruncated{} as distinct arms cascade from this conflict. recommendation: Update §12 [OPEN] #1 in #894 to cite data-error-design.md §3.1.1 and either (a) accept its Occam collapses — revising the §10 table to show BadMagic and PayloadTruncated as sub-forms of EnvelopeTruncated (arm 9) and BufferTooSmall as deferred per that design's §12 [OPEN] — or (b) open a new spike explicitly contesting data-error-design.md §3.1.1 on the BadMagic and PayloadTruncated rows, citing concrete callers that require typed dispatch, before assigning tags 11 and 12. The R2 response's VersionUnsupported disposition (sub-case of DeserializeError, tag 3) is consistent with data-error-design.md §3.1.1 and requires no change.

cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:0e2050f followup:#894

Adopted the `data-error-design.md` §3.1.1 Occam-collapse in full (option (a)). Edits on branch `fix/data-envelope-serdes-design-followup-r1`, commit `0e2050f`:

1. **`specs/data/envelope-serdes-design.md` §10 preamble** (lines ~950–956): Replaced the reference to "this design proposes new arms" with an explicit attribution to `data-error-design.md` as the authoritative owner of the `ErrorTag` closed sum. States that no new arms are added by envelope-serdes.

2. **§10 failure table** (lines ~960, ~962, ~969): `BadMagic` row now annotated "discriminator only — maps to `EnvelopeTruncated` arm 9" with `core::Error mapping` citing `data-error-design.md` §3.1.1 Occam-collapse. `PayloadTruncated` row similarly annotated. `BufferTooSmall` row now annotated "deferred — no ErrorTag arm assigned" and cites `data-error-design.md` §12 [OPEN] for the deferral.

3. **§11.1 malformed-input test table** (lines ~1031, ~1033): `envelope_bad_magic` now asserts `EnvelopeTruncated{at.offset = 0}` (with discriminator note). `envelope_payload_truncated` now asserts `EnvelopeTruncated{at.offset = 148, at.schema = header_decoded}` (with discriminator note). Test names retained — they describe the trigger, not the arm.

4. **§12 [OPEN] #1 disposition record** (lines ~1118–1166): R2 entries proposing tags 11, 12, 13 replaced with revised dispositions citing `data-error-design.md` §3.1.1 as authoritative. Records the Occam-collapses for BadMagic and PayloadTruncated, the deferral for BufferTooSmall, and notes that VersionUnsupported (DeserializeError sub-case) was already consistent. Withdrawal note for the §10.1 enum extension (arms 11/12/13). Implementation note for plan authors confirming test arm names.

Issue #742, opened 2026-05-03
[SPIKE] design-data-data-error-detailed


cjhowe-us:
Parent: #729

## Question

Produce the detailed design for the **data-error** aggregate in the `data` context.

Closed-sum data::Error.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/data/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/core-runtime/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **data-error** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/data/data-error-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(data): detailed design — data-error (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-data-data-error-detailed` is blocked by this issue.

cjhowe-us:
agent:design-data-data-error
status:done
issue:#742
branch:docs/data-data-error-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a64508c8a691f7bed
host:mac
cloud:none
commit:63c6477b9915fe24bd11517bdc20005aaf5952c3
pr:#838
[...]

Issue #738, opened 2026-05-03
[SPIKE] design-data-migration-dispatcher-detailed


cjhowe-us:
Parent: #729

## Question

Produce the detailed design for the **migration-dispatcher** aggregate in the `data` context.

Per-version function chain.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/data/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/core-runtime/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **migration-dispatcher** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/data/migration-dispatcher-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(data): detailed design — migration-dispatcher (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-data-migration-dispatcher-detailed` is blocked by this issue.

cjhowe-us:
agent:design-data-migration-dispatcher
status:done
issue:#738
branch:docs/data-migration-dispatcher-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a34796d0910a323f5
host:mac
cloud:none
commit:0fa951132673ecfd84e74bf10ba36bb1f1606463
pr:#835
[...]

Issue #734, opened 2026-05-03
[SPIKE] design-data-middleman-dylib-detailed


cjhowe-us:
Parent: #729

## Question

Produce the detailed design for the **middleman-dylib** aggregate in the `data` context.

glibre-types.dylib build target.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/data/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/core-runtime/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **middleman-dylib** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/data/middleman-dylib-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(data): detailed design — middleman-dylib (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-data-middleman-dylib-detailed` is blocked by this issue.

cjhowe-us:
agent:design-data-middleman-dylib
status:done
issue:#734
branch:docs/data-middleman-dylib-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a53835c84508103b3
host:mac
cloud:none
commit:b143ef281351c47c9a6d58ba8e221b4eacd74ade
pr:#834
[...]

Issue #732, opened 2026-05-03
[SPIKE] design-data-flatbuffers-codegen-detailed


cjhowe-us:
Parent: #729

## Question

Produce the detailed design for the **flatbuffers-codegen** aggregate in the `data` context.

glbr-sergeant CLI codegen.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/data/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/core-runtime/serialization.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **flatbuffers-codegen** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/data/flatbuffers-codegen-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(data): detailed design — flatbuffers-codegen (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-data-flatbuffers-codegen-detailed` is blocked by this issue.

cjhowe-us:
agent:design-data-fory-codegen
status:done
issue:#732
branch:docs/data-fory-codegen-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a353827da1ffcbd05
host:mac
cloud:none
commit:ded7051d2600dd78c39d0ad3a61a13b2734fd707
pr:#836
[...]

PR #838, opened 2026-05-04
docs(data): detailed design — data-error (refs #742)

cjhowe-us:
## Summary

- Closed sum `specs/data/data-error-design.md` for the data-context error aggregate (refs #742)
- Unifies failure surfaces from schema-registry #730, envelope-serdes #736, migration-dispatcher #738, fory-codegen runtime side #732, middleman-dylib #734
- Nine `ErrorTag` arms locked from SPEC §10.1; Occam-collapse audit demotes fourteen sibling-proposed candidates onto existing arms or to §12 [OPEN] entries (`SourceHashMismatch`, `BufferTooSmall`, `SchemaSourceMismatch` deferred for SPEC §10 amendments)

## Test plan

- [ ] Sibling task-breakdown spike unblocks once this design lands
- [ ] §11 unit-test plan (one Catch2 case per arm + cross-aggregate uniformity + golden-per-source) feeds the implementation plans
- [ ] §10 meta-failure rule (no catch-all `Internal` arm; new surfaces require SPEC amendment) audited at every sibling implementation plan

### DIFF
--- a/specs/data/data-error-design.md
+++ b/specs/data/data-error-design.md
@@ -1,0 +1,1199 @@
+# data — Detailed Design: data-error aggregate
+
+> Detailed design for the closed sum `glibre::types::data::Error`
+> declared in `specs/data/SPEC.md` §5 §error.hpp / §5.3 / §10. Refines
+> those sections in place; cites
+> `reviews/decisions/error-model.md`,
+> `reviews/decisions/fory-codegen.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`,
+> `reviews/decisions/perf-budget.md`, and
+> `reviews/decisions/frame-phases.md`. Style siblings on `main`:
+> `specs/platform/platform-error-design.md`,
+> `specs/data/schema-registry-design.md`,
+> `specs/data/envelope-serdes-design.md`,
+> `specs/data/middleman-dylib-design.md`. Introduces no new public
+> surface beyond what SPEC §5 / §10 already enumerate; deviations from
+> the cited records would require an amendment spike, not an in-place
+> edit.
+>
+> Refs: spike #742 — `[SPIKE] design-data-data-error-detailed`.
+> Parent #729. Sibling task-breakdown spike is blocked by this
+> deliverable. All conclusions re-derived; harmonius prior art
+> (`harmonius/docs/design/core-runtime/`,
+> `harmonius/docs/requirements/data-systems/`) is research input
+> only — re-derived against PHILOSOPHY, never ported.
+
+## 1. Purpose
+
+The data-error aggregate is the **single closed enumeration** that
+unifies every failure surface the `data` context can raise across its
+public API boundaries. Its one responsibility — the only reason it
+changes — is **"a sibling data aggregate gained or lost a failure
+surface"**: a new envelope refusal arrives from envelope-serdes, the
+schema-registry adds a source-hash drift refusal, the migration
+dispatcher learns to detect a chain back-edge, the middleman static
+init wires a second validation pass. Concretely the aggregate owns:
+
+1. The closed enum `glibre::types::data::ErrorTag` and the trivially
+   copyable `glibre::types::data::Error` struct that carries it (SPEC
+   §10.1; restated in §3 below). Adding / removing / renaming an arm
+   is a deliberate central edit gated by SPEC §4.3 inv #3 (SONAME
+   bump) plus a fresh `data::Error` schema version (§7).
+2. The per-source translation guidance — for each sibling data
+   aggregate (schema-registry #730, envelope-serdes #736,
+   migration-dispatcher #738, fory-codegen runtime side #732,
+   middleman-dylib #734), which `ErrorTag` arm fires under which
+   trigger, which payload fields the arm populates, and which
+   `core::Error` arm wraps it at the loader's call site (SPEC §10.2,
+   §10.3).
+3. The construction discipline that keeps `Error` off the success
+   path: trivially copyable POD aggregate, no allocation in
+   construction, no I/O, no virtual dispatch, no TLS write at the
+   value's level (SPEC §10.4 logging discipline; PHILOSOPHY §6
+   "zero runtime reflection").
+4. The Fory schema for the *log/replay carrier* — a small,
+   stable-on-the-wire `(tag, payload-fields-by-arm)` record that
+   `glibre::log_error` and the e2e replay harness use to round-trip
+   a `data::Error` through `spdlog`'s structured field set without
+   losing arm identity (§7).
+5. The arm-count invariant tying `ErrorTag` to the data-context ABI
+   hash via the `DataErrorRecord` schema (§7.3): every arm-count
+   bump rebuilds `glibre-types.dylib` and forces every plugin to
+   re-link, the same gate that catches schema-shape drift
+   (`reviews/decisions/plugin-abi.md` §"Versioning Rules").
+
+What this aggregate explicitly **refuses to own**:
+
+- **`core::Error`.** `PluginAbiHashMismatch`, `PluginInitFailed`,
+  `SchemaMigrationFailed`, `HotReloadRefused`, `FramePhaseMisordered`,
+  `OutOfBudget`, `PluginDlopenFailed`, `PluginMissingEntryPoint`,
+  `PluginManifestInvalid`, `PluginEngineTooOld`, `PluginNameCollision`,
+  `PluginDependencyMissing`, `SystemScheduleCycle` are owned by `core`
+  per `reviews/decisions/error-model.md` Type Sketch, `core/SPEC.md`
+  §10, and `reviews/decisions/plugin-abi.md` §"Failure Modes →
+  core::Error". A data-side translator that fabricated a `core::Error`
+  would violate composition rule 1 ("per-context enums are leaves;
+  nothing nests another context's enum inside its own").
+- **`platform::Error` / `render::Error` / `physics::Error` /
+  any other context's enum.** Same rule. Each context owns its own
+  closed sum and translates at the call site that crosses the
+  boundary, never inward-facing.
+- **The per-aggregate failure semantics.** Which arms
+  `Envelope<T>::deserialize` may return, the recovery contract for
+  `MigrationStepMissing` against an inbound payload from a
+  decommissioned schema, and the static-init death-test discipline
+  for `SchemaRegistryConflict` are owned by the sibling designs
+  (`specs/data/envelope-serdes-design.md` §10,
+  `specs/data/schema-registry-design.md` §10,
+  `specs/data/middleman-dylib-design.md` §10). This aggregate
+  enumerates the typed arm; the sibling decides when to raise it.
+- **The `glibre::log_error` formatter.** Owned by `core`. Data
+  contributes the structured payload shape (§7) and the arm-name
+  table; core's helper does the spdlog dispatch.
+- **Cross-arm retry / chain.** Each data aggregate constructs its own
+  `Error` at the site the failure happens (SPEC §10.4); this aggregate
+  does not auto-collapse one arm into another. A
+  `MigrationStepMissing` and a `SchemaMigrationFailure` collapse onto
+  one `core::Error::SchemaMigrationFailed` only at the loader's call
+  site (`reviews/decisions/plugin-abi.md` §"Failure Modes" row 11),
+  and that collapse lives in `core`'s code, not here.
+- **The choice of `magic_enum` vs hand-written `to_string`** for
+  arm-name logging. Owned by `core/error.hpp` per `error-model.md`
+  Open Q #1; this design consumes whatever core decides without
+  forking. The only data-side requirement is that the chosen approach
+  round-trips every arm name (§7.4).
+- **Domain `Error` enums on top of the spine.** Future game-framework
+  plugins that author persistent types via `.fory` schemas may surface
+  domain-side failures (`InventoryOverflow`, `EffectStackTooDeep`).
+  Those are *that plugin's* enum, not `data::Error`. The spine
+  guarantees byte round-trip and migration; domain semantics are out
+  of scope (SPEC §1, §3.3).
+
+The aggregate's SRP boundary is the sharpest in the data context: if a
+sibling data aggregate gains a refusal it cannot collapse onto an
+existing arm, this design changes (one new `ErrorTag` value, one new
+row in §3.2, one new payload-fields cell in §3.3, one schema version
+bump in §7). Anything else is out of scope.
+
+## 2. Requirements coverage
+
+Mapping of harmonius requirements consulted as research input
+(`PHILOSOPHY.md` §"How harmonius is used" — re-derived, not ported)
+onto the MVP coverage in this aggregate. Every entry is independently
+re-derived; coverage does not imply harmonius's design was correct,
+only that the underlying *requirement* survives the re-derivation.
+
+| Harmonius source clause                                                                       | Re-derived MVP requirement                                                                                                            | Coverage in this aggregate                                                                                                                                  |
+|-----------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `core-runtime/error-handling.md` (engine-wide typed errors, no exceptions across ABI)         | Every public data boundary returns `std::expected<T, data::Error>`; raw OS / library integers never cross the §5 surface.             | §3 closed enum + §4 public surface; §5 hot/cold split (success path never constructs an `Error`).                                                            |
+| `data-systems/attributes-effects.md` (rkyv archive failure modes scattered across contexts)   | Refused: per-domain failure semantics live in their plugin contexts. The data spine surfaces only schema / migration / ABI failures. | §1 Refusals — domain-side error enums are out of scope; §3.2 covers only the spine's seven sibling-aggregate triggers.                                       |
+| `data-systems/data-tables.md` (per-row constraint validation, FK resolution failures)         | Refused: the spine guarantees byte round-trip; domain validation is the consuming context's responsibility.                          | §1 Refusals; §3.2 has no `ConstraintViolation` arm.                                                                                                          |
+| `data-systems/directed-graphs.md` (cycle detection, traversal failures)                       | Re-purposed: `MigrationCycle` is a *meta-cycle* in the migration chain itself, not a domain-graph cycle. Domain cycles stay in plugin code. | §3.2 row `MigrationCycle` — codegen-time / static-init only; runtime cannot observe.                                                                          |
+| `core-runtime/plugin-abi.md` (refuse load on hash mismatch)                                   | The data-error aggregate types the refusal; `core` raises the `core::Error::PluginAbiHashMismatch` wrap.                              | §3.2 row `AbiHashMismatch`; §3.4 wrapping table.                                                                                                              |
+| `core-runtime/hot-reload.md` (drain → swap → migrate → resume; migration failures are typed)  | The migration arms (`SchemaMigrationFailure`, `MigrationStepMissing`, `MigrationCycle`) cover every refusal mode the dispatcher raises. | §3.2 rows 2 / 7 / 8; §3.4 collapses 2 + 7 onto `core::Error::SchemaMigrationFailed` at the loader's call site.                                              |
+| `data-systems/composition.md` (immutable definitions ↔ ECS components binding)                 | Refused: binding lives in `core` (ECS + plugin loader); `data` only types the schema-side failures.                                  | §1 Refusals — `core::Error::PluginInitFailed` is core's, not data's.                                                                                          |
+
+Refusals routed elsewhere (research input from harmonius and from
+sibling decision records that **does not** belong in this aggregate,
+recorded so the boundary is re-traceable):
+
+- **Per-loader-step error arms** (`PluginDlopenFailed`,
+  `PluginMissingEntryPoint`, `PluginManifestInvalid`,
+  `PluginEngineTooOld`, `PluginNameCollision`,
+  `PluginDependencyMissing`, `SystemScheduleCycle`,
+  `PluginInitFailed`). Refused — owned by `core` per
+  `reviews/decisions/plugin-abi.md` §"Failure Modes → core::Error".
+  Data raises only the four arms `core` wraps as
+  `PluginAbiHashMismatch` and `SchemaMigrationFailed` (and the
+  `HotReloadRefused` umbrella from `error-model.md`).
+- **Auto-retry policies.** Refused — caller decision per SPEC §10.2
+  recovery vocabulary; the spine exposes `refuse decode` / `refuse
+  load` / `abort build` / `process abort`, never an in-engine retry.
+- **Localised error messages.** Refused — `data::Error` payloads
+  carry only stable identifiers (FQN, version pair, hex hash, byte
+  offset, tag number); UI-side localisation is a `tools` /
+  future-`l10n` concern. SPEC §10.4 makes the handler responsible for
+  prose, never the constructor.
+
+## 3. Detailed model
+
+### 3.1 The closed `data::Error` aggregate
+
+Restated from SPEC §10.1 for the design audit; this section adds the
+*rationale* per arm, the *invariants* that govern addition / removal /
+renaming, and the *Occam-collapse audit* that pinned the arm count at
+nine.
+
+```cpp
+// glibre-types.dylib: data/runtime/include/glibre/types/error.hpp
+//
+// Public surface; mirrors specs/data/SPEC.md §5 §error.hpp and §10.1
+// verbatim. Adding a payload field to one arm is a SPEC §10
+// amendment + a §7 schema-version bump, not an in-place edit.
+
+namespace glibre::types::data {
+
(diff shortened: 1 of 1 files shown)


specs/data/data-error-design.md
cjhowe-us, line 1: severity:HIGH location:specs/data/data-error-design.md:1 problem:The design doc preamble states "Introduces no new public surface beyond what SPEC §5 / §10 already enumerate." However SPEC.md §5 (the public-interface header stub, line ~607) still declares `enum class Error : std::uint16_t` with exactly five bare arms and no payload fields, while SPEC.md §10.1 (and this design doc §3.1) replace that with a completely different shape: `ErrorTag` enum + `Error` struct + `WireSite` struct. The design doc does not amend the §5 stub to match §10.1. This leaves SPEC.md §5 and §10 internally inconsistent: a consumer reading §5 would implement a 5-arm enum, while §10 and this design mandate a 9-arm tagged-union struct. The design doc should have either (a) included an explicit §5 amendment strike-through showing the old 5-arm enum being superseded, or (b) noted the §5 stub as an obsolete draft that §10 supersedes, with a forward pointer. fix:Add a §0 or top-of-doc note explicitly stating that this design amends SPEC.md §5 (the public interface header stub `error.hpp` section) to align with §10.1. Strike through or annotate the §5 `enum class Error` stub in SPEC.md to mark it as superseded by the `ErrorTag + Error` struct declared in §10.1. A follow-up plan PR should reconcile the two sections in-place.

cjhowe-us, line 1: severity:MED location:specs/data/data-error-design.md:224 problem:The `Error` struct in the design doc §3.1 declares `host_hash` and `plugin_hash` as `eastl::string_view`, but SPEC.md §10.1 (the normative source) declares these same fields as `std::string_view`. The engine-wide PHILOSOPHY §11 rule is that `eastl::` replaces `std::` containers and string types — so `eastl::string_view` is correct per PHILOSOPHY. However SPEC.md §10.1 uses `std::string_view`, meaning either SPEC.md §10.1 has a PHILOSOPHY §11 violation or the design doc disagrees with the spec on which type to use. Both cannot be correct simultaneously, and the spec is the normative surface. fix:Confirm which type is authoritative. If `eastl::string_view` is correct (per PHILOSOPHY §11), amend SPEC.md §10.1 to use `eastl::string_view` for `host_hash` and `plugin_hash`. If `std::string_view` is intentional (e.g. because `glibre-types.dylib` must not expose EASTL across its dylib boundary per PHILOSOPHY §11 final sentence), update the design doc §3.1 to match and add a comment citing the exception. Either resolution requires a SPEC amendment spike or a correction PR.

cjhowe-us, line 1: severity:MED location:specs/data/data-error-design.md:340 problem:In §3.3 the payload-fields-by-arm table marks `at.offset` as YES for `SchemaUnknown` (arm 6), but §3.2 row for `SchemaUnknown` states `at.offset = 0` (the failure is at the envelope header). A YES cell in §3.3 conventionally means the field carries meaningful, non-sentinel data. Populating `at.offset` with a fixed sentinel value of 0 for `SchemaUnknown` means the field is effectively unused — the offset is always 0 and carries no diagnostic information. This contradicts the intent of the YES marking (to distinguish active payload fields from default-constructed ones) and could mislead implementors into thinking the offset will vary. The same issue applies to `at.version` which is marked YES but §3.2 says it is populated from the envelope — this one is genuinely meaningful. fix:Change the `at.offset` cell for `SchemaUnknown` to `YES (=0, sentinel)` (mirroring the `EnvelopeTruncated` pattern that already uses `YES (opt)`) or to a `-` with a footnote that the offset is always 0 and the envelope header carries the diagnostic. Update §3.2 to be explicit that `at.offset = 0` for `SchemaUnknown` is a sentinel, not an opaque byte count.

cjhowe-us, line 1: severity:LOW location:specs/data/data-error-design.md:302 problem:§3.2 "Per-source trigger / payload table" states the table is "exhaustive against the sibling designs as of the spike's read date" and instructs that new sibling aggregates "MUST add a row here in the same PR that introduces the surface." This is an enforcement obligation placed on a doc-only PR — the obligation has no compile-time enforcement, no CI check, and no lint. The instruction means that future plan PRs for schema-registry, envelope-serdes, migration-dispatcher, etc. must amend this table, but there is currently no mechanism to remind or enforce that. The §11 golden-per-source tests (§11.6) are intended to catch drift at CI, but those tests do not exist yet. fix:Add a cross-reference in the design doc §3.2 footer to the specific acceptance criterion or CI gate that enforces the exhaustiveness claim (e.g. "§11.6 golden-per-source tests enforce this; any PR adding a new sibling-aggregate failure surface without a corresponding golden fails CI"). This makes the obligation mechanical rather than normative-prose-only. If the CI gate is not yet designed, note it as an [OPEN] in §12.

cjhowe-us, line 1: severity:LOW location:specs/data/data-error-design.md:591 problem:§5.4 introduces an optional `source_tag` TLS slot that "SHALL NOT introduce a separate TLS allocation" and instead "shares the `__thread` struct platform owns (§6.3)." However, the `platform` context TLS struct is neither cited in this doc with a concrete symbol name nor is the shared ownership formally negotiated via the platform SPEC or a decision record. A data aggregate silently borrowing into another context's TLS struct is a cross-context coupling that violates SRP (PHILOSOPHY §1, §2). fix:Either (a) demote `source_tag` to a data-context-owned TLS slot (separate struct, separate `__thread` variable) with a MAX 8-byte or 16-byte cap documented here, or (b) open a cross-context TLS sharing decision record (`reviews/decisions/tls-layout.md`) that both `platform` and `data` cite, formalizing the shared struct ownership. Option (a) is simpler and avoids the coupling; option (b) is necessary only if the shared struct is a deliberate invariant other contexts must also join.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:2
  LOW:2

carryover:0

notes: The core deliverable — a 1 199-line closed-sum design for `glibre::types::data::Error` with nine `ErrorTag` arms, per-source trigger/payload tables, Occam-collapse audit of 14 candidate arms, `DataErrorRecord` Fory schema, hot/cold path split, concurrency contract, hot-reload disposition table, and a complete §11 test plan — is substantively correct and well-grounded. The nine arms match SPEC.md §10.1 exactly; the §3.4 `core::Error` wrapping table matches `reviews/decisions/plugin-abi.md`; all five referenced sibling aggregates (#730, #732, #734, #736, #738) have documented detection sites; the three deferred arms (`SourceHashMismatch`, `BufferTooSmall`, `SchemaSourceMismatch`) are correctly pushed to §12 rather than added as catch-alls; the Conventional Commit subject `docs(data): detailed design — data-error (refs #742)` is correctly scoped and correctly uses `refs` (not `closes`, since the spike issue stays open). One HIGH finding must be tracked: SPEC.md §5 still carries the old 5-arm `enum class Error` stub with no payload fields while SPEC.md §10.1 (and this design) use the 9-arm `ErrorTag + Error` struct — the design doc does not amend §5, leaving the spec internally inconsistent for any implementor who reads §5 first. One MED finding concerns a `eastl::string_view` vs `std::string_view` type mismatch for `host_hash`/`plugin_hash` between the design doc and SPEC.md §10.1; PHILOSOPHY §11 mandates `eastl::` but the spec uses `std::`, and the public-ABI exception (PHILOSOPHY §11 last sentence) may or may not apply here — a decision is needed. A second MED finding notes that the §3.3 payload table marks `at.offset` as a plain YES for `SchemaUnknown` even though §3.2 states the offset is always 0 (a sentinel), which could mislead implementors. Two LOW findings cover the unenforced §3.2 exhaustiveness obligation (no CI gate yet) and the undocumented cross-context TLS struct sharing (`source_tag` borrowing platform's `__thread` struct). None of these block the design's conclusions; all can be resolved in a follow-up correction PR or SPEC amendment spike. The spike issue #742 should remain OPEN per its closure rule.

specs/data/data-error-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:5b776fa followup:#888

SPEC.md §5 `error.hpp` stub replaced: the obsolete 5-arm `enum class Error` is superseded by the full 9-arm `ErrorTag + Error + WireSite` struct matching §10.1, with inline note citing §10.1 as the normative source and the ABI-boundary rationale for `std::string_view` on `host_hash`/`plugin_hash`.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:5b776fa followup:#888

Resolution: `std::string_view` is authoritative per PHILOSOPHY §11 final sentence ("Public plugin ABI surfaces never expose std:: containers or eastl:: containers — they cross the boundary as POD spans / handles only") and `plugin-abi.md §"Registration Entry-Point Signature"` (both sides compile against the same libc++, making `std::string_view` layout stable at the dylib boundary). `eastl::string_view` is confined to in-process, non-ABI code. `data-error-design.md` §3.1 `host_hash`/`plugin_hash` updated to `std::string_view` with inline ABI rationale comment. SPEC.md §10.1 already used `std::string_view`; SPEC.md §5 stub updated in the same commit uses `std::string_view` consistently.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:5b776fa followup:#888

§3.3 table `SchemaUnknown` `at.offset` cell changed from plain `YES` to `YES (=0, sentinel)`, mirroring the `EnvelopeTruncated` pattern. §3.3 discipline rule 5 added explaining that the field is always exactly 0 and carries no diagnostic byte-count information. §3.2 `SchemaUnknown` row payload column updated with bold sentinel note. The meaning of YES is now unambiguous: variable payload vs. fixed sentinel.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:5b776fa followup:#888

§3.2 footer updated with a CI enforcement paragraph citing §11.6 golden-per-source tests as the mechanical gate and the `tests/data/errors/golden_per_source/` fixture directory. Added §12 [OPEN: exhaustiveness-ci-gate] tracking that the first implementation plan PR (schema-registry or envelope-serdes) must author the §11.6 fixture harness. Until then the obligation remains normative-prose-only, acknowledged as a gap in review.

cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:5b776fa followup:#888

Option (a) taken: §5.4 demotes `source_tag` to a data-context-owned `thread_local` variable (separate `__thread` allocation from any platform-context TLS struct), removing the undocumented cross-context coupling that violated PHILOSOPHY §1/§2 SRP. The slot is reset by the data context's own phase-1 hook registered via `core::SystemRegistry`. A gate condition is documented: if a future spike concludes shared-struct ownership is required, `reviews/decisions/tls-layout.md` must be authored and approved first. §12 [OPEN: tls-layout-decision] captures this. No `tls-layout.md` file is created here since the decision is not needed today.

specs/data/data-error-design.md
```diff
@@ -0,0 +1,1199 @@
...
+    // refused, the missing step in the chain, or the back-edge of
+    // the detected cycle. Default on every other arm.
+    SchemaId      step_schema{};
+    SchemaVersion step_from{0};
+    SchemaVersion step_to{0};
+
+    // Set on tag 1: the host's compiled-in hash and the offending
+    // plugin's compiled-in hash, hex form (SPEC §4.4 inv. 5).
+    // Empty string_views on every other arm.
+    eastl::string_view host_hash{};
```
cjhowe-us, line 219: severity:MED location:specs/data/data-error-design.md:219 problem:The illustrative struct in §3.1 declares host_hash and plugin_hash as eastl::string_view, but SPEC.md §10.1 (normative) correctly uses std::string_view for these same fields. PHILOSOPHY §11 final sentence mandates that public plugin ABI surfaces use std:: (not eastl::) at the boundary — and the design doc's own §3.3 rule 2 confirms these string_views borrow from process-lifetime .rodata, exactly the ABI-boundary exception scenario. The seam inconsistency means an implementor reading only the design doc's §3.1 struct would emit the wrong type. fix:Change the eastl::string_view declarations for host_hash and plugin_hash in the §3.1 illustrative struct to std::string_view, matching SPEC.md §10.1 exactly; add an inline comment citing PHILOSOPHY §11 final sentence and plugin-abi.md "Registration Entry-Point Signature" as the ABI-boundary exception rationale. Tracked for resolution in PR #888.

specs/data/data-error-design.md
```diff
@@ -0,0 +1,1199 @@
...
+the abi-hash arm. There is no TLS prefix slot — the payload itself is
+the discriminator.
+
+When `glibre::log_error` formats a `data::Error` for spdlog, it MAY
+read an optional thread-local `source_tag` slot that the construction
+site can populate with a static string identifying the producing
+sibling (`"schema-registry"`, `"envelope-serdes"`,
+`"migration-dispatcher"`, `"middleman-static-init"`, `"foryc"`). The
+slot is purely informational — no engine code branches on it — and
+exists to give operators a fast-path tag in the log carrier (§7
```
cjhowe-us, line 591: severity:MED location:specs/data/data-error-design.md:591 problem:§5.4 (as merged) states source_tag "shares the __thread struct platform owns (§6.3) and SHALL NOT introduce a separate TLS allocation." From the R2 cohesion/SRP lens: this is a cross-context structural coupling — data has an undocumented runtime dependency on the platform context's TLS struct layout. PHILOSOPHY §1 (SRP) says one module, one reason to change; PHILOSOPHY §2 (cohesion AND completeness, clean seams) says contexts couple through explicit, documented seams only. The platform TLS struct is not cited in any reviewed/decisions/*.md record; the coupling is implicit. If platform's TLS struct layout changes (e.g. to add a crash-capture slot), the data context silently breaks. fix:Demote source_tag to a data-context-owned thread_local variable (a separate allocation from platform's __thread struct) so the seam is self-contained. Document the gate condition for future re-coupling: a reviews/decisions/tls-layout.md record must exist before data borrows platform's struct. Add a §12 [OPEN: tls-layout-decision] tracking this. Tracked for resolution in PR #888.

specs/data/data-error-design.md
```diff
@@ -0,0 +1,1199 @@
...
+    // -Wswitch makes a missed case here a compile error.
+    __builtin_unreachable();
+}
+```
+
+The `wrap_data_to_core` example lives in `core` (per §3.4); data does
+not host it. The example is included here only so the spike audit can
+verify that adding a new arm forces the central wrap to gain a case.
+
+### 4.3 Composition with `glibre::Error`
```
cjhowe-us, line 504: severity:LOW location:specs/data/data-error-design.md:504 problem:§4.3 says data::Error "may receive a data::Error from a data call and propagate it unchanged" into glibre::Error, and the example shows glibre::Error{e} for a data::Error struct value. error-model.md §Type Sketch defines glibre::Error::Variant as an eastl::variant over enum class types (core::Error, render::Error, etc.), with data::Error listed only as a comment stub. The actual data::Error type (as designed here) is a payload-bearing struct, not an enum class. The variant and the constructor template accept any type E, but the seam contract — that the struct data::Error is a valid variant arm — is never explicitly stated or cited. An implementor wiring data::Error into glibre::Error::Variant would have to infer this, and it is not obvious that eastl::variant over a POD struct (rather than an enum) matches the glibre::log_error dispatch path documented in error-model.md §Logging/Telemetry #2 (which formats error.code as an enumerator name). fix:Add a sentence to §4.3 confirming that data::Error (struct, not enum class) is a valid eastl::variant arm in glibre::Error::Variant per the template constructor in error-model.md; cite that the Logging/Telemetry #2 to_string path for data::Error uses to_string(ErrorTag) from §7.4 rather than magic_enum on the struct directly. A one-line note eliminates the inference gap.

specs/data/data-error-design.md
```diff
@@ -0,0 +1,1199 @@
...
+    "DeserializeError",         // 3
+    "ReservedTagViolation",     // 4
+    "SchemaRegistryConflict",   // 5
+    "SchemaUnknown",            // 6
+    "MigrationStepMissing",     // 7
+    "MigrationCycle",           // 8
+    "EnvelopeTruncated",        // 9
+};
+
+[[nodiscard]] constexpr eastl::string_view to_string(ErrorTag t) noexcept {
```
cjhowe-us, line 795: severity:LOW location:specs/data/data-error-design.md:795 problem:§7.4 defines kErrorTagNames[] as a hand-maintained array indexed by ErrorTag integer value. The §11.5 compile-fail tests gate that a new ErrorTag arm without a DataErrorRecord schema bump fails the codegen consistency check — but there is no compile-fail test asserting that the kErrorTagNames array length is synchronized with the ErrorTag max value. A developer could add ErrorTag::NewArm = 10, bump DataErrorRecord correctly, and forget to append "NewArm" to kErrorTagNames[]. The to_string fallback returns "(unknown)" rather than a compile error, making the gap silent at build time. fix:Add a compile-fail / static-assert test case under §11.5: static_assert(eastl::size(kErrorTagNames) == static_cast<std::uint16_t>(ErrorTag::EnvelopeTruncated) + 1, "kErrorTagNames must have one entry per arm plus the invalid-0 slot") so that any arm addition without a corresponding table update is a compile error. Document this static_assert in §7.4 alongside the to_string function.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:2
  LOW:2

carryover:2 (both R1 MED findings — eastl::string_view seam mismatch in §3.1 struct and source_tag cross-context TLS coupling in §5.4 — tracked as addressed in open PR #888; not yet in main)

notes: The cohesion + SRP + seam quality lens finds the design substantively sound. The aggregate's SRP boundary is explicitly stated in §1 ("one reason to change: a sibling data aggregate gained or lost a failure surface"), the refusal list in §1 is thorough and cites the correct decision records for every boundary, and the §3.4 wrapping table cleanly separates what data owns from what core owns at the loader's call site. The DataErrorRecord dual-purpose (log carrier + arm-count ABI gate) was examined and is not an SRP violation — the schema evolving together with the enum is an inherent consequence of the data context's design (schema → ABI hash), not a second independent reason to change. The two R1 MED findings carryover as open because PR #888 is not yet merged: the §3.1 illustrative struct still declares eastl::string_view for host_hash/plugin_hash (comment 3187332960) while SPEC §10.1 uses std::string_view, and §5.4 still documents undocumented cross-context TLS coupling to platform's __thread struct without a reviews/decisions/tls-layout.md anchor (comment 3187334212). Two new LOW findings are raised in this round: the glibre::Error::Variant composition seam does not explicitly document that data::Error is a payload-bearing struct arm (not an enum class) and does not confirm how Logging/Telemetry #2 dispatches to_string for it (comment 3187335410); and the kErrorTagNames hand-maintained array has no compile-time synchronization check against ErrorTag's arm count, leaving a silent gap if a new arm is added but the table is not updated (comment 3187336716). Both LOWs are resolvable in PR #888 or a follow-on correction PR without scope expansion.

specs/data/data-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:NOOP reasoning:Already addressed in R1 commit 5b776fa (PR #888). The §3.1 illustrative struct host_hash/plugin_hash fields were changed from eastl::string_view to std::string_view with an inline comment citing PHILOSOPHY §11 final sentence and plugin-abi.md 'Registration Entry-Point Signature'. The fix is on the open follow-up PR #888 (fix/data-data-error-design-followup-r1) which is not yet merged into main; the carryover flag in R2 is correct. No further action required in this round.

specs/data/data-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:NOOP reasoning:Already addressed in R1 commit 5b776fa (PR #888). §5.4 was rewritten to demote source_tag to a data-context-owned thread_local variable separate from any platform-context __thread struct, adding SRP rationale per PHILOSOPHY §1/§2. The tls-layout-decision gate condition and §12 [OPEN: tls-layout-decision] were also added. The fix is on open follow-up PR #888 pending merge; carryover is expected until #888 merges to main.

specs/data/data-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2e2f80e followup:#888

Added explicit note to §4.3 confirming that data::Error is a payload-bearing struct (not an enum class), that it is a valid eastl::variant arm in glibre::Error::Variant via the template<class E> constructor documented in error-model.md §"Type Sketch", and that the Logging/Telemetry #2 path uses to_string(ErrorTag) from §7.4 (not magic_enum on the struct) to produce the enumerator-name string. The inference gap is now explicit text. Commit 2e2f80e on PR #888.

specs/data/data-error-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2e2f80e followup:#888

Added static_assert to §7.4 kErrorTagNames[] asserting eastl::size(kErrorTagNames) == static_cast<std::uint16_t>(ErrorTag::EnvelopeTruncated) + 1 with a diagnostic message instructing the developer to append the name string before adding the enumerator. Added a corresponding 'name-table-sync' compile-fail test to §11.5 (compile_fail/name_table_sync_test.cpp) documenting the scenario gated. The silent-gap on new arm addition is now a compile error. Commit 2e2f80e on PR #888.

specs/data/data-error-design.md
cjhowe-us, line 1: severity:LOW location:specs/data/data-error-design.md:1 problem:The design doc preamble states it "introduces no new public surface beyond what SPEC §5 / §10 already enumerate." This is correct for the design itself, but the pre-existing SPEC.md §5 `RegisterStatus` enum (line ~698) derives its `SchemaRegistryConflict` enumerator via `static_cast<std::uint16_t>(Error::SchemaRegistryConflict)` where `Error` is the obsolete 5-arm enum stub. Once PR #888 lands and replaces the §5 stub with the 9-arm `ErrorTag + Error` struct, `Error::SchemaRegistryConflict` will no longer exist (renamed to `ErrorTag::SchemaRegistryConflict`). PR #888 does not currently touch this line, creating a forward breakage. fix:Add to PR #888 a change to `specs/data/SPEC.md` line ~698 replacing `static_cast<std::uint16_t>(Error::SchemaRegistryConflict)` with `static_cast<std::uint16_t>(ErrorTag::SchemaRegistryConflict)` so the §5 stub remains self-consistent after the fix lands.

cjhowe-us, line 1: severity:LOW location:specs/data/data-error-design.md:903 problem:§9.2 "Heap impact" states "The `source_tag` TLS slot reuses platform's TLS struct (§5.4)." This cross-reference becomes stale and internally inconsistent once PR #888 lands: #888 rewrites §5.4 to say `source_tag` is a data-context-owned `thread_local` variable separate from any platform-context `__thread` struct. After #888, §9.2 still says "reuses platform's TLS struct" while §5.4 says the opposite. fix:Add to PR #888 a one-sentence update to §9.2 changing "The `source_tag` TLS slot reuses platform's TLS struct (§5.4)" to "The `source_tag` TLS slot is a data-context-owned `thread_local` variable (§5.4); its footprint is one `std::string_view` (16 bytes)." so §9.2 stays consistent with the corrected §5.4.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:2

carryover:3 (HIGH: SPEC.md §5 obsolete 5-arm stub; MED-1: eastl::string_view in §3.1 struct; MED-2: source_tag cross-context TLS coupling in §5.4 — all three tracked as addressed in open PR #888, not yet merged to main)

notes: The round-3 topology + spec alignment + polish lens finds the design substantively sound and closes the retro trail with two new LOW findings, both targeted at PR #888 as incremental fixes. DAG coherence is confirmed: all six cited decision records exist on disk (error-model.md, fory-codegen.md, plugin-abi.md, hot-reload-protocol.md, perf-budget.md, frame-phases.md); all four sibling design files exist (schema-registry-design.md, envelope-serdes-design.md, middleman-dylib-design.md, migration-dispatcher-design.md); the dependency DAG is acyclic and complete. Spec invariants are cited correctly: §4.x inv references checked against SPEC.md content; §10.1/§10.2 cross-references verified against the SPEC §10 section. Issue topology is correct: #742 is type:spike with no pts:* label (per AGENTS.md, spikes carry no estimate), parent #729 is type:sub-epic. The six §12 open questions are well-formed with explicit resolution gates.

LOW-1 (new, comment 3187389310): SPEC.md §5 RegisterStatus enum derives SchemaRegistryConflict via Error::SchemaRegistryConflict; once PR #888 replaces the §5 stub with the 9-arm ErrorTag + Error struct, that reference becomes invalid (the enumerator is now ErrorTag::SchemaRegistryConflict). PR #888 does not patch this line. Fix belongs in PR #888.

LOW-2 (new, comment 3187390332): data-error-design.md §9.2 line 903 still says "The source_tag TLS slot reuses platform's TLS struct (§5.4)." PR #888 rewrites §5.4 to describe a data-context-owned thread_local, making §9.2 a stale dangling assertion about the opposite ownership. Fix belongs in PR #888.

The three prior-round carryover items (SPEC.md stub HIGH, eastl string_view MED-1, TLS coupling MED-2) are all addressed in the PR #888 commit 5b776fa + 2e2f80e. No items remain untracked. This retro trail is complete pending PR #888 merge; implementation work on sibling aggregates should not start until #888 lands.

Issue #755, opened 2026-05-03
[SPIKE] design-shader-shader-cache-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **shader-cache** aggregate in the `shader` context.

BLAKE3 CAS + manifest.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/pipeline-state-cache.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **shader-cache** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/shader-cache-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — shader-cache (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-shader-cache-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-shader-cache
status:done
issue:#755
branch:docs/shader-shader-cache-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a621fd5639aeec547
host:mac
cloud:none
commit:e235c5287119ca7efad7733b5a3eaafa6ef5c896
pr:#843
[...]

Issue #753, opened 2026-05-03
[SPIKE] design-shader-descriptor-layout-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **descriptor-layout** aggregate in the `shader` context.

4 frequency groups.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **descriptor-layout** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/descriptor-layout-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — descriptor-layout (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-descriptor-layout-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-descriptor-layout
status:done
issue:#753
branch:docs/shader-descriptor-layout-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a679886f329502c54
host:mac
cloud:none
commit:4509c31fbb8914e9a439ef6fbc1e8856daa5c3da
pr:#844
[...]

Issue #751, opened 2026-05-03
[SPIKE] design-shader-shader-reflection-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **shader-reflection** aggregate in the `shader` context.

Slangc reflection-record ingest (no backend bytecode container parsing).

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/rendering/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **shader-reflection** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/shader-reflection-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — shader-reflection (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-shader-reflection-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-shader-reflection
status:done
issue:#751
branch:docs/shader-shader-reflection-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a39a0dbe4ed0fb4a0
host:mac
cloud:none
commit:050c1f74b0fbfa5d7a9974bb821ee8e779f64d88
pr:#842
[...]

Issue #749, opened 2026-05-03
[SPIKE] design-shader-compilation-pipeline-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **compilation-pipeline** aggregate in the `shader` context.

slangc subprocess (single hop): metallib for MVP, DXIL for post-MVP.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/render-pipeline.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **compilation-pipeline** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/compilation-pipeline-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — compilation-pipeline (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-compilation-pipeline-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-compilation-pipeline
status:done
issue:#749
branch:docs/shader-compilation-pipeline-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-abd6fbeca1a8d93f5
host:$(hostname -s)
cloud:none
commit:3ceca4e9c60e3f961a752794c73e2517a8690d5e
pr:#841
[...]

PR #839, opened 2026-05-04
docs(shader): detailed design — shader-source (refs #745)

cjhowe-us:
## Summary

- Detailed design for the `shader` context's `ShaderSource` aggregate (SPEC §4.1): Slang translation-unit ingestion — file load, project-rooted include resolution with cycle detection and escape rejection, UTF-8/LF/BOM normalization, entry-point scanning, and BLAKE3 content hash over normalized bytes plus the ordered include closure.
- Twelve-section template followed (Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions).
- Refines SPEC §4.1 / §5 stubs; locks the hash composition to `MAGIC || u32_le-len-prefixed normalized_bytes || u32_le-len-prefixed ordered closure paths || per-file content hashes`. The cache key is composed by `ShaderCache` (#755), not here.
- Scope discipline: explicitly refuses ownership of `PermutationKey` (#747), `CompilationPipeline` (#749), `ReflectionBlob` (#751), `DescriptorLayout` (#753), `ShaderCache` (#755), and the platform `FileWatcher` (#719). Aggregate is shipping-excluded per SPEC §6.1 / §6.5 — runtime resolves through the cooked `ShaderLibrary` keyed by hash.
- Harmonius treated as research input only per PHILOSOPHY (R-12.4.3 hot-reload, R-15.11.2 source-hash cache, R-2.1.16 Slang front-end); every conclusion independently re-derived.

## Test plan

- [x] §11 unit-test list covers every §10 refusal arm (`SourceNotFound`, `IncludeEscape`, `IncludeCycle`, `EncodingInvalid`, `EntryPointMissing`, `EntryPointStageAmbiguous`) plus determinism properties (CRLF/BOM/trailing-LF normalization, byte-stable hash across runs and hosts, hash-changes-on-include-rename, topological closure order).
- [x] §11 integration tests bind watcher-driven invalidation (#719) to `ShaderArtifactReplaced` event publication; refusal preserves prior `ShaderSource` and prior cache state.
- [x] §11 perf microbenchmarks pin the cooker's per-TU budget (<5 ms at MVP scale; BLAKE3 ≥3 GiB/s on M1) and feed the engine `perf-budget.yml` CI gate.

### DIFF
--- a/specs/shader/shader-source-design.md
+++ b/specs/shader/shader-source-design.md
@@ -1,0 +1,1147 @@
+# Shader-Source Detailed Design
+
+> Detailed design for the `shader` context's `ShaderSource` aggregate
+> (SPEC §4.1, §5 `ShaderSource` / `EntryPoint` / `PreprocessedSource`).
+> Refines `specs/shader/SPEC.md` §4.1 and the harmonius-mined
+> ingredients of §3 collapse 1 (Slang-only source language) and §3
+> collapse 4 (no bytecode-container reflection).
+> All conclusions re-derived; harmonius prior art
+> (`/Users/cjhowe/Code/harmonius/docs/requirements/content-pipeline/hot-reload.md`
+> R-12.4.3, `docs/requirements/tools/shared-cache.md` R-15.11.2,
+> `docs/requirements/rendering/gpu-abstraction-layer.md` R-2.1.16,
+> `docs/design/rendering/shader-variants.md`,
+> `docs/design/rendering/render-pipeline.md`) cited as research input
+> only.
+
+Refs: spike #745 — `[SPIKE] design-shader-shader-source-detailed`.
+Parent sub-epic #744 (`[SUB-EPIC] Detailed Designs — shader`). Sibling
+task-breakdown spike blocked-by this deliverable.
+
+## 1. Purpose
+
+`ShaderSource` is the single aggregate in the `shader` context permitted
+to read `.slang` files from disk, expand their `#include` closure, and
+produce the canonical translation-unit value (`PreprocessedSource`)
+that every downstream aggregate keys against. Its one responsibility is
+**ingesting authored Slang text and presenting it as a sealed,
+content-addressed translation unit**: resolve a project-relative path
+against the project source root, parse `#include` directives, walk the
+directed include graph with cycle detection and escape rejection,
+normalize the post-include byte stream, scan stage-tagged entry points,
+and finalise a BLAKE3 content hash over the normalized bytes plus the
+ordered include closure. On any refusal it leaves the disk untouched
+and the caller's prior `ShaderSource` (if any) live and unchanged.
+
+What `ShaderSource` explicitly refuses to own:
+
+- **The permutation key.** `PermutationKey` (SPEC §4.2) is a separate
+  aggregate; ingestion never reads or mutates a key. The detailed
+  design lives in spike #747.
+- **Compilation.** Driving slangc, shaping argv, sandboxing the
+  subprocess, capturing stderr — `CompilationPipeline` (SPEC §4.3,
+  spike #749). `ShaderSource` produces inputs; the pipeline consumes
+  them.
+- **Reflection.** Ingesting slangc's native reflection JSON and tagging
+  bindings with descriptor frequency — `ReflectionBlob` (SPEC §4.4,
+  spike #751).
+- **Descriptor layout.** Projecting reflection onto the four-frequency
+  schema — `DescriptorLayout` (SPEC §4.5, spike #753).
+- **Cache identity arithmetic.** The shader cache key composes the
+  source hash with the resolved key, canonical compile flags, and
+  target (SPEC §2 `Shader Hash`); `ShaderSource` produces only the
+  source-hash component. The composing aggregate is the
+  `ShaderCache` (SPEC §4.6, spike #755).
+- **File-system watching.** The watcher loop, debounce policy, and
+  event coalescing live in the platform `FileWatcher` (spike #719);
+  `ShaderSource` exposes a re-ingestion entry point the watcher calls,
+  but does not poll, schedule, or own the watch.
+- **Slang authoring conventions.** Material-graph codegen into Slang
+  belongs to the `material` context (SPEC §3 refusal 2). `ShaderSource`
+  treats every input as authored bytes, regardless of provenance.
+- **GPU memory, PSO objects, runtime binding.** SPEC §3 refusals 1,
+  4, 5.
+
+If the resolution algorithm, the include syntax, the cycle-detection
+rule, the escape-rejection rule, the entry-point attribute scanner, the
+normalization rule, or the BLAKE3 composition over the closure
+changes, this design changes. Anything else is out of scope.
+
+## 2. Requirements Coverage
+
+Mapping of harmonius requirements to MVP refusal-or-coverage. Every
+entry is independently re-derived; harmonius is research input only
+(PHILOSOPHY §"How harmonius is used").
+
+| Harmonius source                                                                              | Glibre disposition (MVP)              | Coverage site                                                                                                                          |
+|-----------------------------------------------------------------------------------------------|---------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `requirements/content-pipeline/hot-reload.md` R-12.4.3 — detect shader source change, recompile affected permutations, swap at frame boundary, viewport-overlay errors, prior shader remains active | **Covered (ingest half)**             | §8 below: re-ingestion on watcher tick; trigger condition is `total_hash` change; refusal leaves prior `ShaderSource` live (SPEC §8.4). |
+| `requirements/tools/shared-cache.md` R-15.11.2 — cache by **source hash**, platform, permutation flags                                  | **Covered (source-hash half)**        | §7 below pins BLAKE3 composition; the platform-/flag-/key-side composition is `ShaderCache`'s (#755).                                  |
+| `requirements/rendering/gpu-abstraction-layer.md` R-2.1.16 — Slang front-end, no runtime compile in shipping                            | **Covered**                           | §5 surface ships only; `ShaderSource::open` is excluded from shipping link target (SPEC §6.1, §6.5); the runtime path uses cached blobs keyed by hash. |
+| `design/rendering/shader-variants.md` — preprocessor closure feeds permutation enumeration                                              | **Covered (closure half)**            | §3.4 constructs the closure; permutation enumeration consumes it without mutating it.                                                  |
+| `design/rendering/render-pipeline.md` § "Shader Compilation Pipeline" — single Slang front-end, no transpile chain                       | **Covered**                           | §3 below treats Slang as the only input language; no fallback parser is admitted.                                                      |
+| `design/rendering/shader-variants-test-cases.md` — deterministic content hash, bit-stable across hosts                                  | **Covered**                           | §7 + §11: BLAKE3 over normalized bytes + ordered closure; goldens enforce bit-equality across runs.                                    |
+| `design/rendering/render-pipeline.md` § RF-9 — re-run reflection on new bytecode so descriptor layout stays in sync                       | **Refused (out of scope here)**       | RF-9's reflection-side obligation belongs to `ReflectionBlob` (#751); `ShaderSource` only re-publishes a new `total_hash`.              |
+
+Glibre-native requirements added beyond harmonius:
+
+- **Project-rooted include resolution; no absolute paths, no
+  upward-escapes.** SPEC §4.1 invariant 3. Defense against
+  cross-project leakage and against non-deterministic includes that
+  depend on host-side filesystem layout.
+- **Acyclic include graph with explicit cycle reporting.** SPEC §4.1
+  invariant 3. Cycles are a first-class refusal arm (`IncludeCycle`)
+  not an infinite-recursion crash. The cycle path is attached to the
+  diagnostic envelope so the editor can render it.
+- **Total content hash composes normalized bytes + ordered closure.**
+  Two byte-equal authored sources whose include-graph closures differ
+  must produce different `total_hash`. Re-derived: a cache key that
+  ignores the closure aliases two authoring states and breaks
+  R-15.11.2.
+- **Encoding is fixed (UTF-8, LF-normalized, BOM-stripped).** Required
+  for cross-host bit-stability of `total_hash` (PHILOSOPHY §7
+  determinism). Authoring tools that emit CRLF or UTF-8-BOM are
+  normalized at ingestion; the on-disk file is never rewritten.
+- **One translation unit per ingest.** `ShaderSource::open` admits one
+  `.slang` file at a time. Batched ingestion is the cooker's
+  responsibility (parallelism per §6 below); no fan-out lives inside
+  the aggregate.
+
+## 3. Detailed Model
+
+### 3.1 Aggregate composition
+
+```text
+ShaderSource (root, value-shaped)
+├── SourceId                  id_                    // §5
+├── PreprocessedSource        preprocessed_          // §5
+│     ├── eastl::vector<std::byte>     bytes_         // normalized, post-include
+│     ├── eastl::vector<IncludeNode>   include_closure_
+│     └── ShaderHash                 total_hash_
+└── eastl::vector<EntryPoint> entry_points_         // §5
+```
+
+`ShaderSource` is constructed exclusively through the static factory
+`open(project_root, project_relative)`. There is no public
+constructor; copy / move are defaulted (the value is large but cheap
+to move). The aggregate holds no file handles after `open` returns —
+it is a pure in-memory snapshot of one ingestion.
+
+The implementation pipeline is internal and split into four pure
+subroutines, each called exactly once per `open`:
+
+```text
+open(project_root, project_relative)
+  └── 1. resolve(project_root, project_relative) → AbsolutePath
+      └── 2. ingest_closure(AbsolutePath) → (normalized_bytes,
+                                              ordered_includes)
+          └── 3. scan_entry_points(normalized_bytes) → entry_points
+              └── 4. hash_total(normalized_bytes,
+                                ordered_includes) → ShaderHash
+```
+
+Each subroutine has no side effects beyond reading from the project
+source root and allocating against the `ContextTag::shader` arena
+(SPEC §9.4). They are tested in isolation in §11.
+
+### 3.2 Resolver (`resolve`)
+
+Project-rooted resolution against a single canonical project source
+root. The resolver is the sole owner of the path-validation contract.
+
+```text
+resolve(project_root, project_relative):
+    1. Reject if project_relative is absolute             → IncludeEscape
+    2. Reject if project_relative contains "." segments
+       that escape the root (`..`-walk normalisation)     → IncludeEscape
+    3. Compose abs := canonical(project_root / project_relative)
+    4. Reject if commonpath(abs, canonical(project_root))
+       != canonical(project_root)                          → IncludeEscape
+    5. Reject if !exists(abs) or !is_regular_file(abs)    → SourceNotFound
+    6. Return abs
+```
+
+Notes:
+
+- Canonicalisation uses `std::filesystem::weakly_canonical` for
+  include-target paths (the file may not yet exist during a partial
+  edit; the canonicalised form is still well-defined) and
+  `std::filesystem::canonical` for the project root (must exist at
+  ingestion time).
+- Symlinks are followed for resolution but the *canonical* target must
+  still lie inside the project root. A symlink whose target leaves the
+  root is `IncludeEscape`.
+- The project root is captured at `open()` time and threaded through
+  every recursive include resolution; the resolver has no global
+  state.
+- The resolver is identical for the entry-point file and for every
+  `#include` directive encountered during ingestion. There is one
+  algorithm; the SPEC §4.1 invariant "include-graph closure resolves
+  entirely under the project source root" holds for the root file as
+  well.
+
+### 3.3 Include directive parser
+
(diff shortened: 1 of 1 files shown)


specs/shader/shader-source-design.md
```diff
@@ -0,0 +1,1147 @@
...
+  this to highlight the cycle.
+
+### 3.5 Normalizer
+
+Five rules pinned by determinism (PHILOSOPHY §7) and by the SPEC §4.1
+invariant 2 (byte-equal preprocessed content → equal
+`PreprocessedSource`):
+
+1. **UTF-8 only.** Any byte sequence that is not valid UTF-8 is
+   refused with `EncodingInvalid`. Re-derivation: hashing
```
cjhowe-us, line 331: severity:HIGH location:specs/shader/shader-source-design.md:331 problem:EncodingInvalid is used here (and in §8.4 line 832, §10.1 line 942, §11.1 line 1024) but this identifier does not exist in the SPEC §5 closed enum class shader::Error. The §5 enum has SourceNotFound, SourceParseFailed, IncludeEscape, IncludeCycle, EntryPointMissing, EntryPointStageAmbiguous, and 18 other arms — none named EncodingInvalid. The SPEC §10.2 per-enumerator table also has no EncodingInvalid row. The design doc introduces a new error identifier that the closed-sum enum does not declare; any implementation following this design would reference an undeclared enumerator, a compile error. fix:Two options: (a) file a SPEC §5 amendment adding EncodingInvalid (preferred — encoding failure deserves its own arm separate from SourceParseFailed which SPEC §10.2 scopes to slangc parse errors) and cross-reference it in §10.2; or (b) map UTF-8 rejection onto SourceParseFailed with a clear rationale note that SourceParseFailed covers both slangc-surface parse failures and ingestion-time encoding failures. Option (a) is cleaner SRP. Until the amendment lands, the four occurrences of EncodingInvalid in this design doc are dangling identifiers.

specs/shader/shader-source-design.md
```diff
@@ -0,0 +1,1147 @@
...
+    return out
+```
+
+Behaviour pinned by SPEC §4.1 invariant 1 ("every emitted EntryPoint
+has exactly one stage attribute"):
+
+- Zero stage attributes on a function → not enumerated; not an error
+  (the function may be a helper).
+- Two or more stage attributes on the same function →
+  `EntryPointStageAmbiguous`.
```
cjhowe-us, line 397: severity:MED location:specs/shader/shader-source-design.md:397 problem:EntryPointMissing trigger conflict between SPEC §10.2 and this design. SPEC §10.2 defines EntryPointMissing as: 'compile is invoked for an entry point name that the preprocessed ShaderSource does not expose' — meaning a CompilationPipeline-level error raised at compile time when the caller asks for an entry point name the ShaderSource manifest does not contain. This design §3.6 raises EntryPointMissing during open() when a [shader(...)] attribute has no following function declaration — a distinct trigger condition occurring during ingestion. Two different trigger conditions are sharing one enumerator. SPEC §6.4 (reflection step) also references EntryPointMissing for a third trigger context (the reflected entry-point set is absent from the ShaderSource manifest). The enumerator's semantic scope is therefore ambiguous across three sites. fix:Determine whether the ingestion-time condition (orphaned [shader(...)] attribute with no function declaration) should remain as EntryPointMissing or should be mapped to SourceParseFailed (which SPEC §10.2 already scopes to open-time slangc parse failures including 'malformed [shader(...)] annotation'). An orphaned attribute is effectively a parse error and SourceParseFailed is the correct arm by SPEC §10.2. If the ingestion-time scan is kept separate from slangc, then a design note clarifying that EntryPointMissing covers both the compile-time and the scan-time case is required in §10.2.

specs/shader/shader-source-design.md
```diff
@@ -0,0 +1,1147 @@
...
+            # The directive line itself is dropped from out_bytes;
+            # the includee's bytes already entered out_bytes during
+            # the recursive walk. This is the splice — no #line
+            # markers are emitted (§3.5 explains why).
+
+    on_stack.pop()
+    visited.insert(abs)
+    closure_order.append(IncludeNode{
+        project_relative_path = relative_to_root(abs),
+        content_hash          = blake3(bytes_of(abs))   # per-file hash
```
cjhowe-us, line 303: severity:LOW location:specs/shader/shader-source-design.md:303 problem:The pseudocode uses bytes_of(abs) to compute the per-file content_hash but the variable abs is a path, not a byte buffer. The algorithm already computed raw := read_file(abs) and bytes := normalize(raw) immediately above. §7.3 clarifies that content_hash is 'the per-file BLAKE3 of pre-normalization bytes', meaning it should be blake3(raw), not blake3(bytes). The unexplained bytes_of(abs) notation conflicts with the normalizer rule 4 which says 'The file content_hash stored in IncludeNode reflects the post-normalization bytes', creating a direct contradiction with §7.3 which says pre-normalization. fix:Replace blake3(bytes_of(abs)) with blake3(raw) in the pseudocode to make explicit that the per-file hash is BLAKE3 of the raw (pre-normalization) bytes. Then update normalizer rule 4 to remove the sentence 'The file content_hash stored in IncludeNode reflects the post-normalization bytes' which contradicts §7.3. The correct statement per §7.3 is that content_hash is BLAKE3 of pre-normalization bytes; the total_hash's splice uses post-normalization bytes via out_bytes.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:1
carryover:0
notes:The ShaderSource detailed design (specs/shader/shader-source-design.md) is structurally sound and comprehensive. All twelve required sections are present. The scope discipline is correct: ingestion owns only the four pure subroutines (resolve, ingest_closure, scan_entry_points, hash_total) and explicitly refuses every out-of-scope aggregate. The BLAKE3 composition scheme (MAGIC + length-prefixed normalized bytes + length-prefixed closure paths + per-file content hashes) is correct and satisfies PHILOSOPHY §7 determinism. The public surface returns std::expected<T, shader::Error> at every fallible boundary, consistent with reviews/decisions/error-model.md. Concurrency analysis is complete and the no-global-state property is correctly argued. The test plan covers all six refusal arms in §10.1 plus determinism, diamond DAG, cycle, and performance properties. The one blocking issue is that EncodingInvalid (used four times across §3.5, §8.4, §10.1, §11.1) is absent from the SPEC §5 closed enum class shader::Error; any implementation following this design would fail to compile. This requires either a SPEC §5 amendment or a remapping to an existing enumerator before the task-breakdown spike (sibling of #745) can proceed. The EntryPointMissing trigger is ambiguous between open-time (this design §3.6) and compile-time (SPEC §10.2) contexts; clarification or enumerator split is needed. The content_hash pseudocode uses an unexplained bytes_of(abs) that contradicts both the algorithm's own local variables (raw vs bytes) and the pre-/post-normalization intent in §7.3 vs normalizer rule 4; a one-line fix in the pseudocode resolves it. Conventional Commit subject and scope are correct. Issue ref #745 is present. PR is MERGED; findings are backfill record only — no unmerge is expected.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:1
carryover:0
notes:The ShaderSource detailed design (specs/shader/shader-source-design.md) is structurally sound and comprehensive. All twelve required sections are present. The scope discipline is correct: ingestion owns only the four pure subroutines (resolve, ingest_closure, scan_entry_points, hash_total) and explicitly refuses every out-of-scope aggregate. The BLAKE3 composition scheme (MAGIC + length-prefixed normalized bytes + length-prefixed closure paths + per-file content hashes) is correct and satisfies PHILOSOPHY §7 determinism. The public surface returns std::expected<T, shader::Error> at every fallible boundary, consistent with reviews/decisions/error-model.md. Concurrency analysis is complete and the no-global-state property is correctly argued. The test plan covers all six refusal arms in §10.1 plus determinism, diamond DAG, cycle, and performance properties. The one blocking issue is that EncodingInvalid (used four times across §3.5, §8.4, §10.1, §11.1) is absent from the SPEC §5 closed enum class shader::Error; any implementation following this design would fail to compile. This requires either a SPEC §5 amendment or a remapping to an existing enumerator before the task-breakdown spike (sibling of #745) can proceed. The EntryPointMissing trigger is ambiguous between open-time (this design §3.6) and compile-time (SPEC §10.2) contexts; clarification or enumerator split is needed. The content_hash pseudocode uses an unexplained bytes_of(abs) that contradicts both the algorithm's own local variables (raw vs bytes) and the pre-/post-normalization intent in §7.3 vs normalizer rule 4; a one-line fix in the pseudocode resolves it. Conventional Commit subject and scope are correct. Issue ref #745 is present. PR is MERGED; findings are backfill record only.

specs/shader/shader-source-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED followup:#887 commit:ebf8193

Added EncodingInvalid to the SPEC §5 closed enum (after SourceParseFailed, before IncludeEscape), to the to_string switch, and to the §10.2 per-enumerator contract table with its full trigger/recovery/severity row. Rationale for option (a) per the reviewer's preference: encoding rejection fires inside the §4.1 normalizer before slangc is involved, which is a distinct failure site from SourceParseFailed; SRP warrants a dedicated arm.

specs/shader/shader-source-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED followup:#887 commit:b096750

Remapped the ingestion-time orphaned-[shader(...)] trigger from EntryPointMissing to SourceParseFailed across all four occurrence sites in the design: §3.6 pseudocode, §3.6 behaviour notes, §8.4 refusal table, §10.1 table, §10.2 refuse semantics, and §11.1 test table. SPEC §10.2 reserves EntryPointMissing for the compile-time arm (CompilationPipeline raises it when the caller asks to compile a named entry point the ShaderSource manifest does not expose). An orphaned attribute with no following function declaration is a structural source parse error; SourceParseFailed's 'malformed [shader(...)] annotation' scope covers it precisely. Also added a clarifying note in §10.2 refuse semantics explicitly separating the two trigger contexts.

specs/shader/shader-source-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED followup:#887 commit:b096750

Replaced blake3(bytes_of(abs)) with blake3(raw) in the §3.4 walker pseudocode (raw is already defined immediately above as raw := read_file(abs), the pre-normalization bytes). Removed the contradictory sentence from §3.5 rule 4 ('The file content_hash stored in IncludeNode reflects the post-normalization bytes') and replaced it with a clarifying note that content_hash is pre-normalization BLAKE3 per §7.3, with a pointer to §7.3 for rationale. §7.3 is now the single authoritative statement: content_hash = BLAKE3(pre-normalization raw bytes); post-normalization bytes enter out_bytes for the total_hash splice.

specs/shader/shader-source-design.md
```diff
@@ -0,0 +1,1147 @@
...
+admits the same in-process trigger the watcher uses. `ShaderSource`
+exposes a friend-shaped test-only constructor that takes a
+synthesised byte stream in lieu of reading from disk; the
+normalization (§3.5), closure walk (§3.4), scan (§3.6), and hash
+(§3.7) all run identically. This is the entry point §8.6's bullet
+"inject an Slang diff for `source_id`" hooks into.
+
+## 9. Performance
+
+The wall-time and memory budget for `ShaderSource::open`. All numbers
```
cjhowe-us, line 855: severity:MED location:specs/shader/shader-source-design.md:855 problem:§8.5 specifies a 'friend-shaped test-only constructor' on ShaderSource gated by `#if defined(GLIBRE_E2E)`. Using a friend declaration for test hooks embeds a test-framework concern inside the aggregate's class definition, giving the ShaderSource header two reasons to change: (1) ingestion algorithm changes and (2) test-double scaffolding changes. This violates SRP at the header level. fix:Move the test-only factory to a separate test-only header (e.g. `tests/shader/source/shader_source_test_support.hpp`) that defines a free function `glibre::shader::test::make_shader_source_from_bytes(eastl::span<const std::byte>, SourceId)`. This factory can be a friend of ShaderSource in the test-support header only — the class definition in the public header remains clean. The `GLIBRE_E2E` guard moves to the test-support header, not the aggregate header. The §8.6 inject_source_diff path calls into this factory; no change to the aggregate ABI is needed.

specs/shader/shader-source-design.md
```diff
@@ -0,0 +1,1147 @@
...
+(`ShaderArtifactReplaced` event). The watcher integration adapter is
+the sole new code added by this design beyond the four §3 subroutines.
+
+### 8.2 Re-ingestion is the cheapest refresh
+
+An `open` call is bounded by §9 wall-time (<5 ms per TU at MVP
+scale). The watcher's debounce / coalescing (#719) ensures that a
+burst of save events (editors flush in two or three writes) triggers
+one re-ingestion. Idempotence on hash equality (§8.1) ensures that a
+"touch" of the file (mtime bumped, content unchanged) is a no-op past
```
cjhowe-us, line 791: severity:MED location:specs/shader/shader-source-design.md:791 problem:§8.1 places the watcher-reaction adapter 'inside `source/`' but §8.3 places the inverse index (`included_by: AbsolutePath → eastl::vector<SourceId>`) 'in the cooker / editor adapter (the same place that calls `fire_recompile`)' and §6.2 places `fire_recompile` in `cache/cooker.cpp`. This is contradictory: the adapter cannot be both inside `source/` and inside `cache/`. An implementer following §8.1 will put the adapter in `source/`, then find that the inverse index it needs lives in `cache/`, creating a `source/ → cache/` intra-plugin dependency that is not explicitly sanctioned. If the adapter instead lives in `cache/` alongside the cooker, the §8.1 claim that it is in `source/` is wrong. Either way, a reader cannot determine the correct module boundary from the current text. fix:Clarify §8.1 to distinguish the two objects: (1) the ShaderSource re-ingestion call (`ShaderSource::open`) lives in `source/`, (2) the watcher-event handler that calls `open` AND updates the inverse index AND calls `fire_recompile` lives in `cache/cooker.cpp` (or in a dedicated `source_watcher_adapter.cpp` under `cache/`). State explicitly which file owns the adapter and the inverse index, and confirm the dependency direction is `cache/ → source/` (not the reverse).

specs/shader/shader-source-design.md
```diff
@@ -0,0 +1,1147 @@
...
+it is a pure in-memory snapshot of one ingestion.
+
+The implementation pipeline is internal and split into four pure
+subroutines, each called exactly once per `open`:
+
+```text
+open(project_root, project_relative)
+  └── 1. resolve(project_root, project_relative) → AbsolutePath
+      └── 2. ingest_closure(AbsolutePath) → (normalized_bytes,
+                                              ordered_includes)
```
cjhowe-us, line 136: severity:LOW location:specs/shader/shader-source-design.md:136 problem:The §3.1 pipeline diagram uses nested indentation (└──) which visually implies that each step is called from inside the prior step (i.e. that `hash_total` is nested inside `scan_entry_points`, which is nested inside `ingest_closure`, which is nested inside `resolve`). The text describes them as 'called exactly once per `open`' and 'pure subroutines', but the diagram contradicts that by making it look like a chain of nested calls rather than a flat sequential dispatch from `open`. An implementer reading only the diagram would write `resolve` calling `ingest_closure` calling `scan_entry_points` calling `hash_total`, which is incorrect — `open` calls all four sequentially. fix:Replace the nested arrow diagram with a flat sequential form that makes `open` the unambiguous caller:\n```text\nopen(project_root, project_relative)\n  1. abs          := resolve(project_root, project_relative)    → AbsolutePath\n  2. (bytes, inc) := ingest_closure(abs)                       → (normalized_bytes, ordered_includes)\n  3. eps          := scan_entry_points(bytes)                  → entry_points\n  4. hash         := hash_total(bytes, inc)                    → ShaderHash\n```

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0
notes:The ShaderSource detailed design is cohesive and SRP-correct at the aggregate level. ShaderSource's single responsibility — ingest a Slang translation unit from disk and produce a sealed, content-addressed PreprocessedSource — is clearly stated in §1 and honoured throughout: the four subroutines (resolve, ingest_closure, scan_entry_points, hash_total) are pure, each handles one algorithmic step, and nothing inside the aggregate reaches into PermutationKey, CompilationPipeline, ShaderCache, or FileWatcher. The public surface is minimal: three const accessors plus one static factory, all returning types bounded by §5. The std::expected<T, shader::Error> boundary is consistently used at every fallible site per error-model.md. The BLAKE3 composition scheme is clean and correctly length-prefixed. Two medium findings require design-doc amendments before the implementation spike begins: (1) §8.5 specifies a friend-shaped test-only constructor inside the aggregate class, which embeds a test-framework concern in the production header — a dedicated test-support header with a free-function factory is the correct pattern; (2) §8.1 and §8.3 give contradictory answers about where the watcher adapter and its inverse index (included_by) live — §8.1 says inside source/ while §8.3 says in cache/cooker.cpp — an implementer cannot resolve this without a spec clarification that the dependency direction is cache/ → source/ and the adapter and inverse index both live in cache/. One low finding: the §3.1 nested-arrow pipeline diagram implies chained calls (resolve→ingest_closure→scan_entry_points→hash_total) rather than flat sequential dispatch from open — the diagram should be redrawn as a flat numbered list. R1 findings (HIGH EncodingInvalid SPEC gap, MED EntryPointMissing trigger remap, LOW content_hash pseudocode) are fully tracked in open follow-up PR #887 and are not re-raised here.

specs/shader/shader-source-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#887 commit:aa0a57d

§8.5 updated: removed the friend-shaped test-only constructor from the ShaderSource aggregate header. The production class definition now carries no test-specific members. Added a dedicated test-support header path: tests/shader/source/shader_source_test_support.hpp (compiled only under GLIBRE_E2E) declares glibre::shader::test::make_shader_source_from_bytes as a free function. The GLIBRE_E2E guard and the friend declaration move entirely into the test-support TU; the aggregate header is unconditionally clean. §8.6's inject_source_diff bullet calls this factory.

specs/shader/shader-source-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#887 commit:aa0a57d

§8.1 and §8.3 updated to resolve the source/ vs cache/ contradiction. §8.1 now opens with an explicit Module boundary block: ShaderSource::open lives in plugins/shader/source/ (the only object this design contributes to source/); the watcher-event handler, inverse index (included_by), and fire_recompile all live in plugins/shader/cache/ (either cache/cooker.cpp or a dedicated cache/source_watcher_adapter.cpp). The dependency direction cache/ → source/ is stated explicitly in §8.1 and repeated in §8.3. The prior ambiguous claim that the adapter was 'inside source/' has been removed. The code block comment now reads 'calls into source/' to make the cross-boundary call visible.

specs/shader/shader-source-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#887 commit:aa0a57d

§3.1 diagram replaced with a flat numbered list. The nested-arrow (└──) form that implied chained calls has been removed. The new diagram shows open() as the explicit caller dispatching each subroutine sequentially, with named intermediate values (abs, bytes, inc, eps, hash) that make the data flow explicit. Added a prose note immediately after the diagram: 'open is the sole caller of all four subroutines. They are not chained — each subroutine receives its inputs from open directly, not from the return value of its predecessor.'

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:1
carryover:1 (R1 HIGH EncodingInvalid SPEC gap — addressed in open PR #887, not yet merged)
notes:Round-3 topology + spec-alignment review passes cleanly. The aggregate DAG in SPEC §4 (ShaderSource → PermutationKey → CompilationPipeline → ReflectionBlob) is coherent and the design doc's module boundary declaration (cache/ → source/) is consistent with it — no cross-context concerns and no circularity. All R1 and R2 findings are correctly captured in open PR #887 (EncodingInvalid SPEC §5 gap, EntryPointMissing→SourceParseFailed remapping, blake3(raw) pseudocode fix, §8.5 SRP test-support separation, §8.1/8.3 adapter location clarification, §3.1 flat-dispatch diagram); #887 must merge before the sibling task-breakdown spike begins. Issue #745 labels (type:spike, phase:mvp, domain:shader) and parent #744 (type:sub-epic) are correct per AGENTS.md — no estimate on a spike, no pts:* label, sub-epic → spike is a valid hierarchy. SPEC invariant citations in the design doc (§4.1 inv 1 at §3.6, §4.1 inv 3 at §3.2/§3.4, §6.1/§6.5 shipping exclusion) are accurate. The five §12 open questions are properly scoped as post-MVP defers — none leaks cross-context concerns or introduces hidden scope. One new LOW topology concern identified and recommended as DEFER: ShaderHash (SPEC §5 line 666) is a single 32-byte named type used for three semantically distinct hash domains (per-file pre-normalization BLAKE3 in IncludeNode.content_hash, source-content-only total BLAKE3 in PreprocessedSource.total_hash, and the full compound cache-key BLAKE3 defined in the struct comment). SRP warrants distinct named types; renaming is a multi-section SPEC amendment that is out of scope for this spike and should be filed as a follow-up plan or spike when the implementation begins. PR #839 itself is a well-structured spike deliverable; the retro trail is clean; verdict is APPROVE pending #887 merge.

Issue #747, opened 2026-05-03
[SPIKE] design-shader-permutation-key-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **permutation-key** aggregate in the `shader` context.

4D codec + enumeration.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/shader-variants.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **permutation-key** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/permutation-key-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — permutation-key (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-permutation-key-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-permutation-key
status:done
issue:#747
branch:docs/shader-permutation-key-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a32b535f0447f5d69
host:mac
cloud:none
commit:3a74bfbdafb763a29693b53d65b7a342010743dc
pr:#840
[...]

Issue #745, opened 2026-05-03
[SPIKE] design-shader-shader-source-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **shader-source** aggregate in the `shader` context.

Slang parser + include resolver.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/rendering/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **shader-source** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/shader-source-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — shader-source (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-shader-source-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-shader-source
status:done
issue:#745
branch:docs/shader-shader-source-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a02f6c5eaa7bbc131
host:mac
cloud:none
commit:4589517d394afabd416147f9fbbfd59d9cd580ab
pr:#839
[...]

Issue #323, opened 2026-05-03
[STORY] Encode and enumerate the 4-axis PermutationKey deterministically


cjhowe-us:
### Domain

shader

### Phase

phase:mvp

### Persona

engine developer

### User Story

As an engine developer, I want a deterministic 4-axis `PermutationKey` (ShadingModel × FeatureSet × RenderPath × LODTier) with bit-stable encoding and dense `PermutationIndex` ordinal, so that shader artifacts are addressable, hashable, and codegen-enumerable across hosts and runs.

### Acceptance Criteria (Gherkin)

```gherkin
Given a PermutationKey k1 = { ShadingModel::Standard, FeatureSet{Skinned|MotionVectors}, RenderPath::Forward, LODTier::Desktop }
When I call k1.to_bytes()
Then the result is exactly 6 bytes
And k1.is_well_formed() is true
And PermutationKey::from_bytes(k1.to_bytes()) returns a value equal to k1

Given two hosts (macOS arm64, macOS x86_64)
When each host computes k1.to_bytes()
Then the byte arrays are bit-equal

Given the full PermutationKey cross-product (8 × 64 × 6 × 4 = 12288 keys)
When the codegen enumerator walks the cross-product
Then iteration order is fixed and matches ascending tuple order (ShadingModel, FeatureSet bits, RenderPath, LODTier)
And every key maps to a unique PermutationIndex in [0, 12288)
```

### Manual Test Script

1. Run the codegen tool: `glibre-shadercc --emit-permutation-table > out.cpp`.
2. Inspect `out.cpp`; confirm the table has exactly 12288 entries in tuple order.
3. Run the determinism harness: `ctest -L permutation_key_codec`.
4. Confirm both arm64 and x86_64 runners report identical BLAKE3 hash of the encoded table.

### E2E Test Plan

`tests/e2e/shader/permutation-key-codec.glibre-trace` + assertions:
- `assert_eq(k.to_bytes().size(), 6)`
- `assert_eq(from_bytes(to_bytes(k)), k)` for a sampled set of keys
- `assert_eq(host_a_table_hash, host_b_table_hash)`
- `assert_eq(walk_order, expected_ascending_order)`

### Closure Checklist (in order)

- [ ] E2E test authored and passing in CI
- [ ] Manual test script reviewed by a second human
- [ ] Manual test executed; PASS recorded as a comment

### Harmonius Source

[...]

Issue #70, opened 2026-05-03
[SUB-EPIC] Acceptance + Implementation Plan — shader


cjhowe-us:
# Acceptance + Implementation Plan — shader

Draft user-stories, close open questions, produce type:plan backlog.

Parent epic: #66

## Sections owned (in `specs/shader/SPEC.md`)

- §11 Acceptance Criteria
- §12 Open Questions
- type:plan backlog

No estimate; rolls up.

PR #840, opened 2026-05-04
docs(shader): detailed design — permutation-key (refs #747)

cjhowe-us:
## Summary

- 12-section detailed design for the `PermutationKey` aggregate
  (SPEC §4.2) — closed 4-axis enums, 6-byte canonical `PackedBytes`
  encoding with reserved-bit / reserved-byte forward-compat seam,
  dense `PermutationIndex` bijection (cardinality 12 288), hot-
  adjacent equality, allocator-free contract.
- Honors SRP — refuses ingestion (#745), compilation (#749), cache
  lookup (#755). Surfaces exactly two `shader::Error` arms:
  `PermutationKeyMalformed` (refuse) and `PermutationKeyOutOfRange`
  (fatal).
- Routes every harmonius `shader-variants.md` clause to
  covered / collapsed / refused with explicit rationale; pins ABI
  rules, hot-reload survival, performance budgets (cook-time
  < 80 ns; hot-adjacent equality < 3 ns inline), and a Catch2 test
  plan (unit / integration / benchmark) anchored to user story #323.

## Test plan

- [ ] Sibling task-breakdown spike (#747 follow-up) consumes this
      doc to enumerate `type:plan` issues.
- [ ] Implementation plans in sub-epic #70 cite §4 as the public
      surface and §11 as the test contract.
- [ ] Catch2 contract `permutation_key_encoding_is_total_injective_and_bit_stable`
      lands as the conjunction of the §11.1 unit-suite tests when
      #323 closes.

Issue stays OPEN per spike-closure rules.

Refs #747.

### DIFF
--- a/specs/shader/permutation-key-design.md
+++ b/specs/shader/permutation-key-design.md
@@ -1,0 +1,1212 @@
+# `shader` — `PermutationKey` Detailed Design
+
+> Detailed design for the **permutation-key** aggregate in the `shader`
+> context (SPEC §4.2, §5 lines 523–589, §7.1 sub-schema row, §10.2 rows
+> `PermutationKeyMalformed` / `PermutationKeyOutOfRange`). This document
+> refines the SPEC's invariants into an implementation-shaped contract.
+> It is the deliverable of spike #747 and the input to the sibling
+> task-breakdown spike that decomposes implementation work into
+> `type:plan` issues.
+>
+> Authority: this design refines `specs/shader/SPEC.md` §4.2, §5 lines
+> 523–589, §7.1, §7.2 (`PermutationKeyRecord`), §7.3 #2, §7.4 #1, §9
+> (entries marked `PermutationKey`), and §10.2 rows
+> `PermutationKeyMalformed` / `PermutationKeyOutOfRange`. Anything not
+> addressed here defers to those sections; anything that appears to
+> contradict them is a defect in this document.
+
+---
+
+## 1. Purpose
+
+The permutation-key aggregate owns **the rules for naming one point in
+the closed 4-axis shader specialization cross-product** as a single
+trivially-copyable value object with a deterministic packed encoding
+suitable for hashing, ordering, and codegen-table generation. Its sole
+responsibility is to **canonicalize the 4-tuple `(ShadingModel,
+FeatureSet, RenderPath, LODTier)` into a stable byte-key and to
+serve equality and ordering at hash-table speed**.
+
+It refuses to own:
+
+- **Source ingestion.** The Slang translation unit, include closure,
+  and entry-point manifest belong to `ShaderSource` (SPEC §4.1, spike
+  #745). The permutation-key references the ingested source only via
+  the BLAKE3 `source_hash` that the `ShaderHash` envelope carries;
+  the key itself does not store source bytes, paths, or entry-point
+  names.
+- **Compilation.** Driving slangc, building flag lists, and emitting
+  bytecode belong to `CompilationPipeline` (SPEC §4.3, spike #749).
+  The key is a *direction* the pipeline consumes; it never invokes,
+  spawns, or canonicalizes flags.
+- **Cache lookup and CAS storage.** Content-addressable lookup and
+  the cooked `ShaderLibrary` belong to `ShaderCache` (SPEC §4.6,
+  spike #755). The key is the **permutation half** of a `ShaderHash`;
+  the other halves (source-hash, flags-hash, target byte) are joined
+  by the cache when it derives the artifact identity. The key is
+  *not* a `ShaderHash`.
+- **Reflection and descriptor layout.** A `ReflectionBlob` (SPEC §4.4)
+  and `DescriptorLayout` (SPEC §4.5) are derived per
+  `(Backend, PermutationKey)`, but the key holds none of their
+  bytes — it is an input to derivation, not a record of the result.
+- **Permutation enumeration policy.** Which keys a project
+  *materializes* into its build is the offline permutation
+  enumerator's concern (SPEC §6, sub-epic #70). The key knows the
+  cross-product cardinality (`kShadingModelCount × 2^kFeatureBitCount
+  × kRenderPathCount × kLODTierCount`) but does not enumerate it.
+- **Capability validation.** Whether a given backend (`IShaderBackend`,
+  SPEC §4.7) advertises mesh shaders / RT / fp16 etc. is a
+  `Capabilities` query. The key carries no capability bits; mismatch
+  is detected at compile time with `Error::CapabilityNotSupported`,
+  not at key-construction time.
+
+The design lives entirely inside the `shader/permutation/` sub-module.
+The public ABI surface it exports is the four closed enums
+(`ShadingModel`, `FeatureBit`, `RenderPath`, `LODTier`), the
+`FeatureSet` bitfield, the `PermutationKey` POD struct, the
+`PermutationIndex` dense ordinal, and the `to_bytes` /
+`from_bytes` / `is_well_formed` operations declared in SPEC §5.
+
+---
+
+## 2. Requirements Coverage
+
+Harmonius prior art (`harmonius/docs/design/rendering/shader-variants.md`,
+`shader-variants-test-cases.md`) is treated as **research input only**
+(PHILOSOPHY "How harmonius is used", AGENTS.md). Conclusions are
+re-derived against glibre's invariants.
+
+The table below enumerates every harmonius requirement / design
+clause that touches the permutation-key space and routes it to one of:
+**covered** (with this doc's section), **collapsed** (folded into a
+single glibre primitive), or **refused** (out of MVP scope with
+rationale rooted in glibre SPEC §3 collapses or §4.2 invariants).
+
+| Harmonius clause                                                                                            | glibre disposition                                                                                                                               | Where    |
+|-------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|----------|
+| `R-2.3.10.1` — Shaders defined as material × feature × pass × LOD                                           | **Covered.** Re-derived as the closed 4-tuple `(ShadingModel, FeatureSet, RenderPath, LODTier)` (SPEC §2, §4.2).                                 | §3       |
+| `R-2.3.10.2` — `PermutationKey` is hashable, serializable, and stable                                       | **Covered.** `to_bytes()` produces a 6-byte total-injective bit-stable encoding; serialization rides through `PermutationKeyRecord` (§7).        | §3, §7   |
+| `R-2.3.10.3` — Per-mesh max variant count = 64; total per-project = 4096                                    | **Refused for the key.** Cross-product budgets are an enumerator-policy concern (sub-epic #70), not a key-aggregate concern. The key knows total cardinality (§3.4) but does not enforce caps. | n/a (out of scope) |
+| `R-2.3.10.4` — Asset processing precompiles the tracked variant set                                         | **Refused for the key.** Cooker concern (`ShaderCache::cook`, SPEC §4.6 / §6.4); the key is the *input*, not the policy.                          | n/a      |
+| `R-2.3.10.5` — Untracked variants compile on demand, recorded in usage metrics                              | **Refused.** glibre forbids on-demand runtime compile (SPEC §4.6 inv 2, §4.8 inv 3, §10.3). The key has no on-demand fallback path.               | n/a      |
+| `R-2.3.10.6` — Hot reload of compiled variants                                                              | **Collapsed.** Hot-reload survival follows from the key being immutable POD: keys carried as state survive plugin swap byte-equal (§8). No per-key hot-swap state machine. | §8       |
+| `R-2.3.10.7` — CLI report of variant counts, uncompiled misses, and bundle sizes                            | **Refused.** CLI / metrics live in tools; the key exposes `PermutationIndex` and total cardinality so tools can iterate, but does not own reporting. | n/a      |
+| `PermutationKey { shading_model, features, render_path, lod }` (Rust prior art, `shader-variants.md` §"Core Types") | **Covered.** Re-derived in C++23 as a POD struct (SPEC §5 lines 569–583). Field order matches harmonius for cross-walk readability; bit packing is glibre's. | §3       |
+| `ShaderFeatures` 32-bit bitset with 16 used bits (`shader-variants.md` §"ShaderFeatures Bitset")            | **Collapsed.** glibre's `FeatureSet` is 16 bits with 6 currently-defined positions (SPEC §5 lines 530–557). The harmonius set (skinning, morph_targets, vertex_color, normal_map, alpha_cutout, alpha_blend, two_sided, parallax, decal, emissive, clear_coat, anisotropic, refraction, motion_vectors, velocity, stereo) collapses onto SPEC §5's tighter MVP set; growth is additive (§7 / SPEC §7.4 #1). | §3.2     |
+| `PermutationKey::content_hash() → BLAKE3` (`shader-variants.md` §"Core Types")                              | **Collapsed and re-scoped.** The key itself does **not** hash. The BLAKE3 digest that identifies a *cache entry* is `ShaderHash`, computed by `ShaderCache` over `(source_hash ∪ key.to_bytes() ∪ flags_hash ∪ target_byte)` (SPEC §2, §4.6 inv 1, §7.3 #1). The key contributes its 6 packed bytes; nothing more. | §3.5, §7 |
+| `ShadingModel { Pbr, Skin, Hair, Cloth, Foliage, Eye, Water, ClearCoat, ... }` (`shader-variants.md` class diagram) | **Covered.** Re-derived as the SPEC §5 enum `ShadingModel { Standard, Skin, Hair, Cloth, Foliage, Eye, Water, ClearCoat }` (cardinality 8). Harmonius's `Pbr` collapses into glibre's `Standard`. | §3.2     |
+| `RenderPath { Forward, Deferred, ShadowMap, Velocity, ... }` (`shader-variants.md`)                         | **Covered.** Re-derived as SPEC §5 `RenderPath { Forward, Deferred, DepthOnly, Shadow, Velocity, Probe }` (cardinality 6). `ShadowMap` collapses into `Shadow`; `Probe` is glibre-new for IBL bake passes. | §3.2     |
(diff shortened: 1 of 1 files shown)


specs/shader/permutation-key-design.md
```diff
@@ -0,0 +1,1212 @@
...
+
+inline constexpr std::uint32_t kPermutationCrossProductCardinality =
+    static_cast<std::uint32_t>(kShadingModelCount)
+  * (1u << kFeatureBitCount)
+  * static_cast<std::uint32_t>(kRenderPathCount)
+  * static_cast<std::uint32_t>(kLODTierCount);   // = 12288
+
+// Bijection (§3.4). Inverse fails with PermutationKeyOutOfRange.
+PermutationIndex                     to_index(const PermutationKey&)             noexcept;
+std::expected<PermutationKey, Error> from_index(const PermutationIndex&)         noexcept;
```
cjhowe-us, line 576: severity:HIGH location:specs/shader/permutation-key-design.md:575-576 problem:The design's §4 public surface introduces three free functions — `to_index`, `from_index`, and `permutation_key_byte_less` — plus two compile-time constants `kPermutationCrossProductCardinality` and `kFeatureBitMask`, none of which appear in the SPEC §5 authoritative header stub (specs/shader/SPEC.md). The design explicitly claims to refine SPEC §5 (see frontmatter authority block, and §11.4 cross-reference). Introducing public-boundary symbols not reflected in the SPEC's §5 means the SPEC and the design are out of sync; an implementer following the SPEC stub will silently drop these five symbols. fix:The SPEC §5 header stub must be amended to add: (1) the two cardinality-related constants (`kPermutationCrossProductCardinality`, `kFeatureBitMask`); (2) the two bijection free functions (`to_index(const PermutationKey&)`, `from_index(const PermutationIndex&)`); (3) the comparator free function (`permutation_key_byte_less`). Since this doc is the spike deliverable and the SPEC is a separate file, the correct fix is a follow-up SPEC amendment PR that adds these symbols to SPEC §5 and cites this design as the authority. Tag that PR as a blocker on the implementation plan issues before any code lands.

specs/shader/permutation-key-design.md
```diff
@@ -0,0 +1,1212 @@
...
+    out[5] = std::byte{0};   // RESERVED
+    return out;
+}
+```
+
+`from_bytes` is `noexcept` and *fallible*; it is the only validation
+seam in the aggregate:
+
+```cpp
+std::expected<PermutationKey, Error>
```
cjhowe-us, line 347: severity:MED location:specs/shader/permutation-key-design.md:347 problem:The `from_bytes` signature in §3.5 returns `std::expected<PermutationKey, Error>` where `Error` is `shader::Error`. The error-model.md §Decision item 1 requires: "Every public boundary in the engine returns `std::expected<T, glibre::Error>`." The §4 boundary table and the cross-reference section acknowledge the deferral ("the engine-wide glibre::Error variant alias will append this enum when the shader plugin lands"), but the design does not explicitly state under what conditions `from_bytes` is a plugin-ABI-crossing boundary vs. a within-plugin call. If any caller outside the shader plugin calls `from_bytes` (e.g. the cache context loading a manifest), the return type at that crossing is `shader::Error`, not `glibre::Error`, violating error-model.md decision item 1. fix:Clarify in §3.5 and §4 whether `from_bytes` is ever called across a plugin ABI boundary. If yes, add a plugin-seam adapter that wraps `shader::Error` into `glibre::Error` at that crossing and document it in §4's "Errors returned at this surface" table. If `from_bytes` is always within-plugin (called only by the shader plugin's own serialization layer), add an explicit sentence stating this scope so an implementer does not accidentally export the raw `shader::Error` return across the C ABI.

specs/shader/permutation-key-design.md
```diff
@@ -0,0 +1,1212 @@
...
+A `PermutationKey` is a **trivially-copyable value object** of fixed
+size. It carries four scalar/bitfield axes, exposes a 6-byte canonical
+serialization, and supports equality, total ordering, and dense-ordinal
+codegen indexing. It is **not** a class with construction validation
+that may fail: any `PermutationKey` value the code holds is
+*structurally* well-formed (its enumerators are in declared range);
+ill-formed bytes are rejected at the `from_bytes` boundary, never as a
+held value.
+
+Sizing and alignment (SPEC §5 lines 569–583):
```
cjhowe-us, line 128: severity:MED location:specs/shader/permutation-key-design.md:128-143 problem:The sizing table claims `alignof(PermutationKey) = 1 byte (no padding required)` and assigns `features` at offset 1 with size 2. But `FeatureSet` wraps a `std::uint16_t`, which has 2-byte alignment. With `shading_model` (1 byte) at offset 0 and `features` (uint16_t-aligned) at offset 1, the compiler will insert 1 byte of padding before `features` unless the struct is marked `[[gnu::packed]]` or the fields are reordered. The result is `sizeof(PermutationKey) = 6` (as the note states), but the `alignof` is 2, not 1. The `static_assert(sizeof(PermutationKey) <= 8)` in the §4 header will pass, but a compiler-layout assumption of "offset 1 = features" in any memcpy-based code would be wrong — the struct lays out as `[sm | pad | featLo | featHi | path | lod]` not `[sm | featLo | featHi | path | lod | RESERVED]`. The design correctly uses explicit `to_bytes()` to avoid this, but the table creates a misleading alignment claim. fix:Correct the sizing table: state `alignof(PermutationKey) = 2` (driven by `FeatureSet`'s uint16_t member). Update the offset column to show the actual compiler-generated layout including the implicit padding byte at offset 1. Add a note that `to_bytes()` explicitly encodes to the PackedBytes wire format and does NOT memcpy the struct directly; any code that attempts a raw struct memcpy for wire I/O is a defect. Consider adding a `static_assert(offsetof(PermutationKey, features) == 2)` to the unit suite to lock down the actual struct layout.

specs/shader/permutation-key-design.md
```diff
@@ -0,0 +1,1212 @@
...
+  inline bool permutation_key_byte_less(
+      const PermutationKey& a, const PermutationKey& b) noexcept {
+      return a.to_bytes() < b.to_bytes();   // lexicographic on 6 bytes
+  }
+  ```
+
+  This is identical to the canonical iteration order (§3.4) for
+  structurally well-formed keys: nesting `(sm, features, path, lod)`
+  with little-endian features matches the byte order
+  `[sm | f_lo | f_hi | path | lod | 0]` lexicographically, **as long
```
cjhowe-us, line 415: severity:LOW location:specs/shader/permutation-key-design.md:412-421 problem:The note at §3.6 states that when `kFeatureBitCount > 8` the byte-comparator order and the canonical index order will diverge for in-axis adjacency, but this is documented only in passing mid-section. This is a latent correctness risk: any future implementer adding a 9th feature bit may not encounter this note and could produce an enumerator that relies on the orders being equivalent. fix:Promote the divergence warning to a named invariant or a dedicated subsection with a label (e.g. "§3.6 Warning — bit-count growth boundary") and add it as an explicit open question (§12) tracking the concrete steps needed when bit 6 lands — specifically: (a) verify whether the index formula must be updated, (b) test that `permutation_key_byte_comparator_matches_lex_order` is updated to reflect the diverged ordering, (c) confirm that any cooker code that assumed the two orders were equivalent is updated. Alternatively, add a `static_assert(kFeatureBitCount <= 8, "byte and index orders diverge")` in the header so the build fails loudly rather than silently when a 9th bit would be added.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:2
  LOW:1

carryover:0

notes:The design document delivers its declared scope — a 12-section detailed design for the PermutationKey aggregate, covering all required elements from the spike issue: 4-axis closed enums with correct cardinality (8 × 64 × 6 × 4 = 12,288 verified), 6-byte PackedBytes canonical encoding with a sound reserved-bit and reserved-byte forward-compatibility seam, a total-injective PermutationIndex bijection with an explicit mixed-radix codec, hot-adjacent equality contract, allocator-free guarantee, Catch2 test plan (unit + integration + benchmark), and full harmonius requirements routing table with no silent drops. The SRP boundary is well drawn: the key correctly refuses compilation, source ingestion, cache lookup, and capability validation. Error surface is correctly limited to two enumerators (PermutationKeyMalformed / PermutationKeyOutOfRange), both returning std::expected. The Conventional Commit subject is well-formed (docs(shader): ... refs #747), the PR body cites the issue, and the spike closure rules (issue stays OPEN) are followed correctly.

The one HIGH finding is structural: the design's §4 public surface introduces five symbols (to_index, from_index, permutation_key_byte_less, kPermutationCrossProductCardinality, kFeatureBitMask) that are not reflected in the SPEC's §5 authoritative header stub. Since this document is a spike deliverable that refines the SPEC, the SPEC stub is now out of date and will mislead implementers following it. A follow-up SPEC amendment PR is required before implementation plans land.

The two MED findings are: (1) from_bytes returns shader::Error directly without an explicit statement of whether this is ever a plugin-ABI-crossing boundary — clarification is needed to ensure error-model.md §Decision item 1 is honored at the right crossing points; (2) the struct sizing table incorrectly claims alignof = 1 and offset 1 for features, when the actual compiler layout has a padding byte at offset 1 making the struct 6 bytes with alignof 2. The to_bytes() implementation is correct, but the table as written would mislead a memcpy-based reader. The LOW finding is a latent risk around the documented byte-order/index-order divergence at kFeatureBitCount > 8, which is buried mid-section and would benefit from a static_assert guard or an explicit open-question entry.

None of the findings block the merged spike's intent; all are addressable in the follow-up amendment PR or in the first implementation plan.

specs/shader/permutation-key-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e41076a followup:#886

The SPEC §5 header stub now declares all five public symbols introduced by permutation-key-design.md §4: `kFeatureBitMask`, `kPermutationCrossProductCardinality`, `to_index`, `from_index`, and `permutation_key_byte_less`. Each carries a brief doc-comment cross-referencing the design §. Implementers following the SPEC stub will no longer silently drop these five symbols.

specs/shader/permutation-key-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e41076a followup:#886

Added an explicit paragraph in permutation-key-design.md §3.5 stating that `from_bytes` is a within-plugin call only, never exposed directly across the C ABI. The paragraph specifies that external callers must route through the `glibre::Error` adapter per error-model.md §Composition Rules item 1, and that no raw `shader::Error` value crosses the C ABI directly.

specs/shader/permutation-key-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e41076a followup:#886

Corrected the §3.1 sizing table: alignof(PermutationKey) = 2 (driven by FeatureSet's uint16_t member), with an implicit 1-byte padding at offset 1 making offsetof(features) == 2. Added an explicit note that to_bytes() deliberately decouples struct layout from wire format, and that any raw struct memcpy for wire I/O is a defect. Also noted the required static_assert(offsetof(PermutationKey, features) == 2) in the unit suite.

specs/shader/permutation-key-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e41076a followup:#886

Promoted the byte-order/index-order divergence warning to a named §3.6 Warning subsection with: (a) a mandatory static_assert(kFeatureBitCount <= 8) directive in the implementation file, (b) a three-step remediation checklist (verify index formula, update unit-suite comparator test, audit cooker code), and (c) a corresponding [OPEN] item in §12 tracking those steps when the guard trips.

specs/shader/permutation-key-design.md
```diff
@@ -0,0 +1,1212 @@
...
+   `compile` is `if (!key.is_well_formed()) return
+   std::unexpected(Error::PermutationKeyMalformed);`. This catches
+   any caller that constructed a key by `static_cast`-ing arbitrary
+   ordinals (memory corruption, bad codegen-table walker).
+2. **Bridging across the C ABI.** A plugin trait method that takes
+   a `PermutationKey` by value receives it across the C ABI of the
+   plugin dylib; the receiving side checks `is_well_formed()` to
+   refuse silently-garbage-initialized inputs from a misbehaving
+   caller before any further work.
+
```
cjhowe-us, line 458: severity:LOW location:specs/shader/permutation-key-design.md:449-458 problem:Section §3.7 prescribes CompilationPipeline internals ("the very first line of compile is if (!key.is_well_formed()) return ...") rather than documenting PermutationKey's own predicate contract. Dictating a sibling aggregate's internal call order is a mild SRP seam leak: the PermutationKey aggregate's design doc should state what is_well_formed() validates and for whom it is intended, but should not prescribe the calling aggregate's implementation order. SPEC §4.3 owns the compile() invariants; that prescription belongs in the CompilationPipeline design doc (spike #749). fix:Rephrase to guidance: "Callers that construct a PermutationKey from arbitrary ordinals (e.g. CompilationPipeline at entry, plugin-ABI receivers) should call is_well_formed() as a defensive pre-condition check before proceeding." Remove the imperative "the very first line of" prescription. The implementation plan for spike #749 can re-state the invariant in its own context.

specs/shader/permutation-key-design.md
```diff
@@ -0,0 +1,1212 @@
...
+| `permutation_key_contributes_six_bytes_to_shader_hash`                      | The `ShaderHash` BLAKE3 input has the key's `to_bytes()` at the documented offset (SPEC §7.3 #1) and changes when any axis changes. |
+| `permutation_key_change_invalidates_shader_hash`                            | For each axis, mutating that axis in a held key produces a different `ShaderHash` (sensitivity). |
+| `permutation_key_with_reserved_bit_set_fails_manifest_load`                 | Hand-crafted manifest payload with a `FeatureSet` reserved bit set is rejected with `Error::CacheIntegrity` (the loader bubbles `PermutationKeyMalformed` up). |
+| `permutation_key_manifest_sort_is_byte_lex`                                 | `ShaderCacheManifest.enumerated_keys` produced from a shuffled input is sorted ascending by `permutation_key_byte_less`. |
+| `permutation_key_survives_plugin_swap_byte_equal`                           | Held `PermutationKey` on a `render` `MaterialBinding` survives a `shader.dylib` swap with no transformation; bytes are identical pre- and post-swap. |
```
cjhowe-us, line 1105: severity:LOW location:specs/shader/permutation-key-design.md:1105 problem:The integration test permutation_key_survives_plugin_swap_byte_equal (§11.2) places a cross-plugin hot-reload validation — exercising a render MaterialBinding surviving a shader.dylib swap — in tests/shader/permutation/. The PermutationKey aggregate is a passive observer in this test; the behavior under test belongs to the hot-reload protocol (hot-reload-protocol.md Step 2 + SPEC §8.5) and the render context's MaterialBinding, not to the permutation-key aggregate's own contract. This test straddles two context boundaries and the shader/permutation/ test directory. fix:When the hot-reload integration harness materializes under sub-epic #70 (or in a cross-plugin tests/ directory), migrate this test there. Until then the [!hide] gate in §11.2 is a reasonable placeholder. Add a comment to the test stub noting the intended migration destination: tests/cross-plugin/hot_reload/ or similar, once that harness exists.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:2

carryover:0

notes:The design document passes the round-2 cohesion + SRP + seam quality lens cleanly. The document owns exactly one aggregate (PermutationKey), each of the 12 sections covers one concern with no cross-contamination, and the four technical responsibilities — closed-enum axis definitions (§3.2), 6-byte canonical encoding (§3.5), bijection codec (§3.4), and spec-frozen comparator (§3.6) — are cleanly separated into distinct subsections with no responsibility bleed between them. The public surface (§4) is appropriately minimal: 6 cardinality constants, 4 enum types, 2 value-object types, 3 member functions, 2 free functions, and exactly 2 error arms. Nothing crosses the plugin ABI that should not. The round-1 HIGH and both MEDs are fully routed to PR #886 (already open with all four corrections confirmed by the impl-respond comments); carryover is zero. Two LOW findings are raised: (1) §3.7 drifts from documenting the PermutationKey predicate contract into prescribing CompilationPipeline's internal call order ("the very first line of compile is..."), a seam leak that the spike #749 implementation plan should own instead; (2) the integration test permutation_key_survives_plugin_swap_byte_equal in §11.2 exercises a cross-plugin hot-reload contract (render MaterialBinding + shader.dylib swap) that belongs more naturally in a cross-plugin hot-reload harness rather than tests/shader/permutation/; migration is deferred until that harness exists. Neither LOW blocks the merged spike intent or the follow-up #886 corrections.

specs/shader/permutation-key-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#886 commit:e9589e8

§3.7 no longer prescribes CompilationPipeline internals. The imperative 'the very first line of compile is...' phrasing has been replaced with defensive guidance: callers should call is_well_formed() defensively at the earliest point they hold a key from an untrusted source; how and where CompilationPipeline does this is left as an implementation detail owned by issue #749.

specs/shader/permutation-key-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#886 commit:e9589e8

§11.2 permutation_key_survives_plugin_swap_byte_equal row now carries a migration note: the test is a stub in tests/shader/permutation/ until a dedicated cross-plugin hot-reload harness exists; intended destination is tests/cross-plugin/ once that harness lands. No code moved — migration deferred until the harness is available.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:0

carryover:0

notes:Round-3 (topology + spec alignment + polish) finds the retro trail complete and coherent. All R1 findings (1 HIGH: SPEC §5 stub missing five symbols; MED-1: from_bytes error-boundary ambiguity; MED-2: sizing table wrong alignof/offsetof; LOW: kFeatureBitCount > 8 divergence buried) and all R2 findings (LOW-1: §3.7 seam leak prescribing CompilationPipeline internal order; LOW-2: §11.2 test in wrong harness location) are fully addressed in PR #886 across two commits. DAG is coherent: #840 (MERGED) is the spike deliverable; #886 (OPEN) is the designated correction vehicle and carries no scope creep. Spec alignment is verified: the five new SPEC §5 symbols (kFeatureBitMask, kPermutationCrossProductCardinality, to_index, from_index, permutation_key_byte_less) are placed correctly after PermutationIndex and before the Stage/CompileTarget section; the corrected sizing table matches the actual ABI for the SPEC's declared PermutationKey struct field order (alignof=2, offsetof(features)=2, sizeof=6); kFeatureBitMask=0x003F and kPermutationCrossProductCardinality=12288 are arithmetic-verified. Issue #747 labels (type:spike, phase:mvp, domain:shader) are correct per AGENTS.md; no pts:* label is present as required; issue state OPEN matches spike closure rules. No new findings. The retro trail closes here; implementation plans under sub-epic #70 may proceed against the corrected SPEC §5 stub once #886 merges.

Issue #757, opened 2026-05-03
[SPIKE] design-shader-shader-backend-detailed


cjhowe-us:
Parent: #744

## Question

Produce the detailed design for the **shader-backend** aggregate in the `shader` context.

IShaderBackend + impls.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/shader/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/rendering/gpu-abstraction-layer.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **shader-backend** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/shader/shader-backend-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(shader): detailed design — shader-backend (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-shader-shader-backend-detailed` is blocked by this issue.

cjhowe-us:
agent:design-shader-shader-backend
status:done
issue:#757
branch:docs/shader-shader-backend-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a66415cbe8b0f95fb
host:mac
cloud:none
commit:990290a87e86eea086f3fc46e210bb5dad6adf95
pr:#845
[...]

Issue #330, opened 2026-05-03
[STORY] Compile Slang to MetalLib via slangc subprocess


cjhowe-us:
### Domain

shader

### Phase

phase:mvp

### Persona

engine developer

### User Story

As an engine developer, I want `CompilationPipeline::compile(source, key, CompileTarget::MetalLib)` to invoke `slangc` as a single out-of-process subprocess that emits a `.metallib` directly from Slang source, so that Apple platforms ship Metal artifacts produced from a single compile hop — never via a DXIL pivot, never via hand-authored MSL.

### Acceptance Criteria (Gherkin)

```gherkin
Given a valid ShaderSource lit.slang
And a well-formed PermutationKey k
When I call CompilationPipeline::compile(lit, k, CompileTarget::MetalLib)
Then the pipeline invokes slangc exactly once (verified via subprocess-trace test hook)
And returns a sealed ShaderArtifact whose bytecode begins with the Metal library file magic
And ShaderArtifact::shader_hash includes (preprocessed_source ‖ k.to_bytes() ‖ canonical_flags ‖ u8(MetalLib))

Given slangc rejects the input with a non-zero exit
When CompilationPipeline::compile is called for MetalLib
Then the call returns Error::MetalLibEmitFailed
And no Metal artifact is published

Given slangc is unavailable on the host (binary missing)
When CompilationPipeline::compile is invoked for MetalLib
Then the call returns Error::CompilerInvocationFailed
```

### Manual Test Script

1. On macOS, in the editor, open `lit.slang`; right-click → "Compile (Metal)".
2. Inspect the output panel; confirm `slangc` ran exactly once and the resulting `.metallib` is non-empty.
3. Rename the `slangc` binary; retry; confirm `CompilerInvocationFailed` is surfaced.
4. Restore the binary; force the pipeline to feed it deliberately-bad Slang source; confirm `MetalLibEmitFailed` is surfaced.

### E2E Test Plan

`tests/e2e/shader/slangc-metallib.glibre-trace` + assertions:
- `assert_subprocess_chain(["slangc"])`
- `assert_value(compile_result)`
- `assert_metal_library_magic(bytecode)`
- `assert_error(missing_binary, Error::CompilerInvocationFailed)`
- `assert_error(bad_source, Error::MetalLibEmitFailed)`

### Closure Checklist (in order)

- [ ] E2E test authored and passing in CI
- [ ] Manual test script reviewed by a second human
[...]

Issue #325, opened 2026-05-03
[STORY] Emit DXIL from Slang via slangc subprocess (post-MVP, D3D12)


cjhowe-us:
### Domain

shader

### Phase

phase:post-mvp

### Persona

engine developer

### User Story

As an engine developer, I want `CompilationPipeline::compile(source, key, CompileTarget::DXIL)` to invoke `slangc` as an out-of-process subprocess that emits DXIL directly from Slang source, so that on D3D12 / Windows hosts shipping plugins are linked without any compiler dependency in-process.

### Acceptance Criteria (Gherkin)

```gherkin
Given a valid ShaderSource lit.slang with vs_main + ps_main
And a well-formed PermutationKey k
When I call CompilationPipeline::compile(lit, k, CompileTarget::DXIL)
Then the returned std::expected resolves to a ShaderArtifact
And ShaderArtifact::bytecode is non-empty and starts with the DXIL container magic
And ShaderArtifact::shader_hash is the BLAKE3 of (preprocessed_source ‖ k.to_bytes() ‖ canonical_flags ‖ u8(DXIL))
And the slangc subprocess was spawned exactly once (verified via test-hook process counter)

Given slangc exits non-zero with a syntax-error stderr stream
When CompilationPipeline::compile is called on the malformed source
Then the call returns Error::CompilerExitNonZero
And the captured stderr is exposed via the structured diagnostic envelope

Given CompilationPipeline::compile is invoked twice with byte-equal inputs (source + key + flags + target)
When both calls complete
Then both ShaderArtifact::shader_hash values are byte-equal
```

### Manual Test Script

1. In the editor on a Windows host, open `lit.slang`; right-click → "Compile (DXIL)".
2. Inspect the output panel; confirm a non-empty DXIL container is reported and a 32-byte BLAKE3 hash is displayed.
3. Edit `lit.slang` to introduce a syntax error; recompile; confirm the editor surfaces the slangc stderr verbatim under `CompilerExitNonZero`.
4. Revert; recompile twice; confirm both runs report identical hashes.

### E2E Test Plan

`tests/e2e/shader/slangc-dxil-compile.glibre-trace` + assertions:
- `assert_value(compile_result)`
- `assert_dxil_container_magic(bytecode)`
- `assert_eq(shader_hash, expected_blake3)`
- `assert_error(syntax_error_case, Error::CompilerExitNonZero)`
- `assert_eq(subprocess_spawn_count, 1)`

[...]

PR #841, opened 2026-05-04
docs(shader): detailed design — compilation-pipeline (refs #749)

cjhowe-us:
## Summary

Detailed design for the shader context's `CompilationPipeline` aggregate
(`specs/shader/SPEC.md` §4.3, §4.7). Twelve sections per the spec
template covering: SRP-bounded purpose (one job: source + key + target
→ sealed artifact); harmonius requirements coverage with explicit
collapses (transpile chain → one slangc hop; in-process trait → driver
subprocess; cross-reflection lib → slangc native reflection paired
with bytecode); detailed model (per-job `Job` value object,
canonicalized argv, eight-stage subprocess lifecycle, length-prefixed
stdout framing, JSON stderr envelope, BLAKE3 cache-key composition,
sealed artifact assembly); public surface (re-uses `IShaderBackend::compile`
from §5; no new ABI surface); cold-only hot/cold split; concurrency
(self-contained per-job, worker-pool scaled, no re-entry, no driver-
thread runs); persistence + ABI (pipeline persists nothing; the
`ShaderHash` recipe is the cross-aggregate ABI); hot-reload
integration (recompile-on-source-change; no `migrate` function);
performance (≤900ms typical, ≤2.0s ceiling, zero per-frame); failure
modes (every §10 arm the pipeline can emit, RAII subprocess reaping,
shipping refusal blanket rule); test plan (mocked-driver unit tests,
real-driver integration tests for #330 / #325, perf benchmarks); and
five `[OPEN]` questions tracked for sub-epic follow-up.

The deliverable explicitly refuses scope outside this aggregate: source
ingestion (#745), permutation key (#747), reflection extraction (#751),
descriptor layout (#753), cache lookup / insertion (#755), and backend
selection (#757) all remain owned by their sibling aggregates.

Refs: #749

## Test plan

- [ ] Reviewer confirms the twelve-section template is honored
- [ ] Reviewer confirms each harmonius element is either covered or
      refused with rationale (§2 table)
- [ ] Reviewer confirms SRP boundary holds (the pipeline does not
      reach into ingestion / key / reflection / descriptor / cache /
      backend selection)
- [ ] Reviewer confirms the §10 error map is consistent with
      `specs/shader/SPEC.md` §10.2

### DIFF
--- a/specs/shader/compilation-pipeline-design.md
+++ b/specs/shader/compilation-pipeline-design.md
@@ -1,0 +1,1053 @@
+# Compilation-Pipeline Detailed Design
+
+> Detailed design for the `shader` context's `CompilationPipeline`
+> aggregate (`specs/shader/SPEC.md` §4.3, §4.7, §5, §6.2). Refines the
+> SPEC's pinned contract that one Slang translation unit plus one
+> `PermutationKey` plus one `CompileTarget` flow through one `slangc`
+> subprocess invocation and return one sealed `ShaderArtifact`. All
+> conclusions independently re-derived; harmonius `docs/design/
+> rendering/render-pipeline.md` § "Shader Compilation Pipeline" cited as
+> research input only.
+
+Refs: spike #749 — `[SPIKE] design-shader-compilation-pipeline-detailed`.
+Parent #744. Sibling task-breakdown spike blocked-by this deliverable.
+
+## 1. Purpose
+
+`CompilationPipeline` is the single component in the `shader` context
+permitted to spawn the `glibre-shadercc` driver subprocess and to
+translate its outputs into a sealed `ShaderArtifact`. Its one
+responsibility is **driving one offline compile job**: take an already-
+opened `ShaderSource`, an already-resolved `PermutationKey`, and a
+selected `CompileTarget`; build the canonicalized argv; spawn the
+driver; capture stdout (length-prefixed bytecode + reflection bytes)
+and stderr (structured `shader::Error` JSON); and return either a
+sealed `ShaderArtifact` (bytecode + `ReflectionBlob` +
+`DescriptorLayout` paired by construction) or a typed `shader::Error`
+(SPEC §4.3, §10).
+
+What `CompilationPipeline` explicitly refuses to own:
+
+- **Source ingestion** — `ShaderSource::open`, include-graph closure,
+  preprocessor expansion, entry-point scan, and total-hash composition
+  belong to the §4.1 aggregate (sibling `shader-source-design.md`,
+  spike #745). The pipeline accepts an already-validated
+  `PreprocessedSource`.
+- **Permutation enumeration / encoding** — `PermutationKey::to_bytes`,
+  `PermutationKey::from_bytes`, `PermutationIndex` ordinal arithmetic,
+  cross-product walks, and project-pruning belong to the §4.2 aggregate
+  (sibling `permutation-key-design.md`, spike #747). The pipeline
+  accepts an already-resolved key value.
+- **Reflection ingestion** — translating slangc's native reflection
+  record into the canonical `ReflectionBlob` (`reflection/
+  slangc_reflection_ingester.cpp`, frequency-tagging) belongs to the
+  §4.4 aggregate (sibling `reflection-blob-design.md`, spike #751).
+  The pipeline only forwards the raw reflection payload from the
+  driver to the ingester and pairs the result with the bytecode.
+- **Descriptor-layout derivation** — projecting `ReflectionBlob` onto
+  the four-frequency tables and the `RootSignatureSchema` belongs to
+  the §4.5 aggregate (sibling `descriptor-layout-design.md`, spike
+  #753). The pipeline calls `DescriptorLayout::derive` exactly once
+  per artifact and embeds the result.
+- **Cache lookup / insertion** — keyed put / get against the CAS, the
+  manifest, and the cooked `ShaderLibrary` belong to §4.6 (sibling
+  `shader-cache-design.md`, spike #755). The pipeline never touches
+  the CAS; the cooker is the sole `insert` site.
+- **Backend selection** — `IShaderBackend` instance choice and
+  capability negotiation belong to the §4.7 plugin trait dispatcher
+  (sibling spike #757). The pipeline accepts an already-bound
+  `IShaderBackend&` and never selects between backends.
+- **Source-watcher signaling** — filesystem events from the platform
+  watcher (#719) reach the pipeline only as already-decoded
+  recompile requests; the pipeline neither registers nor unregisters
+  watches.
+- **Render-graph topology, PSO objects, GPU resource allocation** —
+  refused (`specs/shader/SPEC.md` §3 refusals 1, 4, 5).
+- **Runtime compilation in shipping** — the entire
+  `CompilationPipeline` translation unit is excluded from the shipping
+  link target (`specs/shader/SPEC.md` §4.3 invariant 3, §4.8 invariant
+  3, §6.5). Shipping has no `compile()` virtual at the ABI surface.
+
+The pipeline's SRP boundary is sharp: if the driver argv schema, the
+stdout / stderr framing, the per-job synchronization model, the
+exit-code → `shader::Error` mapping, the per-job cache-key composition,
+or the in-process pairing of bytecode + reflection + descriptor layout
+changes, this design changes. Anything else is out of scope.
+
+## 2. Requirements Coverage
+
+Mapping of harmonius `docs/design/rendering/render-pipeline.md` §
+"Shader Compilation Pipeline" + `docs/requirements/rendering/
+gpu-abstraction-layer.md` R-2.1.16/.17/.18 to glibre MVP refusal-or-
+coverage. Every entry independently re-derived; harmonius is research
+input only.
+
+| Harmonius element | Glibre disposition (MVP) | Coverage site |
+|-------------------|--------------------------|---------------|
+| `HLSL → dxc → DXIL → metal-shaderconverter → metallib` transpile chain | **Refused — collapsed.** | One subprocess hop per artifact (`Slang → slangc → metallib` MVP, `Slang → slangc → DXIL` post-MVP). SPEC §3 collapse 1; this design §3.4. |
+| `ShaderCompiler::compile(desc) → ShaderCompileResult` (in-process trait) | **Partial — covered by `IShaderBackend::compile` only.** | Slangc subprocess invoked behind the trait via the `glibre-shadercc` driver; never linked in-process (SPEC §3 collapse 3, §4.3 invariant 1). This design §3.5 + §3.6. |
+| `ShaderTarget { Dxil, SpirV, MetalLib }` | **Partial — `MetalLib` MVP, `DXIL` post-MVP, `SpirV` deferred.** | `CompileTarget` enum (SPEC §5). Pipeline rejects `DXIL` until post-MVP support lands (`Error::UnsupportedTarget`). SPIR-V removed from the enum entirely (SPEC §11 acceptance criteria — #327 closed). |
+| `metal-shaderconverter` as a second subprocess hop | **Refused.** | Slangc emits `metallib` directly from Slang; no transpile chain (SPEC §4.3 invariant 2). |
+| Reflection extracted from a separate cross-reflection library | **Refused.** | slangc's native reflection API emits the reflection record paired with the bytecode in the *same* subprocess invocation (SPEC §3 collapse 4, §4.4 invariant 1). The pipeline forwards the paired payload to the §4.4 ingester. |
+| Runtime `OnDemandCompiler` | **Refused for shipping.** | `CompilationPipeline` is `#if !GLIBRE_SHIPPING`-guarded out of the plugin (SPEC §4.3 invariant 3, §4.8 invariant 3). Editor / cooker is the only caller. |
+| GR-2 / GR-4 in-process linkage | **Refused.** | slangc is invoked exclusively via the `glibre-shadercc` driver subprocess (SPEC §3 collapse 3, §6.2). |
+| R-2.1.18 structured errors at every public boundary | **Covered.** | `std::expected<ShaderArtifact, shader::Error>`. Driver stderr is a stable JSON envelope mapped onto the closed §10 enum. This design §3.7 + §10. |
+
+Glibre-native requirements added beyond harmonius:
+
+- **One artifact = one sealed value.** The pipeline's success return is
+  a single `ShaderArtifact` whose `bytecode` + `reflection` +
+  `descriptor_layout` fields are all populated by construction. No
+  partially-constructed artifact escapes the pipeline (SPEC §4.4
+  invariant 1; this design §3.8).
+- **Per-job determinism.** Two pipeline invocations with byte-equal
+  `(PreprocessedSource, PermutationKey, CompileTarget, canonical
+  flags)` produce structurally-equal `ShaderArtifact` values modulo
+  the subprocess non-determinism that the canonicalization layer
+  elides (SPEC §4.7 invariant 3; this design §3.7).
+- **Refusal at every step is logged once and leaves no on-disk
+  state.** The pipeline never writes to the CAS; failure leaves the
+  cache byte-equal to the pre-call snapshot (`error-model.md`).
+  This design §10.
+- **Hot path is empty.** Shipping has no compile path; editor / cook
+  is the sole hot path, and even there the pipeline is a cold-only
+  surface (PHILOSOPHY §6, §8). This design §5 + §9.
+- **Bounded concurrency via worker pool.** Multiple jobs run in
+  parallel; one job's state is self-contained (SPEC §4.3
+  "Stateless"). This design §6.
+
+## 3. Detailed Model
+
+### 3.1 Aggregate composition
+
+```text
+CompilationPipeline (root, owned by editor / cooker)
+├── const IShaderBackend&      backend_   (ref; selected by §4.7 dispatcher)
+├── ArgvBuilder                argv_      (canonicalizer; deterministic)
+├── DriverSpawner              spawner_   (subprocess launcher — fork/exec wrapper)
+├── PayloadDecoder             decoder_   (length-prefixed stdout framer)
+├── ErrorEnvelopeReader        errors_    (stderr JSON → shader::Error)
+├── const reflection::Ingester& ingester_ (ref; §4.4 ingester)
+├── DescriptorLayout::derive   derive_    (free function; §4.5)
+└── LogSink&                   log_       (spdlog-backed)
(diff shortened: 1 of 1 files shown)


specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+The pipeline can emit any of the following enumerators from SPEC §5
+/ §10.2. Any other arm in §5 is emitted by a sibling aggregate
+(ingester, cache, source) and surfaces *through* the pipeline only
+when the pipeline calls into that sibling.
+
+| Enumerator | Trigger | Recovery | Severity |
+|------------|---------|----------|----------|
+| `EntryPointMissing` | The requested job names a non-existent entry point on the source. | refuse compile; pipeline propagates from `IShaderBackend::compile` argument check before spawn. | refuse |
+| `EntryPointStageAmbiguous` | Source carries an entry point with zero or multiple `[shader(...)]` tags (caught by §4.1, surfaced through pipeline). | refuse compile; defer to source aggregate's resolution. | refuse |
+| `PermutationKeyOutOfRange` | `key.is_well_formed()` is `false` at job construction. | refuse compile; codegen-table drift between caller and pipeline. | fatal |
```
cjhowe-us, line 847: severity:HIGH location:specs/shader/compilation-pipeline-design.md:847 problem:§10.1 row for `PermutationKeyOutOfRange` conflates two distinct SPEC §5 enumerators. The design maps `key.is_well_formed() == false` → `PermutationKeyOutOfRange` with severity `fatal`. But per SPEC §10.2: `PermutationKeyMalformed` is the enumerator for `from_bytes` rejection (severity `refuse`), while `PermutationKeyOutOfRange` specifically covers a `PermutationIndex` exceeding the §4.2 cardinality product (severity `fatal`). `is_well_formed()` returning false is the `PermutationKeyMalformed` case (`refuse`), not `PermutationKeyOutOfRange`. The design silently drops `PermutationKeyMalformed` from the §10.1 table entirely. fix:Add a `PermutationKeyMalformed` row to §10.1 with trigger = "`from_bytes()` rejects bytes (enumerator out of declared range at deserialization)"; severity = `refuse`. Correct the `PermutationKeyOutOfRange` trigger to "a `PermutationIndex` exceeds the §4.2 cardinality product; indicates codegen-table drift"; keep severity `fatal`. Remove the current `is_well_formed()` trigger from the `PermutationKeyOutOfRange` row.

specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+pids and leaked fds. Every error path that returns out of `compile`
+runs this destructor by stack-unwinding through `std::expected`'s
+construction.
+
+### 10.4 Logging
+
+Per `error-model.md`:
+
+- `error` severity: `CompilerInvocationFailed`,
+  `CompilerTimedOut`, `MetalLibEmitFailed`, `CacheReadOnlyViolation`
```
cjhowe-us, line 921: severity:MED location:specs/shader/compilation-pipeline-design.md:921 problem:§10.4 lists `CacheReadOnlyViolation` as a pipeline-emitted `error`-severity log entry, but §10.1 (the same section) explicitly states the pipeline does NOT emit `CacheReadOnlyViolation` directly: "The pipeline does not emit `SourceNotFound`, ... `CacheReadOnlyViolation` directly." The two clauses contradict each other. fix:Remove `CacheReadOnlyViolation` from the §10.4 error-severity list. If the design intends to log it defensively (e.g. if the pipeline is ever invoked against a read-only cache), add a parenthetical explaining the pipeline propagates it only when the caller passes a read-only cache store, and clarify whose log responsibility that is (the cooker, which calls pipeline). The §10.1 "does not emit directly" language is the spec-aligned statement; §10.4 must not contradict it.

specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+The pipeline owns this discipline through RAII helpers:
+
+```cpp
+class SubprocessHandle {
+public:
+    ~SubprocessHandle() {
+        if (pid_ != -1) {
+            kill(pid_, SIGKILL);  // belt-and-suspenders
+            int status;
+            ::waitpid(pid_, &status, 0);
```
cjhowe-us, line 897: severity:MED location:specs/shader/compilation-pipeline-design.md:897 problem:The `SubprocessHandle` RAII destructor in §10.3 calls `kill(pid_, SIGKILL)` unconditionally before `waitpid` on the success path. After a normal job completion (stage 5 exit → stage 6+7 drain → stage 8 assemble), `waitpid` has already been called in stage 5 to capture the exit status. The destructor then fires on the same `pid_`, calling `kill` and `waitpid` again on an already-reaped pid. On POSIX, `waitpid` on a reaped pid returns `ECHILD`; `kill` on an already-reaped pid with a recycled pid could signal a different process. fix:Add a `reaped_` bool member (default false) to `SubprocessHandle`; set it to `true` after the stage-5 `waitpid` call (or expose a `release()` method that sets `pid_ = -1`). The destructor`s guard should be: `if (pid_ != -1)` — and the stage-5 code should call `release()` after a successful `waitpid`. Alternatively, expose a `wait() -> int` method that returns the exit status and sets `pid_ = -1`, so the destructor is a pure cleanup guard and never re-waits on an already-reaped pid.

specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+    const ShaderSource&  source,
+    const PermutationKey key,
+    const CompileTarget  target,
+    const CanonicalFlags& flags) {
+  Blake3Hasher h;
+  h.update(source.preprocessed().total_hash.bytes);    // 32 B
+  h.update(key.to_bytes());                            //  6 B
+  h.update(flags.hash.bytes);                          // 32 B
+  std::byte t = static_cast<std::byte>(target);
+  h.update(eastl::span(&t, 1));                        //  1 B
```
cjhowe-us, line 335: severity:LOW location:specs/shader/compilation-pipeline-design.md:335 problem:The illustrative cache-key code in §3.7 uses `eastl::span(&t, 1)` without a template argument. EASTL`s `span<T>` template deduction from a raw `std::byte*` pointer may not unambiguously deduce the `T = std::byte` parameter in all clang versions, and the `const`-ness is dropped (the pointer is non-const). This is a code-example precision issue that could mislead the implementer. fix:Write `h.update(eastl::span<const std::byte>(&t, 1))` with the explicit template argument and the `const`-qualified element type, consistent with how the BLAKE3 hasher`s `update()` signature is expected to receive read-only byte views.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:2
  LOW:1
carryover:0 (round 1, no prior rounds)
notes:The design is structurally sound and substantially complete across all twelve required sections. The SRP boundary is well-drawn: every in-scope responsibility is present and every out-of-scope responsibility is explicitly refused with a citation. The source → key → target → sealed artifact pipeline is coherent and the stage lifecycle (compose_argv, compose_cache_key, posix_spawn, stdin write, kqueue await, stdout/stderr capture, artifact assembly) is well-specified. The Conventional Commit subject is correct (`docs(shader):`) and `Refs: #749` is present in both the commit message and the document header. The error surface uses `std::expected<ShaderArtifact, shader::Error>` consistently, which is correct per SPEC §5 (the engine-wide `glibre::Error` alias appends this enum when the plugin lands; SPEC §5 footnote). Decision record citations are all legitimate and their clause references are accurate. One HIGH finding blocks clean sign-off: §10.1 conflates `PermutationKeyMalformed` and `PermutationKeyOutOfRange`, dropping the former entirely and mislabeling the trigger for the latter. SPEC §10.2 gives them different triggers and different severity classes (refuse vs fatal), and an implementer who reads only the design doc will emit wrong error arms. Two MED findings should be addressed before implementation: (1) the internal contradiction in §10.4 that lists `CacheReadOnlyViolation` as a pipeline log output while §10.1 explicitly says the pipeline does not emit it directly; (2) the `SubprocessHandle` RAII destructor sketch calls `kill+waitpid` in the destructor on a pid that was already `waitpid`'d in stage 5, risking a signal to a recycled pid — the sketch needs a `release()` or `reaped_` guard. One LOW finding: the §3.7 illustrative code uses `eastl::span(&t, 1)` without the template argument, which may not compile and may mislead the implementer. The twelve-section template is honored; harmonius coverage table is complete; shipping-exclusion contract is correctly stated; §9 performance numbers agree with `perf-budget.md`; five open questions are appropriate scope for follow-up.

specs/shader/compilation-pipeline-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b775af9 followup:#885

Split the conflated row into two distinct §10.1 entries: `PermutationKeyMalformed` (`is_well_formed()==false` → refuse) and `PermutationKeyOutOfRange` (`PermutationIndex` exceeds cardinality → fatal), matching SPEC §10.2 exactly.

specs/shader/compilation-pipeline-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b775af9 followup:#885

Removed `CacheReadOnlyViolation` from §10.4's pipeline-emitted error list. Added a parenthetical cross-reference to §10.1's 'not emit' list to make the boundary explicit.

specs/shader/compilation-pipeline-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b775af9 followup:#885

Added `reaped_` boolean field and `mark_reaped()` method to `SubprocessHandle`. Destructor now gates the `kill+waitpid` block on `!reaped_`. Stage 5 calls `mark_reaped()` immediately after its `waitpid` call, so the destructor never signals a recycled pid.

specs/shader/compilation-pipeline-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b775af9 followup:#885

Fixed to `eastl::span<const std::byte>(&t, 1)` — explicit template argument eliminates deduction ambiguity and restores the `const`-ness that the raw pointer lacked.

specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+
+```text
+CompilationPipeline (root, owned by editor / cooker)
+├── const IShaderBackend&      backend_   (ref; selected by §4.7 dispatcher)
+├── ArgvBuilder                argv_      (canonicalizer; deterministic)
+├── DriverSpawner              spawner_   (subprocess launcher — fork/exec wrapper)
+├── PayloadDecoder             decoder_   (length-prefixed stdout framer)
+├── ErrorEnvelopeReader        errors_    (stderr JSON → shader::Error)
+├── const reflection::Ingester& ingester_ (ref; §4.4 ingester)
+├── DescriptorLayout::derive   derive_    (free function; §4.5)
```
cjhowe-us, line 131: severity:MED location:specs/shader/compilation-pipeline-design.md:131 problem:§3.1 aggregate-composition diagram lists `DescriptorLayout::derive   derive_   (free function; §4.5)` as a named line item alongside concrete stored members (`ArgvBuilder argv_`, `DriverSpawner spawner_`, etc.). A free function is not an owned member and has no storage, lifetime, or ABI contribution to the aggregate. The parenthetical `(free function; §4.5)` partially disambiguates but does not prevent an implementer from reading the diagram as implying a stored callable (function pointer, `eastl::function`, or policy template) — especially because every other line item in the tree IS a stored member or reference. This muddies the §4.3/§4.5 seam: §4.5 owns `derive`; §4.3 calls it. fix:Remove `DescriptorLayout::derive derive_` from the §3.1 composition tree entirely. Replace with a prose note below the diagram: "The pipeline calls `DescriptorLayout::derive` (§4.5 free function) inline inside `assemble_artifact` (§3.8). No instance is stored; the function is resolved at compile time." This preserves the implementer signal without suggesting ownership.

specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+```
+
+The pipeline owns the subprocess pid + pipes for the duration of one
+`compile()` call and not one byte longer; on return the artifact is
+the caller's, and on error nothing escapes (`std::expected` carries
+only the error enumerator).
+
+## 4. Public Surface
+
+### 4.1 Types (locked from SPEC §5)
```
cjhowe-us, line 407: severity:LOW location:specs/shader/compilation-pipeline-design.md:407 problem:Section 6.2 documents the cooker and editor recompile dispatcher worker-pool implementation in detail (MPSC ring buffer, 256-job bounded capacity, hardware_concurrency() sizing formula, producer/consumer mutex strategy). These are caller-aggregate internals. The cooker owns the pool that calls pipeline.compile(), not the pipeline itself. Per SRP the pipeline's section 6 should describe only what the pipeline guarantees per-job (self-containment, no cross-job shared mutable state); caller pool strategy belongs in the cooker/editor designs (sibling spikes). fix:Trim section 6.2 to a concise statement of the pipeline's per-invocation concurrency contract: 'Each compile() invocation is self-contained. Callers may invoke compile() concurrently from any number of threads; overlapping arguments duplicate the subprocess invocation (deduplication is the cooker's responsibility per SPEC 6.4 step 2e).' Move the MPSC ring buffer capacity, hardware_concurrency formula, and the worker-loop code snippet to the cooker design spike.

specs/shader/compilation-pipeline-design.md
```diff
@@ -0,0 +1,1053 @@
...
+pipeline's lifetime.
+
+Driver argv shape (locked by SPEC §6.2; this section transcribes it
+for the pipeline implementer):
+
+```text
+glibre-shadercc
+    --schema-version=1
+    --source-path=<absolute path to .slang file>
+    --include-root=<absolute project source root>
```
cjhowe-us, line 214: severity:LOW location:specs/shader/compilation-pipeline-design.md:214 problem:Section 3.4 shows '--source-path=<absolute path>' in the driver argv, while section 3.5 stage 4 says source bytes are piped via stdin to 'avoid leaking the source path through the filesystem.' An implementer reading section 3.4 first sees a --source-path argument and may question why the path is passed when bytes arrive via stdin; without reading the stage-4 paragraph the two look contradictory. The design does not state that --source-path is used only for diagnostic file references in slang error output (not for filesystem reads by the driver or slangc). fix:Add a sentence after the --source-path line in the argv schema: 'The driver uses --source-path for diagnostic file-path attribution in slang error messages only; it does not read the source file from this path. The preprocessed source bytes arrive via stdin (stage 4) — this is what the sandbox model depends on.'

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0 (all four R1 findings addressed in commit b775af9; follow-up in PR #885)
notes:Round-2 cohesion + SRP + seam-quality pass. The document owns a single aggregate (CompilationPipeline) without exception; every responsibility is either claimed with a section and SRP statement, or explicitly refused with a SPEC citation and sibling-spike pointer. The twelve-section template is honored and each section maps to exactly one concern. No invented types or cross-context variants appear. The five-stage pipeline described in SPEC §4.3 is rendered as eight stages in §3.5, which is correct and internally consistent (compose_argv + compose_cache_key are pure-function preambles, not I/O stages). The public surface is minimal: the pipeline re-uses IShaderBackend::compile from §5 and adds no new ABI surface. std::expected<ShaderArtifact, shader::Error> is used consistently at the one public boundary. One MED seam issue: the §3.1 aggregate-composition diagram lists DescriptorLayout::derive as a named tree node alongside stored members and references. A free function has no storage or lifetime and should not appear in the composition diagram; its presence implies an owned callable and blurs the §4.3/§4.5 seam. Two LOW issues: §6.2 documents the cooker/editor's worker-pool internals (MPSC ring buffer capacity, hardware_concurrency formula, worker loop) which belong in the cooker design spike, not in the pipeline's concurrency section; and §3.4 does not explain why --source-path appears in argv alongside stdin-piped source bytes, which creates an apparent contradiction with the sandbox rationale in §3.5 stage 4. All three findings are suitable for PR #885 since they are documentation-layer clarifications with no behavioral impact; the impl-respond agent should route them there rather than opening a new PR.

cjhowe-us:
responder:go-impl-respond round:2 decision:ADDRESSED followup:#885 commit:3408fbc

All three round-2 findings addressed in commit 3408fbc on PR #885 (fix/shader-compilation-pipeline-design-followup-r1):

- MED (§3.1 aggregate composition diagram): Removed DescriptorLayout::derive from the diagram (it is a free function with no storage); added prose note that it is called inline inside assemble_artifact (stage 8, §3.8). decision:ADDRESSED
- LOW-1 (§6.2 worker-pool internals): Replaced the MPSC ring / hardware_concurrency / worker-loop block with a per-invocation self-containment statement; cooker worker-pool mechanics deferred to the cooker design spike. decision:ADDRESSED
- LOW-2 (§3.4 --source-path contradiction): Added explicit prose clarifying that --source-path is for diagnostic attribution only (driver does not open the file; source bytes arrive via stdin); resolves the apparent contradiction with the §3.5 sandbox rationale. decision:ADDRESSED

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:2
carryover:0
notes:Round-3 topology + spec alignment + polish pass on merged PR #841. The dependency DAG is coherent: all R1 findings (HIGH:1 PermutationKey split, MED:2 CacheReadOnlyViolation contradiction + SubprocessHandle double-waitpid, LOW:1 eastl::span template) and all R2 findings (MED:1 DescriptorLayout::derive in composition diagram, LOW:2 §6.2 cooker internals + --source-path contradiction) are addressed in the open follow-up PR #885 with two clean commits routed correctly. Issue #749 carries the right labels (type:spike, phase:mvp, domain:shader; no pts:* which is correct for spikes). SPEC §10.2 per-enumerator table alignment is complete in PR #885: PermutationKeyMalformed (refuse, is_well_formed==false) and PermutationKeyOutOfRange (fatal, PermutationIndex > cardinality) match SPEC §10.2 verbatim. Shipping-exclusion contract (§4.3 invariant 3, §4.8 invariant 3) is correctly documented. Two LOW polish findings for PR #885 to address before merge: (1) PR #885 title credits only "r1 followup" but the second commit (3408fbc) addresses R2 findings — body is accurate but the title will mislead git log; recommend updating the PR title to "r1+r2 followup" or "compilation-pipeline-design corrections". (2) design doc §10.2 cross-reference (line ~870) cites "shader/SPEC.md §10.1" for the per-enumerator severity contract but the normative table lives at SPEC §10.2; the cite should be SPEC §10.2. Neither finding blocks merge of PR #885. The three-round retro trail for PR #841 is complete; this verdict is final.

Issue #79, opened 2026-05-03
[SPIKE] design-shader-failure-modes


cjhowe-us:
Parent: #69

## Question

Typed errors + recovery for shader public boundaries.

## Deliverable

Fill §10 of `specs/shader/SPEC.md`.

## Closure

Deliverable committed; parent sub-epic checkbox flipped.

cjhowe-us:
agent:adv-shader-fail
status:started
issue:#79
branch:chore/spec-shader-§10-failure-modes
worktree:/tmp/glibre-shader-fail
host:mac
cloud:none
commit:2f6c08a
pr:none
notes:Filling §10 of specs/shader/SPEC.md. Will complete the shader::Error closed sum (extending the §5 declaration with the SPIKE-mandated additions: SourceParseError, IncludeResolutionFailed, DxcInvocationFailed, DxcOutputDiagnostic, MetalShaderConverterFailed, ReflectionParseError, DescriptorLayoutInvalid, PermutationOutOfRange, CacheIntegrity, CacheCorrupt, CacheMiss, ArtifactSizeExceeded, ShaderCompileTimeout, ShippingBuildCannotCompile) and document each enumerator's trigger, recovery, and severity. Edit only §10. PR + auto-merge squash; issue stays OPEN.

cjhowe-us:
agent:adv-shader-fail
status:done
issue:#79
branch:chore/spec-shader-§10-failure-modes
worktree:/tmp/glibre-shader-fail
host:mac
cloud:none
commit:1a3c407
pr:#312
notes:Filled §10 of specs/shader/SPEC.md. Established a 3-class severity vocabulary (refuse / fallback / fatal) bound to cache-state and observer-bus contracts (§8.4 projection). Pinned every §5 shader::Error arm to its trigger, recovery, and severity. Encoded the shipping-build blanket rule for DXC / metal-shaderconverter (ShippingCompilationAttempted, fatal). Tracked ArtifactSizeExceeded as a §5 amendment scoped to a follow-up plan, with a temporary CacheIntegrity-mapping rule preserving the closed-sum guarantee. Edited only §10. PR #312 squash-merged via auto-merge. Issue stays OPEN per spike closure rules in AGENTS.md (parent sub-epic checkbox flip is the closure trigger).

cjhowe-us:
Deliverable merged. Closing per AGENTS.md spike-closure rule.

Issue #69, opened 2026-05-03
[SUB-EPIC] Quality — shader


cjhowe-us:
# Quality — shader

Internal architecture sketch, perf budget, failure modes.

Parent epic: #66

## Sections owned (in `specs/shader/SPEC.md`)

- §6 Internal Architecture
- §9 Performance Budget
- §10 Failure Modes & Error Model

No estimate; rolls up.

PR #845, opened 2026-05-04
docs(shader): detailed design — shader-backend (refs #757)

cjhowe-us:
## Summary

- Pin the `IShaderBackend` trait surface (compile / reflect / link / capabilities) and the sole MVP implementation, `SlangMetalBackend`, wrapping the `glibre-shadercc` driver hop already designed in sibling spike #749.
- Introduce the `MetalLibraryLoader` seam that materialises cached `metallib` bytes into `MTL::Library` / `MTL::Function` handles via metal-cpp, so render's `PSOCache` consumes opaque handles without ever touching slangc, Objective-C, or raw bytecode containers. No Obj-C++ in engine code.
- Cover all 12 SPEC-template sections with refusal/coverage tables for harmonius `gpu-abstraction-layer.md` R-2.1.x clauses; refuse what belongs to render (`MetalDevice`, `PSOCache`); propose two new `shader::Error` arms (`MetalLibraryCreateFailed`, `FunctionMissing`) gated behind a §5 amendment plan.

Refs spike #757; sibling task-breakdown blocked-by this deliverable. Issue stays OPEN.

## Test plan

- [ ] Catch2 unit tests at `tests/shader/backend/` exercising every §10 refusal arm via mocked metal-cpp + a fixture `glibre-shadercc` binary (24 named tests in §11.1).
- [ ] Catch2 integration tests at `tests/shader/integration/` exercising the real `glibre-shadercc` driver + a real Metal device acquired via `tests/platform/metal_device_fixture.cpp` (gated by `MACOS_HAS_METAL=1`).
- [ ] Catch2 `BENCHMARK` blocks at `tests/shader/perf/backend_bench.cpp` asserting cold-path budgets (`compile` <100 ms, `load_library` <2 ms, `lookup_function` <0.05 ms, `capabilities` <0.001 ms).
- [ ] E2E traces at `tests/e2e/shader/backend/` covering happy-path, library-create-failure, and entry-point-rename hot-reload — each with a recorded `.glibre-trace` golden.

### DIFF
--- a/specs/shader/shader-backend-design.md
+++ b/specs/shader/shader-backend-design.md
@@ -1,0 +1,1520 @@
+# Shader-Backend Detailed Design
+
+> Detailed design for the `shader` context's `IShaderBackend` aggregate
+> and its sole MVP implementation, the **Slang+Metal backend** (SPEC
+> §4.7, §5 lines 825–849, §6.1 `backend/`, §6.2 driver flow). The
+> aggregate's role at the public boundary is to define and implement the
+> four-operation seam — `compile`, `reflect`, `link`, `capabilities` —
+> through which the rest of the engine consumes one shader source
+> language family. In MVP that family is Slang; the implementation
+> wraps the offline `glibre-shadercc` driver (the `slangc` subprocess
+> hop pinned in SPEC §6.2) for compile / reflect / link, and binds the
+> resulting cached `ShaderArtifact` records into Metal 4
+> `MTL::Library` / `MTL::Function` handles via metal-cpp so that
+> `render`'s `PSOCache` (`specs/render/SPEC.md` §4.1.7) can build
+> pipeline-state objects without ever touching slangc, Objective-C, or
+> raw bytecode containers.
+>
+> Authority: this document refines `specs/shader/SPEC.md` §4.7, §4.8
+> (cross-aggregate invariants), §5 (lines 815–849 + the `IShaderBackend`
+> trait + `Capabilities` + `LinkedModule`), §6.1 `backend/`, §6.2
+> (`glibre-shadercc` driver flow), §6.5 (shipping cut), §8.4 refusal
+> cases, §9.3 (per-aggregate budget for `IShaderBackend`), and §10.2
+> rows `CompilerInvocationFailed`, `CompilerExitNonZero`,
+> `CompilerTimedOut`, `UnsupportedTarget`, `MetalLibEmitFailed`,
+> `LinkFailed`, `SpecializationConstantMissing`, `CapabilityNotSupported`,
+> `ShippingCompilationAttempted`. Anything not addressed here defers to
+> those sections; anything that appears to contradict them is a defect
+> in this document.
+
+Refs: spike #757 — `[SPIKE] design-shader-shader-backend-detailed`.
+Parent sub-epic #744 (`[SUB-EPIC] Detailed Designs — shader`). Sibling
+task-breakdown spike blocked-by this deliverable.
+
+---
+
+## 1. Purpose
+
+The shader-backend aggregate owns **the narrow plugin-trait seam
+through which the engine talks to one shader source-language family**,
+and the sole MVP implementation of that seam: a Slang backend that
+spawns the `glibre-shadercc` driver subprocess (offline) and that
+materializes the resulting cached `ShaderArtifact` records into
+Metal 4 `MTL::Library` / `MTL::Function` handles via metal-cpp
+(offline + dev / editor; the cooked artifact bytes themselves ride
+into shipping through `ShaderCache::Library`).
+
+The aggregate is two responsibilities held in deliberate tension by
+SRP, joined only because they share a single seam:
+
+1. **Trait surface.** Define the `IShaderBackend` v-table the engine
+   compiles against — `compile`, `reflect`, `link`, `capabilities` —
+   with shipping-build link exclusion of the `compile` virtual
+   (SPEC §4.3 inv 3, §4.8 inv 3). The trait is the entire ABI the
+   `shader` plugin exposes for the cooker, the editor adapter, and
+   the `render`-side PSO builder.
+2. **Slang+Metal implementation.** Provide one concrete v-table —
+   `SlangMetalBackend` — that fans out to four sub-services:
+   - `compile` → spawns `glibre-shadercc` (one slangc subprocess hop)
+     to emit a `metallib` payload + paired native reflection JSON;
+     forwards reflection bytes to the §4.4 ingester; embeds the
+     resulting `ReflectionBlob` and the §4.5 `DescriptorLayout` into
+     a sealed `ShaderArtifact`.
+   - `reflect` → re-runs the §4.4 ingester against an
+     already-cached artifact's reflection blob bytes (idempotent;
+     caller is the cooker rebuilding a cache after a schema bump,
+     not a per-frame path).
+   - `link` → emits a single combined `LinkedModule` from a span of
+     compatible `ShaderArtifact`s, baking specialization constants
+     where Metal 4 admits it and otherwise refusing with
+     `Error::SpecializationConstantMissing` /
+     `Error::LinkFailed`.
+   - `capabilities` → returns the static, offline-pinned capability
+     descriptor for the active backend (mesh shaders, RT, work
+     graphs, wave intrinsics, fp16). Pure function; no runtime
+     query against the Metal device.
+3. **Library materialization.** A separate, narrow service —
+   `MetalLibraryLoader` — that takes a sealed `ShaderArtifact`
+   (already in the cache, byte-validated by BLAKE3) plus a
+   `metal_cpp::MTL::Device*` borrowed from `render`'s `MetalDevice`
+   (`specs/render/SPEC.md` §4.1.6) and returns an opaque
+   `MetalLibraryHandle` carrying:
+   - one `MTL::Library*` constructed via
+     `MTL::Device::newLibrary(dispatch_data, NS::Error**)` over the
+     artifact's `bytecode` span, **never** via Slang source or MSL
+     text — shipping has no slangc and no MSL;
+   - one `MTL::Function*` per `EntryPoint` in the artifact, looked
+     up by name through `MTL::Library::newFunction(NS::String*)`.
+
+This third service is the load-bearing addition this design pins
+beyond what SPEC §5 already declares: SPEC §5's `IShaderBackend`
+returns `ShaderArtifact` value-objects whose `bytecode` is opaque to
+the `shader` context; the **materialization seam** turns those bytes
+into the GPU-driver-resident objects that `render`'s `PSOCache`
+consumes when building `MTL::RenderPipelineState`. Without it,
+`render` would have to either (a) call metal-cpp's library-create API
+itself — pulling slangc-aware reflection knowledge into `render` — or
+(b) re-read the cache on the render thread, duplicating the
+shader-context BLAKE3 lookup. Both violate SRP. The materialization
+seam belongs in `shader/backend/` because it pairs the artifact with
+the same `ReflectionBlob` (entry-point names, stage tags) the
+backend produced — the binding from "blob of bytes" to "callable
+function handle" is a `shader`-context concern.
+
+What `IShaderBackend` and the materialization seam **explicitly refuse
+to own**:
+
+- **PSO objects, render-graph passes, command encoders.** The
+  `MTL::RenderPipelineState` / `MTL::ComputePipelineState` objects
+  built from `(MetalLibraryHandle, render-state-hash)` belong to
+  `render`'s `PSOCache` (`specs/render/SPEC.md` §4.1.7, sub-epic
+  #766). The shader-backend hands `render` opaque library + function
+  handles and stops at the v-table boundary.
+- **The Metal device.** `MetalDevice` ownership and lifetime live in
+  `render` (`specs/render/SPEC.md` §4.1.6, §4.8 invariant cited).
+  The shader-backend takes a borrowed `MTL::Device*` per call and
+  never retains it past the call. The shader plugin does not
+  `dlopen` metal-cpp; metal-cpp linkage stays inside the
+  `backend/slang/` translation unit, which is shipping-excluded for
+  the slangc half and shipping-included only for the loader half.
+- **Source ingestion.** `ShaderSource` (SPEC §4.1, sibling
+  `shader-source-design.md`, spike #745) reads `.slang` files,
+  expands `#include`, scans entry-point attributes. The backend
+  consumes a `const ShaderSource&` only.
+- **Permutation enumeration / encoding.** `PermutationKey`
+  (SPEC §4.2, sibling `permutation-key-design.md`, spike #747)
+  defines the 4-axis closed sum. The backend consumes a
+  `const PermutationKey&` and never enumerates the cross-product.
+- **Compilation pipeline orchestration.** `CompilationPipeline`
+  (SPEC §4.3, sibling `compilation-pipeline-design.md`, spike #749)
+  drives the `glibre-shadercc` driver invocation, builds the
+  canonicalized argv, captures stdout/stderr, and translates exit
+  codes onto `shader::Error`. The backend's `compile` *delegates* to
+  the pipeline; it does not duplicate driver-spawning logic.
+- **Reflection ingestion internals.** The §4.4 ingester
+  (`reflection/slangc_reflection_ingester.cpp`, sibling spike #751)
+  parses slangc native reflection JSON. The backend forwards the
+  raw reflection bytes from the driver to the ingester and pairs
+  the result with the bytecode in one `ShaderArtifact`.
+- **Descriptor-layout derivation.** `DescriptorLayout::derive`
+  (SPEC §4.5, sibling `descriptor-layout-design.md`, spike #753)
+  projects a `ReflectionBlob` onto the four-frequency tables. The
+  backend invokes `derive` once per artifact and embeds the result.
+- **Cache lookup / insertion.** The `ShaderCache` (SPEC §4.6,
+  sibling `shader-cache-design.md`, spike #755) owns put / get
+  against the CAS. The backend's `compile` returns a fresh
+  `ShaderArtifact`; the cooker (sole writer) is the call site that
+  inserts.
+- **Backend selection across source-language families.** Today
+  there is one source-language family (Slang). A second family
+  (e.g. a future raw-MSL backend, hypothetical only) would arrive
+  as a second `IShaderBackend` implementation; the trait dispatcher
+  would select between them. MVP has no dispatcher because there
+  is one impl. The trait surface admits the second impl without
+  changing the surface.
+- **Capability negotiation policy.** The cooker pre-walks the
+  enumerated permutation set against `capabilities()` to refuse
+  unsupported permutations early (`Error::CapabilityNotSupported`,
+  SPEC §10.2). That filtering policy is the cooker's; the backend
+  only reports the static descriptor.
+- **Runtime compilation in shipping.** The entire `compile()`
+  virtual is `#if !GLIBRE_SHIPPING` in SPEC §5 (lines 836–840);
+  the slangc subprocess driver is not bundled into shipping
+  (SPEC §6.5). The materialization seam *survives* the shipping
+  cut because shipping must still load `MTL::Library` from cached
+  bytes; nothing in materialization references slangc.
+
+The aggregate's SRP boundary is sharp: if the four-op trait surface
+changes, this design changes; if the Slang+Metal impl's spawn /
+ingest / pair / materialize chain changes, this design changes;
+anything else is out of scope.
+
+---
+
+## 2. Requirements Coverage
+
+Mapping of harmonius requirements to MVP refusal-or-coverage. Every
+entry is independently re-derived (PHILOSOPHY §"How harmonius is
+used"); harmonius is research input only. The cited harmonius file is
+`docs/requirements/rendering/gpu-abstraction-layer.md` (R-2.1.x), with
+cross-walks into `docs/design/rendering/render-pipeline.md` and
+`docs/design/rendering/pipeline-state-cache.md`.
+
+| Harmonius clause                                                                                   | Glibre disposition (MVP)                      | Coverage site                                                                                                                                  |
+|----------------------------------------------------------------------------------------------------|-----------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------|
+| `R-2.1.1` — top-level GPU device / command-buffer / PSO trait with associated types                | **Refused (lives in `render`)**               | The shader-backend is the **shader-source-language** trait, not the GPU-device trait. `MetalDevice` / `MetalQueue` / `MetalCommandBuffer` belong to `render` (`specs/render/SPEC.md` §4.1.6). |
+| `R-2.1.4` — Metal backend implemented via objc2-metal (Rust prior art)                             | **Collapsed to metal-cpp; Slang seam owns library creation** | §3.5 below. Metal API access is **metal-cpp** (`MTL::Device::newLibrary` / `MTL::Library::newFunction`). No Obj-C++ in engine code (`CLAUDE.md`, PHILOSOPHY §"How harmonius is used"). The seam from cached bytecode to `MTL::Library` / `MTL::Function` lives in `shader/backend/slang/metal_library_loader.cpp`. |
+| `R-2.1.16` — descriptor bindings in four frequency groups, mapped to backend-native mechanisms     | **Refused at this aggregate; covered by §4.5**| `DescriptorLayout` (SPEC §4.5) owns the four-frequency partition. The backend embeds a `DescriptorLayout` in every `ShaderArtifact` it returns; the engine-native mapping (Metal argument-buffer schema) is built by `render` from the embedded `RootSignatureSchema`. |
+| `R-2.1.17` — compile HLSL → DXIL / SPIR-V / metallib via dxc + metal-shaderconverter subprocesses; **no runtime shader compilation** | **Re-derived: Slang → metallib via slangc; no runtime compile in shipping** | §3.2 and §3.6 below. One subprocess hop (`glibre-shadercc` → `slangc`); shipping link excludes the subprocess driver entirely (SPEC §4.3 inv 3, §4.8 inv 3, §6.5). |
+| `R-2.1.18` — structured error types at every public boundary                                       | **Covered**                                   | §4.1 surface returns `std::expected<T, glibre::shader::Error>`; `reviews/decisions/error-model.md`; SPEC §10.2 enumerator-by-enumerator pinning. |
+| `R-2.1.11` — feature emulation layer (mesh-shader / RT fallbacks)                                  | **Refused for MVP**                           | §3.3 below. `Capabilities` reports static caps; the cooker filters unsupported permutations against caps. No emulation path in the backend; the M1 baseline supports mesh shaders + RT natively (PHILOSOPHY "macOS 26 / Apple Silicon baseline"). |
+| `R-2.1.10` — GPU work graphs (D3D12 native, emulated on Metal/Vulkan)                              | **Refused for MVP, capability flag reserved** | §3.3 below. `Capabilities::work_graphs == false` for MVP Slang+Metal backend. Post-MVP work-graph support arrives as a capability flip + slangc target flag, not a new trait.                                                                                  |
+| `design/rendering/render-pipeline.md` § "Shader Compilation Pipeline" — single Slang front-end, no transpile chain | **Covered**                                   | §3.1 below. One `IShaderBackend` impl; one `glibre-shadercc` driver hop; `metallib` (MVP) and `DXIL` (post-MVP) emitted directly. |
+| `design/rendering/pipeline-state-cache.md` R-2.3.9.2 — PSO key composes shader hash with device fingerprint | **Refused at this aggregate (lives in `render`)** | The `(shader_hash, state_hash)` PSO key is render's (`specs/render/SPEC.md` §4.1.7). The backend produces the *bytes* and the *handle*; PSO assembly composes them with render-state.                                                                          |
+| `design/rendering/render-pipeline.md` § "Descriptor Layout Inference" — descriptor layout inferred once and cached | **Covered**                                   | The backend embeds a derived `DescriptorLayout` in every `ShaderArtifact`. Re-derivation at runtime is forbidden (SPEC §4.5 inv 1, §4.8 inv 4). |
+| `design/rendering/render-pipeline.md` § RF-9 — re-run reflection on new bytecode after hot-reload  | **Covered (compile-side); reflection-side delegated** | §8 below. After a Slang source change the backend's `compile` re-runs the driver and pairs new bytecode with new reflection in one subprocess hop (SPEC §4.4 inv 1); the actual ingestion call lives in `reflection/` per §4.4. |
+| `requirements/rendering/gpu-abstraction-layer.md` R-2.1.3 — PSO validated at creation              | **Refused at this aggregate (lives in `render`)** | PSO creation lives in `render`'s `PSOCache::insert` path, not the shader-backend. The backend's contribution is "produce a valid `MTL::Library` + `MTL::Function` set"; `render` validates the PSO-state composition.                                          |
+| `design/rendering/pipeline-state-cache.md` R-2.3.9.8 — descriptor layout inferred from slangc reflection once | **Covered (offline, in cache)**               | `DescriptorLayout::derive(reflection)` runs once per `(PermutationKey, target)` inside `compile`; the result is embedded in the `ShaderArtifactRecord` (SPEC §7.2). Runtime is a table lookup.                                                                  |
+
+Glibre-native requirements added beyond harmonius:
+
+- **Materialization seam.** Harmonius treated "compiled bytecode →
+  `MTL::Library`" as render's job; glibre re-derives that the seam
+  belongs in `shader/backend/` because the `EntryPoint` names
+  (the keys for `MTL::Library::newFunction`) are produced by the
+  same Slang front-end that emitted the bytecode. Coupling the two
+  in `render` would replicate `EntryPoint` knowledge across context
+  boundaries; SRP forbids this. Re-derived in §3.5.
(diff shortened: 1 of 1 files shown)


specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+    // Construct a Metal MTL::Library from a sealed ShaderArtifact.
+    // Refuses if artifact.target != CompileTarget::MetalLib.
+    // Refuses with MetalLibraryCreateFailed if metal-cpp's
+    // newLibrary returns null / NS::Error*.
+    [[nodiscard]] static std::expected<MetalLibraryHandle, glibre::shader::Error>
+    load_library(MTL::Device* device, const ShaderArtifact& artifact) noexcept;
+
+    // Resolve one entry point -> MTL::Function. Reflection's
+    // EntryPoint manifest is the source of truth for legitimate
+    // names; lookup of a name not in the manifest is a programmer
```
cjhowe-us, line 481: severity:HIGH location:specs/shader/shader-backend-design.md:481 problem:MetalLibraryHandle stores a raw borrowed `const ReflectionBlob*` pointer taken from the `ShaderArtifact` passed to `load_library`, but §4.2 borrow rules explicitly state "the caller may release the artifact after the call." If the artifact is released while the MetalLibraryHandle is live and `lookup_function` is subsequently called (which accesses `handle.reflection()`), the pointer is dangling — undefined behaviour with no lifetime enforcement at the call site or in the type system. The risk is not theoretical: PSOCache holds the `MetalLibraryHandle` while the ShaderArtifact record in the cache may be evicted or migrated. fix:Transfer ownership of the `ReflectionBlob` into `MetalLibraryHandle` rather than borrowing a pointer. Change the private field to `ReflectionBlob reflection_` (value, not pointer), take it by value or move inside `MetalLibraryLoader::load_library`, and update `MetalLibraryHandle(MTL::Library*, ReflectionBlob&&) noexcept` accordingly. If the borrow-semantics are intentional for perf, document an explicit invariant: "the ShaderArtifact must outlive every MetalLibraryHandle produced from it," enforce it with a GLIBRE_ASSERT in `~MetalLibraryHandle` (debug builds), and update §4.2 borrow rules to remove the contradictory "caller may release the artifact after the call" sentence.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+
+    // shipping: compile_fn = nullptr (link-time absence)
+    std::expected<glibre::shader::ShaderArtifact, glibre::shader::Error>
+        (*compile_fn)(
+            void* impl,
+            const glibre::shader::ShaderSource&,
+            const glibre::shader::PermutationKey&,
+            glibre::shader::CompileTarget) noexcept;
+
+    std::expected<glibre::shader::ReflectionBlob, glibre::shader::Error>
```
cjhowe-us, line 978: severity:HIGH location:specs/shader/shader-backend-design.md:978 problem:The `BackendVTable` and `MetalLibraryLoaderVTable` structs in §7.4 expose `std::expected<glibre::shader::ShaderArtifact, ...>` and `eastl::span<const glibre::shader::ShaderArtifact>` directly as function-pointer return/argument types across the plugin ABI boundary. `ShaderArtifact` contains `eastl::vector<std::byte> bytecode`, `ReflectionBlob` (itself containing multiple `eastl::vector` and `eastl::string` members), and `DescriptorLayout`. `plugin-abi.md` §"Plugin file shape" rule 2 states: "Public plugin ABI surfaces never expose `std::` containers or `eastl::` containers — they cross the boundary as POD spans / handles only." PHILOSOPHY §11 repeats this. Returning a non-trivially-destructible aggregate containing EASTL vectors across a function-pointer table registered in `glibre-types.dylib` and called from the render plugin violates this contract. An ABI mismatch during plugin hot-reload could corrupt the heap without a hash-check safety net, because the hash gates dylib load, not individual function-pointer call conventions. fix:Transport ShaderArtifact across the vtable boundary as an opaque cache key (`ShaderHash`, 32 bytes — POD) plus an integer handle into the ShaderCache. The callee (shader plugin) looks up the artifact in its own CAS by hash; the caller (cooker or render plugin) passes and receives only the hash. Alternatively, introduce a `ShaderArtifactView { const std::byte* bytecode; std::size_t bytecode_size; const std::byte* reflection; std::size_t reflection_size; }` flat POD struct that spans into the CAS-resident bytes without owning them. Either approach satisfies plugin-abi.md. The design needs a §4.3 revision to describe this seam correctly.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+Failure modes:
+
+| Error                                  | Trigger                                                                                                |
+|----------------------------------------|--------------------------------------------------------------------------------------------------------|
+| `Error::CapabilityNotSupported`        | Backend named `name` not registered (defense-in-depth: no other arm fits an unknown source-language family). |
+
+Lifetime: the backend instance is constructed in
+`glibre_plugin_register` and destroyed in
+`glibre_plugin_unregister`. The plugin-ABI decision
+(`reviews/decisions/plugin-abi.md`) makes register / unregister the
```
cjhowe-us, line 617: severity:MED location:specs/shader/shader-backend-design.md:617 problem:`get_backend(name)` is specified to return `Error::CapabilityNotSupported` when the named backend is not registered. `CapabilityNotSupported` means "backend exists but the requested capability is not in its Capabilities struct" — a semantic concept documented in §3.3 and §10.1. Using it for an unknown-name lookup conflates two distinct failure modes: a wiring error (wrong name) vs a capability gap (name correct, feature absent). Any consuming code that matches on the error to decide next action will misinterpret a name-typo as a capabilities shortfall. fix:Either (a) introduce a new `shader::Error::BackendNotFound` arm (requires a SPEC §5 amendment, which the design already proposes for MetalLibraryCreateFailed/FunctionMissing — batch it), or (b) map to `Error::UnsupportedTarget` which is closer in meaning ("this backend / target combo is not available"). Document the chosen mapping in §3.7 and in §10.1. Also add a unit test `backend.get_backend_refuses_unknown_name` to §11.1.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+// across plugin versions, exposed by the shader plugin to other
+// plugins via the type registry. The cooker reads this table at
+// register-time; render reads it once during init.
+struct ShaderBackendVTable {
+    std::expected<ShaderArtifact, Error> (*compile)(
+        IShaderBackend*, const ShaderSource&, const PermutationKey&,
+        CompileTarget) noexcept;        // shipping: nullptr
+    std::expected<ReflectionBlob, Error> (*reflect)(
+        IShaderBackend*, const ShaderArtifact&) noexcept;
+    std::expected<LinkedModule, Error> (*link)(
```
cjhowe-us, line 747: severity:MED location:specs/shader/shader-backend-design.md:747 problem:The `BackendVTable.compile_fn` slot is declared as present in the struct in both shipping and non-shipping builds (only set to `nullptr` at runtime in shipping). The `IShaderBackend` v-table correctly uses `#if !GLIBRE_SHIPPING` to remove the `compile` virtual at compile time, providing a hard link-time enforcement. The vtable struct loses this guarantee: a shipping binary that mistakenly calls `vtable.compile_fn(...)` gets a null-pointer crash instead of a link error. This is weaker than what the SPEC requires (§4.3 inv 3, §4.8 inv 3: "shipping never compiles"). fix:Wrap `compile_fn` in the `BackendVTable` struct with `#if !GLIBRE_SHIPPING` so the slot physically disappears from the struct in shipping builds — matching the SPEC intent. Because the vtable is Fory-schema-versioned (§7.4), this will require two schema versions (one for shipping, one for tooling) or a single schema with the field marked `optional since 0` and absent in shipping serialization. A simpler approach: do not include `compile_fn` in the ABI-crossing vtable at all; instead, expose it only on the in-process `IShaderBackend*` surface (which already has the `#if` guard) and require cooker to hold a direct `IShaderBackend*` pointer obtained in-process, not through the vtable. Document this in §4.3.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+| Span size | Behaviour                                                                                                                         |
+|-----------|-----------------------------------------------------------------------------------------------------------------------------------|
+| 0         | Refuse with `LinkFailed` (degenerate input)                                                                                       |
+| 1         | Trivial identity: copy the artifact's bytecode + reflection into a `LinkedModule`. No subprocess spawn.                            |
+| ≥ 2       | Full link path: re-spawn `glibre-shadercc` in `--link` mode with the N artifact bytecode buffers as inputs; produce one combined `metallib` + reflection. Refuse with `Error::SpecializationConstantMissing` when an unbound spec-constant is referenced; refuse with `Error::LinkFailed` for general slangc-link errors. |
+
+The link path is **shipping-included for size 0/1** (trivial copy)
+and **shipping-excluded for size ≥ 2** (subprocess required). MVP
+is dominated by single-module artifacts; size ≥ 2 is reserved for
+future material-shader composition where the cooker links a base
```
cjhowe-us, line 583: severity:MED location:specs/shader/shader-backend-design.md:583 problem:`link` size ≥ 2 is described in §3.6 as "shipping-excluded for size ≥ 2 (subprocess required)" enforced "by the same `#if !GLIBRE_SHIPPING` guard the `compile` virtual carries." However `link` is a single virtual that IS included in shipping (size ≤ 1 paths). The `#if !GLIBRE_SHIPPING` guard on the virtual level would exclude all of `link`, including the trivial size 0/1 path the design says must survive the shipping cut. So either (a) the shipping exclusion for size ≥ 2 is a runtime branch guarded by an assertion or `Error::ShippingCompilationAttempted`-equivalent, not the same `#if` the compile virtual uses, or (b) a separate `link_multi` virtual is needed. The current wording is self-contradictory and implementers will face this immediately. fix:Clarify §3.6: the size ≥ 2 path is not link-excluded by the `#if` preprocessor guard; instead, in shipping builds the size ≥ 2 branch returns `Error::ShippingCompilationAttempted` (or a new `Error::ShippingLinkAttempted`). Update the "Shipping linkage" column in the §3.2 table for `link` from "included" to "included (size ≤ 1); runtime-refused with ShippingCompilationAttempted for size ≥ 2". Add a test `backend.shipping_link_size_2_refused` to §11.1.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+| `UnsupportedTarget`              | `compile`, `load_library` | The slangc backend does not support the requested `CompileTarget` (e.g. `DXIL` before post-MVP), or `MetalLibraryLoader::load_library` was called with a non-`MetalLib` artifact. | Adjust the target selection in the cooker / loader call site; capability-table mismatch is a build-time wiring bug.                                          | `refuse`              |
+| `MetalLibEmitFailed`             | `compile`        | slangc emitted no metallib, or emitted one whose container shape failed driver validation.                                                            | Re-author the source; the prior metallib (if any) remains the live entry.                                                                                    | `refuse`              |
```
cjhowe-us, line 1258: severity:MED location:specs/shader/shader-backend-design.md:1258 problem:§10.1 failure modes table lists every `shader::Error` arm attributable to the backend aggregate except `ReflectionExtractionFailed`. The SPEC §5 enum (line 513: `ReflectionExtractionFailed`) exists and is a distinct arm from `MetalLibEmitFailed`. If the §4.4 ingester (spike #751) fails to parse the slangc reflection JSON (malformed JSON, missing mandatory field, schema version mismatch), the error should propagate as `ReflectionExtractionFailed` through the `compile` path (step 3 in §3.4). The current table omits this arm entirely, meaning the design has no specified recovery action for that failure and no corresponding test in §11.1. fix:Add a row for `ReflectionExtractionFailed` in the §10.1 table under operation `compile`, trigger "§4.4 reflection ingester rejected the slangc JSON payload (malformed, missing field, or schema mismatch)", severity `refuse`, operator action "Re-run the driver with `--emit-reflection` verbosity flag to inspect the raw JSON; if the schema version drifted, bump and re-cook." Add a test `backend.compile_refuses_reflection_extraction_failure` in §11.1.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+task-breakdown spike blocked-by this deliverable.
+
+---
+
+## 1. Purpose
+
+The shader-backend aggregate owns **the narrow plugin-trait seam
+through which the engine talks to one shader source-language family**,
+and the sole MVP implementation of that seam: a Slang backend that
+spawns the `glibre-shadercc` driver subprocess (offline) and that
```
cjhowe-us, line 41: severity:LOW location:specs/shader/shader-backend-design.md:41 problem:§1 opens with "The aggregate is two responsibilities held in deliberate tension by SRP" and then lists three numbered items (1. Trait surface, 2. Slang+Metal implementation, 3. Library materialization). The framing "two responsibilities" is incorrect after the MetalLibraryLoader was added as a third service. This creates a direct contradiction within the first paragraph that will confuse implementers reading the SRP justification. fix:Change "two responsibilities" to "three responsibilities" in the opening sentence of §1, or restructure to say "two primary responsibilities (trait surface + implementation) with a distinct companion service (library materialization)." Also update "The aggregate's SRP boundary is sharp" closing paragraph to name all three change triggers.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+| **`MetalLibraryCreateFailed`** *(this design's amendment)* | `MetalLibraryLoader::load_library` | `MTL::Device::newLibrary(dispatch_data, NS::Error**)` returned `null`; metal-cpp's `NS::Error*` text is captured into `ErrorContext::detail`. | Re-cook against the current metal-cpp + slangc versions; re-build the shipping binary if metal-cpp was bumped.                                                | `refuse`              |
+| **`FunctionMissing`** *(this design's amendment)* | `MetalLibraryLoader::lookup_function` | `MTL::Library::newFunction` returned `null` for the requested entry-point name; the name was in the artifact's reflection but not in the library. Indicates a slangc emission bug or a stale reflection record.    | Re-cook the artifact; the artifact's reflection and bytecode must agree on entry-point names (SPEC §4.4 inv 1, "reflection paired with bytecode").              | `refuse`              |
+
+### 10.2 Refuse semantics
+
```
cjhowe-us, line 1267: severity:LOW location:specs/shader/shader-backend-design.md:1267 problem:In the §10.1 table the `CapabilityNotSupported` row lists "Operation" as `compile`, `get_backend`. `get_backend` is not one of the four IShaderBackend operations (compile/reflect/link/capabilities) — it is a registry helper function defined in §3.7. The table header and row format imply one of the four trait ops in the "Operation" column. Mixing in a registry helper muddies the boundary between trait-surface errors and registry-surface errors. fix:Split `get_backend` into a separate row clearly labeled as "registry helper `get_backend`" rather than a trait operation, or move the `get_backend` failure mode to a new §10.6 "Registry failure modes" subsection that mirrors §3.7's failure-mode table. This clarifies that `CapabilityNotSupported` from `get_backend` is a registry error, not a trait invocation error, and prevents implementers from handling it in the wrong switch arm.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:4
  LOW:2
carryover:0

notes:The design document is thorough and well-structured: it correctly grounds every harmonius clause with an explicit cover-or-refuse disposition, faithfully re-derives the four-operation IShaderBackend trait surface against SPEC §4.7 and §5, and correctly gate-keeps the compile virtual behind `#if !GLIBRE_SHIPPING`. The MetalLibraryLoader seam is a sound SRP re-derivation — placing metallib materialization in shader/backend/ rather than render is correctly justified. SPEC §10 refusal arms are cited accurately for the nine existing arms, the two proposed new arms (MetalLibraryCreateFailed, FunctionMissing) are properly deferred to a follow-up amendment with an explicit temporary mapping, and the test plan (§11) is uncommonly detailed with 25 named unit tests plus integration + E2E + perf benchmark coverage. Conventional Commit subject `docs(shader):` is correct; Refs #757 is present; issue stays OPEN per spike closure rules.

Two HIGH findings block implementation as-is. First: `MetalLibraryHandle` stores a raw borrowed `const ReflectionBlob*` pointer but §4.2 explicitly permits the caller to release the `ShaderArtifact` after `load_library` returns — this is a latent dangling-pointer defect that will produce UB when `lookup_function` is called after artifact eviction. Fix: own a copy of the `ReflectionBlob` inside the handle, or establish and enforce an explicit lifetime invariant. Second: the `BackendVTable` and `MetalLibraryLoaderVTable` structs in §7.4 carry `std::expected<ShaderArtifact, Error>` and `eastl::span<const ShaderArtifact>` as function-pointer return/argument types across the plugin ABI boundary; `ShaderArtifact` contains `eastl::vector` members which plugin-abi.md forbids at ABI surfaces ("POD spans / handles only"). Fix: transport artifacts across the vtable as `ShaderHash` handles (32-byte POD) or flat POD view structs.

Four MED findings: (1) `get_backend` maps an unknown-name error to `CapabilityNotSupported` — wrong semantic arm, should be a new `BackendNotFound` or `UnsupportedTarget`; (2) `BackendVTable.compile_fn` survives into shipping builds as a runtime-null slot rather than being `#if`-guarded out, weakening the compile-time shipping invariant the SPEC requires; (3) §3.6's claim that the size ≥ 2 link path is excluded by "the same `#if !GLIBRE_SHIPPING` guard" as `compile` is self-contradictory — `link` is a single virtual present in shipping; the exclusion must be a runtime refusal, not a preprocessor guard; (4) `ReflectionExtractionFailed` is absent from §10.1's failure modes table and from §11.1's test list, despite being a reachable arm from the §3.4 `compile` step 3 ingestion path.

Two LOW findings: (1) §1 says "two responsibilities" but lists three; (2) §10.1 table mixes `get_backend` (a registry helper) into the "Operation" column alongside the four trait operations, muddying the boundary. Both are editorial but will confuse implementers.

All findings are within scope of a follow-up design amendment to this spike or the implementation plan that the task-breakdown spike (blocked on this deliverable) will produce. No finding requires reopening the spike itself; the implementation-plan spike is the correct resolution vehicle for the HIGH and MED items.

cjhowe-us (commented):
responder:go-impl-respond round:1 decision:ADDRESSED followup:#880

All 8 findings from review 4225905447 addressed in follow-up PR #880 (branch fix/shader-shader-backend-design-followup-r1, commit de7dacf). Summary:

**HIGH-1** (MetalLibraryHandle dangling pointer): ADDRESSED. `MetalLibraryHandle` now stores `ReflectionBlob reflection_{}` as an owned value copy. After `load_library` returns, the caller may freely release the source `ShaderArtifact`. Pseudocode, borrow-rules prose, §9.2 memory budget, §9.3 allocator integration, and §11.1 test description all updated consistently.

**HIGH-2** (EASTL/std types at plugin ABI): ADDRESSED. `BackendVTable` and `MetalLibraryLoaderVTable` rewritten to carry only POD across the `glibre-types.dylib` boundary per `plugin-abi.md`. `ShaderArtifact`/`ReflectionBlob`/`LinkedModule` (all contain `eastl::vector` members) removed from function-pointer signatures; artifacts travel as `ShaderHash` (32-byte POD array), errors as `uint16_t`, `MetalLibraryHandle` as opaque `MetalLibraryToken` (`uint64_t`). Explicit ABI-vs-in-process split note added.

**MED-1** (get_backend wrong error arm): ADDRESSED. `get_backend` unknown-name failure now maps to new `BackendNotFound` enumerator (proposed §10.5 amendment). `CapabilityNotSupported` reserved for trait operations where the backend exists but refuses a capability check. Temporary `CapabilityNotSupported` mapping documented until §10.5 amendment lands.

**MED-2** (compile_fn as runtime nullptr in shipping): ADDRESSED. `BackendVTable.compile_fn` field wrapped in `#if !GLIBRE_SHIPPING` / `#endif`. The field is absent from the shipping struct at compile time; shipping code cannot reference it at all.

**MED-3** (link size≥2 self-contradictory guard claim): ADDRESSED. §3.6 rewritten to clearly state the size≥2 path is a **runtime branch refusal** (`ShippingCompilationAttempted`) in shipping, NOT a preprocessor guard. The `link` vtable slot is always present (only `compile` disappears via `#if`). Explanation of the distinction added.

**MED-4** (ReflectionExtractionFailed absent from §10.1 and §11.1): ADDRESSED. Added to §10.1a failure-modes table with trigger (compile step 3 ingester failure), operator action, and severity. Added to §11.1 unit test plan as `backend.compile_refuses_reflection_extraction_failure`. Authority preamble updated to cite the arm.

**LOW-1** (§1 count mismatch): ADDRESSED. "two responsibilities" corrected to "three responsibilities" to match the three numbered items.

**LOW-2** (§10.1 mixes registry helper with trait ops): ADDRESSED. §10.1 split into §10.1a (trait operations + MetalLibraryLoader) and §10.1b (registry helper `get_backend`).

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+Cross-context callers (cooker in `tools`, render-side loader shim in
+`render`) reach the trait through an in-process function pointer
+exposed by the `shader` plugin's registration callback:
+
+```cpp
+// glibre-types middleman: a POD function-pointer table, byte-stable
+// across plugin versions, exposed by the shader plugin to other
+// plugins via the type registry. The cooker reads this table at
+// register-time; render reads it once during init.
+struct ShaderBackendVTable {
```
cjhowe-us, line 741: severity:MED location:specs/shader/shader-backend-design.md:741 problem:§4.3 defines `ShaderBackendVTable` with non-POD function-pointer signatures (`std::expected<ShaderArtifact, Error>` return, `eastl::span<const ShaderArtifact>` argument) and then immediately asserts "the table crosses through `eastl::span` / POD value-types only" — a direct self-contradiction. Separately, §7.4 defines the same logical ABI table under a different name (`BackendVTable`) with more detail. Two sections now own the same aggregate-ABI surface under two different names, and §4.3 makes a false POD-compliance claim about its version. This is a SRP/cohesion violation at the document level: the public ABI struct should be defined exactly once (canonically in §7, the Persistence+ABI section) and §4.3 should either omit the struct or cite §7.4 by reference. fix:Remove the `ShaderBackendVTable` code block from §4.3 and replace with a prose reference to §7.4 ("the exact function-pointer layout is specified in §7.4"). The contradictory POD-compliance claim ("the table crosses through `eastl::span` / POD value-types only") must also be corrected or removed — it is false for the as-written struct and is already being corrected by #880 at the §7.4 level.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+```text
+IShaderBackend (trait, abstract)
+└── SlangMetalBackend (impl, single MVP instance)
+    ├── compile/         — delegates to CompilationPipeline (#749)
+    │     └── builds ShaderArtifact { bytecode, ReflectionBlob,
+    │                                 DescriptorLayout }
+    ├── reflect/         — delegates to slangc reflection ingester (#751)
+    ├── link/            — combined-module link via slangc spec-const bake
+    ├── capabilities/    — static descriptor (§3.3)
+    └── MetalLibraryLoader (separate service, narrow surface)
```
cjhowe-us, line 261: severity:LOW location:specs/shader/shader-backend-design.md:261 problem:The §3.1 aggregate composition diagram places `MetalLibraryLoader` as a leaf indented under `SlangMetalBackend` (same indentation level as `compile/`, `reflect/`, `link/`, `capabilities/`), which visually implies it is a sub-service of the Slang+Metal impl. §1 explicitly designates it as responsibility #3, distinct from responsibility #2 (the Slang+Metal implementation). The label "(separate service, narrow surface)" partially mitigates this, but the tree indentation will mislead an implementer into nesting `MetalLibraryLoader` inside `SlangMetalBackend`. fix:Promote `MetalLibraryLoader` to a sibling of `SlangMetalBackend` at the top level of the diagram, reflecting its status as an independent third responsibility. The revised diagram shape should show `IShaderBackend` and `MetalLibraryLoader` as two parallel leaves under a `shader-backend aggregate` root, with `SlangMetalBackend` as an impl of `IShaderBackend`.

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
...
+          ├── load_library(MTL::Device*, ShaderArtifact) → MetalLibraryHandle
+          └── lookup_function(MetalLibraryHandle, EntryPoint) → MTL::Function*
+```
+
+`IShaderBackend` is a v-table abstract base with four pure virtuals
+(SPEC §5 lines 832–849). Concrete instances are constructed by the
+plugin's `glibre_plugin_register` entry point and registered into the
+engine's backend registry by name (`"slang-metal"` in MVP). The trait
+is plugin-internal: no `extern "C"` boundary crosses the v-table — the
```
cjhowe-us, line 270: severity:LOW location:specs/shader/shader-backend-design.md:270 problem:§3.1 states "The trait is plugin-internal: no `extern \"C\"` boundary crosses the v-table" but §7.2 names the cooker and the `render` plugin as two external consumers of the `ShaderBackendVTable`. The intent is correct (the C++ vtable pointer mechanism does not cross the ABI — external callers use a separate POD function-pointer table in `glibre-types.dylib`) but the phrasing "plugin-internal" with no qualification reads as "only the shader plugin calls this" which contradicts §7.2 and §4.3. fix:Clarify the sentence: "The C++ `IShaderBackend` vtable is plugin-internal: no external caller invokes the trait directly via C++ vtable dispatch. Cross-plugin callers (cooker, `render`) reach the trait through the POD `BackendVTable` registered in `glibre-types.dylib` (§7.4)." This removes the apparent contradiction with §7.2 without changing the architectural intent.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0

notes:The design document passes the R2 SRP / cohesion / seam-quality lens well overall. The three-responsibility decomposition (trait surface, Slang+Metal impl, library-materialization seam) is architecturally sound and each responsibility is cleanly owned: `IShaderBackend` owns the four-op trait; `SlangMetalBackend` owns offline compilation delegation; `MetalLibraryLoader` owns the bytecode-to-GPU-object materialization. Refusals at every seam are well-articulated, each sibling aggregate is cited by reference rather than re-defined, and the cross-context shader/render seam rationale (SRP justification for placing `MetalLibraryLoader` in `shader/backend/`) is well-argued. All 8 R1 findings are confirmed routed to PR #880 (open follow-up) — no unresolved carryover. One MED finding: §4.3 introduces a code block (`ShaderBackendVTable`) that is a second, weaker, inconsistently-named definition of the same logical ABI struct that §7.4 calls `BackendVTable`; the prose immediately following the §4.3 code block falsely claims POD compliance while the struct carries non-POD types — this is a cohesion defect (two sections own the same aggregate surface) and a self-contradiction. The §4.3 block should be replaced with a prose reference to §7.4 so the canonical definition lives exactly once. Two LOW findings: (1) the §3.1 composition diagram places `MetalLibraryLoader` as a tree-child of `SlangMetalBackend`, visually contradicting §1's framing of it as a distinct third responsibility; (2) §3.1 describes the C++ vtable as "plugin-internal" with "no extern C boundary crossing it" while §7.2 correctly names two cross-plugin consumers — the intent is correct (callers use the POD function-pointer table, not the C++ vtable mechanism) but as written it reads as a contradiction. Both LOWs are editorial. The MED and two LOWs are suitable for absorption into PR #880 alongside the R1 fixes, as they are all within the design-doc amendment scope.

specs/shader/shader-backend-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#880 commit:ab43df2

§4.3's  struct removed; replaced with prose directing readers to §7.4 as the single canonical definition. The false POD claim and name mismatch are eliminated — the code block no longer exists in §4.3 at all.

specs/shader/shader-backend-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#880 commit:ab43df2

MetalLibraryLoader promoted to a parallel sibling in the §3.1 diagram, no longer a child of SlangMetalBackend. The diagram now reads it as a top-level entry with its own heading line referencing §1 item 3 (third responsibility).

specs/shader/shader-backend-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#880 commit:ab43df2

§3.1 vtable note qualified: now reads 'no external caller invokes the trait via C++ vtable dispatch' and explicitly names cross-plugin callers (cooker, render-side loader shim) and directs to the POD BackendVTable in glibre-types.dylib (§7.4).

specs/shader/shader-backend-design.md
```diff
@@ -0,0 +1,1520 @@
+# Shader-Backend Detailed Design
```
cjhowe-us, line 1: severity:LOW location:specs/shader/shader-backend-design.md:1331 problem:§10.5 anchors the SPEC §5 amendment coordination on "sibling spike #79 ArtifactSizeExceeded" but issue #79 ([SPIKE] design-shader-failure-modes) is CLOSED. Using a closed spike as a coordination anchor leaves implementers with no live tracking point for amendment timing. fix:Replace the co-anchor on spike #79 with a direct reference to the open tracking issue: sub-epic #69 ([SUB-EPIC] Quality — shader) which is already named in the same paragraph. Change the sentence to read "they must be added in the follow-up plan tracked under sub-epic #69 (the SPEC §5 amendment sub-epic)" without referencing the closed spike #79.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0

notes:Round-3 topology + spec-alignment + polish lens. The DAG story is coherent: PR #845 (merged) produced the design doc; PR #880 (open) carries all eight R1 findings and three R2 findings as a single follow-up commit; the routing from go-impl-respond is clean with no unresolved R1/R2 carryover entering R3. Issue #757 labels (type:spike, phase:mvp, domain:shader) are correct; no pts:* label is present — correct per AGENTS.md (spikes carry no estimate). Issue state is OPEN — correct per spike closure rules (design committed, issue stays open, sibling task-breakdown remains blocked-by this deliverable). Parent sub-epic #744 ([SUB-EPIC] Detailed Designs — shader) is OPEN and correctly cited in the doc's preamble. The five open questions in §12 are appropriately deferred: MTLBinaryArchive integration, multi-impl dispatcher, link argv schema, per-device Capabilities, and render cross-include decision all belong in §12 rather than widening this spike's scope. Conventional Commit subject docs(shader): correct type, correct scope. All decision records cited (plugin-abi.md, fory-codegen.md, perf-budget.md, error-model.md) exist under reviews/decisions/. One LOW finding: §10.5 co-anchors the SPEC §5 amendment timing on spike #79 ([SPIKE] design-shader-failure-modes) which is CLOSED; that co-anchor is a stale dead link. The live coordination point is sub-epic #69 ([SUB-EPIC] Quality — shader), which the same paragraph already names — the closed-spike sentence is redundant and misleading. Suitable for absorption into PR #880 alongside the R1/R2 fixes. No HIGH or MED findings; the R3 retro trail closes cleanly pending PR #880 merge.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0

notes:Round-3 topology + spec-alignment + polish lens. The DAG story is coherent: PR #845 (merged) produced the design doc; PR #880 (open) carries all eight R1 findings and three R2 findings as a single follow-up commit; the routing from go-impl-respond is clean with no unresolved R1/R2 carryover entering R3. Issue #757 labels (type:spike, phase:mvp, domain:shader) are correct; no pts:* label is present — correct per AGENTS.md (spikes carry no estimate). Issue state is OPEN — correct per spike closure rules (design committed, issue stays open, sibling task-breakdown remains blocked-by this deliverable). Parent sub-epic #744 ([SUB-EPIC] Detailed Designs — shader) is OPEN and correctly cited in the doc preamble. The five open questions in §12 are appropriately deferred: MTLBinaryArchive integration, multi-impl dispatcher, link argv schema, per-device Capabilities, and render cross-include decision all belong in §12 rather than widening this spike scope. Conventional Commit subject docs(shader): correct type, correct scope. All decision records cited (plugin-abi.md, fory-codegen.md, perf-budget.md, error-model.md) exist under reviews/decisions/. One LOW finding: §10.5 co-anchors the SPEC §5 amendment timing on spike #79 ([SPIKE] design-shader-failure-modes) which is CLOSED; that co-anchor is a stale dead link. The live coordination point is sub-epic #69 ([SUB-EPIC] Quality — shader), which the same paragraph already names — the closed-spike sentence is redundant and misleading. Suitable for absorption into PR #880 alongside the R1/R2 fixes. No HIGH or MED findings; the R3 retro trail closes cleanly pending PR #880 merge.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0

notes:Round-3 topology + spec-alignment + polish lens. The DAG story is coherent: PR #845 (merged) produced the design doc; PR #880 (open) carries all eight R1 findings and three R2 findings as a single follow-up commit; routing is clean with no unresolved R1/R2 carryover entering R3. Issue #757 labels (type:spike, phase:mvp, domain:shader) are correct; no pts:* label is present — correct per AGENTS.md (spikes carry no estimate). Issue state is OPEN — correct per spike closure rules (design committed, issue stays open, sibling task-breakdown blocked-by this deliverable). Parent sub-epic #744 is OPEN and correctly cited. The five open questions in §12 are appropriately deferred. All decision records cited (plugin-abi.md, fory-codegen.md, perf-budget.md, error-model.md) exist. One LOW finding: §10.5 co-anchors the SPEC §5 amendment timing on spike #79 ([SPIKE] design-shader-failure-modes) which is CLOSED — a stale dead link. The live coordination point is sub-epic #69 ([SUB-EPIC] Quality — shader), already named in the same paragraph. Suitable for absorption into PR #880. No HIGH or MED findings; R3 retro trail closes cleanly pending PR #880 merge.

specs/shader/shader-backend-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED followup:#880 commit:87d1cd7

Removed both stale co-anchors on closed spike #79 in §10.5 and §12. Both sentences now reference only sub-epic #69 as the live coordination point. Commit 87d1cd7 on branch fix/shader-shader-backend-design-followup-r1.

Issue #774, opened 2026-05-03
[SPIKE] design-render-render-resources-detailed


cjhowe-us:
Parent: #759

## Question

Produce the detailed design for the **render-resources** aggregate in the `render` context.

Transient/persistent/imported.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/render/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **render-resources** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/render/render-resources-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(render): detailed design — render-resources (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-render-render-resources-detailed` is blocked by this issue.

cjhowe-us:
agent:design-render-render-resources
status:done
issue:#774
branch:docs/render-render-resources-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-af356ad172316608a
host:mac
cloud:none
commit:2148745f0b435c075075b32ecd1649b4597c3495
pr:#852
[...]

Issue #768, opened 2026-05-03
[SPIKE] design-render-rt-accel-detailed


cjhowe-us:
Parent: #759

## Question

Produce the detailed design for the **rt-accel** aggregate in the `render` context.

BLAS / TLAS lifecycle.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/render/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **rt-accel** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/render/rt-accel-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(render): detailed design — rt-accel (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-render-rt-accel-detailed` is blocked by this issue.

cjhowe-us:
agent:design-render-rt-accel
status:done
issue:#768
branch:docs/render-rt-accel-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-ad81b64cbea72392c
host:mac
cloud:none
commit:80d4b96111642d8f19753a920eef7d80f398e86d
pr:#850
[...]

Issue #766, opened 2026-05-03
[SPIKE] design-render-pso-cache-detailed


cjhowe-us:
Parent: #759

## Question

Produce the detailed design for the **pso-cache** aggregate in the `render` context.

Pipeline-state cache.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/render/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/pipeline-state-cache.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **pso-cache** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/render/pso-cache-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(render): detailed design — pso-cache (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-render-pso-cache-detailed` is blocked by this issue.

cjhowe-us:
agent:design-render-pso-cache
status:started
issue:#766
branch:docs/render-pso-cache-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a06fccddad66991e2
host:mac
cloud:none
commit:none
pr:none
[...]

Issue #764, opened 2026-05-03
[SPIKE] design-render-render-passes-detailed


cjhowe-us:
Parent: #759

## Question

Produce the detailed design for the **render-passes** aggregate in the `render` context.

Mesh-shader gbuffer / lighting / rt / post.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/render/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/design/rendering/

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **render-passes** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/render/render-passes-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(render): detailed design — render-passes (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-render-render-passes-detailed` is blocked by this issue.

cjhowe-us:
agent:design-render-render-passes
status:done
issue:#764
branch:docs/render-render-passes-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/agent-a16c3f5b48b0cd0cb
host:mac
cloud:none
commit:1ec881e9735377752fcd89daa3645edc4a68bfc1
pr:#848
[...]

PR #852, opened 2026-05-04
docs(render): detailed design — render-resources (refs #774)

cjhowe-us:
## Summary

Resolves spike #774 — `[SPIKE] design-render-render-resources-detailed`.

Adds `specs/render/render-resources-design.md`: 12-section detailed design for the render-resources aggregate (typed phantom-tagged generational handles, transient pool with alias-driven placement, persistent best-fit allocator, ring buffer manager, four-frequency argument-buffer binder, imported-borrow registry, sampler cache, hot-reload survival, failure-mode mapping, full test plan).

## Scope

- Refines `specs/render/SPEC.md` §4.1.4, §5, §6.1, §6.3, §7, §8.2, §9.5, §9.5.1, §9.6, §10 in place.
- Cites `reviews/decisions/{error-model,perf-budget,plugin-abi,hot-reload-protocol,fory-codegen,frame-phases}.md`.
- Introduces no new public surface beyond SPEC §5.
- SRP boundary explicit: aggregate owns typed-handle catalog + heap-pool allocation + lifetime release + alias-driven placement; refuses graph topology (#760), pass bodies (#764), PSO cache (#766), RT/HZB/cluster-cull state (#768/#769/#771), asset content.

## Test plan

- [ ] Unit tests — handle/slot table, transient pool, persistent allocator, ring buffer manager, argument-buffer binder (§11.1–§11.5)
- [ ] Integration tests — real `MTLDevice` end-to-end on macOS-26/M1 fixture (§11.6)
- [ ] Hot-reload tests — survival inventory, schema-mismatch refusal (§11.7)
- [ ] Perf benchmarks — five contractual gate lines (§11.8) feeding `SPEC.md` §9.6 ceilings
- [ ] User-story coverage — feeds #382 (alias planner), #402 (residency-exceeded recovery), #400 (cost-aware budget cull)

### DIFF
--- a/specs/render/render-resources-design.md
+++ b/specs/render/render-resources-design.md
@@ -1,0 +1,1420 @@
+# render — Detailed Design: render-resources aggregate
+
+> Detailed design for the **render-resources** aggregate (typed handles
+> + lifetime management for textures, buffers, render targets, samplers,
+> and argument buffers) declared in `specs/render/SPEC.md` §4.1.4
+> (`Resource` — virtual + physical; transient / persistent / imported)
+> and consumed by §4.1.2 (`RenderGraph`), §4.1.5 (`ExecutionPlan`),
+> §4.1.6 (`MetalDevice` heap allocator), §4.1.7 (`PSOCache` argument-
+> buffer binding), §4.1.8 (`RTAccelStructures` BLAS / TLAS storage),
+> §4.1.9 (`HZB`) and §4.1.10 (`ClusterCullState`). Refines §4.1.4,
+> §5 (`ResourceDesc` / `ResourceUsage` / `ResourceFormat` /
+> `ResourceLifetime` / `Handle<Tag>` / `GraphBuilder::declare_*`),
+> §6.1 `resources/` directory, §6.3 concurrency, §7.1 / §7.3 (nothing
+> persists at this layer), §8.2 hot-reload survival rows for persistent
+> resources / `TransientPool` heaps, §9.5 heap composition (256 MiB
+> transient + 128 MiB persistent + 16 MiB handle tables + 48 MiB RT
+> + 64 MiB PSO = 512 MiB ceiling), §9.5.1 allocator rules, and §10
+> failure-mode rows `ResourceAllocFailed` / `ResourceResidencyExceeded`.
+> Cites `reviews/decisions/error-model.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`,
+> `reviews/decisions/fory-codegen.md`, and
+> `reviews/decisions/frame-phases.md`. Introduces no new public surface
+> beyond `specs/render/SPEC.md` §5; deviation from the cited records
+> requires an amendment spike, not an in-place edit. Resolves
+> `[SPIKE] design-render-render-resources-detailed` (#774).
+
+## 1. Purpose
+
+The render-resources aggregate is render's **typed-handle catalog +
+lifetime allocator** for every GPU-side and host-side
+allocation that lives inside the render plugin. Per
+`specs/render/SPEC.md` §4.1.4 it owns:
+
+1. The **`VirtualResource`** value object — a logical
+   `(format, extent, mip_levels, array_layers, sample_count, usage,
+   lifetime, debug_name, frame-of-birth, frame-of-last-use)` record
+   referenced by `VirtualResourceHandle`. Issued by
+   `GraphBuilder::declare_transient` / `declare_persistent` /
+   `declare_imported`; consumed by passes' read/write declaration sets.
+2. The **`PhysicalAllocation`** entity — the heap slot or
+   externally-vended `MTL::Texture` / `MTL::Buffer` / sampler /
+   argument-buffer that materialises a `VirtualResource`. Referenced
+   by `PhysicalAllocHandle`. Owns no algorithmic policy; is purely a
+   storage record + GPU-handle holder.
+3. The **`TransientPool`** — render's per-frame placement heap pool.
+   `MTL::Heap` instances pre-sized at init (`SPEC.md` §9.5 transient
+   row = 256 MiB) into which the alias planner places lifetime-disjoint
+   `VirtualResource`s as `MTL::Texture` / `MTL::Buffer` placements.
+   The pool drains at the phase 7→9 boundary every frame.
+4. The **persistent allocator** — the long-lived slab of `MTL::Heap`s
+   carrying resources whose lifetime crosses frames: `HZB` pyramids,
+   shadow atlases, history-color targets, ring-buffer storage,
+   lighting LUTs, IBL probes, font / overlay atlases, RT scratch.
+   Sized at init from the §9.5 persistent row (128 MiB).
+5. The **imported-resource borrow registry** — opaque records that
+   wrap externally-owned allocations (the swapchain `MTLDrawable`,
+   `geometry`-vended BLAS, `vfx`-vended particle buffers,
+   `shader`-vended PSO blobs) without taking ownership. Lifetime
+   stays with the importer (`SPEC.md` §4.1.4 invariant 3).
+6. The **typed-handle tables** — five generation-counted
+   `(index, generation)` tables, one per public handle kind:
+   `VirtualResourceHandle`, `PhysicalAllocHandle`,
+   `ArgumentBufferHandle`, `RingSliceHandle`, plus the read-only
+   import handles (`BLASHandle`, `HZBHandle`,
+   `ClusterCullStateHandle`, `ShadowAtlasHandle`) whose tables are
+   owned by their respective aggregates but whose entries route
+   through this layer's catalog for lookup. Tables sit inside the
+   §9.5 "GPU resource handles" 16 MiB row.
+7. The **ring-buffer surface** — per-frame-in-flight CPU-write /
+   GPU-read rings (constant heaps, instance buffers, indirect-arg
+   buffers). Slices are addressed by `RingSliceHandle`; the slab
+   itself is part of the persistent allocator.
+8. The **argument-buffer frequency-group binder** — the four-level
+   binding-table builder (per-frame / per-pass / per-material /
+   per-draw, `SPEC.md` §3.2 collapse #8) that resolves
+   `(VirtualResourceHandle, PhysicalAllocHandle)` pairs into
+   `MTL::ArgumentEncoder` offsets consumed by passes' execute
+   lambdas via the opaque `Bindings` struct. The binder is hot-path
+   and lives behind a stable per-pass cache.
+
+This aggregate **refuses to own**:
+
+- **Graph topology, declared-edge set, alias-planner colouring
+  algorithm, barrier emission, queue assignment, structural-hash
+  cache.** Owned by the `RenderGraph` / `ExecutionPlan` aggregate
+  (`SPEC.md` §4.1.2, §4.1.5; spike #760 detailed design). Resources
+  feed the planner (lifetime rows + footprint estimates) and consume
+  its output (`AliasPlan` mapping each `VirtualResource` to a
+  `PhysicalAllocation` slot); this layer never decides
+  *which colour* a virtual resource gets.
+- **Pass bodies, declared access sets, `execute()` lambda
+  semantics.** Owned by the pass aggregate (`SPEC.md` §4.1.3; spike
+  #764). Passes call `GraphBuilder::declare_*` and reference the
+  returned handles; resources never look inside an execute lambda.
+- **PSO compilation, `MTL::RenderPipelineState` /
+  `MTL::ComputePipelineState` residency, eviction.** Owned by
+  `PSOCache` (`SPEC.md` §4.1.7; spike #766). Argument buffers built
+  here are consumed by PSOs but the cache itself is sibling: the
+  cache stores compiled pipeline state; the resources aggregate
+  stores the data that flows *through* those pipelines.
+- **`MetalDevice` / `MetalQueue` / `MetalCommandBuffer` lifecycles,
+  surface attach, residency-set construction, fence emission.**
+  Owned by the metal-backend aggregate (`SPEC.md` §4.1.6; spike
+  #762). The resources aggregate calls `MetalDevice::heap_allocator()`
+  and `MetalDevice::residency()` to acquire and attach `MTL::Heap`s
+  but does not construct the device, the queues, or the residency
+  set. Heap *attachment* to the residency set is a backend call;
+  heap *allocation* is this layer's responsibility.
+- **BLAS / TLAS construction or storage layout.** Owned by
+  `RTAccelStructures` (`SPEC.md` §4.1.8; spike #768). BLAS handles
+  are imports (read-only by `SPEC.md` §4.1.8 invariant 3); TLAS
+  storage is a persistent `Resource` whose backing slab lives in
+  this layer's persistent allocator but whose contents are managed
+  by the RT aggregate.
+- **HZB / `ClusterCullState` storage policy.** Owned by `HZB`
+  (#769) and `ClusterCullState` (#771) respectively. Their
+  *backing GPU bytes* sit in this aggregate's persistent allocator
+  (per `SPEC.md` §8.2 row "HZB pyramid, ClusterCullState scratch");
+  their *dispatch policy and triple-buffer cadence* live in their
+  own design specs.
+- **Cull, extract, sort, budget cull, RenderFrame finalisation.**
+  Owned by the cull-extract aggregate inside phase 6 (`SPEC.md`
+  §6.2.1; spike #770). Phase 6 reads no resources from this
+  layer except indirectly through the HZB pyramid pointer (#769).
+- **Asset content, vertex/index/meshlet bytes, texture pixels,
+  Draco decode, FBX import.** Owned by `geometry` and `content`
+  (`SPEC.md` §3.3). Resources receives already-resident GPU bytes
+  via `declare_imported` and never reaches into the asset layer.
+- **`RenderSettings`, `CapabilityMask`, `PSOCacheRecord` Fory
+  schemas.** Owned by `SPEC.md` §7.1.1 / §7.1.3 / §7.1.2
+  respectively. Resources is the aggregate that *consumes* the
+  capability set when sizing the persistent slab (HZB extent,
+  shadow-atlas tier) but persists nothing.
+- **Frame schedule.** Phase 6 / phase 7 entry / exit are owned by
+  `core` (`reviews/decisions/frame-phases.md`); resources is
+  invoked inside those phases and obeys their entry / exit
+  guarantees.
+- **Resource serialisation.** Disk persistence of any
+  `VirtualResource`, `PhysicalAllocation`, alias plan, or argument
+  buffer is forbidden. This layer's only persistent artefact would
+  be a heap-residency hint, which is not part of MVP. PHILOSOPHY
+  anti-pattern reaffirmed.
+
+The aggregate's SRP boundary is sharp: if **the way a render
+resource is named, allocated, looked up, freed, or aliased** changes,
+this design changes. Anything else is out of scope.
+
+## 2. Requirements coverage
+
+This section verifies that every harmonius MVP-scope requirement /
+design clause about render resources is either covered by the design
+below or explicitly refused with rationale. Inputs (research only,
+re-derived per PHILOSOPHY §"How harmonius is used"):
+
+- `harmonius/docs/requirements/rendering/resource-management.md`
+  (R-2.5.* clauses on typed handles, transient pools, persistent
+  resources, imported wrappers, generational invalidation).
+- `harmonius/docs/design/rendering/render-pipeline.md`
+  § "Resources", § "Heap & Argument Buffers".
+- `harmonius/docs/design/rendering/rendering-core.md`
+  § "Allocation", § "Handle Tables".
+- `harmonius/docs/design/rendering/argument-buffer-binding.md`
+  (frequency-group binding, sampler tables).
+
+| Harmonius clause                                                                                  | Glibre disposition                                                                                                                                                                                                                                                                                                       |
+|---------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+> buffer binding), §4.1.8 (`RTAccelStructures` BLAS / TLAS storage),
+> §4.1.9 (`HZB`) and §4.1.10 (`ClusterCullState`). Refines §4.1.4,
+> §5 (`ResourceDesc` / `ResourceUsage` / `ResourceFormat` /
+> `ResourceLifetime` / `Handle<Tag>` / `GraphBuilder::declare_*`),
+> §6.1 `resources/` directory, §6.3 concurrency, §7.1 / §7.3 (nothing
+> persists at this layer), §8.2 hot-reload survival rows for persistent
+> resources / `TransientPool` heaps, §9.5 heap composition (256 MiB
+> transient + 128 MiB persistent + 16 MiB handle tables + 48 MiB RT
+> + 64 MiB PSO = 512 MiB ceiling), §9.5.1 allocator rules, and §10
+> failure-mode rows `ResourceAllocFailed` / `ResourceResidencyExceeded`.
```
cjhowe-us, line 18: severity:HIGH location:specs/render/render-resources-design.md:18 problem:Preamble claims the design refines SPEC §10 failure-mode rows named `ResourceAllocFailed` / `ResourceResidencyExceeded`. `ResourceAllocFailed` and `RtCapabilityMissing` are SPEC §10 class-level pattern labels, NOT members of the `render::Error` enum defined in SPEC §5. The actual `render::Error` variants this aggregate introduces are `HeapOutOfMemory`, `TransientPoolExhausted`, `ResourceImportRefused`, and `CapabilityNotSupported` (see SPEC §5 lines 829-865 and §10.3 mapping table). Using `ResourceAllocFailed` and `RtCapabilityMissing` in the preamble, §10.2 cross-refs (lines 1245-1246), and §3.12 forward-reference commentary (line 618) creates false conflation between the §10 classification label and the `render::Error` enum identifiers — the same class of error flagged in sibling PR #853. fix:Replace every occurrence of `ResourceAllocFailed` used as an enum-identifier reference with `HeapOutOfMemory` (persistent allocator arm) or `TransientPoolExhausted` (transient pool arm); replace `RtCapabilityMissing` used as an enum-identifier reference with `CapabilityNotSupported`. Keep these names only where they appear in the `§10 row` column of §3.12 (which is explicitly cross-referencing the SPEC §10 pattern label, not the enum), and add a prose note at first use explaining the two-level distinction.

specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+was rejected because the underlying slot moved on") with a
+debug-build log line carrying the handle's `(index, generation)`,
+the slot's *live* generation, and the most recent producer call site
+(captured in debug builds via `__builtin_FILE` + `__builtin_LINE`
+threaded through the `declare_*` API; release builds skip the
+capture).
+
+Wrap-around safety:
+
+- 24 bits = 16 777 215 unique generations per slot before wrap.
```
cjhowe-us, line 345: severity:HIGH location:specs/render/render-resources-design.md:345 problem:Generation-counter wrap safety analysis is internally inconsistent. §3.3 (line 299-301) states that the upper 8 bits of the 24-bit generation half are reserved for the SamplerHandle sub-tag, leaving only 16 bits for the live wrap counter (wraps every 65,535 cycles). But §3.4's safety analysis (line 345) resets to the full 24-bit count (16,777,215) and derives a '7 hours' safety margin from it. If the sub-tag occupies the upper 8 bits, the effective generation counter is 16 bits, not 24 bits, and the wrap occurs at 65,535 cycles. With the §3.4 churn rate of ~11 frees/slot/frame, that is 65,535 / 11 ≈ 5,957 frames ≈ ~99 minutes at 60 FPS — still safe per the §4.1.4 invariant (transient handles cannot cross a frame boundary), but the documented analysis presents a number that is 256x too large. This is a correctness error in the specification: any implementer who sizes the generation field based on the '16,777,215' number without reading §3.3 will misunderstand the real effective width. fix:Either (a) reconcile: remove the sub-tag bit-packing from §3.3 and use a separate discriminant field inside the slot record instead of repurposing generation bits, keeping the 24-bit generation field intact and the safety margin at 7 hours; or (b) correct the §3.4 wrap analysis to use 16 bits (65,535 wrap) and verify the revised margin (~99 minutes) still satisfies the safety requirement, adjusting the test fixture generation_bits parameter accordingly. Option (a) is strongly preferred because it avoids aliasing the generation counter with a slot-type discriminant, which is a subtle invariant to maintain.

specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+### 3.12 Failure-mode mapping (forward reference to §10)
+
+The catalog produces five render-error variants:
+
+| Trigger | Variant | §10 row |
+|---------|---------|---------|
+| Persistent allocator exhausted (best-fit failed) | `HeapOutOfMemory` | `ResourceAllocFailed` |
+| Transient pool peak-residency > sub-pool cap | `TransientPoolExhausted` | `ResourceAllocFailed` |
+| Per-frame total residency > 512 MiB | `ResourceResidencyExceeded` | `ResourceResidencyExceeded` |
+| Stale handle / generation mismatch / role mismatch | `ResourceImportRefused` | (debug log + retry loop falls into `abort-frame` in release) |
```
cjhowe-us, line 690: severity:MED location:specs/render/render-resources-design.md:690 problem:`ResourceImportRefused` is semantically overloaded across four distinct trigger classes: (a) stale generational handle lookup, (b) role mismatch (e.g. release_persistent on a transient), (c) imported-handle write declaration on a read-borrow, (d) sampler-cache over-cap (§3.13 line 723). None of triggers (a), (b), or (d) are 'import refused' scenarios — they are stale-pointer, API-misuse, and cache-capacity conditions respectively. The error-model.md requires each context's enum to map enumerators to specific failure modes so they remain self-documenting and log-decodable. Using one enumerator for four structurally different errors violates that design goal: a log line showing ResourceImportRefused is ambiguous without reading ErrorContext::detail. fix:Split into at minimum two enumerators: (1) ResourceImportRefused — preserved for the import-borrow-specific cases (c) and (d-as-borrow); (2) StaleResourceHandle — for generation-mismatch lookups (a); consider a third ResourceRoleMismatch for (b). Sampler over-cap (§3.13) should use a dedicated arm or return HeapOutOfMemory since the sampler cache is a fixed-size resource table. Each new enumerator requires a new entry in the render::Error enum (SPEC §5) and therefore an ABI bump — document this in §12 Open Questions since it affects the SPEC §5 freeze.

specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+
+`SPEC.md` §5 publishes the closed handle catalog. The detailed
+design refines two practical consequences:
+
+1. **`SamplerHandle` is a sub-tag of `tags::argument_buffer`.** The
+   header's `tags::argument_buffer` namespace declares both the
+   per-pass argument-buffer record and the sampler-state record;
+   the sub-discrimination is a private 8-bit field inside the
+   `Handle<tag::argument_buffer>`'s reserved generation byte
+   (`SPEC.md` §5: "24-bit generation"; the upper 8 bits of the
```
cjhowe-us, line 299: severity:MED location:specs/render/render-resources-design.md:299 problem:The SamplerHandle sub-tag scheme repurposes bits 40-47 of the generation field of Handle<tags::argument_buffer> for a runtime discriminant. This is not type-safe: the public Handle<Tag> template (SPEC §5) exposes generation() as a raw uint32_t with no knowledge of sub-tags, so code that calls .generation() on an ArgumentBufferHandle will silently read a value that mixes the 16-bit live counter with the 8-bit sub-tag in the upper byte. There is no compile-time enforcement that a SamplerHandle and an ArgumentBufferHandle cannot be confused — both are Handle<tags::argument_buffer> at the type level. The SPEC §5 phantom-tag invariant ('cross-tag assignment is a compile error') is only enforced between distinct Tag types, not between values of the same tag type with different sub-tag bits. fix:Either (a) give SamplerHandle its own tags::sampler tag type (a new empty struct), making it a distinct Handle<tags::sampler> with no overlap with ArgumentBufferHandle — this is the simplest fix and re-establishes compile-time enforcement; or (b) if sub-tagging within one Tag type is essential, make the sub-tag part of the type system by adding a second template parameter (e.g. Handle<Tag, SubTag>) and specializing the generation() accessor to mask out the sub-tag bits. Option (a) requires adding tags::sampler to SPEC §5 (one-line enum value addition), which is in scope for a design doc amendment since SPEC §5 explicitly admits the resource aggregate as the owner of these handle types.

specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+};
+```
+
+The closed-set design (six × four × eight × four = limited
+combinatorics) gives a known upper bound on distinct samplers
+across the entire MVP shader set. Sixteen entries is sized for the
+union of every MVP pass's sampler use ({linear-clamp, linear-repeat,
+trilinear-anisotropic-repeat for albedo, comparison-less-clamp for
+shadow, nearest-clamp for visID resolve, …}); over-cap returns
+`render::Error::ResourceImportRefused`, which is treated as a SPEC
```
cjhowe-us, line 723: severity:LOW location:specs/render/render-resources-design.md:723 problem:Sampler-cache over-capacity returns `render::Error::ResourceImportRefused` (line 723). A sampler cache that cannot accommodate a new entry because the capacity ceiling was reached is not an 'import refused' condition — it is a resource-table exhaustion condition semantically equivalent to a slot-table capacity overflow. §11.1 already defines a separate test for slot-table overflow returning `ResourceResidencyExceeded`. Using `ResourceImportRefused` here creates a misleading error code and makes it harder for operator tooling to distinguish import-borrow rejections from capacity limits. fix:Return `render::Error::ResourceResidencyExceeded` (consistent with slot-table overflow in §11.1) or document a dedicated `SamplerCacheExhausted` variant if the per-component log differentiation matters operationally. At minimum update §10's `ResourceImportRefused` trigger list to enumerate sampler over-cap so the mapping is explicit and testable.

specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+change must construct one of these errors from a second site, the
+SRP boundary is being violated and a refactor is required.
+
+### 10.2 Cross-references
+
+- `SPEC.md` §10.1 — closed sum (eighteen variants).
+- `SPEC.md` §10.2 — recovery ladder.
+- `SPEC.md` §10.3 — per-variant rows (this design's five
+  contributions appear as `ResourceAllocFailed`,
+  `ResourceResidencyExceeded`, `RtCapabilityMissing`).
```
cjhowe-us, line 1246: severity:HIGH location:specs/render/render-resources-design.md:1246 problem:The §10.2 cross-reference section states 'this design's five contributions appear as `ResourceAllocFailed`, `ResourceResidencyExceeded`, `RtCapabilityMissing`'. This is incorrect: `ResourceAllocFailed` and `RtCapabilityMissing` are SPEC §10 class-level labels, not `render::Error` enum members. The actual five enum contributions this design introduces are `HeapOutOfMemory`, `TransientPoolExhausted`, `ResourceResidencyExceeded`, `ResourceImportRefused`, and `CapabilityNotSupported` (as enumerated correctly in §10's failure-mode table). Listing the wrong names here makes the SPEC §10.3 mapping table non-round-trippable and will cause implementers to search for non-existent enum values. fix:Replace the sentence with: 'this design's five contributions to the `render::Error` enum are `HeapOutOfMemory`, `TransientPoolExhausted`, `ResourceResidencyExceeded`, `ResourceImportRefused`, and `CapabilityNotSupported` (bindless arm); these contribute to SPEC §10.3 rows `ResourceAllocFailed` (for the first two) and `RtCapabilityMissing` (for the bindless arm).' This mirrors the two-level distinction correctly.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0

notes:The design fulfills its declared scope well. All twelve required sections are present with substantive content. The three-role resource taxonomy (transient / persistent / imported) is clean and correctly bounded against SPEC §4.1.4. The transient pool's alias-driven placement flow, the persistent best-fit allocator with eight power-of-two bins, and the ring-buffer lock-free bump vend are all coherent and correctly scoped. Concurrency ownership is explicit (driver-thread inserts, read-only workers, lock-free ring). Hot-reload survival inventory is precise and correctly maps to hot-reload-protocol.md. The test plan (§11) is comprehensive: unit, integration, hot-reload, and perf gate benchmarks are all specified. The public surface introduces no symbols beyond SPEC §5 and cites the six required decision records.

Two HIGH findings require follow-up before implementation PRs proceed. First (lines 18, 618, 1245-1246): the preamble, §3.12 forward-reference, and §10.2 cross-reference section conflate SPEC §10 class-level pattern labels (ResourceAllocFailed, RtCapabilityMissing) with render::Error enum identifiers — the same class of error flagged in sibling PR #853. The actual enum members are HeapOutOfMemory, TransientPoolExhausted, ResourceImportRefused, and CapabilityNotSupported. Second (lines 299-353): the generation-counter wrap safety analysis is internally contradictory. §3.3 reserves 8 of the 24 generation bits for the SamplerHandle sub-tag (leaving 16 live bits, wrap at 65,535 cycles, ~99-minute safety window at 11 frees/frame), but §3.4 uses the full 24-bit count (wrap at 16,777,215 cycles, ~7-hour window). The two-level MED findings are: the SamplerHandle sub-tag scheme bypasses the phantom-tag type-safety invariant (both SamplerHandle and ArgumentBufferHandle share Handle<tags::argument_buffer> with no compile-time discrimination), and ResourceImportRefused is semantically overloaded across four structurally distinct failure classes (stale handle, role mismatch, write-on-read-borrow, sampler over-cap), violating the error-model.md self-documenting enum principle. One LOW: sampler over-cap (§3.13) returns ResourceImportRefused when ResourceResidencyExceeded (consistent with slot-table overflow in §11.1) would be more accurate.

Conventional Commit subject 'docs(render): detailed design — render-resources (refs #774)' is correct: type=docs, scope=render, references the parent issue. PR body is clear and test plan is present. Issue #774 is referenced correctly; it remains open per the closure rules.

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3b7662a

HIGH-1 addressed. The preamble, §3.12 forward-ref table, and §10.2 cross-references now correctly distinguish between §10 design-name labels (ResourceAllocFailed, RtCapabilityMissing — appearing only in the §10-row column and in cross-reference prose explicitly labeled as such) and §5 enum identifiers (HeapOutOfMemory, TransientPoolExhausted, CapabilityNotSupported — used for all C++ identifier references). The §10.2 section was rewritten to enumerate the actual §5 enum contributions and show the mapping to §10 design-names explicitly. followup:#872

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:9a393e7

HIGH-2 addressed. The root cause of the contradiction was the sub-tag scheme (MED-1). By introducing tags::sampler as a distinct phantom type, the 8-bit sub-tag reservation in the generation field is eliminated entirely. §3.4 now states uniformly '24 bits live, no sub-tag reservation' for all five SlotTable instances, with wrap-safety recomputed consistently as 16 777 215 / 11 ≈ 7-hour window. The previously contradictory dual accounting (§3.3: 16 bits; §3.4: 24 bits) is gone. followup:#872

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:9a393e7

MED-1 addressed. SamplerHandle now uses Handle<tags::sampler> (a new distinct phantom type added to SPEC.md §5's tags namespace and handle alias table). This removes the sub-tag bit-packing entirely. Cross-assignment between SamplerHandle and ArgumentBufferHandle is now a compile error enforced by the phantom-tag system (two different Tag types). The §3.3 table updated; §11.1 unit-test assertion extended to cover SamplerHandle vs ArgumentBufferHandle cross-assignment. followup:#872

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3b7662a

MED-2 addressed. ResourceImportRefused is now reserved only for genuine write-on-read-borrow refusal (import-side: a graph structural error, abort-engine recovery). Split into self-documenting variants: StaleResourceHandle (generation mismatch at lookup, abort-frame recovery) and ResourceRoleMismatch (wrong release API for lifetime kind, cold-path warn). Sampler over-cap (case d in original) is addressed separately via LOW fix — returns ResourceResidencyExceeded. All three new variants added to SPEC.md §5 enum with ABI-add annotations and to §10.1 closed-sum table. Single construction site per variant preserved per error-model.md SRP rule. followup:#872

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3b7662a

LOW addressed. Sampler-cache over-cap now returns render::Error::ResourceResidencyExceeded, consistent with the §11.1 slot-table-overflow precedent. The §3.13 prose updated to cite this consistency. §10's trigger description for ResourceResidencyExceeded updated to include case (b): sampler over-cap. followup:#872

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3b7662a

HIGH-1 (§10.2 cross-ref arm) addressed. The §10.2 cross-references section was rewritten: it now enumerates the actual §5 enum contributions (HeapOutOfMemory, TransientPoolExhausted, ResourceResidencyExceeded, StaleResourceHandle, ResourceRoleMismatch, ResourceImportRefused, CapabilityNotSupported) and explicitly maps each to its §10 design-name row. The erroneous 'contributions appear as ResourceAllocFailed, RtCapabilityMissing' sentence is gone. followup:#872

specs/render/render-resources-design.md
```diff
@@ -0,0 +1,1420 @@
...
+| `geometry` | BLAS | Cook output (one BLAS per cooked mesh) | Until `geometry`'s mesh asset is unloaded |
+| `geometry` | Vertex / index / meshlet streams | Asset load | Until asset unloaded |
+| `vfx` (post-MVP) | Particle / cloth GPU buffers | Per-frame VFX submit | Until next VFX submit |
+| `shader` | PSO blobs (consumed by `PSOCache`, not by resources) | (out of scope at this layer) | (out of scope) |
+
+The borrow record fields:
+
+```
+ImportedBorrow
+├── kind_              : enum { Drawable, BLAS, GeomStream, VfxBuffer }
```
cjhowe-us, line 536: severity:MED location:specs/render/render-resources-design.md:536 problem:`ImportedBorrow.kind_` enum includes `VfxBuffer` — a post-MVP arm explicitly marked "(post-MVP)" in the borrow-registry table four lines above. PHILOSOPHY §5 (greatly reduced MVP scope) and the anti-pattern rule ("cross-domain abstractions invented before two concrete users exist") prohibit this. Including a future-plugin enum arm in a closed MVP enum invites premature branching in implementation code. fix:Remove `VfxBuffer` from the `kind_` enum in §3.8; define the closed MVP set as `{Drawable, BLAS, GeomStream}`. Add a note that the `vfx` arm is a future design amendment requiring SPEC.md §4.1.4 update and an ABI bump. Route to #872 sweep.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:1
  LOW:0
carryover:5 (all R1 findings addressed in open PR #872; none re-raised here)

notes:The R2 cohesion lens finds the design strongly coherent in its primary responsibility. One document owns one aggregate; the §1 SRP boundary is explicit and thorough; each of the twelve sections covers one concern; the public surface introduces nothing beyond SPEC §5; no invariants are invented before two concrete users exist — with one exception. Section §3.8 includes `VfxBuffer` in the closed `ImportedBorrow.kind_` enum despite the vfx plugin being explicitly labelled post-MVP in the same section's table. This is the one MED finding: a future-plugin enum arm inserted into a closed MVP enum before the vfx context exists, violating PHILOSOPHY §5 and the anti-pattern rule against cross-domain abstractions invented early. The fix is narrow (remove the arm, note it as a future ABI amendment) and routes to the #872 sweep. All five R1 findings (2 HIGH, 2 MED, 1 LOW) were correctly addressed in PR #872 (open); no carryover finds warrant re-escalation here. The argument-buffer binder colocation inside ResourceCatalog is defensible: the binder's only operation is resolving (VirtualResourceHandle → MTL::*) pairs via catalog slot lookup, making it operationally part of the catalog's lookup seam rather than a separate concern. The ring-buffer manager's inline snippet uses std::atomic, which is correct per PHILOSOPHY §11's explicit carve-out for std::atomic. The sampler-cache over-cap producing ResourceResidencyExceeded (creating a second construction site for that variant) is surfaced for the first time in #872's own diff, where it is acknowledged as an open SRP violation — not present as a new finding in this PR's merged state.

specs/render/render-resources-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED followup:#872 commit:1fb6cb7

VfxBuffer removed from ImportedBorrow.kind_. Closed MVP enum is now {Drawable, BLAS, GeomStream}. A blockquoted note documents the future ABI amendment path: VfxBuffer arm added when the vfx plugin ships, filed as SPEC.md §4.1.4 ABI-add row with a middleman-dylib hash bump.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:0
carryover:0 (all R1: 2 HIGH + 2 MED + 1 LOW addressed in PR #872; all R2: 1 MED addressed in PR #872 commit 1fb6cb7)

notes:Topology, spec alignment, and polish are all sound for the merged state of PR #852. The dependency story is coherent: PR #872 (open) carries the complete fix sweep for every finding raised in R1 and R2, with three logical commits (9a393e7, 3b7662a, 1fb6cb7) each correctly scoped to their finding clusters. The R1 inline comment thread on #852 shows all six findings replied with decision:ADDRESSED before #872 was raised. SPEC.md §5 is consistently updated in #872 with tags::sampler, SamplerHandle alias, StaleResourceHandle, and ResourceRoleMismatch — plus the §10.1 ABI-add rows — so the spec/design/impl dependency chain is clean. Issue #774 carries labels type:spike, phase:mvp, domain:render with no pts:* label; that matches AGENTS.md exactly (spikes carry no estimate). The issue correctly remains OPEN: the spike's deliverable is committed but implementation plan leaves have not yet landed. The §10.2 two-construction-site concern for ResourceResidencyExceeded (sampler-cap arm + residency-ceiling arm) was noted in R2 as a new finding that belongs to PR #872's own review trail, not retroactively to the merged PR #852 — it is not re-raised here. No open questions remain in the §12 section that cross DAG boundaries incorrectly; all deferred items are properly labelled as post-MVP or future-spike candidates. Round 3 closes the retro trail for PR #852.

PR #859, opened 2026-05-05
test(core): add e2e trace for archetype-storage (refs #322)

cjhowe-us:
## Summary

Authors `tests/e2e/core/archetype-storage.glibre-trace` for story
[#322](https://github.com/cjhowe-us/glibre/issues/322), covering the
three Gherkin blocks: archetype creation, chunk capacity, swap-remove
integrity.

- **Block 1 (archetype creation).** Spawns two `{A,B}` entities and
  one `{A,B,C}` entity; asserts `archetype_count == 2`, per-archetype
  `columns_aligned == true` and `columns_contiguous == true` for both
  archetypes (specs/core/SPEC.md §4.2 inv 1, 3).
- **Block 2 (chunk capacity).** Overrides `Chunk::CAPACITY = 4` for
  the `{A,B}` archetype via the cfg(debug_overlay)
  `core::testing::set_chunk_capacity_override` hook, then spawns
  `N+1 = 5` rows; asserts `chunk_count == 2` and
  `row_chunk_continuity == true` (§4.2 inv 2). Privileges clarity
  over stress — the 100k-entity production-capacity case is the
  Manual Test Script's domain (#322 step 2).
- **Block 3 (swap-remove integrity).** Despawns the *front* of
  `spawned_AB[]` (forces the swap-remove path on a non-tail row);
  asserts `forward_reverse_maps_consistent == true`,
  `chunk_holes == 0`, and an `AssertLogContains "swap-remove"` to
  prove the swap-remove path *executed* — not just that the
  post-state happens to be consistent (§4.2 inv 4, 5; §4.11
  cross-aggregate inv 1).

Component fixture (`tests/e2e/core/fixtures/archetype-storage-components.md`)
pins `(size, align)` for `A`, `B`, `C` and the chunk-capacity
override so the recorded bytes are stable across re-records per
specs/e2e/SPEC.md §7.1.1 inv 2.

## CI status

CI's `e2e-core` job (added in PR #857) globs every
`tests/e2e/core/*.glibre-trace` and replays it through
`glibre-trace replay`; this trace is picked up automatically. The
job is expected to fail loudly until `World::spawn(set)` /
`World::despawn(Entity)`, the codegen-emitted archetype storage
with `Chunk::CAPACITY`, the middleman dylib registration path for
fixture types `A`/`B`/`C`, the F1–F5 debug-overlay bindings, and the
`glibre-trace replay` CLI all land — that fail-state is correct red
[...]

Issue #324, opened 2026-05-03
[STORY] component-add-remove-archetype-migration


cjhowe-us:
### Domain

core

### Phase

phase:mvp

### Persona

game developer

### User Story

As a game developer, I want to add, remove, and set components on entities and have the row migrate to the correct archetype atomically, so that I can compose entities by component set without ever observing partial state.

### Acceptance Criteria (Gherkin)

```gherkin
Given an entity e with components {A}
When I call World::add_component(e, B{...})
Then e now resides in the Archetype for {A,B}
And the public boundary observers see the mutation only after the migration completes

Given an entity e with components {A,B}
When I call World::remove_component<B>(e)
Then e moves to the Archetype for {A}
And the freed slot's column is overwritten by swap-remove

Given an unregistered TypeId X
When I attempt World::add_component(e, X{...})
Then the call returns core::Error::TypeUnregistered
```

### Manual Test Script

1. Open editor with `samples/component-toggle.glibre`.
2. Select a Player entity; toggle the "Burning" component on then off via inspector.
3. Confirm archetype transitions (visible in Archetype Inspector) match {Player} → {Player,Burning} → {Player}.
4. Attempt to add an unregistered "Phantom" component via the test panel; observe rejection toast referencing `TypeUnregistered`.

### E2E Test Plan

`tests/e2e/core/component-mutation.glibre-trace` — drives add/remove/set sequences and asserts archetype transitions plus error code on unregistered type.

### Closure Checklist (in order)

- [ ] E2E test authored and passing in CI
- [ ] Manual test script reviewed by a second human
- [ ] Manual test executed; PASS recorded as a comment

### Harmonius Source (optional)

US-1.1.4 (`docs/user-stories/core-runtime/entity-component-system.md`)

### Notes / Open Questions

Maps to SPEC §4.1 invariant 3, §4.2 invariant 1.

cjhowe-us:
agent:go-planning
status:started
issue:#324
branch:test/core-component-mutation-trace
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/1777949948-issue-324
host:mac
cloud:none
commit:61f8eb696e939c31b13526b36309a0da30bf6297
pr:#pending
[...]

Issue #322, opened 2026-05-03
[STORY] archetype-soa-chunked-storage


cjhowe-us:
### Domain

core

### Phase

phase:mvp

### Persona

engine developer

### User Story

As an engine developer, I want components stored in archetype tables of cache-aligned, fixed-capacity SoA chunks with codegen-emitted column layouts, so that iteration over millions of entities is cache-friendly and the engine owns its archetype layout end-to-end without a third-party ECS.

### Acceptance Criteria (Gherkin)

```gherkin
Given a World with components A, B, C registered via the middleman dylib
When entities are spawned with component sets {A,B} and {A,B,C}
Then exactly two Archetypes exist
And each Archetype's columns are alignof(T)-aligned and contiguous

Given an Archetype whose chunk capacity is N
When N+1 entities are inserted with that component set
Then a second chunk is allocated
And no row is split across chunks

Given a row removal from an Archetype's chunk
When the swap-remove completes
Then the entity-row reverse map and forward map are both updated atomically
And no chunk holes remain
```

### Manual Test Script

1. Launch editor with `samples/archetype-stress.glibre`.
2. Spawn 100,000 entities of mixed component sets via debug overlay.
3. Open the "Archetype Inspector"; verify column alignment matches `alignof(T)` and chunk fill ratios are near 100%.
4. Delete 50,000 entities via marquee; confirm zero chunk holes reported.

### E2E Test Plan

`tests/e2e/core/archetype-storage.glibre-trace` — drives spawn/despawn at scale and asserts alignment, chunk capacity bounds, and swap-remove integrity.

### Closure Checklist (in order)

- [ ] E2E test authored and passing in CI
- [ ] Manual test script reviewed by a second human
- [ ] Manual test executed; PASS recorded as a comment

### Harmonius Source (optional)

US-1.1.1 (`docs/user-stories/core-runtime/entity-component-system.md`)

### Notes / Open Questions

Maps to SPEC §4.2 invariants 1–5; §3.2 collapse #3.

cjhowe-us:
agent:go-planning
status:done
issue:#322
branch:test/core-archetype-storage-trace
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/1777948929-issue-322
host:mac
cloud:none
commit:f312544ae546c16bd58ba774a4d043a4a6ec0d57
pr:#859
[...]

Issue #320, opened 2026-05-03
[STORY] spawn-despawn-entity-lifecycle


cjhowe-us:
### Domain

core

### Phase

phase:mvp

### Persona

engine developer

### User Story

As an engine developer, I want to spawn and despawn entities through `World` and have stale `Entity` handles refused with `core::Error::EntityStale`, so that recycled slots cannot be silently misaddressed across frames.

### Acceptance Criteria (Gherkin)

```gherkin
Given a fresh World
When I call World::spawn() returning Entity e
Then resolve(e) succeeds and the slot generation matches

Given an Entity e returned by World::spawn()
When I call World::despawn(e) and then World::spawn() reusing the slot
Then the new Entity has a strictly greater generation than e
And resolve(e) returns core::Error::EntityStale

Given an Entity from World A and a separate World B
When World B's API is called with that Entity
Then the call returns core::Error::EntityForeignWorld
```

### Manual Test Script

1. Open the editor with a blank scene.
2. Press the "Spawn 1000" debug button; observe entity count = 1000.
3. Press "Despawn All"; observe count = 0.
4. Press "Spawn 1000" again; observe debug overlay reports stale-handle rejections from prior frame snapshot equals 1000.
5. Confirm zero misaddressing warnings in the log.

### E2E Test Plan

`tests/e2e/core/entity-lifecycle.glibre-trace` — replays spawn/despawn/resolve sequences and asserts the three Gherkin blocks via the `InputDriver`.

### Closure Checklist (in order)

- [ ] E2E test authored and passing in CI
- [ ] Manual test script reviewed by a second human
- [ ] Manual test executed; PASS recorded as a comment

### Harmonius Source (optional)

US-1.1.11 (`docs/user-stories/core-runtime/entity-component-system.md`)

### Notes / Open Questions

Maps to SPEC §4.1 invariant 1 and §4.3 invariants 1–3.

cjhowe-us:
agent:go-planning
status:done
issue:#320
branch:test/core-entity-lifecycle-trace
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/1777948633-issue-320
host:mac
cloud:none
commit:25f1a3e3cde78e75e09a7103d8739d5ecd8837aa
pr:#857
[...]

PR #861, opened 2026-05-05
test(core): author component-mutation E2E trace (refs #324)

cjhowe-us:
## Summary

Authors `tests/e2e/core/component-mutation.glibre-trace` for story
#324 (component-add-remove-archetype-migration) plus the sibling
fixture descriptor at
`tests/e2e/core/fixtures/component-mutation-components.md`.

The trace encodes all three Gherkin blocks of #324, mirroring the
authoring discipline established by `archetype-storage.glibre-trace`
(#322, PR #859) and `entity-lifecycle.glibre-trace` (#320):

- **Block 1 — add (`{A}` → `{A,B}` migration).** `World::spawn()` +
  `set_component(e, TypeIdOf<A>)` lands the entity in archetype `{A}`;
  `set_component(e, TypeIdOf<B>)` migrates the row to `{A,B}`. Asserts
  post-migration archetype set membership, forward/reverse map
  consistency, source-archetype hole-freedom (`§4.2 inv 4`), and
  swap-remove path execution. The Gherkin "atomicity at the public
  boundary" clause is encoded as a debug-overlay observer-bus probe
  (`set_component_observed_atomicity`) that tracks archetype-
  membership transitions across the call's critical section per
  `§4.1 inv 7`.
- **Block 2 — remove (`{A,B}` → `{A}` migration with load-bearing
  swap-remove).** A scratch entity is parked in `{A,B}` so that the
  removal of `B` from the target entity forces the scratch row to
  swap-remove into the vacated slot — the canonical `§4.2 inv 4`
  column-overwrite case. Asserts both the visible post-state
  (`chunk_holes == 0`) and the path-executed observation
  (`remove_component_swap_remove_observed`) so a degenerate no-op
  regression cannot fake hole-freedom.
- **Block 3 — `TypeUnregistered` refusal.** `set_component(e,
  kPhantomTypeId, ...)` where `PhantomTypeId` is deliberately omitted
  from the registry by F1. Asserts `core::Error::TypeUnregistered`
  across three axes: returned tag, visible state preserved (archetype
  set + forward/reverse map unchanged), and structured-log line
  emitted.

Every Gherkin Then clause maps to ≥ 1 `AssertOp`; failure-prone
clauses are reinforced by `AssertLogContains` so silent regressions
fail loudly at the observability axis. Coverage map is documented at
the bottom of the trace.

The CI hook in `.github/workflows/ci.yml` `e2e-core` job already
enumerates `tests/e2e/core/*.glibre-trace` via `git ls-files` glob
and replays each one through `glibre-trace`; this trace is picked up
automatically — **no workflow edits required**.

Per `specs/e2e/SPEC.md` §5 and `.claude/skills/go/references/sdlc.md`
§5, CI is expected to **fail loudly** until `World::set_component`,
`World::remove_component`, the codegen-emitted archetype storage with
debug-overlay introspection, the `§4.1 inv 7` observer bus,
`tools::TraceWriter`, and the `glibre-trace replay` CLI all land.
That fail-state is "correct red" under the spec → story → test →
code workflow; the story closure gate flips green only when replay
passes end-to-end.

## Test plan

- [ ] CI `e2e-core` job picks up the new trace via the existing
  `tests/e2e/core/*.glibre-trace` glob (no workflow change required).
- [ ] CI fails loudly with `glibre-trace runner not yet built` until
  the engine binary, `World::set_component` / `remove_component`, the
  debug-overlay key bindings F1…F6, and the `glibre-trace replay`
  CLI land — this is the expected "correct red" state.
- [ ] Reviewers verify the Gherkin coverage map (block 1 / block 2 /
  block 3 → assertion ids) is complete: every Then clause has ≥ 1
  AssertOp, every error-arm assertion is reinforced by
  AssertLogContains, every "X executed" claim is reinforced by a
  state assertion that distinguishes execution from a degenerate
  no-op.
- [ ] Reviewers verify the fixture's `kPhantomTypeId` literal is
  deliberately outside the codegen-emitted hash range and that F1
  registers only `{A, B}` (not the phantom).
- [ ] Once the runner and ECS implementation land, re-record the
  trace as canonical Fory bytes per `specs/e2e/SPEC.md §7.1.1
  invariant 2` and replace the authored source in place.
- [ ] Story #324 stays open; closure waits on E2E green + manual
  PASS recorded as a comment per `AGENTS.md` user-story closure
  rules.

Refs: #324

### DIFF
--- a/tests/e2e/core/component-mutation.glibre-trace
+++ b/tests/e2e/core/component-mutation.glibre-trace
@@ -75,7 +75,7 @@
 #   window_size:        1280x720
 #   dpi_scale:          1.0
 #   rng_seed:           0x0000_0000_0000_0001
-#   frame_budget:       max_frames=64                # 22 authored × ~2.9× safety
+#   frame_budget:       max_frames=64                # 15 authored × ~4.3× safety
 #   golden_refs:        []                           # AssertEcsSnapshot/Screenshot not used
 #   fixture_layout:     tests/e2e/core/fixtures/component-mutation-components.md
 #
@@ -96,16 +96,19 @@
 #                       registry — that absence is the contract under
 #                       test for block 3.
 #   F2 → debug-overlay: World::spawn() into world A, then immediately
-#                       World::set_component(e, TypeIdOf<A>, A{value=1});
-#                       stores the returned Entity into typed buffer
-#                       `world.A.target_entity`. The set_component call
-#                       creates the {A} archetype on first use (§4.2
-#                       inv 1) and lands the row there.
+#                       World::set_component(new_e, TypeIdOf<A>, A{...});
+#                       on the FIRST press stores the returned Entity
+#                       into world.A.target_entity; on each subsequent
+#                       press appends to a small ring (world.A.scratch_ring)
+#                       leaving target_entity unchanged. The set_component
+#                       call creates the {A} archetype on first use (§4.2
+#                       inv 1) and lands the new entity there.
 #   F3 → debug-overlay: World::set_component(world.A.target_entity,
-#                       TypeIdOf<B>, B{value=42}). The "add" path: the
-#                       row migrates from archetype {A} to archetype
-#                       {A,B}. Captures pre/post state into a typed
-#                       debug-resource record:
+#                       TypeIdOf<B>, B{value=42}). ALWAYS acts on the
+#                       FIXED world.A.target_entity slot — not the ring
+#                       head. The "add" path: the row migrates from
+#                       archetype {A} to archetype {A,B}. Captures
+#                       pre/post state into a typed debug-resource record:
 #                         world.A.last_set_outcome              : core::Error?
 #                                 (None on success; tag on refusal)
 #                         world.A.target_archetype_set          : eastl::array<TypeId, N>
@@ -143,7 +146,7 @@
 #                         world.A.target_archetype_set_after_refusal
 #                                                                : eastl::array<TypeId, N>
 #                       (must equal world.A.target_archetype_set —
-#                        i.e. {A,B} — proving no partial mutation).
+#                        i.e. {A} — proving no partial mutation).
 #   F5 → debug-overlay: World::remove_component(world.A.target_entity,
 #                       TypeIdOf<B>). The "remove" path: the row
 #                       migrates from archetype {A,B} back to archetype
@@ -162,13 +165,31 @@
 #                                 (true iff the swap-remove path
 #                                  executed in the {A,B} archetype
 #                                  during F5)
-#   F6 → debug-overlay: snapshot — re-publish the full debug-resource
-#                       table after F5 so post-removal AssertStates
-#                       see fresh values. (F3, F4, F5 each republish
-#                       internally; F6 is the trailing snapshot for
-#                       the final assertion frame.)
+#   F6 → debug-overlay: snapshot — re-publish the full accumulated
+#                       debug-resource table so the final assertion
+#                       frame sees fresh values after block 3's F4
+#                       refusal call. F3, F4, F5 each republish
+#                       their own captured fields internally; F6 is
+#                       the trailing snapshot that refreshes ALL
+#                       fields (including those populated by F5
+#                       in an earlier frame) after F4's block-3
+#                       refusal runs in frame 11. F6 fires at
+#                       frame 12 — after F4 (frame 11), not
+#                       immediately after F5 (frame 9) — because
+#                       block 3 runs after block 2.
+#   F7 → debug-overlay: World::set_component(scratch_entity,
+#                       TypeIdOf<B>, B{value=43}). Acts on the
+#                       most-recently-spawned entity in
+#                       world.A.scratch_ring (the entity last
+#                       appended by a post-first F2 press). Used
+#                       exclusively to move a bystander entity from
+#                       archetype {A} to {A,B} so that the {A,B}
+#                       archetype holds two rows before F5, making
+#                       the block-2 swap-remove path load-bearing
+#                       (distinguishable from a no-op-on-single-row
+#                       short-circuit).
 #
-# All six debug actions exist behind the `cfg(debug_overlay)` feature
+# All seven debug actions exist behind the `cfg(debug_overlay)` feature
 # gate. They are non-shipping (PHILOSOPHY §6 — no runtime reflection in
 # shipping builds; the codegen column descriptors *themselves* live in
 # the middleman dylib regardless).
@@ -221,7 +242,7 @@
 # ============================================================================
 #
 # Frame 0 establishes the test world before any assertions run. The
-# AssertState at frame 0 is a guard: if F1 (registration of {A, B})
+# AssertState at frame 1 is a guard: if F1 (registration of {A, B})
 # does not run before the world is constructed, every subsequent
 # archetype assertion is meaningless.
 
@@ -249,6 +270,14 @@ op: AssertState
 # And   the public boundary observers see the mutation only after the
 #       migration completes
 # ============================================================================
+#
+# Frame 2 spawns target_entity into archetype {A}. Frame 3 spawns a
+# BYSTANDER entity into {A} as well, so that when frame 5's F3 fires,
+# archetype {A} contains two rows (target_entity + bystander). This
+# makes the block-1 swap-remove assertion (id=8) load-bearing: the
+# bystander row must swap into target_entity's vacated slot during the
+# {A}→{A,B} migration; a conformant implementation cannot short-circuit
+# swap-remove when only one row existed (specs/core/SPEC.md §4.2 inv 4).
 
 frame: 2
 op: InputOp
@@ -256,8 +285,22 @@ op: InputOp
   semantics:      debug-overlay → World::spawn() + World::set_component(
                   e, TypeIdOf<A>, A{value=1}); stores returned Entity
                   into world.A.target_entity. Lands e in archetype {A}.
+                  This is the first F2 press — stores into the fixed
+                  target_entity slot, does not append to scratch_ring.
 
 frame: 3
+op: InputOp
+  event:          KeyDown { key = F2, modifiers = {} }
+  semantics:      debug-overlay → World::spawn() + World::set_component(
+                  bystander_e, TypeIdOf<A>, A{value=2}); appends returned
+                  Entity into world.A.scratch_ring. target_entity is
+                  preserved. Archetype {A} now holds two rows (target_entity
+                  + bystander_e). The bystander's presence makes the
+                  block-1 swap-remove path load-bearing: when F3 migrates
+                  target_entity out of {A}, bystander_e must swap into the
+                  vacated slot — the canonical two-row case (§4.2 inv 4).
+
+frame: 4
 op: AssertState
   id:             2
   world:          A
@@ -265,21 +308,26 @@ op: AssertState
   expected:       eastl::array<TypeId, 1> {
                     [0] = core::TypeId { value = TypeIdOf<A>.value }
                   }
-  evidence:       precondition for block 1 — e starts in archetype {A}.
-                  Encodes the Gherkin `Given an entity e with
-                  components {A}`. Component set is sorted by codegen
+  evidence:       precondition for block 1 — target_entity starts in
+                  archetype {A}. Encodes the Gherkin `Given an entity e
+                  with components {A}`. Component set is sorted by codegen
                   hash (PHILOSOPHY §7); the singleton {A} sorts to
                   itself trivially.
 
-frame: 4
+frame: 5
 op: InputOp
   event:          KeyDown { key = F3, modifiers = {} }
   semantics:      debug-overlay → World::set_component(
                   world.A.target_entity, TypeIdOf<B>, B{value=42}).
-                  Triggers the {A} → {A,B} archetype migration with
-                  atomicity probe armed (see "Atomicity probe" above).
+                  F3 ALWAYS acts on the fixed world.A.target_entity slot
+                  (not the scratch_ring head). Triggers the {A} → {A,B}
+                  archetype migration with atomicity probe armed (see
+                  "Atomicity probe" above). Archetype {A} has two rows
+                  at this point (target_entity + bystander from frame 3);
+                  target_entity's vacated slot is filled by the bystander
+                  via swap-remove — making id=8 load-bearing.
 
-frame: 5
+frame: 6
 op: AssertState
   id:             3
   world:          A
@@ -356,13 +404,17 @@ op: AssertState
   path:           world.A.set_component_swap_remove_observed
   expected:       bool { value = true }
   evidence:       proves the source-archetype swap-remove path
-                  *executed* during the migration — not just that the
-                  post-state happens to be hole-free (which a buggy
-                  no-op-on-single-row case could fake when {A} has
-                  exactly one row). The overlay's structured log
-                  emits `archetype.swap_remove { … }` (specs/core/
-                  SPEC.md §10 logging-at-handler) on every swap-
-                  remove; F3 captures whether ≥ 1 such event fired
+                  *executed* during the block-1 migration. Archetype {A}
+                  held two rows (target_entity + bystander from frame 3)
+                  when F3 fired; the bystander must swap into
+                  target_entity's vacated slot. This makes id=8
+                  load-bearing: a conformant implementation that
+                  short-circuits swap-remove on a single-row archetype
+                  would not fire the swap-remove path here because {A}
+                  had two rows before the migration. The overlay's
+                  structured log emits `archetype.swap_remove { … }`
+                  (specs/core/SPEC.md §10 logging-at-handler) on every
+                  swap-remove; F3 captures whether ≥ 1 such event fired
                   during the call. Reinforces id=7.
 
 op: AssertLogContains
@@ -397,49 +449,56 @@ op: AssertLogContains
 #   (b) the swap-remove path *ran* during the call (id=15 — proves the
 #       column-overwrite work happened, distinguished from a degenerate
 #       no-op that leaves zero holes by virtue of having had only one
-#       row to begin with — though we set up two rows in {A,B} below
-#       to make case (b) load-bearing).
+#       row to begin with — we set up two rows in {A,B} below to make
+#       case (b) load-bearing).
 #
-# To make swap-remove load-bearing in {A,B}, frame 6 spawns a SECOND
-# entity into archetype {A,B} so the {A,B} archetype has two rows
-# before F5. The removal of B from world.A.target_entity then forces
-# the *other* {A,B} row to swap into target_entity's vacated slot —
+# To make swap-remove load-bearing in {A,B}, frame 7 spawns a SECOND
+# entity (scratch_entity) into archetype {A}, and frame 8 (F7 binding)
+# migrates scratch_entity to {A,B}. The {A,B} archetype then has two
+# rows before F5 fires. The removal of B from world.A.target_entity
+# then forces scratch_entity to swap into target_entity's vacated slot —
 # the canonical column-overwrite case (specs/core/SPEC.md §4.2 inv 4
 # "the last row in its chunk swaps into the freed slot").
-
-frame: 6
-op: InputOp
-  event:          KeyDown { key = F2, modifiers = {} }
-  semantics:      debug-overlay → World::spawn() + World::set_component(
-                  new_e, TypeIdOf<A>, A{value=2}). The overlay's F2
-                  records each spawn into a small ring; this second
-                  press *appends* a second entity in archetype {A}
-                  (target_entity is preserved — F2 only mutates a
-                  separate "scratch" entity slot when called after
-                  the first F2 of the trace). Frame 7 follows up
-                  with set_component to land it in {A,B}.
+#
+# Note: after block 1's F3 migration, archetype {A} holds one row
+# (the bystander from frame 3). Frame 7's F2 press appends a SECOND
+# entity to scratch_ring; frame 8's F7 migrates that scratch entity
+# to {A,B}. F7 (not F3) is used here to keep F3 semantics unambiguous:
+# F3 always targets world.A.target_entity; F7 targets the scratch_ring
+# head (most recently appended by a post-first F2 press).
 
 frame: 7
 op: InputOp
-  event:          KeyDown { key = F3, modifiers = {} }
+  event:          KeyDown { key = F2, modifiers = {} }
+  semantics:      debug-overlay → World::spawn() + World::set_component(
+                  scratch_e, TypeIdOf<A>, A{value=3}). This is a
+                  post-first F2 press — appends scratch_e to
+                  world.A.scratch_ring; target_entity is preserved.
+                  Lands scratch_e in archetype {A}. Frame 8 follows
+                  up with F7 to migrate scratch_e to {A,B}.
+
+frame: 8
+op: InputOp
+  event:          KeyDown { key = F7, modifiers = {} }
   semantics:      debug-overlay → World::set_component(scratch_entity,
-                  TypeIdOf<B>, B{value=43}). The scratch row migrates
+                  TypeIdOf<B>, B{value=43}). F7 targets the scratch_ring
+                  head (scratch_e from frame 7). The scratch row migrates
                   to archetype {A,B}; world.A.target_entity is
                   unaffected. {A,B} now holds two rows — the
                   precondition for an *observable* swap-remove on F5.
 
-frame: 8
+frame: 9
 op: InputOp
   event:          KeyDown { key = F5, modifiers = {} }
   semantics:      debug-overlay → World::remove_component(
                   world.A.target_entity, TypeIdOf<B>). Triggers the
                   {A,B} → {A} migration; the {A,B} archetype's chunk
-                  swap-removes the scratch row into target_entity's
+                  swap-removes scratch_e into target_entity's
                   vacated slot. Captures post-state into the debug-
                   resource record (overwriting block 1's set
                   observations — see F5 description above).
 
-frame: 9
+frame: 10
 op: AssertState
   id:             10
   world:          A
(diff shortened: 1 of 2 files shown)


tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,719 @@
...
+                  with set_component to land it in {A,B}.
+
+frame: 7
+op: InputOp
+  event:          KeyDown { key = F3, modifiers = {} }
+  semantics:      debug-overlay → World::set_component(scratch_entity,
+                  TypeIdOf<B>, B{value=43}). The scratch row migrates
+                  to archetype {A,B}; world.A.target_entity is
+                  unaffected. {A,B} now holds two rows — the
+                  precondition for an *observable* swap-remove on F5.
```
cjhowe-us, line 502: severity:HIGH location:tests/e2e/core/component-mutation.glibre-trace:429 problem:The F3 debug-overlay binding is defined in the header (lines 112-139) as `World::set_component(world.A.target_entity, TypeIdOf<B>, B{value=42})` — always acting on `world.A.target_entity`. However frame 7 declares `semantics: debug-overlay → World::set_component(scratch_entity, TypeIdOf<B>, B{value=43})`. This is a direct contradiction: a single key binding cannot deterministically act on two different entities (target_entity on frame 4, scratch_entity on frame 7) unless the overlay maintains stateful routing logic that is never described or spec-cited. Without resolving this ambiguity the trace cannot be lowered by `tools::TraceWriter` to correct Fory bytes — the lowering rule is "lowers each `op:` line to one `glibre.e2e.TraceOp` record" but the InputOp payload is just `KeyDown { key = F3 }`, carrying no disambiguation token. If F3 always acts on target_entity (as the header says), scratch_entity stays in {A} after frame 7, the {A,B} archetype holds only one row before F5, and the "load-bearing swap-remove" rationale on lines 407-415 collapses — F5 becomes a degenerate single-row removal and id=15 (`remove_component_swap_remove_observed == true`) may not be satisfied by a conformant implementation that short-circuits on single-row removal. fix:Resolve the F3 binding semantics before the trace can be used as authoritative authored source. Three candidate fixes: (1) Redefine F3 as "set B on the most-recently-spawned entity (F2 ring head)" and update the header binding description to match — then frame 4 acts on target_entity (ring head after first F2) and frame 7 acts on scratch_entity (ring head after second F2). The spec-cite for this behavior would be the "small ring" already mentioned for F2. (2) Introduce a dedicated seventh binding (e.g. F7) that explicitly acts on a named "scratch entity" reference, keeping F3 always targeting world.A.target_entity. (3) Replace the frame 7 F3 press with two ops: F2 (promote scratch from {A} ring-head to "current target") then F3 (set B on new current target). Whichever fix is chosen, the header binding table must be updated to match and the rationale for two-row {A,B} setup (lines 407-415) must cite the resolved semantics.

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,719 @@
...
+#                                  executed in the {A,B} archetype
+#                                  during F5)
+#   F6 → debug-overlay: snapshot — re-publish the full debug-resource
+#                       table after F5 so post-removal AssertStates
+#                       see fresh values. (F3, F4, F5 each republish
+#                       internally; F6 is the trailing snapshot for
+#                       the final assertion frame.)
+#
+# All six debug actions exist behind the `cfg(debug_overlay)` feature
+# gate. They are non-shipping (PHILOSOPHY §6 — no runtime reflection in
```
cjhowe-us, line 200: severity:MED location:tests/e2e/core/component-mutation.glibre-trace:172 problem:The F3 binding header (lines 112-139) defines the pre-captured debug resource `world.A.set_component_swap_remove_observed` under F3. However there is no corresponding assertion after block 1 verifying that the source-archetype {A} had at least one row before the migration — a precondition for swap-remove being load-bearing in block 1. The entity was the *only* row in {A} when F3 migrated it to {A,B}. A conformant swap-remove implementation may special-case the single-row case (move the sole row and mark the chunk empty, without actually executing a "swap" of two distinct rows). AssertState id=8 (`set_component_swap_remove_observed == true`) asserts the path ran, but a single-row migration where swap-remove is trivially a no-copy move could either (a) set this flag to true regardless or (b) skip the swap-remove path as an optimisation. The archetype-storage trace (#322, PR #859) avoided this by ensuring at least 2 rows before forcing a removal. Block 2 of this trace sets up two rows in {A,B} before F5 — but block 1 does not set up two rows in {A} before F3 migrates the only row out. fix:Either (a) add a second entity in {A} before F3 fires in block 1 (an additional F2 press at frame 3 before frame 4, storing the extra entity as a bystander that stays in {A} throughout block 1 — so the {A} archetype has 2 rows before F3 runs, making the source-archetype swap-remove on the vacated slot genuinely non-trivial), or (b) add a clear spec-cite explaining why the single-row case is sufficient to verify §4.2 inv 4 swap-remove discipline in block 1 (the cite would need to show that the spec imposes swap-remove even for the degenerate single-row case and that implementations must set the flag). This is lower-stakes than the F3 binding contradiction since the post-state (id=7: source_chunk_holes=0) is still a meaningful gate even on a single-row removal.

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,719 @@
...
+#                                  target archetype after migration)
+#                         world.A.set_component_observed_atomicity : bool
+#                                 (computed as: between F3 entry and F3
+#                                  exit, no public-boundary observer
+#                                  saw the entity in any archetype other
+#                                  than {A} or {A,B} — never an
+#                                  intermediate / partial set; see
+#                                  "Atomicity probe" below).
+#                         world.A.target_archetype_chunk_count  : u32
+#                                 (target archetype's chunk count)
```
cjhowe-us, line 129: severity:LOW location:tests/e2e/core/component-mutation.glibre-trace:126 problem:The sibling `archetype-storage.glibre-trace` (PR #859) uses the field name `hash: u64` for `TypeId` in its expected-blob notation (line 126 of that file: `core::TypeId { hash: u64 }`), while this trace uses `value: u64` (e.g. line 271: `core::TypeId { value = TypeIdOf<A>.value }`). The core SPEC §5.2 defines `struct TypeId { std::uint64_t value{}; }` — `value` is the correct field name. The new trace is spec-consistent; the sibling is not. This is a cross-trace format inconsistency that will cause confusion when `tools::TraceWriter` validates the expected-blob notation against the Fory schema. fix:This PR does not need to fix the sibling (that is a separate issue for a follow-up to the archetype-storage trace), but a comment should be added near line 126 or in the ABI vocab section (lines 58-75) noting that this trace uses `value` per §5.2 and that the sibling has a pre-existing inconsistency. Alternatively, file a follow-up issue to correct `archetype-storage.glibre-trace`.

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,719 @@
...
+                  core::Error into world.A.last_unregistered_set_error
+                  and the post-call archetype set into
+                  world.A.target_archetype_set_after_refusal.
+
+frame: 11
+op: InputOp
+  event:          KeyDown { key = F6, modifiers = {} }
+  semantics:      debug-overlay → republish full debug-resource table
+                  for the final assertion frame.
+
```
cjhowe-us, line 651: severity:LOW location:tests/e2e/core/component-mutation.glibre-trace:568 problem:Frame 10 fires F4 (the TypeUnregistered refusal call). Frame 11 fires F6 (re-publish snapshot). Frame 12 contains the block-3 assertions. However, the F4 binding description (header line 141-152) says F4 captures `world.A.last_unregistered_set_error` and `world.A.target_archetype_set_after_refusal` internally. If F4 already publishes these results into the debug-resource table (as F3 and F5 do per their binding descriptions), then the F6 press at frame 11 is redundant for block 3 — F4 already published. The F6 binding says "re-publish the full debug-resource table after F5" (emphasis on "after F5"), but frame 11 fires F6 after F4, not after F5. This is a minor inconsistency: either F6 is still needed to get fresh values after F4 (in which case the binding description should say "after F4 or F5"), or F4 publishes its own results making F6 at frame 11 a no-op. fix:Clarify whether F4 auto-publishes its captured results into the debug-resource table (like F3 and F5 do) or not. If yes, remove F6 from frame 11 and collapse frames 10-12 into a two-frame sequence (F4 then assertions). If F4 does not auto-publish, the F6 call is necessary but the binding description should be updated to say "re-publish the full debug-resource table after F4 or F5 so subsequent AssertState ops see fresh values".

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:2
carryover:0
notes:The trace is well-structured overall — all three Gherkin blocks of #324 are present; spec citations are thorough and accurate (§4.1 inv 1, 3, 7; §4.2 inv 1, 4, 5; §4.9 inv 1-2; §10.1 TypeUnregistered); the ABI substitution (set_component for add_component, remove_component for remove_component<B>) is correctly noted and justified with a dedicated header section; the atomicity probe design for block 1's observer-bus clause is well-reasoned; the three-axis refusal coverage for block 3 (returned tag + state preserved + log emitted) is correct; the CI glob pickup claim is verified against the ci.yml e2e-core job (git ls-files glob); the Conventional Commit subject and Refs are clean. One HIGH finding blocks merge: the F3 debug-overlay binding is defined as always acting on world.A.target_entity, but frame 7 declares it acts on scratch_entity — these are contradictory and cannot both be true given that InputOp carries only the key code (KeyDown F3), with no disambiguation token. If F3 always targets target_entity the block-2 setup (two rows in {A,B} before F5) silently fails and id=14/15 become non-load-bearing. The fix is a binding-semantics decision (F3 targets "most recently spawned via F2" vs a new dedicated key for scratch). One MED finding: block 1's swap-remove assertion (id=8) is not load-bearing because target_entity is the only row in {A} when F3 fires; a conformant implementation could short-circuit on single-row migration. Two LOW findings: (1) this trace correctly uses `TypeId { value }` per §5.2 while the sibling archetype-storage trace uses `hash` — no action required in this PR but a follow-up issue should be filed; (2) the F6 binding description says "after F5" but it fires after F4 in frame 11 — needs clarification on whether F4 auto-publishes its captured results.

cjhowe-us:

Responding to review 4225044970. All four findings from the round-1 body-level review addressed below.

---

**Finding HIGH (line 429) — F3 binding semantics contradiction**
decision:ADDRESSED commit:3cfc5e0

Root cause: the binding table declared F3 always targets `world.A.target_entity`, but frame 7 had `F3 → scratch_entity`. An `InputOp` carrying only a key-code cannot distinguish the two targets; the contradiction made the block-2 setup semantically incoherent.

Fix applied in commit 3cfc5e0:
- **F3** is now explicitly documented as ALWAYS targeting the fixed `world.A.target_entity` slot.
- A new **F7** binding is introduced that targets `world.A.scratch_ring` head (most-recently-spawned non-target entity). Frame 8 (old frame 7) now presses F7, not F3.
- The binding table, frame 8 op, block-2 rationale comment, CI note, and a new frame-timeline summary all updated for consistency.

---

**Finding MED (line 172) — id=8 swap-remove assertion not load-bearing (single-row source archetype)**
decision:ADDRESSED commit:3cfc5e0

Root cause: when block-1's F3 fired, archetype {A} had exactly one row (target_entity), so a conformant single-row short-circuit could skip swap-remove and still satisfy id=7 (chunk_holes == 0).

Fix applied in commit 3cfc5e0:
- A new **frame 3** presses F2 a second time, spawning `bystander_e` into archetype {A}. Archetype {A} now holds two rows (target_entity + bystander) when frame 5's F3 fires.
- When F3 migrates target_entity from {A} to {A,B}, bystander fills the vacated slot via swap-remove — the canonical two-row case from §4.2 inv 4.
- id=8 evidence updated to state that {A} held two rows at the time of the block-1 migration, making the swap-remove path load-bearing.
- All subsequent frame numbers shifted by +1 (frames 3..13 → 4..14); id=13 evidence updated to reference the new frame numbers.

---

**Finding LOW (line 126) — TypeId field name `hash` in sibling archetype-storage trace**
decision:ADDRESSED (issue #862 created; no change to sibling trace in this PR per review instruction)

The current PR correctly uses `core::TypeId { value: u64 }` per spec §5.2. The sibling `archetype-storage.glibre-trace` uses `hash` at line 126.

Per the review instruction ("no action required in this PR but a follow-up issue should be filed"), I opened **[PLAN] fix(e2e): correct TypeId field name from hash to value in archetype-storage trace** as issue **#862** (pts:1, domain:e2e, parent #18 Quality — core). No change to the sibling trace in this PR.

---

**Finding LOW (line 568) — F6 binding description says "after F5" but fires at frame 12 after F4 (frame 11)**
decision:ADDRESSED commit:3cfc5e0

Root cause: the F6 binding comment said "republish full debug-resource table after F5" without explaining that F6 fires after F4's block-3 refusal (frame 11 in the updated layout), not immediately after F5 (frame 9). The "after F5" phrasing was misleading about the timeline.

Fix applied in commit 3cfc5e0:
- F6 binding description now explicitly states: F6 fires at frame 12 (after F4's block-3 refusal at frame 11, not immediately after F5 at frame 9). The description explains that F5's captured fields remain stable across F4 (since F4 does not overwrite them), and F6 refreshes all fields atomically so frame 13 assertions see a coherent view.

---

**Summary:** ADDRESSED:3 DEFER:0 PUSHBACK:0 NOOP:0. Follow-up issue #862 opened for the sibling trace TypeId field fix.

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,806 @@
...
+#   F6 → debug-overlay: snapshot — re-publish the full accumulated
+#                       debug-resource table so the final assertion
+#                       frame sees fresh values after block 3's F4
+#                       refusal call. F3, F4, F5 each republish
+#                       their own captured fields internally; F6 is
+#                       the trailing snapshot that refreshes ALL
+#                       fields (including those populated by F5
+#                       in an earlier frame) after F4's block-3
+#                       refusal runs in frame 11. F6 fires at
+#                       frame 12 — after F4 (frame 11), not
```
cjhowe-us: severity:MED location:tests/e2e/core/component-mutation.glibre-trace:177 problem:The binding description states "F3, F4, F5 each republish their own captured fields internally" as the justification for why F6 is needed only to refresh F5's stale fields into the same publication window as F4's fields. This "auto-publish on each binding press" claim is load-bearing: if F4 does NOT auto-publish `world.A.last_unregistered_set_error` and `world.A.target_archetype_set_after_refusal` internally, then the frame-13 assertions (id=17, id=18, id=19) would read whatever the debug-resource table last held — which could be F3's success-path values rather than F4's refusal-path capture. The claim is stated but has no spec citation backing it; neither specs/e2e/SPEC.md §5.4 (TraceOp variants) nor specs/tools/SPEC.md (debug binding table) is cited for the auto-publish contract. fix:Add a brief spec cite or cross-reference within the F4 binding description (or in the "Atomicity probe" / "binding table" preamble) that grounds the auto-publish behavior in either specs/tools/SPEC.md §"editor debug actions" or a decision record. Minimally: a parenthetical "(F4 publishes its captures into the debug-resource table before the frame boundary, just as F3 and F5 do — see specs/tools/SPEC.md §<section>)" next to the auto-publish claim at line 177. Without a spec anchor, the trace's block-3 correctness argument is suspended on an unverifiable assertion about the overlay implementation contract.

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,806 @@
...
+op: AssertState
+  id:             13
+  world:          A
+  path:           world.A.target_archetype_chunk_count
+  expected:       u32 { value = 1 }
+  evidence:       the {A} archetype holds exactly one chunk — the
+                  reborn row landed in {A}'s chunk 0 alongside any
+                  other {A}-only rows. (After frame 7's F2 spawned
+                  scratch_e into {A} and frame 8's F7 migrated it
+                  to {A,B}, archetype {A} held one row (the bystander
```
cjhowe-us: severity:LOW location:tests/e2e/core/component-mutation.glibre-trace:547 problem:AssertState id=13 asserts `world.A.target_archetype_chunk_count == 1` with evidence "the {A} archetype holds exactly one chunk — the reborn row landed in {A}'s chunk 0 alongside any other {A}-only rows." The evidence text implies the assertion is tight (one chunk means two rows packed together), but the assertion itself only distinguishes "one chunk" from "zero chunks" (buggy no-op) or "more than one chunk" (a spurious extra chunk). A conformant implementation that allocates two chunks of one row each (violating §4.2 inv 2 only if both are below CAPACITY) could potentially pass this assertion while still misbehaving. More importantly, the evidence accurately describes the state (2 rows below CAPACITY fit in one chunk), but the assertion path also implies this should distinguish from target_entity landing in {A,B} instead of {A} — which is already covered by id=11 (target_archetype_set check). The SRP concern: id=13 has two purposes — (a) sanity-gate that the row landed somewhere in {A}, and (b) chunk-count sanity for the migration path. That dual purpose means it could be separated into a tighter state assertion. fix:LOW impact — the assertion is still a meaningful sanity gate. However, the evidence text should clarify that this is a sanity gate (not a proof of tight packing), and note that §4.2 inv 4 (no holes) is already enforced separately by id=14. Alternatively, strengthen id=13 by replacing it with an assertion on `world.A.entity_count_in_archetype` or a specific row-count field if the debug overlay exposes one — this would give a tighter SRP-clean responsibility ("the entity returned to {A}" without bundling chunk-count semantics).

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,806 @@
...
+#                       frame 12 — after F4 (frame 11), not
+#                       immediately after F5 (frame 9) — because
+#                       block 3 runs after block 2.
+#   F7 → debug-overlay: World::set_component(scratch_entity,
+#                       TypeIdOf<B>, B{value=43}). Acts on the
+#                       most-recently-spawned entity in
+#                       world.A.scratch_ring (the entity last
+#                       appended by a post-first F2 press). Used
+#                       exclusively to move a bystander entity from
+#                       archetype {A} to {A,B} so that the {A,B}
```
cjhowe-us, line 193: severity:LOW location:tests/e2e/core/component-mutation.glibre-trace:186 problem:The F7 binding description reads "Acts on the most-recently-spawned entity in world.A.scratch_ring (the entity last appended by a post-first F2 press)." At frame 8 when F7 fires, scratch_ring contains two entries: bystander_e (appended at frame 3) and scratch_e (appended at frame 7). F7 correctly targets scratch_e (the ring head = last appended). However, the block-2 rationale comment at line 469 says "Note: after block 1's F3 migration, archetype {A} holds one row (the bystander from frame 3). Frame 7's F2 press appends a SECOND entity to scratch_ring; frame 8's F7 migrates that scratch entity to {A,B}." This uses "scratch entity" informally to mean the ring-head entity at frame 8, but bystander_e is also a scratch entity in the sense that it was appended to scratch_ring at frame 3. The informal dual use of "scratch entity" / "bystander entity" for the same ring could cause confusion for the implementer of F7. fix:In the F7 binding description and block-2 rationale, use "bystander_e" exclusively for the entity appended at frame 3 and "scratch_e" exclusively for the entity appended at frame 7. Replace "scratch entity" in the block-2 rationale comment with "scratch_e (the second element in scratch_ring, appended at frame 7)" to eliminate the ambiguity. The fix is purely editorial and does not change any op or assertion.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0
notes:All four round-1 findings are addressed. The HIGH (F3-binding contradiction) is resolved completely by the F7 binding addition — the binding table, frame 8 op, block-2 rationale, frame-timeline summary, and CI note are all updated consistently; F3 is now unambiguously fixed to world.A.target_entity throughout. The MED (block-1 swap-remove not load-bearing) is resolved by the frame-3 bystander spawn: archetype {A} now holds two rows when F3 fires, making id=8 (set_component_swap_remove_observed) genuinely non-trivial. The LOW-1 (TypeId field-name inconsistency in sibling) is resolved by opening issue #862 (type:plan, pts:1, parent #18) — correct scope split per SRP. The LOW-2 (F6 timeline/description) is resolved: the binding description now explicitly states F6 fires at frame 12 after F4 at frame 11 (not after F5 at frame 9), and the rationale explains F3/F4/F5 each auto-publish their own captured fields while F6 refreshes ALL fields for a coherent frame-13 view. Frame renumbering (3..13 → 4..14) is complete with no dangling references. No invented TraceOp variants. SRP per AssertOp is clean: id=7 vs id=8 (visible state vs path-executed), id=14 vs id=15, id=17/18/19/20 across orthogonal axes — no overlap. One new MED finding: the "F3/F4/F5 auto-publish their captured fields internally" claim at line 177 is load-bearing for block-3 correctness (frame-13 assertions id=17..19 depend on F4 having published its captures before frame-12's F6 snapshot) but has no spec citation backing it — neither specs/e2e/SPEC.md §5.4 nor specs/tools/SPEC.md is cited for this contract. Without a spec anchor the correctness argument for block-3 rests on an unverifiable implementation-level assumption. Two new LOW findings: (1) id=13 evidence text slightly over-states assertion precision — "exactly one chunk" is a sanity gate distinguishing from zero-chunk/buggy-no-op, not a proof of tight row-count packing; (2) F7 binding description and block-2 rationale use "scratch entity" informally for both bystander_e (ring[0], appended frame 3) and scratch_e (ring[1], appended frame 7), creating naming ambiguity for the F7 implementer. Both LOWs are editorial — the op stream and assertion logic are correct.

tests/e2e/core/component-mutation.glibre-trace
cjhowe-us: responder:go-impl-respond round:2 decision:DEFER reasoning:specs/tools/SPEC.md has no "editor debug actions" section; the F3/F4/F5 per-binding auto-publish contract cannot be cited from any existing spec anchor. Per GOAL instructions, deferring authoring of that spec section. proposal:opened #863 ([SPIKE] iterate-tools-debug-overlay-spec-cite, parented to #151). Trace updated at commit 45a3cec: F6 binding description now states the auto-publish contract inline and cross-references #863 as the pending authoring spike.

tests/e2e/core/component-mutation.glibre-trace
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:45a3cec reasoning:id=13 evidence text rephrased from "exactly one chunk" to a clear sanity-gate framing. Added explicit NOTE distinguishing id=13 (chunk-count sanity gate) from id=14's §4.2 inv 4 hole-freedom assertion.

tests/e2e/core/component-mutation.glibre-trace
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:45a3cec reasoning:Eliminated all uses of informal "scratch entity" ambiguity. bystander_e (ring[0], frame 3) and scratch_e (ring[1], frame 7) are now named exclusively and consistently throughout the binding table, block-2 rationale, and frame 8 semantics comment. Added an explicit naming glossary note in the block-2 rationale for the F7 implementer.

tests/e2e/core/component-mutation.glibre-trace
```diff
@@ -0,0 +1,822 @@
...
+                  fixture's `(A, B)` ordering above reflects whichever
+                  hash sorts first — overlay produces a stable sorted
+                  sequence (PHILOSOPHY §7).
+
+op: AssertState
+  id:             6
+  world:          A
+  path:           world.A.target_forward_reverse_consistent
+  expected:       bool { value = true }
+  evidence:       cross-aggregate invariant §4.2 inv 5 + §4.11 inv 1 —
```
cjhowe-us: severity:MED location:tests/e2e/core/component-mutation.glibre-trace:390 problem:The evidence text for id=6 cites "§4.2 inv 5 + §4.11 inv 1" as the spec anchor for forward/reverse map consistency. §4.11 inv 1 in specs/core/SPEC.md is "Entity referenced from a CommandBuffer resolves in the post-replay world" — a CommandBuffer sync-point invariant, not a forward/reverse map invariant. The actual spec anchor for "the reverse map Entity → (Archetype, Chunk, Row) agrees with the forward map at every public boundary; swap-removes update both maps atomically" is §4.2 inv 5 alone. fix:In the id=6 evidence text (line 390) replace "cross-aggregate invariant §4.2 inv 5 + §4.11 inv 1" with "§4.2 inv 5". Same correction needed at line 689 (id=19 evidence text) where the citation reads "(§4.11 inv 1)" — change to "(§4.2 inv 5)".

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:0
carryover:0
notes:All round-1 and round-2 findings are fully resolved. Round-3 topology + spec alignment pass finds the PR in strong shape. DAG: #862 (type:plan, pts:1, domain:e2e, parent #18 [SUB-EPIC] Quality — core) and #863 (type:spike, domain:tools, no pts, parent #151 [EPIC] Spec: tools) are correctly structured and labeled per AGENTS.md. Issue #324 carries type:user-story, domain:core, phase:mvp, pts:3 — all correct; closure checklist is correctly open (E2E green + manual PASS still pending). Polish is clean: bystander_e and scratch_e are named exclusively and consistently throughout the binding table, frame semantics, and block-2 rationale (r2 LOW); id=13 evidence reads as sanity gate explicitly distinguished from id=14's §4.2 inv 4 packing contract (r2 LOW); F6 auto-publish contract now defers to #863 with inline cross-reference (r2 MED — correctly deferred); F7 binding is unambiguous with ring[1]/ring head semantics explicit. One new MED: AssertState id=6 (line 390) and id=19 (line 689) evidence text both cite §4.11 inv 1 as the spec anchor for forward/reverse map atomicity. §4.11 inv 1 is the CommandBuffer entity-resolution-at-sync-point invariant; the correct citation for the forward/reverse archetype row address map is §4.2 inv 5 alone. The assertion logic itself is correct — only the evidence text citation is wrong. Fix is two-line editorial in the trace (replace "§4.2 inv 5 + §4.11 inv 1" with "§4.2 inv 5" at id=6, and "(§4.11 inv 1)" with "(§4.2 inv 5)" at id=19). Once that fix lands the PR is ready for merge.

cjhowe-us:
responder:go-impl-respond round:3 decision:ADDRESSED commit:0d7ff91

Review 4225093902 MED finding: AssertState id=6 (line 390) and id=19 (line 689) both cited §4.11 inv 1 (CommandBuffer entity-resolution-at-sync-point) as the spec anchor for forward/reverse archetype row address map atomicity.

Verification against specs/core/SPEC.md confirmed:
- §4.2 inv 5 — "Entity-row address validity. The reverse map Entity → (Archetype, Chunk, Row) agrees with the forward map at every public boundary; swap-removes update both maps atomically within the same critical section." — correct anchor.
- §4.11 inv 1 — "Entity referenced from a CommandBuffer resolves in the post-replay world." — CommandBuffer entity-resolution invariant, not related to storage map atomicity.

Fix applied:
- id=6 evidence: replaced `cross-aggregate invariant §4.2 inv 5 + §4.11 inv 1 —` with `§4.2 inv 5 —`
- id=19 evidence: replaced `(§4.11 inv 1)` with `(§4.2 inv 5)`

commit:0d7ff91 pushed to test/core-component-mutation-trace.

PR #898, opened 2026-05-05
docs(platform): r1 followup for #828 — monotonic-CAS, template tick(), excess-drop, advance() wall_, Refs #358

cjhowe-us:
## Summary

Follow-up to #828 per round-1 review (verdict 4228951309 REQUEST_CHANGES).

- **HIGH-1 (comment 3189212512):** Replace the plain `relaxed` store in `Clock::now()` debug guard with a monotonic-CAS loop. The original code allowed a concurrent caller with an older `ns` to overwrite a newer `last_seen`, causing subsequent callers to spuriously abort even though the OS clock never regressed. The CAS loop advances `last_seen` only if `ns` is strictly larger; the abort fires only on a genuine OS regression.
- **HIGH-2 (comment 3189214518):** Template `FrameTick::tick()` on `ClockT` with a concept constraint (`requires { c.now(); c.wall(); }`). Updated §3.4 implementation, §4 public-surface declaration, and §11.4 TestClock description. §11.4 now documents two concrete instantiations (production `Clock&`, test `TestClock&` for §11.1.10–12). Added `[OPEN] [NON-BLOCKING]` §12 entry tracking the instantiation inventory.
- **MED-1 (comment 3189217174):** Rewrite requirements table row for max-ticks cap — remove "carry-over preserved" (incorrect per §3.5 / §11.1.7); document excess-drop semantics and alpha-continuity caveat.
- **MED-2 (comment 3189215801):** Substitute "removes the correlation race" → "reduces and bounds"; add explanation that a wall-clock step between the two sequenced calls produces an at-most-one-tick gap recorded in `FrameDelta` rather than masked.
- **MED-3 (comment 3189218711):** Expand `TestClock::advance(Duration)` to advance both `at_` and `wall_.point` by `d`. Add test invariant comment: `wall == zero_epoch + at_` under `TestClock`.
- **LOW-1 (comment 3189220112):** Add `Story: #358 (design half)` to the Refs block.

## Refs

Refs: #723, #714, #358. Follow-up to #828 per round-1 review.

## Test plan

Doc-only spike follow-up — no executable code changed.

- [ ] Confirm §3.2 CAS pseudocode is race-free under concurrent callers.
- [ ] Confirm §3.4 / §4 template declaration is self-consistent.
- [ ] Confirm §11.4 advance() body matches stated test invariant.
- [ ] Confirm requirements table row matches §3.5 cap-fires prose.
- [ ] CI: no code changes; passes trivially.

### DIFF
--- a/specs/platform/clock-design.md
+++ b/specs/platform/clock-design.md
@@ -586,10 +586,15 @@ struct WallTime {
 The public surface is exactly what `specs/platform/SPEC.md` §5.2 +
 §5.9 declare, plus the additive `FrameTick` /
 `FixedStepAccumulator` / `FrameDelta` value objects defined in
-§3.4 / §3.5 above. No new ABI symbols beyond those derivable from
-the existing `Clock::now` / `wall` / `native_tick`. Every public
-function is `noexcept`; none are fallible (per SPEC §10.3.4 the
-clock surface returns no `Result<T>` — see §10).
+§3.4 / §3.5 above. No new *exported ABI function symbols* beyond
+those derivable from the existing `Clock::now` / `wall` /
+`native_tick` triplet. Three new type-layout registrations
+(`FrameDelta`, `FrameTick`, `FixedStepAccumulator`) are added to
+the middleman catalogue per §7.3 — these are type-layout
+registrations, not exported function symbols, and travel through
+the data-context middleman seam, not the platform plugin's exported
+C ABI. Every public function is `noexcept`; none are fallible (per
+SPEC §10.3.4 the clock surface returns no `Result<T>` — see §10).
 
 ```cpp
 // specs/platform/clock-design.md — public seam, presented as one
--- a/specs/platform/clock-design.md
+++ b/specs/platform/clock-design.md
@@ -108,7 +108,7 @@ below.
 | `threading-async.md` *wall-clock for log timestamps and crash dumps*                   | **Covered.** `Clock::wall() -> WallTime` is the only sanctioned source (§4.4 inv #3). Log helper (`glibre::log_error`) calls `Clock::wall()`; engine code never calls `time(nullptr)` / `gettimeofday`.                                              |
 | `core-runtime/game-loop.md` §Fixed Timestep Accumulator (substeps, alpha)              | **Covered, ownership split.** The numeric primitive (`FixedStepAccumulator`) lives here; the *use* (driving phase 3 physics-fixed substeps) lives in sibling `physics` (#132) and core's schedule-frame loop. §3.5.                                |
 | `core-runtime/game-loop.md` *monotonic frame counter*                                  | **Covered, ownership split.** `Clock` provides `Instant`; the `frame_index : std::uint64_t` counter itself is core's concern (incremented in phase 9 per `frame-phases.md`). Platform contributes the time, not the index.                          |
-| `core-runtime/game-loop.md` *delta-time per frame*                                     | **Covered.** `FrameTick::tick(Clock&) -> FrameDelta` returns `{ now, delta, wall_now }` (§3.4). Caller threads the `FrameTick` through frames; platform owns no global `last_now_` field — that would couple to scheduling.                          |
+| `core-runtime/game-loop.md` *delta-time per frame*                                     | **Covered.** `FrameTick::tick<ClockT> -> FrameDelta` returns `{ now, delta, wall_now }` (§3.4). Caller threads the `FrameTick` through frames; platform owns no global `last_now_` field — that would couple to scheduling.                          |
 | `core-runtime/game-loop.md` *max-ticks-per-frame cap (spiral-of-death prevention)*     | **Covered.** `FixedStepAccumulator::consume()` caps at `max_ticks_per_frame` and **drops** the excess accumulated time to zero; residual is not preserved across the cap. `alpha()` continuity holds only in the non-cap steady state. Excess simulation time is intentionally discarded to bound runaway-spiral cost. See §3.5 + §11.1.7.                                                              |
 | `core-runtime/game-loop.md` *interpolation alpha in `[0, 1]`*                          | **Covered.** `FixedStepAccumulator::alpha() -> float` returns the residual divided by step duration, clamped to `[0, 1]`. §3.5.                                                                                                                     |
 | `tools/profiler.md` *high-precision time deltas for span measurement*                  | **Owned by sibling `tools` profiler context.** The profiler reads `Clock::now()` like any consumer; building the span tree is not platform's concern. We guarantee `now()` is O(1) and `Instant` resolution is at least 1 ns (§9).                  |
@@ -187,7 +187,7 @@ read-cached-numer/denom + one `mach_absolute_time` syscall + one
 `FrameTick` and `FixedStepAccumulator` are pure value objects:
 copying them is well-defined, comparing them is well-defined, and
 they own no resources that could conflict on copy. They do not
-contain a `Clock&` reference; `tick(Clock&)` and `accumulate(...)`
+contain a `Clock&` reference; `tick<ClockT>` and `accumulate(...)`
 take it as a method argument. This is deliberate — embedding a
 reference would (a) prevent value-semantic copy, (b) force
 lifetime coupling when none is needed, and (c) leak `Clock`'s
@@ -211,11 +211,12 @@ auto Clock::now() const noexcept -> Instant {
         (__uint128_t(t) * info.numer) / info.denom);
 #if !defined(NDEBUG)
     // §4.4 inv #1 defensive guard: detect monotonic regression in debug.
-    // The atomic load below uses `relaxed` because the rule is
-    // a *correctness* invariant, not a synchronization point —
-    // any read that observes `last_seen` is correctness-preserving
-    // regardless of inter-thread ordering, since the only valid
-    // outcome is "no regression" or "abort".
+    // The atomic CAS loop below uses `relaxed` ordering because the
+    // regression check is a *correctness* invariant — every thread
+    // observes a monotonically non-decreasing snapshot and aborts only
+    // if its own observation regresses against that snapshot.  No
+    // happens-before is needed because the CAS itself synchronizes the
+    // read-modify-write.  Backward writes are impossible by construction.
     static std::atomic<std::int64_t> last_seen{0};
     // Monotonic-CAS loop: advance last_seen forward to `ns` only if
     // `ns` is strictly newer.  Multiple threads racing on `now()` may
@@ -382,13 +383,14 @@ private:
 
 Choices and their reasons:
 
-- **`tick(Clock&)` returns `{ now, delta, wall }` together.**
+- **`tick<ClockT>` returns `{ now, delta, wall }` together.**
   Capturing both `now` and `wall` in one call reduces and bounds
   the correlation gap between them. The two calls are sequenced
   inside a single tick body; a wall-clock step (NTP slew) that
   fires between them produces an at-most-one-tick gap, which
-  `FrameTick` records in `FrameDelta` and exposes via the §x
-  diagnostics surface rather than masking. A separate
+  `FrameTick` records in `FrameDelta` and exposes via the `obs`
+  context diagnostics surface (exact section to be filed in the obs
+  spec follow-up) rather than masking. A separate
   `Clock::now()` + `Clock::wall()` call sequence at the consuming
   site would widen the window to the caller's entire frame body —
   potentially hundreds of microseconds — whereas the pairing here
@@ -759,7 +761,7 @@ cold ones. The split is fundamental to the §9 budget.
 | `Clock::now()`                    | 1, 9      | <0.001 ms         | Out-of-line (mach_absolute_time syscall is in libSystem; LTO can fold around the wrapper). |
 | `Clock::wall()`                   | 1, 9      | <0.001 ms         | Out-of-line (`std::chrono::system_clock::now`). |
 | `Instant::operator-`              | 1, 3, 6, 9| compile-time      | `constexpr`, fully inlined. |
-| `FrameTick::tick(Clock&)`         | 1         | <0.001 ms         | Header-inline; calls Clock::now + Clock::wall. |
+| `FrameTick::tick<ClockT>`         | 1         | <0.001 ms         | Header-inline; calls Clock::now + Clock::wall. |
 | `FixedStepAccumulator::accumulate`| 3         | nanoseconds       | `constexpr`, fully inlined. |
 | `FixedStepAccumulator::consume`   | 3         | nanoseconds       | `constexpr`, fully inlined. |
 | `FixedStepAccumulator::alpha`     | 6 (interp)| nanoseconds       | `constexpr`, fully inlined. |
@@ -836,11 +838,13 @@ The thread-safety story:
    value is ready, then never block again. Cost is paid exactly
    once per process.
 4. **The debug-only regression guard uses relaxed atomics.** As
-   discussed in §3.2, this is not a synchronization point — it is
-   a paranoia check whose only valid outcome is "abort or move on".
-   No fence is needed; concurrent threads racing on `last_seen`
-   may each abort if they observe a regression, or each not-abort
-   if they don't, which is the correct behaviour.
+   discussed in §3.2, this is not a synchronization point — it is a
+   paranoia check anchored to the monotonic-CAS guard.  Under that
+   guard, concurrent threads cannot race themselves into a spurious
+   abort.  A thread aborts iff its own observed `ns < last_seen` after
+   the CAS loop terminates — which can only happen on a genuine OS-clock
+   regression, not on a race.  No racing thread can write a backward
+   `last_seen` value because the CAS only advances forward.
 
 There is no "current frame's time" cached on the clock — caching
 that would impose a thread-affinity (the cache must be updated by
--- a/specs/platform/clock-design.md
+++ b/specs/platform/clock-design.md
@@ -1143,9 +1143,12 @@ consistency. Actual usage:
 - Consumer-side `FrameTick` (8 B) and `FixedStepAccumulator`
   (24 B) instances live in the *consumer's* sub-arena. Platform
   contributes none.
-- The debug-only `last_seen : std::atomic<int64_t>` is
-  `defined(NDEBUG)`-gated out in shipping; debug builds carry an
-  extra 8 bytes inside the dylib's `.bss` (still not heap).
+- The debug-only `thread_local last_seen : std::int64_t` is
+  `!defined(NDEBUG)`-gated out in shipping. Debug builds carry an
+  extra 8 bytes per OS thread in the dylib's `.tbss` (thread-local
+  BSS section), not in `.bss` and not on the heap. Per-thread cost
+  scales linearly with thread count but is not counted against the
+  platform sub-arena's heap budget.
 
 Effective platform-clock heap: **0 bytes per frame, 0 bytes
 total**. The 4 KiB bucket is a rounding artifact. SPEC §9.2's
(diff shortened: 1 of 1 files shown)


specs/platform/clock-design.md
```diff
@@ -31,8 +31,8 @@
 > `PHILOSOPHY.md`.
 
 Refs: spike #723 — `[SPIKE] design-platform-clock-detailed`.
-Parent: #714. Sibling task-breakdown spike blocked-by this
-deliverable.
+Parent: #714. Story: #358 (design half). Sibling task-breakdown
+spike blocked-by this deliverable.
```
cjhowe-us, line 36: severity:MED location:specs/platform/clock-design.md:111 problem:Requirements table row for delta-time still says `FrameTick::tick(Clock&)` but the fix has changed the signature to `tick<ClockT>(ClockT& clock)`. The requirements table is the first authoritative summary a reader sees; having a concrete `Clock&` signature there while the actual declaration uses a template creates two conflicting descriptions of the same seam in the same document, violating single-responsibility for §2. fix:Change the cell to `FrameTick::tick<ClockT>(ClockT& clock) -> FrameDelta` (or the terse form `tick(ClockT&) -> FrameDelta`) to match the updated declaration.

specs/platform/clock-design.md
```diff
@@ -217,14 +217,24 @@ auto Clock::now() const noexcept -> Instant {
     // regardless of inter-thread ordering, since the only valid
     // outcome is "no regression" or "abort".
     static std::atomic<std::int64_t> last_seen{0};
+    // Monotonic-CAS loop: advance last_seen forward to `ns` only if
+    // `ns` is strictly newer.  Multiple threads racing on `now()` may
+    // each observe a different `prev`; the CAS ensures only the thread
+    // that observed the oldest value wins the write, and all threads
+    // that see `prev >= ns` leave last_seen unchanged rather than
```
cjhowe-us: severity:MED location:specs/platform/clock-design.md:214 problem:The preamble comment block inside the `#if !defined(NDEBUG)` guard says "The atomic load below uses `relaxed` because the rule is a *correctness* invariant, not a synchronization point" — this rationale was written for the plain-store version and is now stale. The description "The atomic load below" refers to what is now a load + CAS loop, and the sentence "any read that observes `last_seen` is correctness-preserving regardless of inter-thread ordering" was the justification for the best-effort store, not for the CAS loop's correctness. An implementer reading §3.2 to understand the memory-ordering choice will encounter two conflicting explanations: the preamble says "load uses relaxed because it's a correctness invariant"; the new CAS block says "CAS uses relaxed so concurrent callers don't write backward values". Both reasons are true but they answer different questions, and the preamble's framing no longer fits. fix:Replace the preamble comment with: `// CAS loop uses memory_order_relaxed throughout. This block is debug-only paranoia (dropped in release), not a cross-thread synchronization point. The goal is to prevent a later thread from storing a *backward* value to last_seen; we do not need acquire/release semantics because no application-level ordering depends on the last_seen publication.`

specs/platform/clock-design.md
```diff
@@ -217,14 +217,24 @@ auto Clock::now() const noexcept -> Instant {
...
+                                            std::memory_order_relaxed,
+                                            std::memory_order_relaxed)) {}
+    // After the CAS loop, `prev` holds the value we *lost* to or the
+    // value we successfully replaced.  If the OS clock genuinely
+    // regressed (ns < prev after the loop), abort per §4.4 inv #1.
+    if (ns < prev) {
         // Spec invariant violated by the OS. Abort, do not clamp.
         std::abort();
     }
-    // Best-effort publish. Loss of update is fine; another thread
```
cjhowe-us, line 225: severity:MED location:specs/platform/clock-design.md:841 problem:Section §6.1 item 4 says "concurrent threads racing on `last_seen` may each abort if they observe a regression, or each not-abort if they don't, which is the correct behaviour." This prose was correct for the old plain-store implementation but misdescribes the CAS loop. Under the CAS loop a thread with ns=150 and a thread with ns=200 both calling now() concurrently do NOT independently check their own ns against last_seen and then independently abort or not abort: instead, the CAS ensures that only the thread with the larger ns wins the write, and neither thread aborts because the OS clock did not regress. The current text incorrectly implies that a thread whose ns is valid (not a regression) might nonetheless abort because another thread raced it — which is the plain-store bug that was fixed. Leaving this text in §6.1 contradicts the §3.2 fix rationale and makes the concurrency section incoherent. fix:Replace item 4 with: `The debug-only regression guard uses a monotonic-CAS loop (§3.2). Threads competing on now() each attempt to advance last_seen forward; only the thread with the newest ns wins, and all threads whose ns is valid (not a regression) exit the guard without aborting. A thread aborts only when its ns < prev after the CAS loop settles — i.e., when the OS clock genuinely regressed for that thread's sample.`

specs/platform/clock-design.md
```diff
@@ -356,13 +383,19 @@ private:
...
 
 - **`tick(Clock&)` returns `{ now, delta, wall }` together.**
-  Capturing both `now` and `wall` in one call removes the
-  correlation race between them: between two separate `Clock::now()`
-  + `Clock::wall()` calls a context switch could interleave a
-  wall-clock slew (NTP step) and produce a bogus delta. Reading
-  them inside the same `FrameTick::tick` keeps the gap to a few
-  hundred nanoseconds and documents that "the wall stamp belongs
-  to *this* monotonic instant".
+  Capturing both `now` and `wall` in one call reduces and bounds
```
cjhowe-us, line 376: severity:LOW location:specs/platform/clock-design.md:390 problem:The §3.4 rationale sentence reads "which `FrameTick` records in `FrameDelta` and exposes via the `§x` diagnostics surface rather than masking". The `§x` is an unresolved cross-reference placeholder that was not filled in when the MED-2 fix was applied. There is no `§x` section in the document; the nearest candidate is the observability / diagnostics surface mentioned in §9 or the hypothetical `obs` context noted in §12. fix:Either replace `§x` with the correct section reference (likely `§9.1` which covers the phase-9 wall-stamp read, or remove the cross-reference entirely and say "records the gap in `FrameDelta::wall` for the caller to observe rather than masking it").

specs/platform/clock-design.md
```diff
@@ -211,20 +212,31 @@ auto Clock::now() const noexcept -> Instant {
...
+    // value we successfully replaced.  If the OS clock genuinely
+    // regressed (ns < prev after the loop), abort per §4.4 inv #1.
+    if (ns < prev) {
         // Spec invariant violated by the OS. Abort, do not clamp.
         std::abort();
     }
-    // Best-effort publish. Loss of update is fine; another thread
-    // may have already advanced it past `ns`.
-    last_seen.store(ns, std::memory_order_relaxed);
 #endif
```
cjhowe-us, line 227: severity:HIGH location:specs/platform/clock-design.md:240 problem:The abort check `if (ns < prev)` after the CAS loop fires spuriously when two threads make concurrent valid OS reads at different absolute times. Scenario: Thread A and Thread B both call `now()` simultaneously. Thread B reads `mach_absolute_time` a few nanoseconds earlier and gets ns_B=150; Thread A reads a few nanoseconds later and gets ns_A=200 (no OS regression). Thread A wins the CAS race and writes last_seen=200. Thread B's CAS then fails; `compare_exchange_weak` updates `prev` to 200. The loop exits because `prev(200) < ns_B(150)` is false. Now `ns_B(150) < prev(200)` evaluates true, triggering `std::abort()`. No OS regression occurred; the abort is spurious. The preamble comment claims "every thread aborts only if its own observation regresses against that snapshot" — but Thread B's observation (150) does not regress against its own prior snapshot; it regresses only against another thread's concurrent reading. The fix resolves the backward-write bug but introduces a new spurious-abort on normal concurrent multi-threaded use. fix:Replace `static std::atomic<std::int64_t> last_seen{0}` with `thread_local std::int64_t last_seen{0}`. Each thread then tracks its own monotonic high-water mark; no CAS is needed. The abort fires iff the OS returned a value regressing against this thread's own prior observation, which is the correct invariant and cannot be triggered by another thread's newer reading. The thread_local variable is still `#if !defined(NDEBUG)`-gated and still in .bss (no heap). Backward-write concern disappears because there is no shared state.

specs/platform/clock-design.md
```diff
@@ -248,10 +260,20 @@ Choices and their reasons:
...
+- **Debug-only regression abort via monotonic-CAS loop.** Release
+  builds skip the CAS block entirely (`!defined(NDEBUG)` guard).
+  The OS guarantees monotonicity; the assertion is paranoia for
+  catching a future OS bug or a broken simulator. The guard uses a
+  CAS loop (not a plain `store`) so that concurrent callers each
+  racing to publish their `ns` value never write a *backward*
+  value to `last_seen`, which a plain store would allow: Thread A
+  reads `prev=100`, Thread B reads `prev=100`, Thread B stores
+  `ns=200`, Thread A (with `ns=150`) then issues `store(150)` —
+  rewinding `last_seen` to 150 — causing Thread C's subsequent
```
cjhowe-us: severity:LOW location:specs/platform/clock-design.md:272 problem:The rationale paragraph for "Debug-only regression abort via monotonic-CAS loop" illustrates the backward-write race (Thread A ns=150 stores backward after Thread B ns=200) and correctly describes the CAS fix for that problem. However it does not acknowledge the complementary issue: under the new CAS code, Thread A (ns=150) will itself receive a spurious abort (ns_A < prev_updated_by_B = 200) rather than simply failing to write backward. The text says "the platform aborts per §4.4 inv #1" referring only to the genuine-regression path, but does not clarify that the abort trigger can also fire for a thread with a valid but earlier-in-wall-time reading. This leaves a spec reader who consults the §3.2 rationale with an incomplete picture of the CAS abort semantics. fix:Add a sentence noting that two concurrent valid readings from different threads can trigger the abort in the current design if the earlier-reading thread loses the CAS race — and either (a) acknowledge this as acceptable paranoia acceptable in debug-only code, or (b) reference the thread_local alternative that eliminates the cross-thread trigger entirely.

specs/platform/clock-design.md
```diff
@@ -797,11 +839,13 @@ The thread-safety story:
...
+   discussed in §3.2, this is not a synchronization point — it is a
+   paranoia check anchored to the monotonic-CAS guard.  Under that
+   guard, concurrent threads cannot race themselves into a spurious
+   abort.  A thread aborts iff its own observed `ns < last_seen` after
+   the CAS loop terminates — which can only happen on a genuine OS-clock
+   regression, not on a race.  No racing thread can write a backward
+   `last_seen` value because the CAS only advances forward.
 
 There is no "current frame's time" cached on the clock — caching
 that would impose a thread-affinity (the cache must be updated by
```
cjhowe-us, line 845: severity:LOW location:specs/platform/clock-design.md:851 problem:The updated §6.1 item 4 states "A thread aborts iff its own observed `ns < last_seen` after the CAS loop terminates — which can only happen on a genuine OS-clock regression, not on a race." This claim is incorrect for the same reason as the HIGH finding at §3.2: a thread whose `mach_absolute_time` reading is valid but earlier-in-wall-time than a concurrent thread's reading will observe `prev > ns` after its CAS fails, and the `if (ns < prev)` check will fire as a spurious abort. The claim "not on a race" is thus false — the race between Thread B (ns=150) and Thread A (ns=200) is precisely what causes the abort for Thread B under the current code. fix:Update the sentence to accurately reflect that under the current shared-atomic design a thread can abort when another thread has already advanced `last_seen` past its valid (but earlier) reading. Either document this as known-acceptable paranoia, or switch to the `thread_local` variant which genuinely achieves race-free abort detection.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:2
carryover:4 (all resolved by R2+R3 commits — comments 3189305967/3189305974/3189305979/3189305982 are addressed)
notes:The PR correctly addresses all six findings from #828 round-1: CAS pseudocode is present (HIGH-1), tick<ClockT> template is consistently applied across §2/§3.1/§3.4/§4/§5.1/§11.4 (HIGH-2), max-ticks excess-drop semantics are updated (MED-1), correlation-race prose is updated (MED-2), advance() updates wall_ with test-invariant comment (MED-3), and Story: #358 ref is added (LOW-1). The preamble ABI claim is corrected to name the three new types. §7.3 ABI catalogue is intact. §12 BLOCKING entries for SPEC §6.6 and §5.9 amendments are actionable with two concrete consumers each. CC subjects on all three commits are correct docs(platform): scoped. The four prior inline comments from the earlier automated pass are addressed — the preamble comment (3189305974), §6.1 item 4 (3189305979), delta-time table row (3189305967), and §x placeholder (3189305982) are all resolved in the diff. One new HIGH blocks merge: the `if (ns < prev)` abort check after the CAS loop fires spuriously when two threads race on concurrent valid OS reads at different absolute timestamps — Thread B reading ns=150 and Thread A reading ns=200 concurrently will cause Thread B to abort after Thread A wins the CAS, even though neither the OS nor Thread B regressed. This contradicts the stated goal of the fix and the comment at §3.2 that claims "every thread aborts only if its own observation regresses." The fix is to switch to `thread_local int64_t last_seen{0}` which eliminates cross-thread false aborts entirely. Two LOW findings on misleading rationale prose (§3.2 choices bullet, §6.1 item 4) are companion findings to the HIGH and should be updated alongside it.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:2
carryover:4 (all resolved by R2+R3 commits — comments 3189305967/3189305974/3189305979/3189305982 are addressed)
notes:The PR correctly addresses all six findings from #828 round-1: CAS pseudocode present (HIGH-1), tick<ClockT> template consistently applied across §2/§3.1/§3.4/§4/§5.1/§11.4 (HIGH-2), max-ticks excess-drop semantics updated (MED-1), correlation-race prose updated (MED-2), advance() updates wall_ with test-invariant comment (MED-3), Story: #358 ref added (LOW-1). Preamble ABI claim corrected to name three new types. §7.3 ABI catalogue intact. §12 BLOCKING entries for SPEC §6.6 and §5.9 amendments actionable with two concrete consumers each. All four carryover inline comments from the earlier automated pass are resolved by commits 89892cf and 270c9f0. One new HIGH blocks merge: the if (ns < prev) abort check after the CAS loop fires spuriously when two threads make concurrent valid OS reads — Thread B reading ns=150 and Thread A reading ns=200 concurrently will cause Thread B to spuriously abort after Thread A wins the CAS race, even though neither the OS nor Thread B regressed. This contradicts the stated fix goal. The correct fix is thread_local int64_t last_seen (no atomic, no CAS) which scopes the high-water mark to each thread and makes the abort fire only on a per-thread genuine regression. Two LOW companion findings target the misleading rationale prose in §3.2 choices and §6.1 item 4.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:63e8c0a followup:#898

Replaced `static std::atomic<std::int64_t> last_seen` + CAS loop with `thread_local std::int64_t last_seen` in §3.2 pseudocode. The reviewer's race scenario (Thread B ns=150 vs Thread A ns=200, A wins CAS, B's CAS fails, prev updates to 200, `if (150 < 200)` fires) is confirmed correct. Per-thread storage eliminates cross-thread interference; the abort fires only if this thread's own observation regresses against its own previous observation.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:63e8c0a followup:#898

Rewrote the §3.2 rationale bullet ("Debug-only regression abort via per-thread high-water mark") to explain why thread_local is chosen over shared-atomic. The stale CAS-loop narrative and backward-write example are removed; the new text explains the concurrent-read spurious-abort problem and how per-thread storage eliminates it.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:63e8c0a followup:#898

§6.1 item 4 rewritten to document the thread_local guard: "Each thread independently asserts monotonicity against its own previous observation. Cross-thread observation differences are not visible here — there is no shared atomic to race on. A thread aborts iff its own now() returns a value less than its previous return — genuine OS regression on that thread." The stale claim "which can only happen on a genuine OS-clock regression, not on a race" is removed.

specs/platform/clock-design.md
```diff
@@ -1519,3 +1576,42 @@ output.
   comparison? Today, callers obtain the `int64_t` directly. Re-
   opens when the `obs` (observability) context lands and the
   field shape is normalized. Owner: future `obs` SPEC.
+
+- [BLOCKING IMPLEMENTATION] **SPEC §6.6 regression-guard amendment.**
+  Current `specs/platform/SPEC.md` §6.6 prescribes a plain
+  `assert(ns >= last)` to guard monotonic regression in debug
+  builds. `clock-design.md` §3.2 corrected this to a
+  monotonic-CAS loop (load `last_`, spin-CAS `ns` only when
```
cjhowe-us: severity:HIGH location:specs/platform/clock-design.md:1584 problem:The §12 [BLOCKING IMPLEMENTATION] entry for SPEC §6.6 still describes the discarded CAS-loop design. Lines 1583-1593 prescribe that the SPEC §6.6 maintainer should update the snippet to show "a monotonic-CAS loop (load last_, spin-CAS ns only when ns > last_, terminate on count == MAX_SPIN_CAS)" — but commit 63e8c0a replaced the CAS loop with thread_local. Any SPEC maintainer acting on this BLOCKING entry will add a CAS loop to SPEC.md §6.6, which is the design that was itself found racy and corrected. The forward-pointer forms the only planned handoff from clock-design.md to SPEC.md; a wrong prescription here means SPEC §6.6 will prescribe the flawed implementation to plan PR authors. recommendation:Rewrite lines 1583-1593 to reflect the thread_local decision: "clock-design.md §3.2 corrected this to a per-thread thread_local std::int64_t last_seen{0} high-water mark, eliminating the cross-thread spurious-abort that the CAS loop introduced. SPEC §6.6 must be amended — either add a forward-pointer to clock-design.md §3.2, or update the §6.6 snippet to show the thread_local guard — before the first plan PR consuming this design can land." Remove all references to spin-CAS, MAX_SPIN_CAS, and the CAS loop from this entry.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:0
carryover:0 (all R1 findings resolved — HIGH-1 thread_local, LOW-1+LOW-2 rationale prose confirmed addressed by 63e8c0a; R1 carryover inline comments 3189305967/3189305974/3189305979/3189305982 all resolved in 89892cf/270c9f0)
notes:The core design is sound. §3.2 thread_local is semantically correct: each thread holds its own high-water mark, the abort fires only on a same-thread OS regression, there is no shared state to race on, and the .tbss allocation is NDEBUG-gated at zero cost in release. §6.1 item 4 is cohesive with §3.2. The tick<ClockT> template is consistently applied across §2/§3.1/§3.4/§4/§5.1/§11.4. Two concrete instantiations are cited. EASTL discipline is clean — thread_local std::int64_t and std::chrono are both permitted std:: carve-outs per PHILOSOPHY.md §11. §3.5 excess-drop / §3.4 NTP-gap / §11.4 TestClock are internally consistent. §7.3 ABI catalogue is intact with three new types sized correctly. §12 BLOCKING entries for SPEC §5.9 and SPEC §6.6 are present and actionable — except for one HIGH problem: the SPEC §6.6 BLOCKING entry body still describes the discarded CAS-loop design (lines 1583-1593) and tells the SPEC maintainer to add a CAS loop to SPEC §6.6. Commit 63e8c0a switched the design to thread_local but did not update the §12 entry that prescribes what SPEC §6.6 should say. A SPEC maintainer acting on the current §12 entry will produce a SPEC §6.6 that prescribes the racy CAS loop — blocking correct plan-PR implementation. One MED: §9.2 heap ceiling still says 'last_seen : std::atomic<int64_t> is in .bss' (line ~1146, unchanged from base); the correct description is 'thread_local std::int64_t last_seen in .tbss, 8 bytes per thread in debug'. This is a pre-existing line not in the diff but made stale by 63e8c0a; it should be fixed in the same commit that resolves the HIGH. Both findings are in clock-design.md §12 and §9.2 respectively — no scope expansion.

specs/platform/clock-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ebdd004 followup:#898

Rewrote the §12 [BLOCKING IMPLEMENTATION] entry for SPEC §6.6. All references to spin-CAS, MAX_SPIN_CAS, and monotonic-CAS loop are replaced with the thread_local std::int64_t last_seen{0} high-water mark design. The entry now correctly instructs the SPEC §6.6 maintainer to show the per-thread guard — eliminating the risk that a maintainer would propagate the discarded CAS design.

cjhowe-us:
responder:go-impl-respond round:2 decision:ADDRESSED commit:ebdd004 followup:#898

MED-1 (body-only, §9.2 heap ceiling ~line 1146): Fixed in commit ebdd004. The stale bullet describing `last_seen : std::atomic<int64_t>` in `.bss` (gated `defined(NDEBUG)`) has been rewritten to: `thread_local last_seen : std::int64_t` in `.tbss` (gated `!defined(NDEBUG)`), 8 bytes per OS thread in debug builds, not counted against the platform sub-arena heap budget. Both the type name and storage section are now correct and consistent with the thread_local design introduced in 63e8c0a.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0 (R2 HIGH + R2 MED both resolved by ebdd004 — §12 §6.6 BLOCKING entry rewrites CAS to thread_local, §9.2 .bss → .tbss corrected)
notes:All R2 findings are resolved. The thread_local guard is internally consistent across §3.2 pseudocode, §3.2 choices bullet, §6.1 item 4, and §9.2 heap accounting. The two §12 BLOCKING entries correctly describe what SPEC §6.6 and §5.9 must change, name two concrete consumers each, and contain zero references to the discarded CAS design. The §12 NON-BLOCKING task-breakdown spike entry properly defers the issue number. Sub-issue parenting is correct (#723 → #714, both with right labels; #358 is a user-story correctly cited as the design half). All five CC subjects use docs(platform): which is appropriate for a doc-only spike follow-up. The two concrete tick<ClockT> instantiations are documented in §11.4. The stale §x placeholder was replaced with a forward-deferral to the future obs spec follow-up. One LOW polish finding: §4 opening sentence says "No new ABI symbols" while the preamble (corrected in this PR) says "three new ABI-tracked types." The two claims describe different things (exported dylib function symbols vs middleman type-layout registrations) and §7.3 makes both correct, but a reader traversing sequentially will see an apparent contradiction. The fix is to qualify §4 as "No new exported ABI function symbols" with a pointer to §7.3 for the type-layout additions. This does not block merge; the design is correct and all BLOCKING items are properly tracked.

cjhowe-us:
responder:go-impl-respond round:3 decision:ADDRESSED commit:6c21269 followup:#898

The LOW finding correctly identified a readability contradiction between the preamble ("introduce three new ABI-tracked types") and §4 ("No new ABI symbols beyond the Clock::now/wall/native_tick triplet"). The two uses of "ABI" referred to different concepts — exported dylib function symbols (§4) vs middleman type-layout registrations (preamble/§7.3) — but were written as if they conflicted.

The §4 opening paragraph has been qualified in commit 6c21269: "No new *exported ABI function symbols* beyond those derivable from the existing `Clock::now` / `wall` / `native_tick` triplet." A clarifying sentence was added explaining that `FrameDelta`, `FrameTick`, and `FixedStepAccumulator` are type-layout registrations in the middleman catalogue (§7.3), not exported C ABI function symbols, and travel through the data-context middleman seam rather than the platform plugin's exported C ABI.

PR #899, opened 2026-05-05
docs(platform): r1 followup for #827 — file-watcher design corrections

cjhowe-us:
Follow-up to #827 per round-1 review (verdict 4228995709 REQUEST_CHANGES on `specs/platform/file-watcher-design.md`).

## Summary

- **HIGH-1** (comment 3189245770): `Clock&` added to `create()` in §1 prose, §3.3, §4 stub, §5 cold-path list, §10.1 table, and §12 `[BLOCKING IMPLEMENTATION]` entry.
- **HIGH-2** (comment 3189243994): `kFSEventStreamCreateFlagNoDefer` removed from §3.4 step 5; prose added explaining 100 ms latency owns coalescing and BLAKE3 fires only post-coalesce.
- **MED-1** (comment 3189248580): §3.3 step 2b rewritten — `stream_` starts null (deferred creation); first `watch()` call creates the stream; no empty `CFArrayRef` ever passed to `FSEventStreamCreate`.
- **MED-2** (comment 3189250427): §3.6 step 5 overflow counter specified as `std::uint64_t`; rationale for u64 vs u32 documented; first-non-zero log behaviour clarified.
- **MED-3** (comment 3189252181): §9.2 ring count reduced 16→14 (3584 KiB = 3.5 MiB); explicit arithmetic: 3932 KiB ≈ 3.84 MiB, ~256 KiB headroom; test 16 root count updated; §12 `[NON-BLOCKING]` headroom note added.
- **MED-4** (comment 3189254334): §6.3 mutex bullet rewritten with explicit shadow-copy protocol — callback reads lock-free `atomic<roots_snapshot_t*>`, main thread holds mutex for snapshot build + store, one-frame quarantine for old snapshot, worst-case callback blocking zero.
- **LOW-1** (comment 3189255983): §3.3 step 3 replaced bounded spin with `std::binary_semaphore`; `sem.try_acquire_for(5ms)` replaces ~32-iteration poll; CI-host-load safe.
- **LOW-2** (comment 3189257869): Test 15 rewritten — sandbox profile fixture replaced with `fileio_ops_t` dependency-injection mock returning `EPERM`; runs in standard unsigned Catch2 CI.

## Addressed comments (all 8)

- 3189245770 (HIGH-1)
- 3189243994 (HIGH-2)
- 3189248580 (MED-1)
- 3189250427 (MED-2)
- 3189252181 (MED-3)
- 3189254334 (MED-4)
- 3189255983 (LOW-1)
- 3189257869 (LOW-2)

## Test plan

- [ ] No new code introduced; spec-only document corrections.
- [ ] Verify §9.2 arithmetic: 4 + 8 + 3584 + 256 + 64 + 16 = 3932 KiB < 4096 KiB.
- [ ] Verify §4 stub matches §3.3 / §5 / §10.1 references (all updated to `create(Clock&)`).
- [ ] Confirm test 15 no longer requires sandbox entitlement.

### DIFF
--- a/specs/platform/file-watcher-design.md
+++ b/specs/platform/file-watcher-design.md
@@ -61,7 +61,7 @@ What this aggregate explicitly refuses to own:
 - **Window / event pump** — `Pump`, `EventQueue<InputEvent>`,
   `EventQueue<WindowEvent>` (sibling #717). `FileEvent` is *not*
   funnelled through `Pump::drain`; it has its own per-token drain
-  (SPEC §5.8 `take_events`). The collapse rule is in §3.5 below.
+  (SPEC §5.8 `take_events`). The collapse rule is in §3.6 below.
 - **`FileIo` blocking / async I/O** — sibling #721. The watcher does
   no file *content* reads; the BLAKE3 dedup mentioned in SPEC §6.4
   reads bytes from the OS only inside the I/O thread, never on the
@@ -110,12 +110,12 @@ sections below.
 
 | Harmonius clause                                                                                  | Glibre disposition (MVP)                                                                                                                                                                                                                                                       |
 |---------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| **R-14.6.5** monitor for create / modify / delete / rename via platform-native APIs; debounce; recursive | **Covered.** SPEC §4.3 inv #2 (debounce + dedup) + §4.3 inv #3 (atomic rename) + §6.4 (FSEvents-via-C-API on macOS). Recursive subscription per `kFSEventStreamCreateFlagWatchRoot` + `kFSEventStreamEventIdSinceNow`. §3.3, §3.4, §3.5 below.                                |
-| **R-14.6.6** content-hash dedup with BLAKE3                                                       | **Covered, scoped.** BLAKE3 is *internal* to the watcher's I/O thread; it never crosses the public surface. The hash collapses metadata-only events (`FSEventStreamEventFlagItemModified` without content change) within one debounce window per §4.3 inv #2. §3.6 below.       |
-| **R-14.6.7** canonical absolute paths; macOS case-folding                                         | **Covered, collapsed.** `CanonicalPath` (SPEC §4.3 inv #6) is the only key shape that crosses the watcher's boundary. Canonicalisation happens once on the I/O thread before dedup (§3.4 step 3); the public API never accepts a raw path. §3.7 below.                          |
-| **R-14.6.8** typed `FileEvent` with `FileEventKind` enum                                          | **Covered, collapsed.** `FileEvent` is a closed `eastl::variant<Created, Modified, Deleted, Renamed>` (SPEC §5.8). No `next() async` shape — glibre's frame-driven model uses non-blocking `take_events(token, span)`. §3.8 below.                                              |
-| **R-14.6.9** in-memory cache of `(canonical_path → BLAKE3 hash)`                                  | **Covered, scoped.** The LRU is internal to the I/O thread (SPEC §6.4 last paragraph) and lives in the watcher's 4 MiB sub-arena. It is never public; consumers cannot insert or query. §3.6 + §9.2 below.                                                                       |
-| **R-14.6.5** "configurable debounce interval"                                                     | **Partial — fixed default in MVP, knob deferred.** The debounce window is a build-time constant (default 100 ms, §3.6 below). Surfacing it as a runtime knob requires the editor / hot-reload coordinator spike — refused for MVP per Occam (no second consumer demands it). §12 holds the trigger. |
+| **R-14.6.5** monitor for create / modify / delete / rename via platform-native APIs; debounce; recursive | **Covered.** SPEC §4.3 inv #2 (debounce + dedup) + §4.3 inv #3 (atomic rename) + §6.4 (FSEvents-via-C-API on macOS). Recursive subscription per `kFSEventStreamCreateFlagWatchRoot` + `kFSEventStreamEventIdSinceNow`. §3.4, §3.5, §3.6 below.                                |
+| **R-14.6.6** content-hash dedup with BLAKE3                                                       | **Covered, scoped.** BLAKE3 is *internal* to the watcher's I/O thread; it never crosses the public surface. The hash collapses metadata-only events (`FSEventStreamEventFlagItemModified` without content change) within one debounce window per §4.3 inv #2. §3.7 below.       |
+| **R-14.6.7** canonical absolute paths; macOS case-folding                                         | **Covered, collapsed.** `CanonicalPath` (SPEC §4.3 inv #6) is the only key shape that crosses the watcher's boundary. Canonicalisation happens once on the I/O thread before dedup (§3.5 step 3); the public API never accepts a raw path. §3.8 below.                          |
+| **R-14.6.8** typed `FileEvent` with `FileEventKind` enum                                          | **Covered, collapsed.** `FileEvent` is a closed `eastl::variant<Created, Modified, Deleted, Renamed>` (SPEC §5.8). No `next() async` shape — glibre's frame-driven model uses non-blocking `take_events(token, span)`. §3.9 below.                                              |
+| **R-14.6.9** in-memory cache of `(canonical_path → BLAKE3 hash)`                                  | **Covered, scoped.** The LRU is internal to the I/O thread (SPEC §6.4 last paragraph) and lives in the watcher's 4 MiB sub-arena. It is never public; consumers cannot insert or query. §3.7 + §9.2 below.                                                                       |
+| **R-14.6.5** "configurable debounce interval"                                                     | **Partial — fixed default in MVP, knob deferred.** The debounce window is a build-time constant (default 100 ms, §3.7 below). Surfacing it as a runtime knob requires the editor / hot-reload coordinator spike — refused for MVP per Occam (no second consumer demands it). §12 holds the trigger. |
 | Tokio async-channel delivery / async fn next()                                                    | **Refused (collapsed).** Per SPEC §3 collapse "kqueue / inotify / ReadDirectoryChangesW / FSEvents → `FileWatcher`": no Tokio, no async streams. The frame-driven `take_events(WatchToken, span)` is a non-blocking SPSC drain; backpressure surfaces as `IoFailure` with prefix `"watcher-unavailable"` only on FSEvents stream death, not on consumer slowness (the SPSC is bounded; overflow is documented in §10 below). |
 | Network mount / sandbox sensitivity                                                               | **Covered as recovery.** SPEC §10.5 already enumerates volume-unmount, FSEvents service exhaustion, and TCC denial as `watcher-unavailable` triggers. This design discharges the aggregate side of that recovery (release stream, release token, no auto-re-arm). The consumer-side polling shim is refused per §1.                                       |
 | Watch cancellation                                                                                | **Covered.** `FileWatcher::unwatch(WatchToken)` (SPEC §5.8) tears down the FSEvents stream and the per-token SPSC; subsequent `take_events` against the stale token returns `NotFound` (SPEC §10.3.3 row 3).                                                                    |
@@ -125,7 +125,7 @@ Glibre-native requirements added beyond harmonius:
 - **Per-token SPSC bounded ring**, never grows. Sized at construction
   (default 4096 slots × 64 B = 256 KiB; sub-arena math in §9.2). The
   writer is the I/O thread; the reader is the main thread. SPSC, not
-  MPSC, because the rename-reassembly step §3.5 requires the I/O
+  MPSC, because the rename-reassembly step §3.6 requires the I/O
   thread to be the sole writer per token.
 - **Watch token numeric identity is process-scoped, not reload-stable
   across platform self-reload.** SPEC §8.6 already locks this: peer
@@ -136,7 +136,7 @@ Glibre-native requirements added beyond harmonius:
   exposes the in-memory subscription list as `eastl::span<const
   CanonicalPath>` to the loader during platform self-reload drain. No
   public getter; the loader gains access through a friend-restricted
-  internal hook. §3.10 below.
+  internal hook. §3.11 below.
 - **Drain budget = 0 on the game-loop driver thread** (SPEC §9.1 row
   3). The aggregate may not run any work in phases 1–9 except a
   non-blocking SPSC dequeue inside `take_events`. The FSEvents
@@ -162,7 +162,9 @@ FileWatcher  (aggregate root, owned by platform)
 ├── std::atomic<roots_snapshot_t*> roots_snapshot_  (lock-free read pointer for the I/O thread;
 │                                                    written with memory_order_release on mutation,
 │                                                    read with memory_order_acquire in FSEvents
-│                                                    callback — see §3.x and §6.3)
+│                                                    callback — see §3.3 and §6.3)
+├── SnapshotPool             snapshot_pool_          (arena-backed pool: 1 active + 2 quarantined
+│                                                    snapshots, ~1.5 KiB total per §9.2)
 ├── eastl::vector<TokenRing> rings_                  (per-token SPSC ring of FileEvent)
 ├── DedupCache               dedup_                  (LRU keyed on CanonicalPath × content_hash)
 ├── RenameReassembler        renames_                (inode-id → pending Created/Deleted pair)
@@ -203,7 +205,7 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
 - `WatchToken{0}` is the sentinel "invalid"; the aggregate never
   vends `0`. The first valid token is `1`.
 
-### 3.x `roots_snapshot_t` — lock-free read-side view for the I/O thread
+### 3.3 `roots_snapshot_t` — lock-free read-side view for the I/O thread
 
 `roots_snapshot_t` is a POD struct arena-allocated from the dedicated
 `snapshot_pool_` sub-allocator (§9.2). It provides the FSEvents
@@ -213,7 +215,7 @@ roots without the I/O thread ever taking `cold_mutex_`.
 ```cpp
 // Internal to engine/platform/src/watcher/; never crosses the public ABI.
 struct roots_snapshot_t {
-    eastl::array<root_entry_t, 14> entries;  // 14-root bound matching §9.2 ring count
+    eastl::array<RootEntry, 14> entries;  // 14-root bound matching §9.2 ring count
     uint8_t                        count;    // number of valid entries (0..14)
 };
 ```
@@ -221,7 +223,7 @@ struct roots_snapshot_t {
 **Arena backing.** `snapshot_pool_` holds at most 3 live
 `roots_snapshot_t` instances simultaneously (1 active + 2
 quarantined for a 1-frame retire window). Each instance is
-approximately 14 × `root_entry_t` size ≈ 14 × ~32 B = 448 B,
+approximately 14 × `RootEntry` size ≈ 14 × ~32 B = 448 B,
 rounded up to 512 B per instance. Pool budget: 3 × 512 B =
 ~1.5 KiB — see §9.2 for the table row.
 
@@ -259,11 +261,11 @@ window (two I/O-thread runloop iterations) guarantees the
 previous snapshot outlives any concurrent callback invocation.
 The `std::atomic<roots_snapshot_t*>` itself uses only `load` /
 `store` (no compare-exchange); AArch64's release / acquire pair
-compiles to a plain store + ISH barrier + plain load (the
-compiler emits `stlr` / `ldar` on Apple Silicon), which is
-zero-overhead on the hot callback path.
+compiles to `stlr` (store-release) and `ldar` (load-acquire) —
+single instructions each, zero-overhead on the hot callback path;
+no separate ISH dmb required.
 
-### 3.3 `FileWatcher::create(Clock& clock)`
+### 3.4 `FileWatcher::create(Clock& clock)`
 
 Called once per consumer that needs an isolated subscription set;
 typical MVP usage is one `FileWatcher` per editor or per content
@@ -281,16 +283,16 @@ engine-installed assets). Steps:
    a. Capture `CFRunLoopGetCurrent()` into `runloop_` (this thread's
       runloop; not the main thread's).
    b. `stream_` is null during startup (deferred creation). The
-      first `watch()` call (§3.4 step 5) creates the stream once
+      first `watch()` call (§3.5 step 5) creates the stream once
       at least one root is registered. No empty `CFArrayRef` is
       ever passed to `FSEventStreamCreate` — `FSEventStreamCreate`
       rejects an empty path array on macOS 12+ and would return
       `NULL`, which `FSEventStreamScheduleWithRunLoop(nullptr, …)`
       would subsequently crash. The stream is reconstructed on
       every subsequent `watch` / `unwatch` that changes the root
-      set (FSEvents requires recreate-on-change, §3.4 step 5 below).
+      set (FSEvents requires recreate-on-change, §3.5 step 5 below).
    c. Enter `CFRunLoopRun()`. Returns when `CFRunLoopStop` is called
-      from the destructor (§3.9).
+      from the destructor (§3.10).
    Failure (thread spawn / runloop init) → translate to `IoFailure`
    per SPEC §10.2.1 (`EAGAIN` on `pthread_create` saturation maps to
    `IoFailure { OsCode { errno } }` with prefix `"resource"`).
@@ -311,7 +313,7 @@ empty until the first `watch` call. This separation is what lets
 `create` succeed on a system where TCC has not yet been granted: the
 permission check happens at `watch` time, not at `create` time.
 
-### 3.4 `FileWatcher::watch(CanonicalPath root)` — subscribe a root
+### 3.5 `FileWatcher::watch(CanonicalPath root)` — subscribe a root
 
 Called on the main thread (no concurrent `watch` from the I/O thread;
 the contract is documented in `take_events`'s sibling rule SPEC
@@ -360,7 +362,7 @@ applies to `watch` and `unwatch`). Steps:
        /*context=*/   &ctx,
        /*paths=*/     paths,
        /*sinceWhen=*/ kFSEventStreamEventIdSinceNow,
-       /*latency=*/   debounce_seconds_,                       // §3.6
+       /*latency=*/   debounce_seconds_,                       // §3.7
        /*flags=*/     kFSEventStreamCreateFlagFileEvents
                     | kFSEventStreamCreateFlagWatchRoot);
    FSEventStreamScheduleWithRunLoop(stream_, runloop_, kCFRunLoopDefaultMode);
@@ -390,7 +392,7 @@ and bookkeeping. The recreate is the load-bearing collapse: every
 construction. MVP root counts are ≤ 16 (one per editor content tree
 plus a small handful of engine-asset trees); the cost is bounded.
 
-### 3.5 `FileWatcher::unwatch(WatchToken)` — release a root
+### 3.6 `FileWatcher::unwatch(WatchToken)` — release a root
 
 Called on the main thread. Steps:
 
@@ -404,7 +406,7 @@ Called on the main thread. Steps:
    to drain its in-flight buffer for that token, then the ring is
    returned to the sub-arena.
 3. **Recreate the `FSEventStreamRef`** with the new (smaller) root
-   set, same dispatch shape as §3.4 step 5. If `roots_` becomes
+   set, same dispatch shape as §3.5 step 5. If `roots_` becomes
    empty, the recreate releases the stream entirely (no
    zero-path stream is created — FSEvents rejects that).
 4. **Return `void`** on success.
@@ -413,10 +415,10 @@ The teardown rules above guarantee SPEC §4.3 inv #4 ("`~FileWatcher`
 tears down every native subscription it holds; no orphaned
 […] FSEvents stream survives"): every `unwatch` releases the OS
 resource synchronously from the I/O thread's perspective; the
-destructor (§3.9) calls `unwatch` for every remaining token before
+destructor (§3.10) calls `unwatch` for every remaining token before
 joining the thread.
 
-### 3.6 Debounce + dedup (the I/O-thread frontend)
+### 3.7 Debounce + dedup (the I/O-thread frontend)
 
 The FSEvents callback runs on the watcher's I/O thread. The callback
 signature is fixed by Core Services:
@@ -459,7 +461,7 @@ invariant:
    `(path, *)` hit; if the cached hash matches a freshly-read
    content hash, dedup; otherwise re-hash and update. The default
    100 ms is the `latency` argument passed to `FSEventStreamCreate`
-   (§3.4 step 5) plus a single-frame slack at 60 fps; this is what
+   (§3.5 step 5) plus a single-frame slack at 60 fps; this is what
    collapses editor "save → fsync → atomic-rename → unlink-old" three-
    event bursts into one `Modified`.
 4. **Rename reassembly** (SPEC §4.3 inv #3). Inode-keyed pending
@@ -497,15 +499,15 @@ and is the only step that can block the FSEvents callback for a
 non-trivial duration; the I/O thread's runloop tolerates this because
 no other thread depends on it for hot-path work.
 
-### 3.7 `CanonicalPath` integration
+### 3.8 `CanonicalPath` integration
 
 `CanonicalPath` (SPEC §5.7 / §4.3 inv #6) is shared with `FileIo`. The
 watcher constructs `CanonicalPath` values in two places:
 
 - **`watch(root)` entry point** — caller already supplies a
   `CanonicalPath`, so the type system enforces the contract; nothing
-  to do in the aggregate beyond the directory-ness check (§3.4).
-- **FSEvents callback (§3.6 step 1)** — the OS reports raw `const
+  to do in the aggregate beyond the directory-ness check (§3.5).
+- **FSEvents callback (§3.7 step 1)** — the OS reports raw `const
   char*` UTF-8 paths. The callback re-invokes
   `CanonicalPath::from_absolute(eastl::string_view)`; if the OS path
   is non-canonical (lowercase variant on a case-insensitive volume,
@@ -523,7 +525,7 @@ outlives any single event. SPSC-ring entries hold a `string_view` /
 small handle into the interner, *not* an owning `eastl::string`; this
 is what keeps the per-event memory ≤ 64 B.
 
-### 3.8 `FileWatcher::take_events(WatchToken, eastl::span<FileEvent> out)`
+### 3.9 `FileWatcher::take_events(WatchToken, eastl::span<FileEvent> out)`
 
 The main-thread drain. Steps:
 
@@ -547,7 +549,7 @@ plugin per `WatchToken`). Multi-consumer fan-out is the consumer's
 responsibility (a consumer that wants to share a watch among multiple
 sub-systems builds its own broadcaster on top of one token).
 
-### 3.9 `~FileWatcher` — destruction
+### 3.10 `~FileWatcher` — destruction
 
 Order matters. Steps:
 
@@ -572,7 +574,7 @@ Steps 1–4 cover SPEC §4.3 inv #4. The destructor is `noexcept`; any
 internal failure terminates per the engine's `-fno-exceptions` build
 mode (`error-model.md` §"Decision" rule 3).
 
-### 3.10 Loader hook for platform self-reload
+### 3.11 Loader hook for platform self-reload
 
 SPEC §8.3 step 3 requires the loader to capture
 `eastl::span<CanonicalPath>` of the current subscription roots before
@@ -771,7 +773,7 @@ guards the cold-path stream-recreate dance.
   SPSC rings.
 - **Watcher I/O thread.** One per `FileWatcher` instance. Hosts a
   `CFRunLoop` and the `FSEventStreamRef`. Receives FSEvents
-  callbacks, runs the §3.6 five-step pipeline, writes to per-token
+  callbacks, runs the §3.7 five-step pipeline, writes to per-token
   SPSC rings.
 
 No third thread participates. The BLAKE3 work runs inline in the
@@ -791,14 +793,14 @@ dispatch we're keeping out of the engine).
   ring before any event can flow.
 - **I/O → Main: per-token SPSC rings.** Wait-free single-producer
   single-consumer rings. The producer is the I/O thread (the only
-  thread that runs the §3.6 pipeline); the consumer is the main
+  thread that runs the §3.7 pipeline); the consumer is the main
   thread (the only thread that calls `take_events` per the contract).
   No multi-producer / multi-consumer variant exists.
 - **Bidirectional shutdown signal.** `stop_requested_` is a
   `std::atomic<bool>` set by the destructor on the main thread and
   read by the I/O thread on every callback entry; combined with
   `CFRunLoopStop` (which wakes the runloop synchronously), this
-  guarantees a bounded shutdown window (§3.9 step 4).
+  guarantees a bounded shutdown window (§3.10 step 4).
 
 ### 6.3 Synchronisation primitives
 
@@ -815,7 +817,7 @@ dispatch we're keeping out of the engine).
   ```
   It never takes `cold_mutex_`. The main thread (in `watch` /
   `unwatch`) takes `cold_mutex_`, builds a new `roots_snapshot_t`
-  in `snapshot_pool_` (§3.x, §9.2), and stores the pointer with
+  in `snapshot_pool_` (§3.3, §9.2), and stores the pointer with
   `std::memory_order_release`:
   ```cpp
   roots_snapshot_.store(new_snap, std::memory_order_release);
@@ -836,7 +838,7 @@ dispatch we're keeping out of the engine).
   spinlock semantics outside the kernel; we use `std::mutex` and
   accept the rare cold-path contention.
 - **No condition variables.** Shutdown uses `CFRunLoopStop` + thread
-  join; SPSC backpressure is the drop-oldest rule from §3.6 step 5.
+  join; SPSC backpressure is the drop-oldest rule from §3.7 step 5.
 
 ### 6.4 Frame-phase positioning
 
@@ -909,7 +911,7 @@ each frame. When the peer plugin reloads at phase 8:
 
 No event is lost: events arriving during the swap window land in the
 SPSC ring, the I/O thread keeps writing, and Q reads them once it
-resumes. The SPSC ring sizing (4096 slots, §3.4) tolerates a swap
+resumes. The SPSC ring sizing (4096 slots, §3.5) tolerates a swap
 window up to ≈ 41 ms of 100-event-per-ms-burst storm — well past the
 0.40 ms budget for phase 8 reload (`perf-budget.md` reload-frame).
 
@@ -929,7 +931,7 @@ memory which is being unmapped. The replacement protocol is
 documented in SPEC §8.3 step 3 (drain) + step 4.3 (resume); this
 design contributes the two specific aggregate-side hooks:
 
-- **§3.10 capture hook** — exposed to the loader for snapshotting the
+- **§3.11 capture hook** — exposed to the loader for snapshotting the
   subscription list into `WatcherSurvival`.
 - **Subscriber re-validation contract** — peer plugins re-acquire the
   *new* `FileWatcher` handle from the platform API after register,
@@ -1024,7 +1026,7 @@ arena. This design partitions:
 | `CanonicalPath` interner                 | 256 KiB        | Bounded; shared with the roots and with in-flight events. Old entries reclaimed when no live event references them.                            |
 | BLAKE3 dedup LRU                         | 64 KiB         | 1024 `(canonical_path × 32 B hash)` slots; ring-replacement, no growth.                                                                      |
 | Rename-reassembly buffer                 | 16 KiB         | At most ~64 inflight half-renames; bounded by debounce window × event rate.                                                                   |
-| `snapshot_pool_` (`roots_snapshot_t` ×3) | ~2 KiB         | 1 active + 2 quarantined snapshots (§3.x). Each ~512 B (14 × `root_entry_t` ≈ 448 B, rounded). Absorbed within existing 256 KiB headroom.    |
+| `snapshot_pool_` (`roots_snapshot_t` ×3) | ~2 KiB         | 1 active + 2 quarantined snapshots (§3.3). Each ~512 B (14 × `RootEntry` ≈ 448 B, rounded). Absorbed within existing 256 KiB headroom.        |
 
 Arithmetic: 4 + 8 + 3584 + 256 + 64 + 16 + 2 = 3934 KiB ≈ 3.84 MiB,
 leaving ~254 KiB headroom within the 4 MiB sub-arena cell. The
@@ -1036,7 +1038,7 @@ arena budget amendment. The ring count was reduced from 16 to 14
 (ring row previously read "16 × 256 KiB = 4 MiB exactly", leaving
 zero room for interner + dedup + reassembly + pimpl overhead).
 14 tokens covers all MVP consumer root-counts with margin (editor
-content tree ≤ 10 roots + engine-asset trees ≤ 4; see §3.4 step 5).
+content tree ≤ 10 roots + engine-asset trees ≤ 4; see §3.5 step 5).
 The overflow behaviour for each sub-component is uniform: hitting the
 bound returns `IoFailure { OsCode { ENOBUFS } }` with prefix
 `"out-of-budget"` per `perf-budget.md` Allocator Rule #2 (strict
@@ -1066,7 +1068,7 @@ priority and does not contend on shared locks.
 If a future SPEC change wants to bound BLAKE3 work per second, the
 knob is `kFSEventStreamCreateFlagFileEvents` (which fires more
 events but at the file granularity we want) plus a per-second
-rate-limiter on hash work in §3.6 step 3. That knob is not exposed
+rate-limiter on hash work in §3.7 step 3. That knob is not exposed
 in MVP; SPEC §12 holds its trigger.
 
 ## 10. Failure modes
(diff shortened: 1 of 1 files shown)


specs/platform/file-watcher-design.md
```diff
@@ -699,14 +728,20 @@ dispatch we're keeping out of the engine).
...
-  `roots_` to walk for path-prefix matching during step 2 of §3.6 —
-  but the read is rare in practice (only after a recreate, before the
-  steady state starts hitting cached prefix entries; an LRU cache of
-  the last-prefix-match keeps the per-event lookup amortised
-  lock-free in the steady state).
+- **One `std::mutex` for cold-path `roots_` mutation plus shadow-copy
+  protocol for the I/O thread.** The FSEvents callback reads a
+  lock-free snapshot pointer (`std::atomic<roots_snapshot_t*>`) —
+  it never takes the mutex. The main thread (in `watch` / `unwatch`)
+  takes the mutex, builds a new `roots_snapshot_t` in the sub-arena,
```
cjhowe-us: severity:HIGH location:specs/platform/file-watcher-design.md:§6.3 problem:§6.3 introduces `std::atomic<roots_snapshot_t*>` as the lock-free snapshot pointer but never defines `roots_snapshot_t`, never places it in the §3.1 aggregate composition diagram, and never documents its layout in the sub-arena. The callback-side `atomic_load` memory order is also unspecified — the text documents `atomic_store(..., memory_order_release)` on the write side but is silent on whether the callback uses `memory_order_acquire`. An implementer cannot know (a) what fields `roots_snapshot_t` contains, (b) where in the sub-arena it is allocated, (c) how it relates to `eastl::vector<RootEntry> roots_` in §3.1, or (d) which memory order to use for the load. The one-frame quarantine retirement rule relies on the I/O thread being the sole reader — correct — but the full read-ordering discipline must be explicit to be implementable. fix:Add `roots_snapshot_t` to the §3.1 composition tree (e.g. `std::atomic<roots_snapshot_t*> snapshot_` alongside `roots_`). Add a minimal definition: `roots_snapshot_t` is an arena-backed POD holding a copy of the `roots_` entries (size bounded by the 14-root cap, allocated from the §9.2 sub-arena). Add to §6.3: "The callback loads the snapshot with `memory_order_acquire` to pair with the main thread's `memory_order_release` store, establishing happens-before for the snapshot's content."

specs/platform/file-watcher-design.md
```diff
@@ -549,7 +578,7 @@ using FileEvent = eastl::variant<
 
 class FileWatcher {
 public:
-    [[nodiscard]] static auto create() noexcept -> Result<FileWatcher>;
+    [[nodiscard]] static auto create(Clock& clock) noexcept -> Result<FileWatcher>;
 
     FileWatcher(FileWatcher&&) noexcept;
```
cjhowe-us, line 703: severity:MED location:specs/platform/file-watcher-design.md:§4 (public surface) problem:`create(Clock& clock)` is now the declared signature in the §4 stub (correctly fixed by this PR), but the §4 surface invariant 2 "closed list of headers" still reads `<EASTL/variant.h>`, `<EASTL/span.h>`, `<cstdint>`, `<glibre/error.hpp>`, and `<glibre/platform/canonical_path.hpp>` — with no entry for a Clock header. Since `Clock` is now a parameter type, the implementation header must include `<glibre/platform/clock.hpp>` (or the equivalent forward-declaration header per SPEC §5.9). The closed-list prose is a seam contract: implementers and plugin authors copy it verbatim as their include list. An incomplete list causes a compile error on the first consumer. fix:Add `<glibre/platform/clock.hpp>` (or the minimal forward-declaration header for `Clock`) to the §4 surface invariant 2 closed list. Verify the §5.8 section of SPEC.md has the same addition once the BLOCKING IMPLEMENTATION stub amendment lands.

specs/platform/file-watcher-design.md
```diff
@@ -1081,13 +1123,18 @@ maps directly to SPEC §11 acceptance criteria #356 / #357.
     `NotFound`.
 14. **`watch_returns_unsupported_on_file_root`** — supply a regular
     file; assert `Unsupported`.
-15. **`watch_returns_permission_denied_on_sandbox_blocked_root`** —
-    test fixture installs a sandbox profile that denies a known
-    path; assert `PermissionDenied`.
+15. **`watch_returns_permission_denied_on_eperm_from_stat`** —
+    mock the internal `stat()` / `open()` seam via the
+    dependency-injected `fileio_ops_t` to return `EPERM` for the
```
cjhowe-us: severity:MED location:specs/platform/file-watcher-design.md:§11 test 15 problem:Test 15 now says "mock the internal `stat()` / `open()` seam via the dependency-injected `fileio_ops_t`" — but `fileio_ops_t` is not defined anywhere in this design document, does not appear in the §3.1 aggregate composition diagram, and is not documented in file-io-design.md either (grep returns nothing). The seam quality principle requires that any dependency-injection interface crossing the watcher aggregate boundary be named and documented. Without a definition, an implementer cannot know: (a) which methods `fileio_ops_t` exposes, (b) where it is declared (same header as FileWatcher? internal only?), (c) whether it is injected into `create()` alongside `Clock&` or via a separate mechanism. This is a SRP concern: if `fileio_ops_t` is FileIo's internal type, coupling the FileWatcher test to it would violate the §1 refusal "FileIo blocking/async I/O — sibling #721". fix:Either (a) define `fileio_ops_t` (or a similarly named `WatcherPosixOps`) as a test-only internal seam in the design's §3 or §11 with a minimal signature (`stat`, `open`, `read`), or (b) rename it to a design-local type (e.g. `detail::posix_ops_t`) and add a one-paragraph definition in §3.4 (the step that calls `stat` for directory-ness check). If it is borrowed from FileIo, add a cross-reference and justify the coupling against §1's SRP boundary.

specs/platform/file-watcher-design.md
```diff
@@ -211,21 +215,29 @@ engine-installed assets). Steps:
    `watcher::detail::io_thread_main(FileWatcher::Impl*)`. Inside it:
```
cjhowe-us, line 307: severity:LOW location:specs/platform/file-watcher-design.md:§3.3 heading problem:The §3.3 section heading still reads `### 3.3 \`FileWatcher::create\`` (without the `Clock&` parameter), which is now inconsistent with the updated §4 stub, §1 prose, §5 cold-path, and §10.1 table — all of which were updated to `create(Clock&)` by this PR. The heading is the first entry point for a reader navigating to this section. fix:Change the §3.3 heading to `### 3.3 \`FileWatcher::create(Clock&)\`` to match every other occurrence updated in this PR.

specs/platform/file-watcher-design.md
```diff
@@ -623,10 +652,10 @@ no allocation, no syscalls beyond the SPSC dequeue):
 **Cold path** (called rarely; allocation, syscalls, OS-level resource
 juggling permitted):
```
cjhowe-us, line 782: severity:LOW location:specs/platform/file-watcher-design.md:§5 cold-path (near line 655) problem:The §5 cold-path section was updated to show `FileWatcher::create(Clock&)` correctly. However the ABI shape paragraph in §4 (just below the code block) still cites the mangled symbol as `_ZN6glibre8platform11FileWatcher6createEv` — the no-argument form. After the `Clock&` parameter is added the mangled name changes; the old name is stale and will mislead implementers comparing `nm` output against the design. fix:Update the §4 ABI shape paragraph's example mangled name or annotate it as "approximate — verify against `nm` after the header lands".

specs/platform/file-watcher-design.md
```diff
@@ -194,7 +203,67 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
 - `WatchToken{0}` is the sentinel "invalid"; the aggregate never
   vends `0`. The first valid token is `1`.
 
-### 3.3 `FileWatcher::create`
+### 3.x `roots_snapshot_t` — lock-free read-side view for the I/O thread
```
cjhowe-us: severity:MED location:specs/platform/file-watcher-design.md:206 problem:Section heading uses `3.x` as a placeholder rather than assigning a proper section number. All cross-references throughout the document (§3.1 composition tree comment, §6.3 mutex bullet, §9.2 table row) reference `§3.x` which is opaque to navigation and cannot be resolved by a section-TOC scan. The section sits between `### 3.2` and `### 3.3`; inserting it without renumbering leaves every subsequent section (`3.3` through `3.10`) with stale headings relative to the actual sequence. recommendation:Renumber the new section to `### 3.3` and shift the current `### 3.3` through `### 3.10` to `### 3.4` through `### 3.11`. Update all cross-references from `§3.x` to `§3.3`. This is a documentation topology fix with no semantic change.

specs/platform/file-watcher-design.md
```diff
@@ -194,7 +203,67 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
...
+
+`roots_snapshot_t` is a POD struct arena-allocated from the dedicated
+`snapshot_pool_` sub-allocator (§9.2). It provides the FSEvents
+callback with a stable, lock-free view of the current subscription
+roots without the I/O thread ever taking `cold_mutex_`.
+
+```cpp
+// Internal to engine/platform/src/watcher/; never crosses the public ABI.
+struct roots_snapshot_t {
+    eastl::array<root_entry_t, 14> entries;  // 14-root bound matching §9.2 ring count
```
cjhowe-us: severity:MED location:specs/platform/file-watcher-design.md:216 problem:`roots_snapshot_t` defines its entries array as `eastl::array<root_entry_t, 14>` using snake_case with `_t` suffix, but the §3.1 composition tree, §3.4 step 4, §3.5 step 1, and §3.6 step 2 consistently use `RootEntry` (CamelCase) for the same concept. If these are the same C++ type (which the design clearly intends), the inconsistency will cause an implementer to believe two distinct types exist, or to wonder which naming convention the actual header will use. Neither `RootEntry` nor `root_entry_t` is declared in any public stub, so the convention choice is internal — but it must be consistent within the design document. recommendation:Replace `root_entry_t` in the `roots_snapshot_t` struct snippet with `RootEntry` to match the composition tree and all prose references, OR add a clarifying parenthetical after the first use of `RootEntry` in §3.1 that says '(also written `root_entry_t` in internal code; same type)'. The former is cleaner.

specs/platform/file-watcher-design.md
```diff
@@ -194,7 +203,67 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
 - `WatchToken{0}` is the sentinel "invalid"; the aggregate never
   vends `0`. The first valid token is `1`.
 
-### 3.3 `FileWatcher::create`
+### 3.x `roots_snapshot_t` — lock-free read-side view for the I/O thread
+
+`roots_snapshot_t` is a POD struct arena-allocated from the dedicated
+`snapshot_pool_` sub-allocator (§9.2). It provides the FSEvents
```
cjhowe-us, line 214: severity:MED location:specs/platform/file-watcher-design.md:209 problem:`snapshot_pool_` is introduced in §3.x as a named sub-allocator of `FileWatcher::Impl` (holding up to 3 live `roots_snapshot_t` instances), but it does not appear in the §3.1 aggregate composition tree. The tree includes `roots_snapshot_` (the atomic pointer) but is silent on the allocator that backs it. An implementer reading §3.1 to understand the full state of `FileWatcher::Impl` will not know that a separate pool allocator field exists. §9.2 lists `snapshot_pool_` as a budget row and §3.x/§6.3 reference it by name — but neither of those sections is the normative composition definition. recommendation:Add `snapshot_pool_` to the §3.1 composition tree, e.g. `├── SnapshotPool snapshot_pool_  (arena-backed; holds ≤3 roots_snapshot_t instances; §3.x, §9.2)`. Place it adjacent to `roots_snapshot_` since the two fields are co-designed.

specs/platform/file-watcher-design.md
```diff
@@ -194,7 +203,67 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
...
+further synchronisation is needed for reading the snapshot
+members — they are plain struct fields behind the atomic pointer.
+
+The callback never dereferences a stale pointer: the quarantine
+window (two I/O-thread runloop iterations) guarantees the
+previous snapshot outlives any concurrent callback invocation.
+The `std::atomic<roots_snapshot_t*>` itself uses only `load` /
+`store` (no compare-exchange); AArch64's release / acquire pair
+compiles to a plain store + ISH barrier + plain load (the
+compiler emits `stlr` / `ldar` on Apple Silicon), which is
```
cjhowe-us: severity:LOW location:specs/platform/file-watcher-design.md:262 problem:The description of the AArch64 emit says "plain store + ISH barrier + plain load (the compiler emits `stlr` / `ldar` on Apple Silicon)". This is imprecise: `stlr` (Store-Release) is a single instruction that atomically combines the store with its release barrier — it is not a two-instruction sequence of 'plain store' + 'ISH DMB'. Similarly `ldar` (Load-Acquire) is a single instruction. The phrasing 'plain store + ISH barrier' could mislead a reader who knows AArch64 well into thinking the compiler emits a `str` + `dmb ish` pair rather than the single `stlr` instruction, or who might consult ARM's programmer's guide and find the terminology does not match. recommendation:Change the sentence to: 'AArch64's release/acquire pair compiles to `stlr` (store-release) and `ldar` (load-acquire) on Apple Silicon — single instructions each, zero-overhead on the hot callback path.' Drop the 'plain store + ISH barrier + plain load' characterization.

specs/platform/file-watcher-design.md
```diff
@@ -893,17 +1020,27 @@ arena. This design partitions:
...
+| BLAKE3 dedup LRU                         | 64 KiB         | 1024 `(canonical_path × 32 B hash)` slots; ring-replacement, no growth.                                                                      |
+| Rename-reassembly buffer                 | 16 KiB         | At most ~64 inflight half-renames; bounded by debounce window × event rate.                                                                   |
+| `snapshot_pool_` (`roots_snapshot_t` ×3) | ~2 KiB         | 1 active + 2 quarantined snapshots (§3.3). Each ~512 B (14 × `RootEntry` ≈ 448 B, rounded). Absorbed within existing 256 KiB headroom.        |
+
+Arithmetic: 4 + 8 + 3584 + 256 + 64 + 16 + 2 = 3934 KiB ≈ 3.84 MiB,
+leaving ~254 KiB headroom within the 4 MiB sub-arena cell. The
+`snapshot_pool_` addition (~2 KiB) is absorbed well within the
+existing 256 KiB headroom; no arena budget amendment is required.
+The headroom is intentional: it absorbs future overhead growth (e.g. a
+widened dedup LRU, additional per-root metadata) without requiring an
```
cjhowe-us, line 1092: severity:HIGH location:specs/platform/file-watcher-design.md:1031 problem:The arithmetic text states '4 + 8 + 3584 + 256 + 64 + 16 + 2 = 3934 KiB' and then 'leaving ~254 KiB headroom within the 4 MiB sub-arena cell.' The actual headroom is 4096 - 3934 = 162 KiB, not 254 KiB. The error was introduced by computing headroom against the pre-snapshot_pool_ total (~3.84 MiB ≈ 3932 KiB, giving ~164 KiB headroom) but the arithmetic now includes +2 for snapshot_pool_ without recalculating the headroom. An implementer relying on ~254 KiB available for future growth will overrun the 4 MiB arena cell by ~92 KiB and hit the out-of-budget error arm. The §12 [NON-BLOCKING] note also says '~256 KiB headroom' which is inconsistent with both the correct value (162 KiB) and the §9.2 body value (254 KiB). fix:At line 1032 change '~254 KiB headroom' to '~162 KiB headroom'. At §12 [NON-BLOCKING] change '~256 KiB headroom' to '~162 KiB headroom'. If 162 KiB is deemed insufficient margin, reduce snapshot_pool_ to 1 active + 1 quarantined (1 KiB) to recover 1 KiB; or accept 162 KiB as sufficient and document that.

specs/platform/file-watcher-design.md
```diff
@@ -61,15 +61,19 @@ What this aggregate explicitly refuses to own:
 - **Window / event pump** — `Pump`, `EventQueue<InputEvent>`,
   `EventQueue<WindowEvent>` (sibling #717). `FileEvent` is *not*
   funnelled through `Pump::drain`; it has its own per-token drain
-  (SPEC §5.8 `take_events`). The collapse rule is in §3.5 below.
+  (SPEC §5.8 `take_events`). The collapse rule is in §3.6 below.
 - **`FileIo` blocking / async I/O** — sibling #721. The watcher does
   no file *content* reads; the BLAKE3 dedup mentioned in SPEC §6.4
   reads bytes from the OS only inside the I/O thread, never on the
   main thread, and never through `FileIo`. The two aggregates share
   the `CanonicalPath` value object and nothing else.
```
cjhowe-us, line 69: severity:MED location:specs/platform/file-watcher-design.md:64 problem:The §1 prose says 'The collapse rule is in §3.6 below.' After the renumbering that inserted the new §3.3 roots_snapshot_t section, §3.6 is now FileWatcher::unwatch() — the debounce/dedup pipeline (which owns the collapse rule for overlapping events) is in §3.7. A reader following this cross-reference arrives at the wrong section. fix:Change '§3.6 below' to '§3.7 below' to match the renumbered debounce+dedup section.

specs/platform/file-watcher-design.md
```diff
@@ -132,7 +136,7 @@ Glibre-native requirements added beyond harmonius:
   exposes the in-memory subscription list as `eastl::span<const
```
cjhowe-us, line 136: severity:MED location:specs/platform/file-watcher-design.md:128 problem:The §2 native-requirements bullet says 'SPSC, not MPSC, because the rename-reassembly step §3.6 requires the I/O thread to be the sole writer per token.' After the renumbering, §3.6 is FileWatcher::unwatch(). The rename-reassembly logic lives in §3.7 (debounce + dedup, step 4). A reader verifying the SPSC single-writer rationale is sent to the wrong section. fix:Change '§3.6 requires the I/O thread' to '§3.7 requires the I/O thread'.

specs/platform/file-watcher-design.md
```diff
@@ -153,7 +157,14 @@ FileWatcher  (aggregate root, owned by platform)
...
 ├── CFRunLoopRef             runloop_                (the io thread's CFRunLoop, opaque)
 ├── FSEventStreamRef         stream_                 (one stream per FileWatcher; multi-root)
-├── eastl::vector<RootEntry> roots_                  (token → canonical root + flags + ring index)
+├── eastl::vector<RootEntry> roots_                  (main-thread authoritative subscription list;
+│                                                    guarded by cold_mutex_)
+├── std::atomic<roots_snapshot_t*> roots_snapshot_  (lock-free read pointer for the I/O thread;
+│                                                    written with memory_order_release on mutation,
+│                                                    read with memory_order_acquire in FSEvents
+│                                                    callback — see §3.3 and §6.3)
+├── SnapshotPool             snapshot_pool_          (arena-backed pool: 1 active + 2 quarantined
```
cjhowe-us, line 169: severity:MED location:specs/platform/file-watcher-design.md:161 problem:The §3.1 aggregate composition tree shows 'roots_' annotated with 'guarded by cold_mutex_' but cold_mutex_ itself has no line in the composition tree. An implementer reading §3.1 to understand the full state of FileWatcher::Impl will see a mutex name referenced in a comment without being able to identify its type or position in the struct layout. §6.3 does name it ('One std::mutex (cold_mutex_)…') but §3.1 is the canonical inventory. The type (std::mutex) and a brief note are needed here so §3.1 is self-contained. fix:Add a line to the §3.1 tree between 'roots_' and 'roots_snapshot_' — e.g.: '├── std::mutex               cold_mutex_             (guards roots_ + snapshot_pool_ writes; never held on the hot callback path)'.

specs/platform/file-watcher-design.md
```diff
@@ -1162,8 +1314,23 @@ story. Neither this design nor the spike issue closes those stories.
 
 ## 12. Open questions
 
+- `[BLOCKING IMPLEMENTATION]` **SPEC §5.8 + §10.3.3 amendment — `FileWatcher::create()` Clock injection.**
+  Must file `[SPIKE] amend-platform-spec-filewatcher-clock-injection` issue parented to #714,
+  scope: replace `create()` with `create(Clock&)` in both SPEC §5.8 and §10.3.3 sections. Two
+  concrete consumers: editor hot-reload coordinator, shipping runtime asset reload. Issue number
+  to be backfilled into this entry once filed; the spike must close before the first plan PR
+  consuming this design.
```
cjhowe-us, line 1378: severity:MED location:specs/platform/file-watcher-design.md:§12 problem:The §12 [BLOCKING IMPLEMENTATION] entry instructs 'Must file [SPIKE] amend-platform-spec-filewatcher-clock-injection issue parented to #714' with 'Issue number to be backfilled into this entry once filed.' As of this review the spike has not been filed (gh issue list search returns no matching issue). A BLOCKING gate with no corresponding issue number is nominal — nothing in the project tracker prevents the first plan PR from proceeding without the SPEC amendment. fix:File the spike issue against #714 before this PR merges, then backfill the issue number into the §12 entry. The spike deliverable is a one-line change to SPEC §5.8 and a matching row update in §10.3.3. The filing should use the spike.yml template per AGENTS.md.

specs/platform/file-watcher-design.md
```diff
@@ -1162,8 +1314,23 @@ story. Neither this design nor the spike issue closes those stories.
...
+
+- `[NON-BLOCKING]` **Ring count set to 14 (reduced from 16).**
+  The §9.2 table was updated to 14 tokens × 256 KiB = 3584 KiB,
+  leaving ~256 KiB headroom within the 4 MiB cell. The headroom
+  protects against future overhead growth (e.g. dedup LRU widening,
+  additional per-root metadata). If a future consumer requires more
+  than 14 simultaneous roots, the §9.2 arithmetic must be re-evaluated
+  and the arena budget amended.
+
```
cjhowe-us, line 1391: severity:LOW location:specs/platform/file-watcher-design.md:§12 problem:The §12 [NON-BLOCKING] ring-count note says 'leaving ~256 KiB headroom within the 4 MiB cell' but the §9.2 arithmetic body (which now includes snapshot_pool_ at +2 KiB) says '~254 KiB headroom' and the correct value is ~162 KiB (4096 - 3934). Having two different wrong headroom figures in the same section makes the spec self-contradictory and cannot both be correct. fix:Update '~256 KiB headroom' in the §12 [NON-BLOCKING] entry to '~162 KiB headroom', consistent with the corrected §9.2 arithmetic (per the HIGH finding on line 1031).

cjhowe-us (commented):

findings:
  HIGH:1
  MED:3
  LOW:1
carryover:0

notes:The PR fulfils its declared scope: all 8 round-1 #827 findings (2H 4M 2L) are verifiably present in commit 8d71258 (kFSEventStreamCreateFlagNoDefer removed with correct rationale; deferred stream creation; binary_semaphore runloop readiness; shadow-copy snapshot protocol; uint64_t overflow counter; 14-ring arithmetic; test-15 fileio_ops_t DI mock); all 5 round-2 #899 findings (1H 2M 2L) are addressed in commit 9787c5d (roots_snapshot_t defined with POD struct, placed in §3.1, acquire load specified, snapshot_pool_ budget row in §9.2, clock.hpp in closed header list, fileio_ops_t scoped to test-local via create_for_test, ABI mangled symbol updated); all 5 round-3 #899 findings (0H 4M 1L) are addressed in commit 20dcd3c (§3.x → §3.3 renumbering, RootEntry uniform, snapshot_pool_ in §3.1 tree, stlr/ldar corrected, §12 BLOCKING spike entry added). Section numbering §3.3–§3.11 is self-consistent and the majority of cross-references were updated correctly. RootEntry naming is uniform. Clock& propagation is consistent across §1, §3.4, §4, §5, §10.1, §12. The stlr/ldar characterisation is now accurate. EASTL discipline is respected throughout (std:: retained only for thread, atomic, mutex per PHILOSOPHY §11). CC subjects on all 3 commits follow docs(platform): convention correctly for spec-only changes. Issue refs #719, #827, #899 all resolve.

One HIGH finding newly introduced in commit 20dcd3c: the §9.2 arithmetic correctly sums 4 + 8 + 3584 + 256 + 64 + 16 + 2 = 3934 KiB but the stated headroom is ~254 KiB rather than the correct 4096 − 3934 = 162 KiB. This error propagates to the §12 [NON-BLOCKING] note which independently states ~256 KiB. An implementer reading the spec to size future sub-arena additions will be misled by ~92 KiB.

Three MED findings: (1) two stale §3.6 cross-references in §1 and §2 (the debounce/collapse rule and the rename-reassembly SPSC rationale) that point to unwatch() rather than §3.7 Debounce+dedup after the renumbering; (2) cold_mutex_ is referenced in the §3.1 tree annotation for roots_ but has no own entry in the composition tree, leaving its type and position implicit. One LOW finding: §12 [NON-BLOCKING] headroom figure ('~256 KiB') contradicts §9.2 body ('~254 KiB') and both are wrong. The arithmetic finding (HIGH) blocks merge; the two stale §3.6 refs (MED) also block since they will navigate readers to the wrong section during implementation. The cold_mutex_ omission (MED) and the §12 figure inconsistency (LOW) are editorial but clean.

cjhowe-us (commented):

findings:
  HIGH:1
  MED:3
  LOW:1
carryover:0

notes:The PR fulfils its declared scope: all 8 round-1 #827 findings (2H 4M 2L) are verifiably present in commit 8d71258 (kFSEventStreamCreateFlagNoDefer removed with correct rationale; deferred stream creation; binary_semaphore runloop readiness; shadow-copy snapshot protocol; uint64_t overflow counter; 14-ring arithmetic; test-15 fileio_ops_t DI mock); all 5 round-2 #899 findings (1H 2M 2L) are addressed in commit 9787c5d (roots_snapshot_t defined with POD struct, placed in §3.1, acquire load specified, snapshot_pool_ budget row in §9.2, clock.hpp in closed header list, fileio_ops_t scoped to test-local via create_for_test, ABI mangled symbol updated); all 5 round-3 #899 findings (0H 4M 1L) are addressed in commit 20dcd3c (§3.x→§3.3 renumbering, RootEntry uniform, snapshot_pool_ in §3.1 tree, stlr/ldar corrected, §12 BLOCKING spike entry added). Section numbering §3.3–§3.11 is self-consistent and the majority of cross-references were updated correctly. RootEntry naming is uniform. Clock& propagation is consistent across §1, §3.4, §4, §5, §10.1, §12. The stlr/ldar characterisation is now accurate. EASTL discipline is respected (std:: retained only for thread, atomic, mutex per PHILOSOPHY §11). CC subjects on all 3 commits follow docs(platform): convention correctly for spec-only changes. Issue refs #719, #827, #899 all resolve.

One HIGH finding newly introduced in commit 20dcd3c: the §9.2 arithmetic correctly sums 4 + 8 + 3584 + 256 + 64 + 16 + 2 = 3934 KiB but the stated headroom is ~254 KiB rather than the correct 4096 - 3934 = 162 KiB. This error propagates to the §12 [NON-BLOCKING] note which says ~256 KiB. An implementer reading the spec to size future sub-arena additions will be misled by ~92 KiB.

Three MED findings: (1)(2) two stale §3.6 cross-references in §1 and §2 — the event-pump collapse-rule ref and the SPSC rename-reassembly rationale ref — both point at unwatch() rather than at §3.7 Debounce+dedup after the renumbering; (3) cold_mutex_ is referenced in the §3.1 tree annotation for roots_ but has no own entry in the composition diagram, leaving its type implicit. One LOW finding: §12 [NON-BLOCKING] headroom figure contradicts §9.2 body and both are wrong.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6aa06a2 followup:#899

Corrected §9.2 arithmetic: 4096 − 3934 = 162 KiB, replaced "~254 KiB" → "~162 KiB" in body text.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6aa06a2 followup:#899

Fixed §1 line 64 cross-ref: "collapse rule is in §3.6 below" → "collapse rule is in §3.7 below".

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6aa06a2 followup:#899

Fixed §2 line 128 cross-ref: "rename-reassembly step §3.6" → "rename-reassembly step §3.7 step 4".

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6aa06a2 followup:#899

Added std::mutex cold_mutex_ as an explicit field in the §3.1 composition tree, immediately after roots_ (which it guards), with annotation: "// guards roots_ on the cold path; never taken on the FSEvents callback (§6.3 shadow-copy snapshot)".

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6aa06a2 followup:#899

Added STATUS note to the §12 [BLOCKING IMPLEMENTATION] entry: acknowledges spike not yet filed as of #899, names the orchestrator / first plan PR author as responsible, and states plan PRs cannot land until the spike closes. Did not file the issue itself (out of scope for impl-respond per dispatch instructions).

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6aa06a2 followup:#899

Fixed §12 [NON-BLOCKING] headroom figure: "~256 KiB headroom" → "~162 KiB headroom", consistent with corrected §9.2 arithmetic.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:NOOP reasoning:snapshot_pool_ was added to the §3.1 composition tree in commit 20dcd3c (R3 followup for #827). Verified present at line 166 of current spec: 'SnapshotPool snapshot_pool_ (arena-backed pool: 1 active + 2 quarantined snapshots, ~1.5 KiB total per §9.2)'. No further action required.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:NOOP reasoning:stlr/ldar wording was corrected in commit 20dcd3c (R3 followup for #827). Verified present at line 264 of current spec: 'compiles to stlr (store-release) and ldar (load-acquire)'. No further action required.

specs/platform/file-watcher-design.md
```diff
@@ -283,13 +365,23 @@ applies to `watch` and `unwatch`). Steps:
        /*context=*/   &ctx,
```
cjhowe-us: severity:HIGH location:specs/platform/file-watcher-design.md:360 problem:The recreate block dispatched via `CFRunLoopPerformBlock` in §3.5 step 5 calls `make_cfarray_from_roots_(roots_)` on the I/O thread without holding `cold_mutex_`. The shadow-copy protocol introduced by this PR (§3.3 / §6.3) explicitly protects the FSEvents callback from ever touching `cold_mutex_`. However the recreate block is not the callback — it is a separately-dispatched I/O-thread closure. The main thread can take `cold_mutex_`, mutate `roots_` (via a second rapid watch/unwatch call), release the mutex, and dispatch another block — all before the first pending block has executed. When that first block finally runs, it reads `roots_` without any synchronization, racing the main thread and producing a CFArrayRef built from a stale or partially-mutated root set. This is a data race on `roots_` on the cold recreate path; the shadow-copy only protects the callback hot path. fix:In §3.5 step 5, the recreate block should capture the path array at dispatch time rather than reading `roots_` at execution time. The main thread is already inside `cold_mutex_` when it dispatches the block; it should call `make_cfarray_from_roots_(roots_)` there (before releasing the mutex), capture the resulting `CFArrayRef` as a block variable, and pass it to `FSEventStreamCreate` when the block executes on the I/O thread. Explicitly add a note to §6.3 clarifying that the recreate block reads a pre-captured snapshot, not live `roots_`.

specs/platform/file-watcher-design.md
```diff
@@ -194,7 +208,67 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
...
+
+**Arena backing.** `snapshot_pool_` holds at most 3 live
+`roots_snapshot_t` instances simultaneously (1 active + 2
+quarantined for a 1-frame retire window). Each instance is
+approximately 14 × `RootEntry` size ≈ 14 × ~32 B = 448 B,
+rounded up to 512 B per instance. Pool budget: 3 × 512 B =
+~1.5 KiB — see §9.2 for the table row.
+
+**Write side (main thread, inside `cold_mutex_`).** After any
+`watch` / `unwatch` that mutates `roots_`:
```
cjhowe-us, line 256: severity:HIGH location:specs/platform/file-watcher-design.md:229 problem:`roots_snapshot_t` is defined in §3.3 as `eastl::array<RootEntry, 14>` with the snapshot size math given as `14 × ~32 B = 448 B`, rounded to 512 B per instance, pool budget ~1.5 KiB. But §9.2 'roots_ capacity' row says `14 RootEntry × ~512 B each = 8 KiB`, and the 'snapshot_pool_' row says 'Each ~512 B (14 × RootEntry ≈ 448 B, rounded)'. These two claims cannot both be true: if `RootEntry ≈ 512 B` (as implied by the roots_ row, which holds a full canonical root path string), then a `roots_snapshot_t` holding 14 of them is ~7 KiB per instance, not 512 B, and the pool is ~21 KiB — not ~1.5 KiB. If `RootEntry ≈ 32 B` (as implied by the snapshot math), then the roots_ capacity row is wrong (14 × 32 B = 448 B, not 8 KiB). An implementer cannot size the snapshot pool or the arena correctly from this contradiction. fix:Clarify the actual layout of `RootEntry`. If it holds an owning `CanonicalPath` (a string up to ~512 B), the snapshot either (a) must hold a smaller proxy type (e.g. `RootEntryRef { WatchToken token; uint8_t ring_index; }` ≈ 16 B) instead of a full `RootEntry`, or (b) the pool budget must be revised to ~21 KiB and §9.2 arithmetic updated accordingly. The design document must use a consistent `RootEntry` size throughout §3.3, §9.2 (both rows), and §3.1 annotations.

specs/platform/file-watcher-design.md
```diff
@@ -106,12 +110,12 @@ sections below.
...
+| **R-14.6.8** typed `FileEvent` with `FileEventKind` enum                                          | **Covered, collapsed.** `FileEvent` is a closed `eastl::variant<Created, Modified, Deleted, Renamed>` (SPEC §5.8). No `next() async` shape — glibre's frame-driven model uses non-blocking `take_events(token, span)`. §3.9 below.                                              |
+| **R-14.6.9** in-memory cache of `(canonical_path → BLAKE3 hash)`                                  | **Covered, scoped.** The LRU is internal to the I/O thread (SPEC §6.4 last paragraph) and lives in the watcher's 4 MiB sub-arena. It is never public; consumers cannot insert or query. §3.7 + §9.2 below.                                                                       |
+| **R-14.6.5** "configurable debounce interval"                                                     | **Partial — fixed default in MVP, knob deferred.** The debounce window is a build-time constant (default 100 ms, §3.7 below). Surfacing it as a runtime knob requires the editor / hot-reload coordinator spike — refused for MVP per Occam (no second consumer demands it). §12 holds the trigger. |
```
cjhowe-us, line 118: severity:MED location:specs/platform/file-watcher-design.md:113 problem:The §2 R-14.6.5 coverage row still lists `§3.4, §3.5, §3.6 below` as its coverage sites. After the §3.x→§3.3 renumber from commit 20dcd3c, §3.4 = create, §3.5 = watch, §3.6 = unwatch, §3.7 = debounce+dedup. The R-14.6.5 clause 'monitor for create / modify / delete / rename via platform-native APIs; debounce; recursive' is covered by create (§3.4), watch (§3.5), AND debounce (§3.7) — not unwatch (§3.6). The stale `§3.6` reference points a reader to unwatch, which is irrelevant to monitoring. fix:Update the R-14.6.5 coverage cell to cite `§3.4, §3.5, §3.7 below` (replacing §3.6 with §3.7).

specs/platform/file-watcher-design.md
```diff
@@ -194,7 +208,67 @@ struct WatchToken { std::uint64_t value{0}; constexpr bool operator==(const Watc
 - `WatchToken{0}` is the sentinel "invalid"; the aggregate never
```
cjhowe-us, line 208: severity:MED location:specs/platform/file-watcher-design.md:170 problem:The §3.1 composition tree annotation for `snapshot_pool_` says `~1.5 KiB total per §9.2` but the §9.2 table row says `~2 KiB`. The correct arithmetic from §3.3 is 3 × 512 B = 1536 B = 1.5 KiB. The §9.2 row rounds up to `~2 KiB` without explanation. An implementer sizing the pool from one section and verifying against the other will find a discrepancy. fix:Align both occurrences to the same value. The §3.3 derivation (3 × 512 B = ~1.5 KiB) is the load-bearing calculation; update the §9.2 table row from `~2 KiB` to `~1.5 KiB` and update the §9.2 body text accordingly. If conservative rounding to `~2 KiB` is intentional for budget headroom, add a note explaining the rounding and update §3.1 to match.

specs/platform/file-watcher-design.md
```diff
@@ -528,6 +627,7 @@ boundary this design ships. Recapitulated for self-containment:
 #include <EASTL/variant.h>
 #include <glibre/error.hpp>
 #include <glibre/platform/canonical_path.hpp>
+#include <glibre/platform/clock.hpp>  // Clock& parameter of create(); use clock_fwd.hpp if clock-design §4 uses a forward-decl header
```
cjhowe-us, line 680: severity:LOW location:specs/platform/file-watcher-design.md:638 problem:The §4 'closed list of headers' includes `<glibre/platform/clock.hpp>` with a conditional note: 'use clock_fwd.hpp if clock-design §4 uses a forward-decl header'. But clock-design §4 says normative declarations live in `glibre/platform/platform.hpp` (not in a `clock.hpp`). There is no `clock_fwd.hpp` referenced anywhere in clock-design.md. The speculative hedge about `clock_fwd.hpp` adds noise and the real answer (`platform.hpp`) is already available from the sister spec. If FileWatcher only needs a forward declaration of `Clock` (since it takes `Clock&` by reference), a forward declaration suffices and avoids pulling in the full clock header transitively. fix:Verify against clock-design.md §4 and SPEC §5. If the canonical header is `<glibre/platform/platform.hpp>`, update the closed list to use that. If a dedicated per-aggregate header (`clock.hpp`) is planned (per §12 [BLOCKING] of clock-design), add that as a conditional note referencing the clock-design spike rather than speculating about `clock_fwd.hpp`.

cjhowe-us (commented):

findings:
  HIGH:2
  MED:2
  LOW:1
carryover:0

notes:All 6 R1 findings on #899 were correctly addressed in commit 6aa06a2 (§9.2 headroom 254→162, §1/§2 §3.6→§3.7 cross-refs, cold_mutex_ in §3.1 tree, §12 STATUS note, §12 NON-BLOCKING headroom fix). The prior-round carryover is zero. Two new HIGH findings emerge from the Round-2 seam-quality lens:

HIGH-1 (line 360 / §3.5 step 5): The stream-recreate block dispatched via CFRunLoopPerformBlock reads roots_ on the I/O thread via make_cfarray_from_roots_(roots_) without holding cold_mutex_. The shadow-copy protocol introduced by this PR covers the FSEvents callback hot path, but the recreate block is a separately-dispatched closure that can race with a second watch/unwatch on the main thread. The fix — capture the CFArrayRef at dispatch time (inside cold_mutex_) and close over the captured value in the block — is narrow and in-scope.

HIGH-2 (line 229 / §3.3): The snapshot_pool_ size math uses RootEntry ≈ 32 B (14 × 32 B = 448 B → 512 B/snapshot → ~1.5 KiB pool), but the §9.2 roots_ capacity row uses 14 RootEntry × ~512 B each = 8 KiB. These are inconsistent: if RootEntry is 512 B (because it holds a full CanonicalPath string), the snapshot holds 14 × 512 B = 7 KiB per instance and the pool is ~21 KiB, not 1.5 KiB. If RootEntry is 32 B, the roots_ capacity row is wrong. The design must either (a) define a separate smaller proxy type for the snapshot, or (b) revise the pool budget and §9.2 arithmetic.

Two MED findings: the §2 R-14.6.5 row still lists §3.6 (unwatch) as a coverage site after the §3.x renumber — the correct citation is §3.7 (debounce+dedup); and snapshot_pool_ size is stated as ~1.5 KiB in §3.1/§3.3 but ~2 KiB in §9.2 (contradicting the table's own arithmetic note).

One LOW finding: the §4 closed-list comment speculates about clock_fwd.hpp, but clock-design.md §4 establishes that normative declarations live in glibre/platform/platform.hpp with no separate clock_fwd.hpp mentioned; the hedge should reference the actual sister spec outcome.

The shadow-copy snapshot mechanics (§3.3 / §6.3), the binary_semaphore runloop gate (§3.4 step 3), the NoDefer-removal rationale (§3.5 step 5 flags prose), the create_for_test seam for test 15, and the §12 gate entries are all well-formed and cohesive. The section numbering (§3.3–§3.11) and cold_mutex_ annotation are clean.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:8966e7c followup:#899

§3.5 step 5 recreate closure rewritten: CFArrayRef is now built on the main thread inside cold_mutex_ before CFRunLoopPerformBlock dispatch. The block captures pending_paths by value (retained before dispatch, released after FSEventStreamCreate). The I/O thread block never reads roots_ directly. §6.3 updated with a clarifying note.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:8966e7c followup:#899

§3.3 introduces RootProxy {eastl::string_view path (16 B, into CanonicalPath interner); uint32_t root_id (4 B); uint32_t generation (4 B)} = 24 B per proxy. roots_snapshot_t now holds eastl::array<RootProxy,14> + uint8_t count = ~344 B/instance, rounded to ~360 B. Pool: 3 × ~360 B ≈ ~1 KiB. §3.1 annotation, §3.3 struct + math, §9.2 table row + body arithmetic all updated. The path string_view points into the CanonicalPath interner (lifetime ≥ quarantine window); callback uses it only for prefix matching within its stack frame.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:8966e7c followup:#899

§2 R-14.6.5 coverage cell updated from '§3.4, §3.5, §3.6 below' to '§3.4, §3.5, §3.7 below'. §3.6 = unwatch (irrelevant to monitoring); §3.7 = debounce+dedup (the correct site).

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:8966e7c followup:#899

§9.2 table snapshot_pool_ row updated from ~2 KiB to ~1 KiB, now consistent with §3.1 annotation and §3.3 derivation. The ~1 KiB figure follows from the HIGH-2 RootProxy fix: 3 × ~360 B per instance ≈ ~1 KiB. §9.2 body text updated to match.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:8966e7c followup:#899

Verified clock-design.md §4: normative Clock declarations live in glibre/platform/platform.hpp per SPEC §5. There is no clock.hpp or clock_fwd.hpp. Updated the #include in §4 code snippet, the closed-list invariant §4 #2, and the ABI mangle note to use <glibre/platform/platform.hpp>. The speculative hedge about clock_fwd.hpp is removed.

specs/platform/file-watcher-design.md
```diff
@@ -1162,8 +1370,27 @@ story. Neither this design nor the spike issue closes those stories.
...
+  to be backfilled into this entry once filed; the spike must close before the first plan PR
+  consuming this design.
+  STATUS: spike issue not yet filed as of #899 commit. The orchestrator (or first plan PR
+  author) must file `[SPIKE] amend-platform-spec-filewatcher-clock-injection` parented to #714
+  and backfill its issue number into this entry. Plan PRs consuming this design cannot land
+  until that spike closes.
+
+- `[NON-BLOCKING]` **Ring count set to 14 (reduced from 16).**
+  The §9.2 table was updated to 14 tokens × 256 KiB = 3584 KiB,
+  leaving ~162 KiB headroom within the 4 MiB cell. The headroom
```
cjhowe-us: severity:LOW location:specs/platform/file-watcher-design.md:1386 problem:The §12 [NON-BLOCKING] entry says 'leaving ~162 KiB headroom within the 4 MiB cell' but after commit 8966e7c reduced snapshot_pool_ from ~2 KiB to ~1 KiB, the §9.2 arithmetic now reads '4 + 8 + 3584 + 256 + 64 + 16 + 1 = 3933 KiB', giving 4096 - 3933 = 163 KiB headroom, not 162 KiB. The §9.2 body prose was correctly updated to '~163 KiB' but the §12 [NON-BLOCKING] note was not updated to match. Two different headroom figures now appear in the same document. recommendation:Change '~162 KiB headroom' in the §12 [NON-BLOCKING] entry to '~163 KiB headroom', consistent with the §9.2 body text at line 1086 ('leaving ~163 KiB headroom within the 4 MiB sub-arena cell').

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0

notes:Round-3 topology + spec-alignment + polish pass against commit 8966e7c (the R2 followup for #899). All five R2 findings — HIGH-1 (recreate-closure cold_mutex_ capture), HIGH-2 (RootProxy snapshot proxy introduction), MED-1 (§2 R-14.6.5 coverage cell §3.7 ref), MED-2 (§9.2 snapshot_pool_ row aligned to ~1 KiB after RootProxy math), and LOW-1 (clock header path → platform.hpp) — are verifiably resolved. Section topology is sound: §3.3 = roots_snapshot_t; §3.4–§3.11 renumbered consistently; every surviving cross-reference checked (§3.5 step 5, §3.7 five-step pipeline, §3.10 step 4, §3.11 capture hook, §6.3 → §3.3 and §3.5, §9.4 → §3.7 step 3). No stale §3.6 references remain in the final head. §12 amendment coverage is correct: BLOCKING entry names SPEC §5.8 and §10.3.3 as the two normative surfaces to amend (verified against SPEC §5.8 line 920 and SPEC §10.3.3 lines 2494-2498), names two concrete consumers (editor hot-reload coordinator, shipping runtime asset reload path per §1 prose), and gates plan PRs on spike closure. §4 closed-header list matches clock-design.md §4 (`glibre/platform/platform.hpp`; no clock.hpp or clock_fwd.hpp). CC subjects on all 5 commits use docs(platform): correctly for spec-only changes; issue refs #719, #827, #899 all resolve; #714 parent correct. The RootProxy struct is internal-only (comment confirms this). EASTL discipline maintained throughout. One LOW finding: §12 [NON-BLOCKING] headroom figure "~162 KiB" is stale by 1 KiB after commit 8966e7c changed snapshot_pool_ from ~2 KiB to ~1 KiB — §9.2 body correctly says "~163 KiB" but §12 was not updated. This is a trivial polish fix that does not block merge.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0

notes:Round-3 topology + spec-alignment + polish pass against commit 8966e7c (the R2 followup for #899). All five R2 findings — HIGH-1 (recreate-closure cold_mutex_ capture), HIGH-2 (RootProxy snapshot proxy introduction), MED-1 (§2 R-14.6.5 coverage cell §3.7 ref), MED-2 (§9.2 snapshot_pool_ row aligned to ~1 KiB after RootProxy math), and LOW-1 (clock header path → platform.hpp) — are verifiably resolved. Section topology is sound: §3.3 = roots_snapshot_t; §3.4–§3.11 renumbered consistently; every surviving cross-reference checked (§3.5 step 5, §3.7 five-step pipeline, §3.10 step 4, §3.11 capture hook, §6.3 → §3.3 and §3.5, §9.4 → §3.7 step 3). No stale §3.6 references remain in the final head. §12 amendment coverage is correct: BLOCKING entry names SPEC §5.8 and §10.3.3 as the two normative surfaces to amend (verified against SPEC §5.8 line 920 and SPEC §10.3.3 lines 2494-2498), names two concrete consumers, and gates plan PRs on spike closure. §4 closed-header list matches clock-design.md §4 (platform.hpp). CC subjects on all 5 commits use docs(platform): correctly; issue refs #719, #827, #899 all resolve. EASTL discipline maintained. One LOW finding: §12 [NON-BLOCKING] headroom figure shows ~162 KiB but §9.2 body (updated by 8966e7c) correctly shows ~163 KiB — trivial 1 KiB carry from the R2 snapshot_pool_ size reduction. Does not block merge.

cjhowe-us (commented):

findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0

notes:Round-3 topology + spec-alignment + polish pass against commit 8966e7c (the R2 followup for #899). All five R2 findings — HIGH-1 (recreate-closure cold_mutex_ capture), HIGH-2 (RootProxy snapshot proxy introduction), MED-1 (§2 R-14.6.5 coverage cell §3.7 ref), MED-2 (§9.2 snapshot_pool_ row aligned to ~1 KiB after RootProxy math), and LOW-1 (clock header path → platform.hpp) — are verifiably resolved. Section topology is sound: §3.3 = roots_snapshot_t; §3.4–§3.11 renumbered consistently; every surviving cross-reference checked (§3.5 step 5, §3.7 five-step pipeline, §3.10 step 4, §3.11 capture hook, §6.3 → §3.3 and §3.5, §9.4 → §3.7 step 3). No stale §3.6 references remain. §12 amendment coverage correct: BLOCKING entry names SPEC §5.8 and §10.3.3 (verified against SPEC lines 920 and 2494-2498), names two concrete consumers, gates plan PRs on spike closure. §4 closed-header list matches clock-design.md §4 (platform.hpp confirmed). CC subjects on all 5 commits use docs(platform): correctly for spec-only changes; issue refs #719, #827, #899 resolve; parent #714 correct. EASTL discipline maintained. One LOW finding: §12 [NON-BLOCKING] headroom says ~162 KiB but §9.2 body says ~163 KiB — 1 KiB carry from R2 snapshot_pool_ reduction. Does not block merge.

specs/platform/file-watcher-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:aad163b followup:#899

PR #900, opened 2026-05-05
docs(platform): r1 followup for #826 — event-pump-design amendments

cjhowe-us:
## Summary

Follow-up to #826 per round-1 review (verdict 4229442136). Addresses all 9 inline findings on `specs/platform/event-pump-design.md`.

- **HIGH-1** (comment 3189617027): Added mandatory `detail::error::set_prefix(prefix::queue_full)` call before `IoFailure{OsCode{0}}` construction in §3.4 and §10.3. Added note: "TLS prefix stamp is mandatory before any `IoFailure{OsCode{0}}` where the prefix is the only discriminator. The stamp is cold-path-only."
- **HIGH-2** (comment 3189618448): Added `alignas(128)` to `head` and `tail` in `RingImpl` struct with annotation explaining Apple Silicon 128B cache line false-sharing.
- **MED-1** (comment 3189619844): Removed "non-power-of-two beyond cap" from §10.1 `Unsupported` triggers; `with_capacity` silently rounds up. Added §11.1 test #8b asserting `with_capacity(100).capacity() == 128`.
- **MED-2** (comment 3189621074): Added one-caller audit trail note in §3.6 and §12 [NON-BLOCKING] entry for `ingest_drop` two-concrete-users gate.
- **MED-3** (comment 3189622189): Fixed §6.4 dangling cross-ref: `§"Severity Conventions"` → `§"Logging/Telemetry"` with explanatory note.
- **MED-4** (comment 3189623443): Rewrote §3.3 mouse coordinate paragraph distinguishing absolute x/y (logical points) from relative dx/dy (physical pixels requiring `dpi_scale()` divide). Added §11.1 test #3b at `DpiScale != 1.0`.
- **LOW-1** (comment 3189624874): Added explicit trigger rescale formula `axis = (raw / 32767.0f) * 2.0f - 1.0f` to §3.3 and requirements table. Added §11.1 test #6b asserting `raw=0 → -1.0f`. Updated R-6.1.6 table row.
- **LOW-2** (comment 3189626239): Expanded §11.3 to explicitly name drag-drop (S1 trace step 7) and quit-synthesis (integration-test-only at MVP) E2E coverage status. Added §12 [NON-BLOCKING] for quit-synthesis gap.
- **LOW-3** (comment 3189627441): Corrected §3.3 TextInput: SDL3's `text[32]` buffer is fixed 32 bytes; the four-byte look-ahead scanner is dead code for SDL3 and is not implemented in the production path. Updated §11.1 test #7 to match.

## Addressed comments

- 3189617027 (HIGH-1)
- 3189618448 (HIGH-2)
- 3189619844 (MED-1)
- 3189621074 (MED-2)
- 3189622189 (MED-3)
- 3189623443 (MED-4)
- 3189624874 (LOW-1)
- 3189626239 (LOW-2)
- 3189627441 (LOW-3)

## Test plan

- All changes are documentation-only to `specs/platform/event-pump-design.md` — no implementation code exists yet.
- New unit tests #3b, #6b, #8b are specified in §11.1 for implementation when the platform-event-pump implementation plan executes.
- No CI failure expected (docs-only).

### DIFF
--- a/specs/platform/event-pump-design.md
+++ b/specs/platform/event-pump-design.md
@@ -89,16 +89,18 @@ The aggregate **refuses to own**:
 The SRP boundary is sharp: if the **SDL3 event vocabulary**, the
 **SDL3 → typed-variant translation table**, the **per-device
 ordering rule**, the **bounded SPSC ring protocol**, the
-**per-frame drain trigger**, or the **drop-event routing to the
-`FileWatcher` aggregate** change, this design changes. Anything
-else — input action mapping, window aggregate state, display
-pacing — is out of scope.
+**per-frame drain trigger**, or the **drop-event routing policy**
+change, this design changes. Anything else — input action mapping,
+window aggregate state, display pacing — is out of scope.
 
-Drop-event routing is listed as a sixth responsibility for full SRP
-transparency. The §12 one-caller gate (per PHILOSOPHY §Anti-patterns)
-governs whether `ingest_drop` becomes a stable cross-aggregate API;
-the SRP boundary stays unchanged whether routing goes via
-`ingest_drop` or via a future `WindowEvent::FileDropped` variant.
+Responsibility 6 (owned, in-scope): the *policy* of drop-event
+routing — that the pump forwards `SDL_EVENT_DROP_FILE` to the
+file-watcher aggregate. This policy stays owned regardless of which
+*mechanism* implements it (current: `FileWatcher::ingest_drop` direct
+call; future option: `WindowEvent::FileDropped` via the queue, gated
+on §12 [NON-BLOCKING] one-caller audit). The SRP boundary tracks the
+policy, not the mechanism — switching mechanisms does NOT trigger an
+SRP-list edit, but adding or removing the policy itself would.
 
 ## 2. Requirements coverage
 
@@ -403,7 +405,7 @@ the event itself triggers a side effect inside the pump:
 |-------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------|
 | `SDL_EVENT_GAMEPAD_ADDED` / `_REMOVED`                                  | Update SDL3's internal gamepad table by calling `SDL_OpenGamepad` / `SDL_CloseGamepad` so subsequent axis/button events resolve a valid `device` index. No glibre `InputEvent` emitted; consumers query gamepad presence at use time. | None (R-6.1.11 partial).                     |
 | `SDL_EVENT_DISPLAY_ADDED` / `_REMOVED` / `_ORIENTATION_CHANGED`         | Cause the next `Window::display()` call to re-query SDL3's display list (`Display` is a snapshot per SPEC §4.1 inv #5, not cached). Emitted as `WindowEvent::DisplayChanged { window, display }` for the focused window.              | `WindowEvent::DisplayChanged`.                |
-| `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`                                | Update the owning `Window::Impl`'s cached `(LogicalSize, PhysicalSize, DpiScale)` snapshot via `surface::detail::query_layer_metrics` *before* the event becomes visible to the engine (§6.3 cross-module note in SPEC). Also updates `Pump::Impl`-cached `DpiScale_` from the new `Window::Impl` value; the updated scale is passed to `translate_mouse` on the next `SDL_EVENT_MOUSE_MOTION` in this frame (consumed by the divide-by-scale path per §3.3). | `WindowEvent::DpiChanged`.                    |
+| `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`                                | Update the owning `Window::Impl`'s cached `(LogicalSize, PhysicalSize, DpiScale)` snapshot via `surface::detail::query_layer_metrics` *before* the event becomes visible to the engine (§6.3 cross-module note in SPEC). Also updates the Pump::Impl-cached DpiScale value from the new `Window::Impl` value; the updated scale is passed to `translate_mouse` on the next `SDL_EVENT_MOUSE_MOTION` in this frame (consumed by the divide-by-scale path per §3.3). | `WindowEvent::DpiChanged`.                    |
 | `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` / `_RESIZED`                      | Update the owning `Window::Impl`'s cached `LogicalSize` / `PhysicalSize`.                                                                                                                                                            | `WindowEvent::Resized`.                       |
 | `SDL_EVENT_WINDOW_FOCUS_GAINED` / `_LOST`                               | None at the pump (the window aggregate's focus state is the queue contents themselves).                                                                                                                                              | `WindowEvent::FocusGained` / `FocusLost`.     |
 | `SDL_EVENT_WINDOW_MINIMIZED` / `_RESTORED`                              | None.                                                                                                                                                                                                                                | `WindowEvent::Minimized` / `Restored`.        |
@@ -1325,3 +1327,13 @@ hot-reload survival path; runs against the same S1 trace with a
   `u32::max`" before the first plan PR consuming this design lands.
   Two concrete consumers: shipping runtime input pump, editor input pump.
   Owner: platform-SPEC amendment — must land before plan PRs open.
+  STATUS: The amendment is a one-line edit to replace "zero / wildly
+  oversized capacity" with "capacity overflow beyond `u32::max`" in the
+  SPEC §10.3.2 trigger column. The orchestrator (or first plan PR
+  author) must either (a) file
+  `[SPIKE] amend-platform-spec-event-pump-with-capacity-trigger`
+  parented to #714 and backfill the issue number into this entry, OR
+  (b) land the one-line SPEC edit directly in the first plan PR
+  consuming this design. Plan PRs cannot land until SPEC §10.3.2 is
+  corrected via either route. Issue: #TBD (to be filed at plan-PR
+  authoring time).
--- a/specs/platform/event-pump-design.md
+++ b/specs/platform/event-pump-design.md
@@ -403,7 +403,7 @@ the event itself triggers a side effect inside the pump:
 |-------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------|
 | `SDL_EVENT_GAMEPAD_ADDED` / `_REMOVED`                                  | Update SDL3's internal gamepad table by calling `SDL_OpenGamepad` / `SDL_CloseGamepad` so subsequent axis/button events resolve a valid `device` index. No glibre `InputEvent` emitted; consumers query gamepad presence at use time. | None (R-6.1.11 partial).                     |
 | `SDL_EVENT_DISPLAY_ADDED` / `_REMOVED` / `_ORIENTATION_CHANGED`         | Cause the next `Window::display()` call to re-query SDL3's display list (`Display` is a snapshot per SPEC §4.1 inv #5, not cached). Emitted as `WindowEvent::DisplayChanged { window, display }` for the focused window.              | `WindowEvent::DisplayChanged`.                |
-| `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`                                | Update the owning `Window::Impl`'s cached `(LogicalSize, PhysicalSize, DpiScale)` snapshot via `surface::detail::query_layer_metrics` *before* the event becomes visible to the engine (§6.3 cross-module note in SPEC).            | `WindowEvent::DpiChanged`.                    |
+| `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`                                | Update the owning `Window::Impl`'s cached `(LogicalSize, PhysicalSize, DpiScale)` snapshot via `surface::detail::query_layer_metrics` *before* the event becomes visible to the engine (§6.3 cross-module note in SPEC). Also updates `Pump::Impl`-cached `DpiScale_` from the new `Window::Impl` value; the updated scale is passed to `translate_mouse` on the next `SDL_EVENT_MOUSE_MOTION` in this frame (consumed by the divide-by-scale path per §3.3). | `WindowEvent::DpiChanged`.                    |
 | `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` / `_RESIZED`                      | Update the owning `Window::Impl`'s cached `LogicalSize` / `PhysicalSize`.                                                                                                                                                            | `WindowEvent::Resized`.                       |
 | `SDL_EVENT_WINDOW_FOCUS_GAINED` / `_LOST`                               | None at the pump (the window aggregate's focus state is the queue contents themselves).                                                                                                                                              | `WindowEvent::FocusGained` / `FocusLost`.     |
 | `SDL_EVENT_WINDOW_MINIMIZED` / `_RESTORED`                              | None.                                                                                                                                                                                                                                | `WindowEvent::Minimized` / `Restored`.        |
@@ -987,7 +987,7 @@ failure is a `glibre::Error` arm (with the platform-private
 
 | Entry point                     | Returnable arms                                  | Trigger                                         |
 |---------------------------------|--------------------------------------------------|-------------------------------------------------|
-| `EventQueue<T>::with_capacity`  | `Unsupported`                                    | zero capacity / wildly oversized capacity (> 2^24 — SPEC sub-arena cap); non-power-of-two is silently rounded up (not an error) |
+| `EventQueue<T>::with_capacity`  | `Unsupported`                                    | zero capacity / capacity overflow beyond u32::max (the sub-arena ceiling); non-power-of-two is silently rounded up (not an error) |
 | `EventQueue<T>::drain`          | total — never returns `unexpected`               | drain is infallible by design (SPSC pop)        |
 | `Pump::create`                  | `IoFailure { OsCode }`                           | `SDL_InitSubSystem(SDL_INIT_VIDEO \| SDL_INIT_GAMEPAD)` failure |
 | `Pump::drain`                   | `IoFailure { OsCode { 0 } }` with prefix `"queue-full"` (fatal); `Unsupported` (off-main-thread) | Queue full at `try_push` site (§4.2 inv #3 fatal); drain called from non-main thread (§6.4) |
@@ -1123,7 +1123,7 @@ the platform user-story issues).
     whose relative `dx/dy` is 200 physical pixels; assert the emitted
     `MouseMove::dx/dy` is `100.0f` (logical points). Confirms the
     divide-by-dpi_scale invariant documented in §3.3 and that the
-    `DpiScale` parameter is the operative divisor. (§R-6.1.4.)
+    `DpiScale` parameter is the operative divisor. (R-6.1.4)
 4. **`event_pump_drops_unknown_sdl_tag`** — feed an
    `SDL_EVENT_USER` and an out-of-MVP-range tag; assert the
    queues remain empty and a `debug`-level log line was emitted.
--- a/specs/platform/event-pump-design.md
+++ b/specs/platform/event-pump-design.md
@@ -85,10 +85,17 @@ The aggregate **refuses to own**:
 
 The SRP boundary is sharp: if the **SDL3 event vocabulary**, the
 **SDL3 → typed-variant translation table**, the **per-device
-ordering rule**, the **bounded SPSC ring protocol**, or the
-**per-frame drain trigger** change, this design changes. Anything
-else — input action mapping, window aggregate state, file watching,
-display pacing — is out of scope.
+ordering rule**, the **bounded SPSC ring protocol**, the
+**per-frame drain trigger**, or the **drop-event routing to the
+`FileWatcher` aggregate** change, this design changes. Anything
+else — input action mapping, window aggregate state, display
+pacing — is out of scope.
+
+Drop-event routing is listed as a sixth responsibility for full SRP
+transparency. The §12 one-caller gate (per PHILOSOPHY §Anti-patterns)
+governs whether `ingest_drop` becomes a stable cross-aggregate API;
+the SRP boundary stays unchanged whether routing goes via
+`ingest_drop` or via a future `WindowEvent::FileDropped` variant.
 
 ## 2. Requirements coverage
 
@@ -246,7 +253,7 @@ namespace glibre::platform::event::detail {
 // Returns nullopt for unknown / consumed-internally tags.
 [[nodiscard]] auto translate_keyboard(const SDL_Event&) noexcept
     -> eastl::optional<InputEvent>;
-[[nodiscard]] auto translate_mouse(const SDL_Event&) noexcept
+[[nodiscard]] auto translate_mouse(const SDL_Event&, DpiScale) noexcept
     -> eastl::optional<InputEvent>;
 [[nodiscard]] auto translate_wheel(const SDL_Event&) noexcept
     -> eastl::optional<InputEvent>;
@@ -283,8 +290,11 @@ Key normalisation rules realised in the table:
   is set — window-relative so coordinate (0, 0) is top-left of the
   client area regardless of compositor placement. Relative `dx/dy`
   from `SDL_SetRelativeMouseMode` are physical pixel deltas on macOS
-  and must be divided by `window.dpi_scale()` (from
-  window-surface-design.md) before being placed in `MouseMove::dx/dy`.
+  and must be divided by the current `DpiScale` before being placed
+  in `MouseMove::dx/dy`. `Pump::Impl` caches the current `DpiScale`
+  (initialised `1.0f` at construction; updated in §3.5 on
+  `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`). Each `translate_mouse`
+  call passes the cached `DpiScale` by value as its second argument.
   Documented invariant: every `MouseMove` payload reports x, y, dx, dy
   in logical points. A unit test at `DpiScale != 1.0` asserts the
   divide-by-scale path (§11.1 test #3b).
@@ -367,7 +377,10 @@ allocation from the platform sub-arena tagged
 Full-on-`try_push` returns `false`; the pump:
 1. Stamps the TLS prefix slot via
    `detail::error::set_prefix(prefix::queue_full)`.
-2. Constructs `unexpected(IoFailure { OsCode{0} })`.
+2. Emits an `error`-severity log with the prefix `"queue-full"` and
+   the queue family name. The stamp must precede this log call so
+   `log_error` reads the stamped prefix (not `prefix::none`).
+3. Returns `unexpected(IoFailure { OsCode{0} })`.
 
 The TLS prefix stamp is mandatory before any `IoFailure{OsCode{0}}`
 construction where the prefix is the only discriminator; without the
@@ -691,9 +704,8 @@ is a contract violation. Detection in MVP:
   abort in shipping builds because a misbehaving plugin should
   surface as an error, not crash the engine.
 
-This is the canonical detection pattern from `error-model.md`
-§"Logging/Telemetry" (see hot-reload refusal severity discussion
-under that heading) — assert in debug, demote in shipping.
+This is the canonical detection pattern consistent with SPEC §4.1
+inv #2 and §4.2 inv #1 — assert in debug, demote in shipping.
 
 ### 6.5 Hot-reload barrier interaction
 
@@ -1007,10 +1019,12 @@ This is the same recovery contract as SPEC §10.3.2's
 
 When `EventQueue<T>::try_push` returns false during `Pump::drain`:
 
-1. The pump emits a `error`-severity log with the prefix
-   `"queue-full"` and the queue family name.
-2. The pump calls `detail::error::set_prefix(prefix::queue_full)`
-   immediately before constructing the failure value.
+1. The pump calls `detail::error::set_prefix(prefix::queue_full)`.
+   The stamp must precede the log so `log_error` reads the stamped
+   prefix. Reorder is cold-path-only — no impact on hot-path performance.
+2. The pump emits an `error`-severity log with the prefix
+   `"queue-full"` and the queue family name (reads the stamped
+   prefix set in step 1).
 3. The pump returns `unexpected(IoFailure { OsCode { 0 } })`
    from the current `Pump::drain` call. The TLS prefix stamp is the
    load-bearing discriminator: callers that receive `IoFailure` from
@@ -1092,11 +1106,12 @@ the platform user-story issues).
    `(LogicalSize, PhysicalSize, DpiScale)` snapshot on
    `Window::Impl` is updated *before* the typed event is enqueued
    (§3.5 / §6.3 SPEC cross-module note).
-3b. **`event_pump_mouse_relative_motion_divided_by_dpi_scale`** — feed
-    a `SDL_MouseMotionEvent` with relative `dx/dy` of 200 physical
-    pixels at `DpiScale = 2.0`; assert the emitted
+3b. **`event_pump_mouse_relative_motion_divided_by_dpi_scale`** — call
+    `translate_mouse(ev, DpiScale{2.0f})` with a `SDL_MouseMotionEvent`
+    whose relative `dx/dy` is 200 physical pixels; assert the emitted
     `MouseMove::dx/dy` is `100.0f` (logical points). Confirms the
-    divide-by-dpi_scale invariant documented in §3.3. (§R-6.1.4.)
+    divide-by-dpi_scale invariant documented in §3.3 and that the
+    `DpiScale` parameter is the operative divisor. (§R-6.1.4.)
 4. **`event_pump_drops_unknown_sdl_tag`** — feed an
    `SDL_EVENT_USER` and an out-of-MVP-range tag; assert the
    queues remain empty and a `debug`-level log line was emitted.
@@ -1221,7 +1236,7 @@ hot-reload survival path; runs against the same S1 trace with a
 
 | Concern                                    | Test                                                                                  |
 |--------------------------------------------|---------------------------------------------------------------------------------------|
-| SDL3 → typed translation correctness       | unit #1, #2, #3, #6, #7                                                               |
+| SDL3 → typed translation correctness       | unit #1, #2, #3, #3b, #6, #7                                                          |
 | Mouse dx/dy DPI divide (§3.3, R-6.1.4)    | unit #3b                                                                              |
 | Trigger released-state formula (§3.3)      | unit #6b                                                                              |
 | with_capacity round-up (§3.4, §10.1)      | unit #8b                                                                              |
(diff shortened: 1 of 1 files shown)


specs/platform/event-pump-design.md
```diff
@@ -277,24 +277,42 @@ Key normalisation rules realised in the table:
...
   preference per `SDL_MouseWheelEvent::direction`.
-- **Text input UTF-8.** SDL3 delivers UTF-8 already. The pump copies
-  up to 32 bytes into `input::TextInput::utf8` and sets
-  `length`. If SDL3 delivers > 32 bytes, the pump emits N successive
-  `TextInput` events covering complete UTF-8 codepoints (split is
-  always on a codepoint boundary, never mid-byte; the codepoint
-  scanner is a four-byte look-ahead).
+- **Text input UTF-8.** SDL3 delivers UTF-8 already via
+  `SDL_TextInputEvent::text[32]` — a fixed 32-byte null-terminated
+  buffer. SDL3 truncates at that boundary before the event reaches
```
cjhowe-us, line 311: severity:HIGH location:specs/platform/event-pump-design.md:296 problem:The R1 fix documents that relative `dx/dy` from `SDL_SetRelativeMouseMode` are physical pixel deltas that "must be divided by `window.dpi_scale()`" before being placed in `MouseMove::dx/dy`. However, `translate_mouse` is declared as a pure free function taking only `const SDL_Event&` — it has no access to a `Window` or its cached DPI scale. Either (a) the function must receive `DpiScale` as a second parameter (making the seam explicit and the function signature wrong as written), or (b) the pump must hold a cached `DpiScale` field (updated via §3.5 `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`) and pass it into `translate_mouse` at call time. Option (b) is architecturally motivated: §3.5 already shows the pump updating `Window::Impl` cached metrics on `DpiChanged` before enqueueing the event. If the pump instead caches the scale itself, `translate_mouse` becomes `translate_mouse(const SDL_Event&, DpiScale) noexcept`. This seam is unresolved: the translate function signature is inconsistent with the documented divide requirement.

recommendation:Update `translate_mouse` signature in §3.3 to accept a `DpiScale` second parameter — `auto translate_mouse(const SDL_Event&, DpiScale) noexcept -> eastl::optional<InputEvent>` — and add a sentence in §3.3 explaining where the pump gets the DPI scale (cached field in `Pump::Impl`, initialised to 1.0 at construction and updated every time `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED` is processed by §3.5). This keeps `translate_mouse` a pure function over its inputs and makes the seam explicit without adding a window-surface dependency to the translation table.

specs/platform/event-pump-design.md
```diff
@@ -396,6 +430,13 @@ file queue for editor-managed assets; routing drop through a
 separate `WindowEvent` would force every importer to drain two
 queues. Single-consumer ⇒ single channel.
 
+`ingest_drop` has one concrete caller at MVP (this branch — the
+`SDL_EVENT_DROP_FILE` handler above). Per PHILOSOPHY §Anti-patterns,
+promotion to a stable cross-aggregate internal contract is gated on
+a second caller materialising. Until then, `ingest_drop` is
+event-pump's private extension to file-watcher and may be removed if
```
cjhowe-us, line 455: severity:MED location:specs/platform/event-pump-design.md:437 problem:§3.6 documents `ingest_drop` as "event-pump's private extension to file-watcher" with a one-caller note and a §12 gate. However, the SRP statement in §1 enumerates exactly five reasons the pump design changes: SDL3 event vocabulary, typed-variant translation table, per-device ordering rule, bounded SPSC ring protocol, per-frame drain trigger. Drop-event routing to a sibling aggregate is a *sixth* unlisted responsibility — and a genuinely independent one (if drop events should go to a `WindowEvent::FileDropped` variant instead, the pump changes for that reason alone, orthogonal to the other five). The §12 gate acknowledges the one-caller problem but does not address the SRP gap.

recommendation:Either (a) add "drop-event routing to the sibling FileWatcher aggregate" explicitly to the §1 SRP boundary statement as a sixth responsibility with rationale ("we own drop routing because the single consumer is always the file-event path"), making the boundary documented and auditable, or (b) move the routing decision out of the pump into the file-watcher design: the pump emits a `WindowEvent::FileDropped { canonical_path }` variant (adding one WindowEvent arm) and the FileWatcher aggregate subscribes to that variant. Option (b) eliminates the pump→file-watcher dependency entirely and keeps the pump's five responsibilities intact, at the cost of one new WindowEvent arm. Option (a) is lower risk at MVP; option (b) is architecturally cleaner. Both are in scope for a design-doc fix.

specs/platform/event-pump-design.md
```diff
@@ -967,9 +1009,16 @@ When `EventQueue<T>::try_push` returns false during `Pump::drain`:
...
 1. The pump emits a `error`-severity log with the prefix
    `"queue-full"` and the queue family name.
-2. The pump returns `unexpected(IoFailure { OsCode { 0 } })`
-   from the current `Pump::drain` call.
-3. The engine frame loop catches the failure, logs at `error`,
+2. The pump calls `detail::error::set_prefix(prefix::queue_full)`
+   immediately before constructing the failure value.
+3. The pump returns `unexpected(IoFailure { OsCode { 0 } })`
+   from the current `Pump::drain` call. The TLS prefix stamp is the
+   load-bearing discriminator: callers that receive `IoFailure` from
```
cjhowe-us, line 1044: severity:MED location:specs/platform/event-pump-design.md:1016 problem:§10.3 (queue-full recovery shape) lists the steps as: (1) emit error-severity log with prefix "queue-full" and queue family name, (2) call `set_prefix(prefix::queue_full)` immediately before constructing the failure value, (3) return `unexpected(IoFailure { OsCode { 0 } })`. The step ordering is wrong for the log: step 1 emits the log BEFORE the TLS prefix is stamped (step 2). If `glibre::log_error` reads `current_prefix()` to populate the structured "prefix" field, it will see `prefix::none` at step 1 (the stamp has not happened yet). The log will be missing its `"queue-full"` prefix field, defeating the purpose of the stamp. The correct ordering is: stamp prefix first, then log (so the log captures the stamped prefix), then construct and return the error.

recommendation:Reorder §10.3 steps: (1) call `detail::error::set_prefix(prefix::queue_full)` — stamp first so the log can read it; (2) emit error-severity log with the queue family name (the log now reads `current_prefix()` and records `"queue-full"`); (3) return `unexpected(IoFailure { OsCode { 0 } })`. Mirror this reordering in §3.4 prose where the same two-step sequence appears.

specs/platform/event-pump-design.md
```diff
@@ -651,7 +692,8 @@ is a contract violation. Detection in MVP:
   surface as an error, not crash the engine.
 
 This is the canonical detection pattern from `error-model.md`
-§"Severity Conventions" — assert in debug, demote in shipping.
+§"Logging/Telemetry" (see hot-reload refusal severity discussion
```
cjhowe-us: severity:LOW location:specs/platform/event-pump-design.md:695 problem:§6.4 cross-reference was updated by R1 to cite `error-model.md §"Logging/Telemetry"` with the note "(see hot-reload refusal severity discussion under that heading)". However, the Logging/Telemetry section of error-model.md discusses severity for *hot-reload refusals* (warn vs error), not the assert-in-debug / demote-in-shipping pattern for *programming errors* (off-thread drain calls). The §6.4 use case is a contract violation by a caller — not a hot-reload refusal. The relevant pattern (assert in debug, downgrade to non-fatal in release) is not formally documented in error-model.md at all; it is a convention described in SPEC §4.1 inv #2 ("debug builds assert, release builds return `PlatformError::Unsupported`") and in SPEC §4.2 inv #1. The citation is still slightly wrong: it points to a section whose severity rule does not apply to the §6.4 scenario.

recommendation:Replace the citation with a SPEC forward-reference: "consistent with SPEC §4.1 inv #2 (main-thread-only Window operations) and §4.2 inv #1 (single-threaded pump contract): debug builds assert, release builds return `Unsupported` without abort." No amendment to error-model.md is required; the SPEC already has the right text.

specs/platform/event-pump-design.md
```diff
@@ -1147,6 +1222,9 @@ hot-reload survival path; runs against the same S1 trace with a
...
 |--------------------------------------------|---------------------------------------------------------------------------------------|
 | SDL3 → typed translation correctness       | unit #1, #2, #3, #6, #7                                                               |
+| Mouse dx/dy DPI divide (§3.3, R-6.1.4)    | unit #3b                                                                              |
+| Trigger released-state formula (§3.3)      | unit #6b                                                                              |
+| with_capacity round-up (§3.4, §10.1)      | unit #8b                                                                              |
 | Sealed-variant boundary (unknown drop)     | unit #4, #5                                                                           |
 | Per-device monotone ordering (§4.2 inv #2) | unit #10; integration #4                                                              |
 | Bounded-channel never-drop (§4.2 inv #3)   | unit #8, #9                                                                           |
```
cjhowe-us, line 1259: severity:LOW location:specs/platform/event-pump-design.md:1230 problem:The §11.4 coverage matrix row "SDL3 → typed translation correctness" lists "unit #1, #2, #3, #6, #7" but does not include #3b (mouse dx/dy DPI divide). Test #3b is listed in its own separate row "Mouse dx/dy DPI divide (§3.3, R-6.1.4) → unit #3b". This is not technically wrong (two rows can share coverage), but a reader of the "SDL3 → typed translation correctness" row will miss #3b, since the divide IS part of translation correctness (wrong divide → wrong translated value). The matrix suggests coverage is complete at #1,#2,#3,#6,#7 for translation, but #3b provides a translation-correctness assertion at non-unit DPI scale.

recommendation:Add "#3b" to the "SDL3 → typed translation correctness" row so that the row reads "unit #1, #2, #3, #3b, #6, #7". The dedicated "Mouse dx/dy DPI divide" row can remain for discoverability. This is a two-word doc fix.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:2
  LOW:2
carryover:0

notes:All nine R1 findings were addressed in commit 89cb6a3 and PR #900. The R1 carryover count is zero. Round-2 cohesion + SRP + seam-quality pass surfaces one new HIGH and two new MEDs.

The HIGH is a seam incompleteness introduced by the R1 MED-4 fix: §3.3 now documents that translate_mouse must divide dx/dy by `window.dpi_scale()`, but the translate_mouse free-function signature `(const SDL_Event&) -> eastl::optional<InputEvent>` has no DPI parameter and no access to window state. The function cannot execute the documented divide without either (a) a new DpiScale parameter, or (b) a cached DpiScale field in Pump::Impl passed at call time. Neither option is specified; the design is internally inconsistent on this seam. Fix is in scope for impl-respond: add a DpiScale parameter to translate_mouse and describe where Pump::Impl acquires the scale (cached on SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED, same event already handled in §3.5).

MED-1: The §1 SRP boundary lists five and only five reasons the pump design changes; drop-event routing to FileWatcher is a sixth, independent reason that is unlisted. The one-caller note and §12 gate from R1 address the PHILOSOPHY §Anti-patterns concern but not the SRP gap. Fix: either add drop routing explicitly to the §1 SRP boundary with rationale, or move the routing out of the pump (emit WindowEvent::FileDropped, let FileWatcher subscribe) — both are design-doc-only changes.

MED-2: §10.3 step ordering is inverted: the log (step 1) fires before set_prefix (step 2), so the log reads prefix::none and the structured "queue-full" prefix field is absent from the telemetry record. Fix: reorder to stamp first, log second, return third. Same reorder in §3.4 prose.

LOW-1: §6.4 cross-ref still cites the wrong section of error-model.md for the assert-in-debug/demote-in-shipping pattern. Logging/Telemetry covers hot-reload refusals, not programming-error contract violations. The correct source is SPEC §4.1 inv #2 and §4.2 inv #1.

LOW-2: Coverage matrix §11.4 "SDL3 → typed translation correctness" row omits #3b (DPI divide test), which IS a translation-correctness assertion.

What works well: alignas(128) RingImpl fix is correct and well-annotated; the capacity round-up semantic is clean; the TLS prefix stamp prose in §3.4 is now explicit and load-bearing; trigger-axis formula is documented with the consumer-remap invariant; TextInput/SDL3 buffer correction is accurate; §11.3 E2E gap tracking is honest. EASTL discipline is sound throughout the public surface stub (eastl::variant, eastl::optional, eastl::span — no std:: containers on public or internal surfaces). The §12 open questions are all [NON-BLOCKING] or [OPEN] and are actionable; none require amendment spikes before implementation begins, with the exception of the new HIGH finding which should be resolved in this PR before implementation proceeds.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:caa746d followup:#900

Added `DpiScale` parameter to `translate_mouse` signature: `auto translate_mouse(const SDL_Event&, DpiScale) noexcept -> eastl::optional<InputEvent>`. Added §3.3 sentence documenting that `Pump::Impl` caches the current `DpiScale` (initialised `1.0f` at construction; updated in §3.5 on `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`), and that each `translate_mouse` call passes the cached `DpiScale` by value. Updated §11.1 test #3b to call `translate_mouse(ev, DpiScale{2.0f})` directly.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:caa746d followup:#900

Added explicit responsibility 6 to §1 SRP boundary statement: drop-event routing to the FileWatcher aggregate. Added rationale paragraph: listed for full SRP transparency; §12 one-caller gate (PHILOSOPHY §Anti-patterns) governs ingest_drop stability; SRP boundary stays unchanged whether routing goes via ingest_drop or a future WindowEvent::FileDropped variant.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:caa746d followup:#900

Reordered §10.3 steps: (1) set_prefix(prefix::queue_full), (2) log_error (now reads stamped prefix), (3) return unexpected. Applied same reorder to §3.4 parallel prose, adding the log step between stamp and return. Added brief rationale note: stamp must precede log so log_error reads the stamped prefix; reorder is cold-path-only.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:caa746d followup:#900

Replaced §6.4 cross-reference: removed error-model.md §"Logging/Telemetry" citation entirely; substituted "consistent with SPEC §4.1 inv #2 and §4.2 inv #1" which are the invariants that actually govern the assert-in-debug/demote-in-shipping pattern for off-thread drain.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:caa746d followup:#900

Added #3b to §11.4 "SDL3 → typed translation correctness" row. New row reads: "unit #1, #2, #3, #3b, #6, #7".

specs/platform/event-pump-design.md
```diff
@@ -930,7 +987,7 @@ failure is a `glibre::Error` arm (with the platform-private
 
 | Entry point                     | Returnable arms                                  | Trigger                                         |
 |---------------------------------|--------------------------------------------------|-------------------------------------------------|
-| `EventQueue<T>::with_capacity`  | `Unsupported`                                    | zero / non-power-of-two beyond cap / wildly oversized capacity (> 2^24 — SPEC sub-arena cap) |
+| `EventQueue<T>::with_capacity`  | `Unsupported`                                    | zero capacity / wildly oversized capacity (> 2^24 — SPEC sub-arena cap); non-power-of-two is silently rounded up (not an error) |
```
cjhowe-us: severity:HIGH location:specs/platform/event-pump-design.md:990 problem:The §10.1 failure-surface table defines the Unsupported trigger for EventQueue<T>::with_capacity as '> 2^24 — SPEC sub-arena cap', but the §12 [BLOCKING IMPLEMENTATION] note (added in commit 672e287) and the inline paragraph below the table both state the intended ceiling is 'capacity overflow beyond u32::max'. These are different values by a factor of 256 (2^24 = 16 M slots; u32::max approximately 4.3 B slots). An implementer reading §10.1 would reject calls with capacity > 16 M; reading §12 they would allow values up to 4.3 B. The table is the authoritative failure-surface specification for this entry point; the inconsistency makes the design non-implementable at this threshold. Neither 2^24 nor u32::max is justified by the actual sub-arena size: the EventQueue sub-arena is 2 MiB (SPEC §9.2), which at the largest slot size (64 B) supports at most ~32 K slots — so both numbers are far above any physically reachable capacity. fix:Update the §10.1 table trigger column for EventQueue<T>::with_capacity to read: 'zero capacity / capacity overflow beyond u32::max (the sub-arena ceiling); non-power-of-two is silently rounded up (not an error)'. Remove the '> 2^24' figure (stale value carried from the old SPEC wording) and align the design's own authoritative table with the §12 BLOCKING note and the amendment it prescribes for SPEC §10.3.2.

specs/platform/event-pump-design.md
```diff
@@ -277,24 +287,45 @@ Key normalisation rules realised in the table:
...
-  window's client area regardless of compositor placement.
+- **Mouse coordinates: all four fields in logical points.**
+  Absolute `x/y` from `SDL_MouseMotionEvent` are already logical
+  points (matching `LogicalSize`) when `SDL_WINDOW_HIGH_PIXEL_DENSITY`
+  is set — window-relative so coordinate (0, 0) is top-left of the
+  client area regardless of compositor placement. Relative `dx/dy`
+  from `SDL_SetRelativeMouseMode` are physical pixel deltas on macOS
+  and must be divided by the current `DpiScale` before being placed
+  in `MouseMove::dx/dy`. `Pump::Impl` caches the current `DpiScale`
+  (initialised `1.0f` at construction; updated in §3.5 on
```
cjhowe-us, line 300: severity:MED location:specs/platform/event-pump-design.md:298 problem:The new §3.3 text (added in commit caa746d) states Pump::Impl caches the DpiScale 'updated in §3.5 on SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED', and the R-6.1.4 coverage row says the same. However the §3.5 internally-consumed events table (pre-existing, not in this diff) describes only one side effect for that SDL3 event: updating Window::Impl's cached metrics. The Pump::Impl DpiScale refresh is never mentioned in §3.5. An implementer following §3.5 as the implementation specification for event handlers would update Window::Impl only, leaving Pump::Impl.dpi_scale stale and breaking the divide-by-scale path in translate_mouse until the next DPI change. fix:Add a sentence to the SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED side-effect column in §3.5 documenting the Pump::Impl DpiScale cache update. Suggested addition after the existing Window::Impl update sentence: 'Also updates Pump::Impl-cached DpiScale from the new Window::Impl value; the updated scale is passed to translate_mouse on the next SDL_EVENT_MOUSE_MOTION in this frame.'

specs/platform/event-pump-design.md
```diff
@@ -1043,6 +1118,12 @@ the platform user-story issues).
    `(LogicalSize, PhysicalSize, DpiScale)` snapshot on
    `Window::Impl` is updated *before* the typed event is enqueued
    (§3.5 / §6.3 SPEC cross-module note).
+3b. **`event_pump_mouse_relative_motion_divided_by_dpi_scale`** — call
+    `translate_mouse(ev, DpiScale{2.0f})` with a `SDL_MouseMotionEvent`
+    whose relative `dx/dy` is 200 physical pixels; assert the emitted
+    `MouseMove::dx/dy` is `100.0f` (logical points). Confirms the
+    divide-by-dpi_scale invariant documented in §3.3 and that the
+    `DpiScale` parameter is the operative divisor. (§R-6.1.4.)
```
cjhowe-us: severity:LOW location:specs/platform/event-pump-design.md:1126 problem:Test #3b's rationale parenthetical reads '(§R-6.1.4.)' — the section-symbol prefix '§' is reserved for section numbers in this design document (§3.3, §10.1, etc.). Requirement identifiers use 'R-6.1.4' without the '§' prefix, as in every other requirement reference in the document (§2 table headers, R-14.1.9, R-14.2.4, etc.). The '§R-6.1.4' form is a cross-reference style error that a future cross-reference linter would flag. fix:Change '(§R-6.1.4.)' to '(R-6.1.4)' to match the requirement-reference convention used throughout the rest of the document.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:1
carryover:0

notes:All prior findings from the #826 three-round pipeline (9 R1 + 5 R2 + 3 R3 = 17 total) were correctly addressed across commits 89cb6a3, caa746d, and 672e287. Carryover is zero. Round-1 coverage + correctness pass on #900 surfaces one new HIGH, one new MED, and one new LOW.

The HIGH (comment 3189836081) is an internal inconsistency introduced by the R3 MED-2 fix: the §10.1 failure-surface table retains the old SPEC wording '> 2^24 — SPEC sub-arena cap' as the Unsupported trigger threshold for EventQueue::with_capacity, while the §12 [BLOCKING IMPLEMENTATION] note (added in commit 672e287) and the inline note paragraph below the same table both prescribe 'capacity overflow beyond u32::max' as the correct ceiling. These are different values by a factor of 256. The §10.1 table is this design's authoritative per-entry-point specification; leaving the old value in place makes the design non-implementable at this threshold. Fix: replace '> 2^24 — SPEC sub-arena cap' in the §10.1 table trigger column with 'capacity overflow beyond u32::max (the sub-arena ceiling)' to align the table with the §12 note and the SPEC amendment it intends.

The MED (comment 3189841807) is a documentation coverage gap: §3.3 and the R-6.1.4 coverage row both state the Pump::Impl DpiScale cache is 'refreshed in §3.5 on SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED', but the §3.5 internally-consumed events table (pre-existing, not in this diff) documents only the Window::Impl cache update for that SDL3 event. An implementer following §3.5 as the authoritative handler spec would omit the Pump::Impl refresh, breaking translate_mouse's divide-by-scale path. Fix: extend the §3.5 side-effect column for SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED to include the Pump::Impl DpiScale update.

The LOW (comment 3189842953) is a cross-reference style error: test #3b uses '(§R-6.1.4.)' where requirement IDs in this document never carry the '§' prefix. Fix: change to '(R-6.1.4)'.

Positive observations: CC subjects on all three commits are well-formed (docs(platform): scope, body messages descriptive, issue refs present). TLS prefix stamp ordering in §3.4 and §10.3 is now consistent: stamp-before-log-before-return across both sections. alignas(128) annotation on RingImpl head/tail is present and rationalized. Mouse dx/dy DPI divide path (translate_mouse signature, §3.3 rule, test #3b) is coherent end-to-end. Trigger formula and released-state invariant are documented with test #6b. §12 BLOCKING entry for SPEC §10.3.2 amendment is actionable with two named concrete consumers. EASTL discipline is maintained throughout.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:2340657 followup:#900

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:2340657 followup:#900

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:2340657 followup:#900

specs/platform/event-pump-design.md
```diff
@@ -85,10 +88,17 @@ The aggregate **refuses to own**:
...
+**per-frame drain trigger**, or the **drop-event routing to the
+`FileWatcher` aggregate** change, this design changes. Anything
+else — input action mapping, window aggregate state, display
+pacing — is out of scope.
+
+Drop-event routing is listed as a sixth responsibility for full SRP
+transparency. The §12 one-caller gate (per PHILOSOPHY §Anti-patterns)
+governs whether `ingest_drop` becomes a stable cross-aggregate API;
+the SRP boundary stays unchanged whether routing goes via
+`ingest_drop` or via a future `WindowEvent::FileDropped` variant.
```
cjhowe-us: severity:MED location:specs/platform/event-pump-design.md:100 problem:The new §1 rationale paragraph is internally contradictory. The SRP boundary statement (lines 89-103 in the PR branch) lists 'drop-event routing to the FileWatcher aggregate' as a sixth reason the design changes — meaning it is an owned responsibility. The very next paragraph then states 'the SRP boundary stays unchanged whether routing goes via ingest_drop or via a future WindowEvent::FileDropped variant.' These two claims contradict each other: the first says routing changes drive design changes; the second says they do not. A reader will not be able to determine whether a change in routing mechanism (ingest_drop → WindowEvent variant) constitutes an in-scope design change or not.

fix:Clarify the distinction between the routing *policy* (who receives drop events — this IS an owned responsibility and a reason the design changes) and the routing *mechanism* (how the policy is implemented — ingest_drop vs WindowEvent variant — this does NOT change the SRP boundary). Suggested replacement for the rationale paragraph: 'Drop-event routing to a file-event consumer is listed as a sixth responsibility because the pump decides which aggregate owns each SDL3 event family. The SRP boundary changes if the consumer policy changes (e.g. drop events routed to a window-event consumer instead). The SRP boundary does NOT change solely because the implementation mechanism shifts from ingest_drop to a WindowEvent::FileDropped variant — that is a §3.6 implementation detail, not a new responsibility.'

specs/platform/event-pump-design.md
```diff
@@ -353,7 +403,7 @@ the event itself triggers a side effect inside the pump:
...
+| `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED`                                | Update the owning `Window::Impl`'s cached `(LogicalSize, PhysicalSize, DpiScale)` snapshot via `surface::detail::query_layer_metrics` *before* the event becomes visible to the engine (§6.3 cross-module note in SPEC). Also updates `Pump::Impl`-cached `DpiScale_` from the new `Window::Impl` value; the updated scale is passed to `translate_mouse` on the next `SDL_EVENT_MOUSE_MOTION` in this frame (consumed by the divide-by-scale path per §3.3). | `WindowEvent::DpiChanged`.                    |
 | `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` / `_RESIZED`                      | Update the owning `Window::Impl`'s cached `LogicalSize` / `PhysicalSize`.                                                                                                                                                            | `WindowEvent::Resized`.                       |
```
cjhowe-us, line 409: severity:LOW location:specs/platform/event-pump-design.md:406 problem:The §3.5 side-effect column for SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED now includes 'Also updates Pump::Impl-cached DpiScale_'. The trailing underscore in DpiScale_ is a C++ private-member naming convention leaking into the spec document. Spec documents should describe logical roles and seam contracts, not C++ identifier conventions. The identifier will likely diverge: other design documents use snake_case without trailing underscore (dpi_scale), and the actual field name chosen at implementation time may differ.

fix:Replace 'DpiScale_' with 'the Pump::Impl-cached DpiScale value' (or 'Pump::Impl::dpi_scale' if the field name must be named). The spec should name the logical concept ('the pump's cached DPI scale'), not the C++ identifier with its naming-convention suffix.

specs/platform/event-pump-design.md
```diff
@@ -1197,3 +1302,26 @@ hot-reload survival path; runs against the same S1 trace with a
...
+  event-pump's private extension to file-watcher and may be removed if
+  the second caller never appears. Per PHILOSOPHY §Anti-patterns.
+- [NON-BLOCKING] **Quit-synthesis E2E coverage gap.** `SDL_EVENT_QUIT`
+  → `WindowEvent::CloseRequested` (§3.5) is not exercised by any E2E
+  trace at MVP. The S1 trace exits via window close, not Cmd-Q.
+  Integration test #6 (`event_pump_quit_synthesises_close_requested`)
+  is the acceptance gate; no E2E trace covers menu-driven quit.
+  Trigger to close: the user-story for Cmd-Q app quit lands in scope.
+- [BLOCKING IMPLEMENTATION] **SPEC §10.3.2 `EventQueue::with_capacity`
+  trigger column amendment.** The current SPEC §10.3.2 entry describes
```
cjhowe-us, line 1320: severity:MED location:specs/platform/event-pump-design.md:1317 problem:The §12 [BLOCKING IMPLEMENTATION] entry for SPEC §10.3.2 amendment says 'Two concrete consumers: shipping runtime input pump, editor input pump' and 'must land before plan PRs open'. However, no GitHub issue or spike is linked or created to track this SPEC amendment. The BLOCKING label without an issue reference is not actionable — the impl-respond agent or any future plan executor reading this entry cannot verify the amendment was done, cannot link a plan PR against it, and cannot check a dependency gate. This is a seam-quality gap: the §12 entry describes a required action but provides no tracking vehicle.

fix:Add a GitHub issue reference to the BLOCKING entry, or create one. The entry should read: 'Owner: platform-SPEC amendment spike — open a spike issue against SPEC §10.3.2, land it before plan PRs open. Issue: #<N> (to be created).' Alternatively, if the SPEC amendment is simple enough to land in this PR (it is a one-line table-cell edit in SPEC.md §10.3.2), land it here and remove the BLOCKING note.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0

notes:All three R1 findings (HIGH §10.1 with_capacity threshold inconsistency, MED §3.5 Pump::Impl DpiScale refresh gap, LOW §R-6.1.4 prefix style) are confirmed ADDRESSED in commit 2340657. Carryover is zero.

Round-2 cohesion + SRP + seam-quality pass surfaces zero HIGHs, two MEDs, and one LOW. No implementation blocker exists that would prevent the design from being implementable — both MEDs are documentation-clarity issues rather than seam incompleteness.

MED-1 (comment 3189884542): The new §1 rationale paragraph introduced in commit caa746d is internally contradictory. The SRP boundary statement lists drop-event routing as a sixth reason the design changes (i.e., an owned responsibility). The following sentence then states the SRP boundary is stable regardless of whether routing goes via ingest_drop or WindowEvent::FileDropped. These two claims cannot both be true in the same sentence pair. The fix is to clarify the policy/mechanism distinction: the pump owns the policy of routing drop events to a file-event consumer (reason to change); it does not own the specific mechanism (ingest_drop vs variant), which is a §3.6 implementation detail. One paragraph, one concept.

MED-2 (comment 3189886666): The §12 [BLOCKING IMPLEMENTATION] entry for the SPEC §10.3.2 amendment is not actionable without a GitHub issue reference. The entry describes a required SPEC edit that must land before any plan PR consuming this design, but does not provide a tracking vehicle. An implementer cannot gate their plan PR against an unlinked requirement. Fix: either create the spike issue and add its reference, or land the one-line §10.3.2 SPEC.md edit directly in this PR and remove the BLOCKING note.

LOW-1 (comment 3189885537): The §3.5 side-effect column for SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED uses 'DpiScale_' — a C++ private-member naming convention that should not appear in a spec document. Replace with 'the Pump::Impl-cached DpiScale value' or plain 'Pump::Impl::dpi_scale'.

What works well: The translate_mouse(SDL_Event, DpiScale) seam is clean and the R1 HIGH is fully resolved. The alignas(128) annotation is present and rationalised. The §10.3 stamp-before-log-before-return ordering is correct and internally consistent with §3.4. The §10.1 trigger column (u32::max ceiling, silent round-up) is consistent with §12 BLOCKING and §3.4. The §11.4 coverage matrix now includes #3b in both the "SDL3 → typed translation" row and the dedicated DPI divide row. The §12 NON-BLOCKING entries for ingest_drop and quit-synthesis are honest, complete, and correctly labelled. EASTL discipline is sound throughout.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:69539e1 followup:#900

Rewrote the §1 paragraph to distinguish policy from mechanism. The SRP boundary list entry now names 'drop-event routing policy' (not 'routing to FileWatcher aggregate'). The following paragraph is replaced with: 'Responsibility 6 (owned, in-scope): the *policy* of drop-event routing — that the pump forwards SDL_EVENT_DROP_FILE to the file-watcher aggregate. This policy stays owned regardless of which *mechanism* implements it (current: FileWatcher::ingest_drop direct call; future option: WindowEvent::FileDropped via the queue, gated on §12 [NON-BLOCKING] one-caller audit). The SRP boundary tracks the policy, not the mechanism — switching mechanisms does NOT trigger an SRP-list edit, but adding or removing the policy itself would.' The contradiction is resolved: the boundary is sensitive to policy changes, not mechanism changes.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:69539e1 followup:#900

Replaced 'Pump::Impl-cached DpiScale_' with 'the Pump::Impl-cached DpiScale value' in the §3.5 side-effect column for SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED. The trailing underscore C++ private-member convention no longer leaks into the spec document.

specs/platform/event-pump-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:69539e1 followup:#900

Added a STATUS note within the §12 [BLOCKING IMPLEMENTATION] entry (filing the spike issue is explicitly out of impl-respond scope per the dispatch). The status note reads: 'STATUS: The amendment is a one-line edit to replace "zero / wildly oversized capacity" with "capacity overflow beyond u32::max" in the SPEC §10.3.2 trigger column. The orchestrator (or first plan PR author) must either (a) file [SPIKE] amend-platform-spec-event-pump-with-capacity-trigger parented to #714 and backfill the issue number into this entry, OR (b) land the one-line SPEC edit directly in the first plan PR consuming this design. Plan PRs cannot land until SPEC §10.3.2 is corrected via either route. Issue: #TBD (to be filed at plan-PR authoring time).' The tag remains [BLOCKING IMPLEMENTATION] and is now actionable for any plan PR author.

cjhowe-us (commented):
round:4 reviewer:r4-manual verdict:APPROVE
findings:
  HIGH:0
  MED:0
  LOW:0
carryover:0

notes: All three R2 open findings on SHA 2340657 are resolved at HEAD 69539e1. MED-1 (§1 SRP self-contradiction on drop-event routing) is addressed: §1's SRP boundary now lists "drop-event routing policy" alongside the other five sharp triggers, and a new closing paragraph explicitly distinguishes policy (forwarding `SDL_EVENT_DROP_FILE` to the file-watcher aggregate) from mechanism (current direct call to `FileWatcher::ingest_drop`; future option `WindowEvent::FileDropped`), with the rule that swapping mechanisms does NOT trigger an SRP-list edit while adding/removing the policy itself would. MED-2 (§12 [BLOCKING IMPLEMENTATION] missing GH issue ref) is addressed via in-PR resolution: the entry now defines an explicit STATUS block with two acceptable resolution paths — (a) file `[SPIKE] amend-platform-spec-event-pump-with-capacity-trigger` parented to #714 and backfill, OR (b) land the one-line SPEC §10.3.2 edit directly in the first plan PR — and asserts plan PRs cannot land until either route is taken. LOW-1 (§3.5 used C++ private-member name `DpiScale_`) is addressed: the table row now reads "Pump::Impl-cached DpiScale value", trailing-underscore convention removed.

Re-scan: §1 SRP coherence holds; §3.5 side-effect column is consistent with §3.3 / §3.6 / R-6.1.4; §10.1 trigger column matches §3.4 round-up + u32::max ceiling and is back-referenced by §10.4 / §12; §10.3 stamp-before-log is explicit (step 1 sets prefix, step 2 emits log reading the stamped prefix, step 3 cites the TLS prefix as the load-bearing IoFailure discriminator); §11.1 has #3b (DPI divide), #6b (trigger released = -1.0f), #8b (with_capacity round-up) all present; §11.4 coverage matrix cites them. EASTL discipline is clean (`eastl::variant`, `eastl::span`, `eastl::optional`; `std::function` and `eastl::function` both negated; only `std::size_t` as fundamental type). Requirement-ID style is uniform (R-6.1.1..R-6.1.15, R-14.1.x, R-14.2.x). `error-model.md` citation at line 564 references "Composition Rules" which exists at line 137 of `reviews/decisions/error-model.md`.

Recommendation: APPROVE for the human gate. Spec is internally coherent, requirement-complete, test-plan-rooted, and SPEC §10.3.2 amendment path is explicit. The only residual external dependency (the SPEC §10.3.2 trigger-column edit) is correctly captured as in-PR-resolvable at first plan-PR authoring time and is non-blocking for landing this design spike PR.

Issue #807, opened 2026-05-03
[SPIKE] design-content-fbx-importer-detailed


cjhowe-us:
Parent: #806

## Question

Produce the detailed design for the **fbx-importer** aggregate in the `content` context.

FBX SDK integration.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/content/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/content-pipeline/asset-import.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **fbx-importer** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/content/fbx-importer-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(content): detailed design — fbx-importer (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-content-fbx-importer-detailed` is blocked by this issue.

cjhowe-us:
agent:go-design
status:started
issue:#807
branch:docs/content-fbx-importer-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/1778171348-issue-807
host:$(hostname -s)
cloud:none
commit:c9bef46
pr:none
[...]

Issue #811, opened 2026-05-03
[SPIKE] design-content-font-importer-detailed


cjhowe-us:
Parent: #806

## Question

Produce the detailed design for the **font-importer** aggregate in the `content` context.

FreeType font import.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/content/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/content-pipeline/asset-import.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **font-importer** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/content/font-importer-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(content): detailed design — font-importer (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-content-font-importer-detailed` is blocked by this issue.

cjhowe-us:
agent:go-design
status:done
issue:#811
branch:docs/content-font-importer-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/1778176654-issue-811
host:mac
cloud:none
commit:2582be771dc8c920712e0474877f34467d1e98d4
pr:#910
[...]

Issue #809, opened 2026-05-03
[SPIKE] design-content-texture-importer-detailed


cjhowe-us:
Parent: #806

## Question

Produce the detailed design for the **texture-importer** aggregate in the `content` context.

FreeImage texture import.

## Required reads

- /Users/cjhowe/Code/glibre/PHILOSOPHY.md
- /Users/cjhowe/Code/glibre/AGENTS.md
- /Users/cjhowe/Code/glibre/specs/content/SPEC.md (FULLY filled — §1-§12)
- /Users/cjhowe/Code/glibre/reviews/decisions/*.md
- Harmonius coverage source(s) — input only, re-derive everything:
- harmonius/docs/requirements/content-pipeline/asset-import.md

Verify that every harmonius requirement / design clause that belongs in MVP scope for the **texture-importer** aggregate is either covered in the new design or explicitly refused with rationale.

## Deliverable

`/Users/cjhowe/Code/glibre/specs/content/texture-importer-design.md` with 12 sections: Purpose, Requirements coverage, Detailed model, Public surface, Hot/cold path split, Concurrency, Persistence + ABI, Hot-reload, Performance, Failure modes, Test plan, Open questions.

## Closure

Design committed via PR with Conventional Commit subject `docs(content): detailed design — texture-importer (refs #N)`. Status comment posted; issue stays OPEN.

## Follow-up

Sibling `[SPIKE] task-breakdown-content-texture-importer-detailed` is blocked by this issue.

cjhowe-us:
agent:go-design
status:done
issue:#809
branch:docs/content-texture-importer-design
worktree:/Users/cjhowe/Code/glibre/.claude/worktrees/1778175079-issue-809
host:mac
cloud:none
commit:fa0a5f3ce389c000970e07da390484f3df6a1bcc
pr:#909
[...]

PR #905, opened 2026-05-07
docs(content): detailed design — fbx-importer (refs #807)

cjhowe-us:
## Summary

- Detailed design for the `FbxImporter` cook-time aggregate inside the content context, refining `specs/content/SPEC.md` §4.1.2 / §5.4 / §6.1 / §6.2 stage 3 / §6.5 cut row / §7.1.2 / §8.5 Class B / §9.3.1 / §10.1 / §10.3 in place. Introduces no new public surface beyond the §5.4 stub.
- Re-derives every harmonius asset-import requirement (R-12.1.1..R-12.1.5) per `PHILOSOPHY.md` against SOLID/SRP, EASTL, no-Obj-C++, determinism, no-runtime-reflection. Refuses audio (R-12.1.3 → `audio` plugin) and routes mesh-internal geometry processing (R-12.2.*) to the `geometry` plugin per SPEC §3.3.
- Twelve sections per the spec template: Purpose, Requirements coverage, Detailed model (per-thread `FbxManager`, `FbxMemoryAllocator` hook routed to `glibre::PerContextAllocator`, seven-stage normalize pipeline, in-memory `MeshArtifactPrecursor` layout), Public surface, Hot/cold path split (build-time-only; runtime cut by §6.5), Concurrency (off-thread worker pool; 0 ms driver-thread cost; cancellation polled per `FbxNode`), Persistence + ABI (FBX SDK = OS-side dependency the runtime never links; `importer_version` BLAKE3 participates in `CookKey` §4.1.3 component #2), Hot-reload (manifest-swap path = no importer code; plugin-code reload drains/re-spawns the worker pool), Performance (cell unaffected at driver thread; 64 MiB soft sub-ceiling per-cook arena, SPEC §9.3.1), Failure modes (six `ImporterError::*` arms; SDK-exception translation table at the unique `-fexceptions` carve-out; `std::bad_alloc` / non-`std::exception` terminate per SPEC §10.7 OQ-2), Test plan (Catch2 under `tests/content/import/fbx/`: dispatch, refusal, cancellation, determinism, off-thread soundness), Open questions (typed `NormalizeParamsView`, animation/morph imports, material parameter values — each routed to a post-MVP epic).

Resolves the deliverable for spike #807. Sibling `[SPIKE] task-breakdown-content-fbx-importer-detailed` is unblocked once this design lands.

## Test plan

- [ ] Round 1 review (parent /go skill): SOLID / SRP boundary, refusal completeness against harmonius `R-12.1.*`, EASTL discipline, `-fexceptions` scope, determinism rules, and frame-phase invariants.
- [ ] Round 2 review: cohesion + duplication against sibling content / data / platform / geometry design docs (FBX SDK seam, `FbxMemoryAllocator` hook routing, `importer_version` participation in `CookKey`).
- [ ] Round 3 review: topology / estimation sanity for the follow-up `task-breakdown-content-fbx-importer-detailed` spike (it will derive `type:plan` issues from §11's Catch2 cases and §3 / §6 / §9 contracts).

### DIFF
--- a/specs/content/fbx-importer-design.md
+++ b/specs/content/fbx-importer-design.md
@@ -1,0 +1,1866 @@
+# content — Detailed Design: fbx-importer aggregate
+
+> Detailed design for the `FbxImporter` aggregate (cook-time entity)
+> declared in `specs/content/SPEC.md` §4.1.2, with public surface
+> locked in §5.4 and lifecycle / scratch / SDK seams pinned in §6.1
+> (`importers/fbx_importer.{hpp,cpp}`), §6.2 stage 3, §6.5 (build-time
+> cut row), §7.1.2 (`importer_version` participation in `CookKey`),
+> §8.5 Class B (importer / cook-step refusal classes), §9.3.1
+> (off-thread soft-ceiling scratch arena), §10.1 (`ImporterError::*`
+> per-arm contract), and §10.3 (`-fexceptions` carve-out + SDK-exception
+> translation table). Refines those sections in place; introduces no
+> new public surface beyond `specs/content/SPEC.md` §5.4. Deviations
+> from the cited records require an amendment spike, not an in-place
+> edit.
+>
+> Cites `reviews/decisions/error-model.md`,
+> `reviews/decisions/perf-budget.md`,
+> `reviews/decisions/plugin-abi.md`,
+> `reviews/decisions/hot-reload-protocol.md`,
+> `reviews/decisions/fory-codegen.md`,
+> `reviews/decisions/frame-phases.md`. No public surface is introduced
+> beyond the §5.4 stub locked in `specs/content/SPEC.md`; the §5.4
+> normalize-parameter vocabulary refinement below is documented as a
+> pending §5.4 amendment (§12, OQ-1) and routed through a follow-up
+> task-breakdown spike, not a unilateral header edit.
+>
+> Harmonius prior art (`harmonius/docs/requirements/content-pipeline/
+> asset-import.md`, R-12.1.1 / R-12.1.2 / R-12.1.3 / R-12.1.4 /
+> R-12.1.5) cited as research input only — every conclusion below was
+> independently re-derived per `PHILOSOPHY.md` §"How harmonius is
+> used". No glTF / Alembic / USD coverage; the §3.2 collapse #1
+> commitment to one mesh path (FBX SDK only) is preserved, not
+> revisited.
+>
+> Refs: spike #807 — `[SPIKE] design-content-fbx-importer-detailed`.
+> Parent: #806 (sub-epic — Detailed Designs — content). Sibling
+> `[SPIKE] task-breakdown-content-fbx-importer-detailed` is blocked by
+> this deliverable.
+
+## 1. Purpose
+
+The `fbx-importer` aggregate is the single component in the content
+context permitted to link the Autodesk FBX SDK and to call into its
+C++ surface. Its one responsibility is **decoding one
+`SourceAsset { kind = Mesh, format = Fbx }` value object into the
+normalized in-memory bytes of a `glibre.content.MeshArtifact`
+precursor**, packaged as `eastl::span<const std::byte>` allocated
+inside a caller-supplied `ImporterArena`, with every SDK-thrown
+exception translated at first ingress into a typed
+`ImporterError::*` arm and every allocation routed through the
+importer's per-cook arena (PHILOSOPHY §11). Concretely the aggregate
+owns:
+
+1. The **`FbxImporter` final entity** (SPEC §4.1.2) — the closed-sum
+   member that serves `SourceKind::Mesh`, constructed at cook-session
+   start through `FbxImporter::create()` (SPEC §5.4) and destroyed at
+   session end. One `FbxImporter` instance per importer worker thread
+   in the cook worker pool (§6.5 below); the FBX SDK's `FbxManager`
+   is **not** thread-safe across `FbxScene` traversal, so the design
+   pins a `FbxManager`-per-thread topology rather than a process-wide
+   singleton.
+2. The **first-ingress SDK seam** — every Autodesk FBX SDK call site
+   in the engine lives inside `plugins/content/cook/import/
+   fbx_importer.cpp` (SPEC §10.3, the unique `-fexceptions` carve-out
+   for FBX). The seam catches `FbxStatus`-typed soft failures, raw
+   `std::exception` derivatives from the SDK's libxml2 / zlib internal
+   paths, and the unbounded `std::bad_alloc` case, and translates each
+   into a typed `ImporterError::*` arm with a structured-log
+   `error.detail` prefix per the §10.3 classification table.
+3. The **scene-graph normalize pipeline** — the deterministic walk
+   over `FbxScene` → `FbxNode` → `FbxMesh` / `FbxSkeleton` /
+   `FbxCluster` that produces the `MeshArtifact` precursor's
+   in-memory layout (vertex streams in engine-canonical layout, index
+   streams, skeleton + scene-hierarchy nodes, per-vertex skinning
+   weights, animation curves *deferred* per §3.6, materials *delegated
+   to texture importer* per §3.6). The walk is pure-of-effect outside
+   the per-cook arena (SPEC §4.1.2 inv #3) and respects the
+   `CancellationToken` polled at every node boundary (SPEC §4.1.2
+   inv #5).
+4. The **`importer_version` digest** — a 32-byte BLAKE3 over the
+   importer's compiled-in identity (FBX SDK version string + major
+   normalize-params-vocabulary version + post-process options
+   vocabulary), embedded as a string literal in the importer's
+   translation unit at build time and returned through
+   `Importer::version()` (SPEC §5.4). The digest is one of the five
+   ingredients of `CookKey` (SPEC §4.1.3 component #2) and the rule
+   that bumping the SDK forces re-cook of every dependent.
+5. The **per-cook arena lifetime** — `ImporterArena` (the typed alias
+   over `glibre::PerContextAllocator` per `perf-budget.md` Allocator
+   Rule 1, SPEC §9.3.1) holds every byte the importer allocates: the
+   FBX SDK's internal allocations (re-routed through the SDK's
+   `FbxMemoryAllocator` hook), the normalized vertex / index streams,
+   the skeleton table, the scene-hierarchy node array, and any
+   transient scratch the normalize pipeline needs. The arena drains
+   once `CookSession`'s end-of-session rollback / publish path runs
+   (SPEC §4.1.9 inv #1, #5) and never bleeds into the runtime heap.
+
+This aggregate **refuses to own**:
+
+- **Mesh-internal geometry processing** — meshlet partitioning,
+  vertex-cache reorder, automatic LOD chain authoring, BLAS
+  construction, lightmap UV unwrapping, mikktspace tangent
+  re-derivation. SPEC §3.3 routes these to the `geometry` plugin;
+  the cook-step orchestrator (§6.2 stage 3) calls into geometry's
+  cook-step API after the importer returns its precursor. The
+  importer hands geometry the canonical streams and walks away.
+- **CAS write** — `CAS::put(CookedAsset)` (SPEC §4.1.5, §5.7) is the
+  cook-step's stage 5 (SPEC §6.2). The importer does not touch
+  `cooked/<prefix>/<hash>` and never sees a `ContentHash`.
+- **Manifest publish** — `Manifest::publish` (SPEC §4.1.6) is the
+  cook-session's end-of-session step (SPEC §6.2 manifest-publish
+  paragraph). The importer does not see an `AssetId` outside the
+  `SourceAsset`'s identity-on-cook context; it never resolves
+  `AssetId → ContentHash`.
+- **Cook-key derivation** — `make_cook_key` (SPEC §5.5, §4.1.3) is
+  authored in `cook/cook_key.cpp` and consumes the importer's
+  `version()` value as one of five ingredients. The importer does
+  not author the digest; it only contributes its compiled-in
+  identity. This is the SRP split that keeps "what counts as the
+  same cook" (`CookKey`) separate from "the SDK seam"
+  (`FbxImporter`).
+- **Fory serialization of the precursor** — the importer returns
+  pre-Fory bytes (an in-memory layout the next cook-step stage
+  Fory-encodes per SPEC §6.2 stage 4). Fory codegen and the
+  `glibre-types.dylib` middleman are owned by `data`
+  (`reviews/decisions/fory-codegen.md`); the importer is a producer
+  of bytes that the next stage Fory-encodes. The §3.4 below pins
+  the precursor's **in-memory** layout — distinct from the Fory
+  envelope shape that lands in the CAS.
+- **Source-file I/O mechanics** — the importer reads through
+  `platform::FileIo` via the `SourceAsset.path` resolved by
+  `cook/session.cpp` stage 1 (SPEC §6.2). The watcher seam
+  (SPEC §4.1.10), the FSEvents callback, and the OS-level `read(2)`
+  mechanics live in `platform`. The importer does not touch any
+  path other than what the cook session hands it through
+  `SourceAsset`.
+- **Source debounce / dedup** — the watcher's debounce window
+  (SPEC §4.1.10 inv #2) collapses rapid resaves into one
+  `RecookRequest` before any importer runs. The importer never
+  sees back-to-back imports of the same `(path, source_hash)` pair
+  inside one cook session.
+- **Audio source decode** — explicitly refused per SPEC §3.3
+  (R-12.1.3 → `audio` plugin). The importer rejects any non-`Mesh`
+  `SourceKind` at construction (the closed-sum dispatch in
+  `importers/dispatch.cpp` makes this a compile-time-exhaustive
+  branch; SPEC §4.1.2 inv #1).
+- **Texture / image / font ingest** — those are the
+  `FreeImageImporter` and `FreeTypeImporter` siblings (SPEC §4.1.2,
+  detailed designs are sibling spikes #810 / #816). The FBX path
+  may reference embedded texture sub-assets (§3.6 below) but
+  delegates their decode to `FreeImageImporter` rather than
+  inlining the decode here.
+- **Schema authoring and codegen** — `glibre.content.MeshArtifact`
+  the Fory schema is authored under `data/schemas/content/
+  MeshArtifact.fory` per `reviews/decisions/fory-codegen.md`. The
+  importer is a producer of bytes conforming to that schema; it
+  never defines the schema.
+- **Plugin loader / hot-reload bodies** — `core::PluginLoader` and
+  the manifest-pointer flip protocol (SPEC §8.2) are the loader's;
+  the importer participates as a build-time-only TU under
+  `tools/glibre-cook` (SPEC §6.1 lifecycle seam, SPEC §6.5
+  shipping-cuts table) and ships **no runtime symbol** in
+  `glibre-content.dylib`.
+- **Obj-C / Obj-C++ glue** — CLAUDE.md "No Obj-C++ in engine code".
+  The Autodesk FBX SDK is plain C++ (linked from Autodesk's
+  pre-built shared library); no `.mm` translation unit appears in
+  this aggregate's source.
+
+The aggregate's SRP boundary is sharp: if the FBX SDK version moves,
+if the normalize-params vocabulary mutates (axis-convention canonical
+form, unit-scale conversion rule, UV-layer selection policy), or if
+the per-vertex skinning encoding changes, **this design changes**.
+Anything else — manifest layout, cook-session orchestration, residency
+policy, Fory schema body, geometry meshlet algorithm — is out of scope.
+
+## 2. Requirements coverage
+
+Mapping of harmonius asset-import requirements
+(`harmonius/docs/requirements/content-pipeline/asset-import.md`,
+R-12.1.1 .. R-12.1.5) onto MVP coverage in this aggregate. Every
+entry is independently re-derived; coverage sites refer to sections
+of `specs/content/SPEC.md` and to the design sections below.
+
+| Harmonius clause                                                                              | Glibre disposition (MVP)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+|-----------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **R-12.1.1** native binary format ingestion with magic / version / hash validation            | **Refused for the FBX importer; collapsed.** R-12.1.1 names a glibre-native pre-cooked binary format (DCC plugin export → engine ingest). Glibre §3.2 collapse #1 explicitly removes that intermediate format: the FBX SDK ingests `.fbx` directly. The validation contract R-12.1.1 names — magic / version / hash — is preserved on the **`SourceAsset → CookedAsset`** seam: SPEC §4.1.1 inv #3 (`source_hash = BLAKE3(bytes)`), §4.1.4 inv #1 (`content_hash = BLAKE3(payload)`), and SPEC §4.1.2 inv #2 (SDK reports `MagicMismatch` / `UnsupportedVersion` at first ingress; §3.7 step 1 below). The "DCC plugin native export" surface re-enters post-MVP behind the same `Importer` seam if needed; the FBX importer covers MVP. |
+| **R-12.1.2** texture source import (PNG / JPEG / EXR / HDR / TIFF) with sRGB / linear decode | **Out of scope (sibling).** Texture decode is the `FreeImageImporter` aggregate (sibling spike #810). The FBX importer **does not** decode embedded texture sub-assets; it emits the texture's source-tree path (resolved per §3.6 below) into the `MeshArtifact` precursor's `material_refs` field, and the cook-session orchestrator schedules the texture cook as a separate `RecookRequest` driven by the `DependencyEdge` graph (SPEC §4.1.10 inv #4).                                                                                                                                          |
+| **R-12.1.3** audio source import (WAV / FLAC / Ogg Vorbis)                                    | **Refused (out of context).** SPEC §3.3 routes audio source decode to the `audio` plugin. The FBX importer's `SourceKind` is closed-sum on `Mesh`; the dispatch table (SPEC §6.1 `importers/dispatch.cpp`) compile-time-rejects any other kind.                                                                                                                                                                                                                                                                |
+| **R-12.1.4** schema validation; errors include source path + byte offset + fix suggestion     | **Covered, scoped.** Path is the structured-log field `source_path` (SPEC §10.5); byte offset is recovered from the FBX SDK's `FbxStatus` line/column when available and embedded in `error.detail` (§3.7 step 2 below); fix-suggestion is the per-arm "Operator action" column from SPEC §10.1 (Class B refusal) — re-export at supported version, restore source under `assets/source/`, break dependency cycle, etc. The structured-log handler formats these uniformly (§3.7 step 5 below).             |
+| **R-12.1.5** parallel batch import with progress + cancellation + rollback                    | **Covered, partitioned.** Parallelism and progress tracking are the `CookSession` aggregate's concerns (SPEC §4.1.9, §6.2): the worker pool runs one `FbxImporter::import_one` invocation per scheduling slot; `CookSession::commit()` returns a `CookReport { cooks_executed, cache_hits }` (SPEC §5.12). Cancellation is observed at every `FbxNode` traversal boundary in the importer (§3.7 step 4 below) per SPEC §4.1.2 inv #5; rollback semantics live with the session (SPEC §4.1.9 inv #1).      |
+
+Glibre-native requirements added beyond harmonius:
+
+- **Per-cook arena allocation discipline** (PHILOSOPHY §11; SPEC
+  §9.3.1; `perf-budget.md` Allocator Rule #1). Every byte the
+  importer allocates — including the FBX SDK's internal allocations
+  rerouted via `FbxMemoryAllocator` (§3.5 step 2 below) — comes from
+  the per-cook arena tagged `ContextTag::content`. The runtime heap
+  is never touched by importer code paths; raw `new` / `malloc` is
+  rejected by `-Wglibre-no-raw-alloc`. The 64 MiB soft sub-ceiling
+  (SPEC §9.3.1) bounds the largest single FBX import.
+- **Single `-fexceptions` carve-out, narrow boundary** (SPEC §10.3,
+  `error-model.md` Decision rule 3 + Consequences last bullet). The
+  carve-out is `plugins/content/cook/import/fbx_importer.cpp` only;
+  the corresponding header `fbx_importer.hpp` and every other TU in
+  the engine compile with `-fno-exceptions`. The `try / catch` block
+  inside `import_one` is the **single** translation site (§3.7
+  step 6 below); no exception ever crosses the importer's public
+  boundary (SPEC §4.1.2 inv #2).
+- **Deterministic normalize pipeline** (PHILOSOPHY §7,
+  cross-aggregate invariant SPEC §4.2 #1). Two cooks of the same
+  source bytes with the same `importer_version` and the same
+  `NormalizeParams` produce **byte-equal** precursor bytes — and
+  hence byte-equal cooked `MeshArtifact` payloads, byte-equal
+  `ContentHash`. The normalize pipeline is straight-line code with
+  no platform intrinsics, no `std::unordered_*` (replaced by
+  `eastl::vector_map` for deterministic iteration order), no PRNG,
+  no clock reads, no thread-local caches. §3.7 step 3 below pins
+  the determinism rule per-stage.
+- **Importer is stateless across cooks; per-cook arena is the only
+  state** (SPEC §4.1.2 identity rule). The `FbxImporter::Impl`
+  pimpl carries only the `FbxManager*` + `FbxIOSettings*` pair
+  (constructed at `create()` time, owned for the importer's
+  lifetime); every `import_one` call resets its scratch view of the
+  arena and produces a fresh precursor. No memoization, no caches
+  keyed off prior cooks.
+- **Cancellation is polled, not preemptive** (SPEC §4.1.2 inv #5).
+  The importer polls `CancellationToken::is_cancelled()` at every
+  `FbxNode` boundary (the natural traversal granularity). Bound:
+  ≤ ~1 ms latency to observe cancellation on an MVP-scale FBX
+  (~200 nodes per ~5 MiB FBX, S3 fixture; §9 below).
+- **No animation-curve / morph-target ingest at MVP** (§3.6 refusal
+  list below; §12 OQ-2). FBX scenes carry `FbxAnimStack` /
+  `FbxAnimLayer` / `FbxBlendShape` data; MVP imports the bind-pose
+  skeleton + skinning weights only. Animation curves enter post-MVP
+  through a `data/schemas/content/AnimationClipArtifact.fory`
+  authored under a future spike behind the same importer seam.
+
+Coverage rule: every harmonius MVP-scope requirement above either
+lands in this design (with a coverage site) or is refused with a
+one-line rationale routed to the owning context. No silent drops.
+
+## 3. Detailed model
+
+### 3.1 Aggregate composition
+
+```text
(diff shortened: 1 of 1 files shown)


specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1866 @@
...
+arena via a generated emitter
+(`glibre/content/mesh_artifact_precursor.hpp`, the Fory codegen
+output's pre-Fory CPU-side struct). The emitter is a straight-line
+walk over the canonicalized arrays from stage 5; no allocation outside
+the arena. The precursor's CPU-side layout is **distinct** from the
+Fory envelope shape (which the cook step's stage-4 produces — SPEC
+§6.2 stage 4); the importer never emits Fory bytes itself.
+
+The returned `eastl::span<const std::byte>` points into the arena;
+the span is valid for the cook session's lifetime per SPEC §9.3.1
```
cjhowe-us, line 719: severity:MED
location:specs/content/fbx-importer-design.md:668
problem:Inconsistent include path for MeshArtifactPrecursor. Stage 6 cites `glibre/content/mesh_artifact_precursor.hpp`, but §3.9 (lines 726-727) and §7.1 correctly cite `data/codegen-output/include/glibre/types/content/mesh_artifact_precursor.hpp`. The fory-codegen.md decision places all generated headers under `${CMAKE_BINARY_DIR}/generated/glibre-types/include/glibre/types/<ctx>/`, so the Stage 6 path is wrong.
fix:Change the Stage 6 sentence to `glibre/types/content/mesh_artifact_precursor.hpp` and audit every other occurrence in the document to use the same single path.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1866 @@
...
+
+```cpp
+// Sketch — final shape lives in fbx_importer.cpp.
+auto FbxImporter::import_one(const SourceAsset& src,
+                             const NormalizeParams& params,
+                             ImporterArena& out_arena,
+                             const CancellationToken& cancel) noexcept
+    -> Result<eastl::span<const std::byte>>
+{
+    try {
```
cjhowe-us, line 1680: severity:MED
location:specs/content/fbx-importer-design.md:1591
problem:The SDK-exception catch sketch at §10.3 includes `catch (const ::fbxsdk::FbxStatus& st)` but FbxStatus is NOT an exception class in the Autodesk FBX SDK — it is a status-code struct returned from SDK calls (e.g. via `FbxManager::GetStatus()` or as a pass-by-reference out-parameter on `FbxImporter::Import`). This catch arm would never fire at runtime; the FbxStatus translation described in §10.2 rows 1-4 must instead be implemented by inspecting the return value of SDK calls rather than catching a thrown exception. If left as-is in the sketch, an implementer reading this document would write unreachable code that silently drops all FbxStatus-based errors.
fix:Remove `catch (const ::fbxsdk::FbxStatus& st)` from the sketch. Replace it with a note that FbxStatus error codes are checked on SDK-call return values (e.g. `fbx_importer_->GetStatus().GetCode()` after `Import(scene_)` returns false), and that those checks are in the stage-1/stage-2 inline code rather than the catch chain. Retain the std::exception and catch(...) arms unchanged.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1866 @@
...
+- **No raw `new` / `malloc`**. Per `perf-budget.md` Allocator Rules
+  header + SPEC §9.4 rule 7. The build rejects raw allocations in
+  `plugins/content/cook/import/`. The FBX SDK's vendor-internal
+  `malloc` is the third-party-wrap exemption (§4.1.2 inv #3), routed
+  through the `FbxMemoryAllocator` hook; this is the only third-party
+  allocator carve-out in the engine.
+
+### 6.4 Cancellation propagation
+
+SPEC §4.1.2 inv #5 + SPEC §4.1.9 inv #5 require cancellation to
```
cjhowe-us, line 1177: severity:MED
location:specs/content/fbx-importer-design.md:1116
problem:§6.3 notes that `FbxMemoryAllocator` routes SDK allocations through `glibre::PerContextAllocator`, and perf-budget.md Allocator Rule 2 says that in `GLIBRE_ALLOC_STRICT=1` mode the allocator returns `std::unexpected{OutOfBudget}` when the ceiling is hit. However `FbxMemoryAllocator` (the Autodesk callback hook) receives a C-style allocate signature that returns `void*` — a null pointer on failure, not a Result type. The design does not address what the `FbxMemoryAllocator` hook does when `PerContextAllocator` refuses a strict-mode allocation: returning null from the hook causes the SDK to invoke its null-pointer handling path (undefined behavior or an internal error path, not a thrown `std::bad_alloc`). The `std::bad_alloc` catch in §10.3 therefore cannot cover the strict-mode over-budget case as described.
fix:Add a sentence in §6.3 clarifying the `FbxMemoryAllocator` null-return protocol: when the allocator returns null (GLIBRE_ALLOC_STRICT exhausted), the importer must check the allocator's status and return `ImporterError::MalformedPayload` with a new `error.detail = "fbx-alloc-exhausted"` before any SDK code dereferences the null. Alternatively, document that the strict-mode ceiling is chosen conservatively enough that null can never occur (which would require a hard-coded minimum to exceed the 64 MiB soft ceiling). Either way, pin the behavior in the §10.2 translation table.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1866 @@
...
+- **`ResidencyError::*`** (SPEC §5.3) are emitted by the runtime's
+  residency manager (SPEC §4.1.7), not by the importer. The cook
+  tool may emit `ResidencyError::IoFailure` from `cas/atomic_write.cpp`
+  during stage 5 of the cook step (CAS write), but that is the cook
+  step's emit site, not the importer's.
+- **`core::Error::*`** arms (`PluginAbiHashMismatch`,
+  `SchemaMigrationFailed`, etc.) are loader-side and only relevant
+  to the runtime plugin reload (§8.2 above). The importer never
+  emits these.
+
```
cjhowe-us, line 1787: severity:LOW
location:specs/content/fbx-importer-design.md:1658
problem:Test `byte_equal_across_macos_and_linux_ci` implies CI has a Linux runner and that the FBX SDK is available on Linux with vcpkg-pinned versions. The project is macOS-first (CLAUDE.md; macOS 26 / Apple Silicon baseline) and the FBX SDK has no upstream vcpkg port (acknowledged in §7.3, which notes an overlay port is needed). The test assumes both conditions will be met without documenting who owns provisioning the Linux CI runner and the FBX SDK overlay port on Linux.
fix:Either (a) tag this test as deferred/conditional until the FBX SDK Linux overlay port is authored and a Linux runner is provisioned, or (b) add a note in §11.4 naming the prerequisite spike that creates the Linux CI infrastructure before this gate can be made mandatory.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1866 @@
...
+    }
+    catch (const std::exception& e) {
+        return std::unexpected{ Error{
+            ImporterError::MalformedPayload,
+            ErrorContext{ __FILE__, __LINE__, "fbx-unclassified" } }};
+    }
+    catch (...) {
+        std::terminate();                 // §10.2 row 7
+    }
+    // Stage 7 (SDK teardown) is in a finally-equivalent guard
```
cjhowe-us, line 1695: severity:LOW
location:specs/content/fbx-importer-design.md:1609
problem:`eastl::scoped_exit` does not exist in EASTL. The EASTL library provides no scope_exit or scoped_guard utility. The design references it as `eastl::scoped_exit` and immediately parenthesizes '(or a hand-written equivalent)', but leaves the implementation completely unspecified, meaning the task-breakdown spike has no concrete guidance on what RAII shape to use.
fix:Replace the parenthetical with a concrete recommendation: either (a) name an existing EASTL mechanism such as `eastl::finally` (if available in the pinned EASTL version — check vcpkg.json), or (b) specify a hand-written `ScopeExit` local helper with the signature `struct ScopeExit { std::function<void()> fn; ~ScopeExit() { fn(); } };` (noting that PHILOSOPHY §11 bans `std::function`; so a no-capture lambda + template parameter would be the correct approach), or (c) route to a `GLIBRE_DEFER` macro to be authored in `core/include/glibre/defer.hpp`. One approach must be picked to avoid scope creep in the implementation plan.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1866 @@
...
+This is asserted by the §9.3 gate #2 (`s3_off_thread_no_driver_spike`)
+and structurally by SPEC §9.3.2's exhaustive driver-thread hot-path
+list excluding the importer. Any future change that would invoke
+`import_one` from the driver thread is rejected at SPEC §9.3.2 and
+SPEC §6.5 cut review.
+
+## 10. Failure modes
+
+The importer's failure surface is the closed sum
+`glibre::content::ImporterError` declared in SPEC §5.3 and §10.1.
```
cjhowe-us, line 1564: severity:LOW
location:specs/content/fbx-importer-design.md:1503
problem:§10 opens by citing `glibre::content::ImporterError` as declared in SPEC §5.3. However, the SPEC §5 public surface (as visible in SPEC.md) declares `ImporterError` inside namespace `glibre::content` and assembles it into `content::Error` = `eastl::variant<ImporterError, ResidencyError>`. The error-model.md decision says the engine-wide `glibre::Error` is the tagged union at public boundaries. The design doc uses `ImporterError::*` arms throughout correctly for the internal cook surface, but never explicitly states how `ImporterError` is wrapped into `glibre::Error` before crossing the `content` public boundary (the `glibre::content::Error` → `core::Error::ContentImport` lift). This leaves the cross-boundary error wrapping implicit.
fix:Add one sentence in §10 (or the §10.1 table header) citing SPEC §5.3: `ImporterError` values are wrapped into `glibre::content::Error` (the `eastl::variant<ImporterError, ResidencyError>` declared in SPEC §5.3), which is itself a leaf of the engine-wide `glibre::Error` variant per error-model.md. The Result<T> alias in the public surface already carries this; making the wrapping chain explicit closes the gap.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:3
  LOW:3
carryover:0
notes:The fbx-importer detailed design satisfies its primary deliverable contract (#807 spike) and demonstrates strong coverage across all twelve required sections. R-12.1.1 through R-12.1.5 harmonius requirements are either covered with concrete rationale or explicitly refused with routing decisions. The §5.4 public surface correctly uses Result<T> (= std::expected<T, glibre::Error>) at every boundary, carries no Obj-C++, and uses eastl::unique_ptr / eastl::span / eastl::vector<T, ImporterArenaAllocator> throughout — EASTL discipline holds. The §11 test plan names 20+ concrete Catch2 test case names covering dispatch, refusal, cancellation, determinism, off-thread soundness, and SDK-exception translation; the test naming convention and test-to-invariant mapping is good. CC subject "docs(content): detailed design — fbx-importer (refs #807)" is correct Conventional Commit format with issue reference. Six decision records cited in the preamble all exist on disk. Three MED findings require fixes before the task-breakdown spike (#807 follow-up) can treat this document as authoritative: (1) the §10.3 catch sketch incorrectly catches FbxStatus as a thrown exception when it is a return-value struct — the catch arm would be dead code and the status-to-arm table in §10.2 rows 1-4 would be silently unreachable; (2) two inconsistent paths for the MeshArtifactPrecursor header (glibre/content/ vs glibre/types/content/ — only the latter is consistent with fory-codegen.md); (3) the FbxMemoryAllocator null-return protocol when PerContextAllocator is exhausted in strict mode is not pinned, leaving a gap between the std::bad_alloc termination claim and the actual allocator-null path. Three LOW findings (Linux CI prereq not gated, eastl::scoped_exit non-existence, ImporterError-to-glibre::Error wrapping chain implicit) are advisory. No HIGH findings. The MED items are in-scope fixes for the impl-response agent without scope expansion.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3bb23bc

Stage 6 include path corrected from `glibre/content/mesh_artifact_precursor.hpp` to `glibre/types/content/mesh_artifact_precursor.hpp` (canonical generated-header path per fory-codegen.md §Pipeline). §3.9 and §7.1 already used the correct path; Stage 6 prose now matches.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3bb23bc

Removed the dead `catch (const ::fbxsdk::FbxStatus& st)` arm. FbxStatus is a return-value struct (not throwable); the catch arm would have been unreachable. Replaced with a comment block inside the sketch explaining that FbxStatus error codes are read via `fbx_importer_->GetStatus().GetCode()` after SDK calls that return false, and that `classify_fbx_status` is called inline at those check sites (stages 1–2). Relabeled §10.2 as 'SDK error-code and exception translation table' and added a clarifying paragraph at the section head distinguishing rows 1-4 (return-value checks) from rows 5-7 (thrown exceptions).

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3bb23bc

Added a new bullet to §6.3 pinning the FbxMemoryAllocator null-return protocol for GLIBRE_ALLOC_STRICT=1. Protocol: the hook records the refusal via an atomic error flag on FbxImporterImpl and returns nullptr; import_one checks the flag before each SDK entry point (Initialize, Import) and returns ImporterError::MalformedPayload with error.detail="fbx-alloc-exhausted" before any null dereference can occur. Also documented that the 64 MiB sub-ceiling is sized above the S3 peak (~30-50 MiB) so this path is a defect guard rather than a routine production path, and that the GLIBRE_ALLOC_STRICT CI gate validates this property.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3bb23bc

Tagged the byte_equal_across_macos_and_linux_ci test as '(deferred — conditional on two prerequisites)': (1) Linux CI runner provisioning, and (2) FBX SDK vcpkg overlay port authored and validated on Linux. Added a note naming the task-breakdown-content-fbx-importer-detailed planning spike as the prerequisite task owner, and clarifying that byte_equal_across_two_runs_same_host is the mandatory determinism gate until both prerequisites land.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3bb23bc

Replaced the eastl::scoped_exit reference (EASTL has no such utility) with a concrete recommendation: a GLIBRE_DEFER macro backed by a zero-capture template ScopeExit guard, to be authored in core/include/glibre/defer.hpp. Provided the exact struct+macro shape in the design doc. Noted that std::function is banned by PHILOSOPHY §11 and that the lambda+template approach has zero overhead. Noted that authoring defer.hpp is a prerequisite task for the implementation plan.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3bb23bc

Added an explicit cross-boundary wrapping chain paragraph at the opening of §10. The paragraph states: ImporterError values are leaves of the content context's error sum; the cook session wraps them into glibre::content::Error (the eastl::variant<ImporterError, ResidencyError> from SPEC §5.3), which is itself a leaf of the engine-wide glibre::Error variant per error-model.md Composition Rules 1-2; import_one returns the inner arm value and the cook session performs the wrapping at the public boundary.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1941 @@
...
+
+Impl  (private; lives in fbx_importer.cpp)
+├── manager_           FbxManager*                       (Autodesk root; one per worker thread)
+├── io_settings_       FbxIOSettings*                    (configured at create())
+├── memory_allocator_  FbxMemoryAllocator                (routes SDK allocs to per-cook arena)
+├── progress_callback_ FbxProgress                       (cancellation-polling shim, §3.7 step 4)
+└── version_string_    eastl::string_view                (e.g. "FBX SDK 2025.0/normalize-v1/post-v1";
+                                                          the source from which version_ is BLAKE3'd)
+```
+
```
cjhowe-us, line 265: severity:MED
location:specs/content/fbx-importer-design.md:253
problem:The Impl struct diagram at §3.1 lists five members (manager_, io_settings_, memory_allocator_, progress_callback_, version_string_) but does NOT include the alloc_exhausted_ atomic error flag introduced by the MED-3 fix in §6.3 and commit 3bb23bc. The MED-3 fix states the hook records the refusal in an atomic error flag on the FbxImporterImpl but that field is absent from the §3.1 structural diagram. Any implementer using §3.1 as the authoritative layout guide will produce an Impl that has no field to record the refusal, causing the import_one pre-call flag check in §6.3 to reference a non-existent member.
fix:Add alloc_exhausted_ to the §3.1 Impl struct diagram: alloc_exhausted_  std::atomic<bool>   (set by FbxMemoryAllocator hook on null-return; checked before every SDK entry point per §6.3). Per PHILOSOPHY §11, std::atomic is one of the retained std:: utilities. The field belongs in the Impl diagram because it is part of the aggregate persistent per-importer-instance state, not per-call scratch.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1941 @@
...
+```cpp
+// core/include/glibre/defer.hpp (to be authored)
+template <typename F>
+struct ScopeExit {
+    F fn;
+    ~ScopeExit() { fn(); }
+};
+#define GLIBRE_DEFER(expr) \
+    ScopeExit GLIBRE_DEFER_ANON(__LINE__){[&]() noexcept { expr; }}
```
cjhowe-us: severity:LOW
location:specs/content/fbx-importer-design.md:1673
problem:The GLIBRE_DEFER macro sketch introduced by the LOW-5 fix is incomplete and internally contradictory. (1) GLIBRE_DEFER_ANON(__LINE__) is used but never defined — GLIBRE_DEFER_ANON is not a standard macro; the conventional two-level paste trick requires two helper macros (e.g. GLIBRE_DEFER_CAT(x,y) x##y and GLIBRE_DEFER_ANON(n) GLIBRE_DEFER_CAT(defer_guard_,n)). Without those definitions, the sketch produces a compile error on every use. (2) The surrounding prose says zero-capture template guard but the macro body uses [&]() capture-by-reference — this is a contradiction. A capture-by-reference lambda is not zero-capture; if the scope exits before the destructor fires (unlikely here but architecturally wrong), the reference dangling is silent UB. The canonical zero-capture pattern either passes the expression as a non-capturing lambda or wraps specific pointers by value.
fix:Either (a) define GLIBRE_DEFER_ANON in the defer.hpp sketch with the two-level paste helpers and change the prose to say capture-by-reference lambda (accurate) not zero-capture, or (b) change [&] to explicit captures of the two SDK pointers by value: [fbx_importer_ = fbx_importer_, scene_ = scene_]() noexcept { ... }, and keep the zero-capture prose. Option (b) is architecturally safer and makes the guard independent of the enclosing scope lifetime. The GLIBRE_DEFER_ANON helper must be shown in the sketch regardless.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,1941 @@
...
+
+**Memory allocator hook.** The FBX SDK exposes
+`FbxSetMemoryAllocator(const FbxMemoryAllocator*)` (a static, on the
+SDK side). The importer installs a process-wide allocator at first
+`create()` call that routes every SDK allocation through
+`glibre::PerContextAllocator` tagged `ContextTag::content`. The
+allocator records the requesting `FbxManager` pointer so the soft
+sub-ceiling (SPEC §9.3.1, 64 MiB) can be enforced per-importer-cook.
+The hook is set exactly once per process; subsequent `create()` calls
+verify the existing hook is the engine's and skip re-installation.
```
cjhowe-us, line 506: severity:MED
location:specs/content/fbx-importer-design.md:497
problem:The FbxMemoryAllocator seam has a hidden SRP coupling that is not fully specified. FbxSetMemoryAllocator() is a process-wide C function accepting a single FbxMemoryAllocator callback struct — the SDK only supports one installed allocator per process. With four worker threads each holding their own FbxImporter instance (each with their own memory_allocator_ Impl member), only the first worker's FbxMemoryAllocator struct is live in the SDK; the remaining three are installed then discarded. The design acknowledges this (hook set once per process) and claims the allocator routes per-cook by recording the FbxManager pointer. However the design never pins how the process-wide allocator struct maintains the FbxManager* -> per-thread arena mapping: the struct must carry either (a) a thread-local lookup or (b) a thread-safe map from FbxManager* to ImporterArena*. Without specifying this, the alloc_exhausted_ flag approach from §6.3 is incomplete — if the process-wide struct does not know which FbxImporterImpl to set the flag on, it cannot set it on the correct worker's flag.
fix:Add a §3.5 paragraph specifying the internal dispatch mechanism. The simplest correct approach is: the process-wide FbxMemoryAllocator struct holds a thread_local ImporterArena* pointer that each worker sets to its own per-call arena immediately before calling any SDK entry point and clears after stage 7 teardown. The alloc_exhausted_ flag can then be thread_local too or live in the FbxImporterImpl looked up through a thread-local pointer. This must be specified before the task-breakdown spike authors the implementation plan.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0

notes:All six round-1 findings were addressed in commit 3bb23bc and the impl-respond replies confirm ADDRESSED decisions for every one; no round-1 carryover is escalated. The round-2 cohesion + SRP + seam quality pass found three new findings. MED-1: the Impl struct diagram at §3.1 was not updated when the MED-3 alloc_exhausted_ atomic flag was added to §6.3, leaving the authoritative layout diagram inconsistent with the §6.3 protocol it is supposed to describe — any implementer will produce a missing field. MED-2: the process-wide FbxSetMemoryAllocator seam is structurally under-specified; the design correctly identifies that one allocator hook serves all four worker threads but does not pin the thread-to-arena dispatch mechanism (thread_local pointer, per-FbxManager* map, or similar) that the alloc_exhausted_ per-Impl flag depends on for correctness under parallel cooks. LOW-1: the GLIBRE_DEFER macro sketch is incomplete (GLIBRE_DEFER_ANON is undefined) and internally contradictory (prose says zero-capture but macro uses [&] capture-by-reference). No HIGH findings. SRP boundaries are otherwise clean: the aggregate refuses CAS write, manifest publish, residency, GPU upload, Fory schema authoring, and audio correctly; the -fexceptions carve-out is a single .cpp TU; EASTL discipline holds throughout (eastl::unique_ptr, eastl::span, eastl::vector with arena allocators, no std:: containers); the error wrapping chain paragraph added by LOW-6 fix is accurate against error-model.md; the public surface at §4 is minimal and unchanged from the SPEC §5.4 stub; hot/cold path split is coherent and enforced at link time per §6.5. The two MED findings are in-scope design-document fixes with no scope expansion; the LOW fix is a single-paragraph clarification.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dc8eeb1

Added `alloc_exhausted_  std::atomic<bool>` to the §3.1 Impl struct diagram with inline comment citing §6.3 as the authority for the field (set by FbxMemoryAllocator hook on null-return; checked before every SDK entry point per §6.3; std::atomic is a retained std:: utility per PHILOSOPHY §11). The diagram is now consistent with the §6.3 alloc-exhausted protocol so implementers following §3.1 as the authoritative layout guide will include the field.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dc8eeb1

Added a §3.5 thread_local dispatch paragraph specifying the mechanism: two thread_local pointers (tl_active_arena and tl_active_impl) that each cook worker sets immediately before calling any SDK entry point and clears after stage 7 teardown (via the GLIBRE_DEFER guard). The process-wide FbxMemoryAllocator callback reads tl_active_arena to dispatch the allocation and tl_active_impl->alloc_exhausted_ to record a null-return. Null tl_active_arena in the callback indicates a spurious SDK call outside of an active cook and returns nullptr without setting the flag. This mechanism is safe under the four-worker parallel topology because each worker thread has its own TLS slot; no cross-thread synchronization is needed for dispatch.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dc8eeb1

Fixed both defects in the GLIBRE_DEFER sketch: (1) Added the two-level token-paste helpers (GLIBRE_DEFER_CAT2, GLIBRE_DEFER_CAT, GLIBRE_DEFER_ANON) so GLIBRE_DEFER_ANON(__LINE__) now compiles correctly. (2) Changed the macro body to accept an expression (the lambda is now passed explicitly by the caller), updated the prose from 'zero-capture' to 'explicit-capture', added a usage example showing [fbx_importer_ = fbx_importer_, scene_ = scene_]() noexcept capture-by-value pattern. The guard is now independent of the enclosing scope lifetime — reference dangling UB is eliminated.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,2006 @@
...
+| **R-12.1.2** texture source import (PNG / JPEG / EXR / HDR / TIFF) with sRGB / linear decode | **Out of scope (sibling).** Texture decode is the `FreeImageImporter` aggregate (sibling spike #810). The FBX importer **does not** decode embedded texture sub-assets; it emits the texture's source-tree path (resolved per §3.6 below) into the `MeshArtifact` precursor's `material_refs` field, and the cook-session orchestrator schedules the texture cook as a separate `RecookRequest` driven by the `DependencyEdge` graph (SPEC §4.1.10 inv #4).                                                                                                                                          |
```
cjhowe-us: severity:MED
location:specs/content/fbx-importer-design.md:187
problem:sibling spike #810 is cited as the FreeImageImporter design spike, but #810 = `[SPIKE] task-breakdown-content-texture-importer-detailed` (the task-breakdown, not the design). The design spike is #809 = `[SPIKE] design-content-texture-importer-detailed`. Referring to the task-breakdown as the 'FreeImageImporter aggregate' design produces a dangling cross-reference that will confuse readers of the task-breakdown spike (#808) that inherits this design.
fix:Change 'sibling spike #810' on line 187 to 'sibling spike #809' to reference the actual design spike.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,2006 @@
...
+```
+plugins/content/cook/import/fbx_importer.cpp     # this design
+plugins/content/cook/import/freeimage_importer.cpp # sibling spike #810
+plugins/content/cook/import/freetype_importer.cpp  # sibling spike #816
+```
+
+The header `fbx_importer.hpp` and **every other TU in the engine**
+compiles with `-fno-exceptions`. The build system enforces this via
+a per-TU compile-flag override in the cook tool's CMakeLists; raw
+`add_executable(glibre-cook ...)` configures `-fno-exceptions`
```
cjhowe-us, line 1658: severity:MED
location:specs/content/fbx-importer-design.md:1658
problem:Line 1652 (diff position 1658) cites 'sibling spike #816' for freetype_importer.cpp, but #816 = `[SPIKE] task-breakdown-content-manifest-detailed` — the manifest task-breakdown, not the font importer. #811 = `[SPIKE] design-content-font-importer-detailed` is the correct design spike. Also line 1651 references '#810' for freeimage_importer.cpp (same issue as line 187: #810 is the texture task-breakdown, #809 is the texture design spike). These wrong numbers will mislead implementers working from spike #808 (the fbx task-breakdown).
fix:Change 'sibling spike #810' to 'sibling spike #809' (texture importer design) and 'sibling spike #816' to 'sibling spike #811' (font importer design) in the -fexceptions carve-out TU list.

specs/content/fbx-importer-design.md
```diff
@@ -0,0 +1,2006 @@
...
+The precursor (the bytes `import_one` returns) is the input to
+the cook step's Fory-encode stage. Its in-memory layout is **not**
+the Fory envelope shape; it is a packed CPU-side struct emitted by
+the codegen rule under `data/codegen-output/include/glibre/types/
+content/mesh_artifact_precursor.hpp` (per `fory-codegen.md` rule —
+the codegen tool emits a pre-Fory CPU struct alongside the Fory
+struct for any persistent type the engine wants a non-Fory in-memory
+form of). The shape:
+
+```text
```
cjhowe-us, line 775: severity:LOW
location:specs/content/fbx-importer-design.md:775
problem:The path `data/codegen-output/include/glibre/types/content/` used here (and again at lines 978 and 1242) is inconsistent with `fory-codegen.md` §Pipeline, which specifies the generated output directory as `${CMAKE_BINARY_DIR}/generated/glibre-types/include/glibre/types/<ctx>/`. The same design doc uses the correct form at line 712. Using two different path forms for the same generated header creates confusion for the task-breakdown spike (#808) implementer determining where to include the generated header.
fix:Replace all three occurrences of `data/codegen-output/include/glibre/types/` with `${CMAKE_BINARY_DIR}/generated/glibre-types/include/glibre/types/` (or the prose equivalent 'the build-time generated header at `generated/glibre-types/include/glibre/types/content/`') to be consistent with fory-codegen.md §Pipeline and with line 712 in this document.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0

notes:All round-1 and round-2 findings are confirmed addressed in the head commit dc8eeb1. The round-2 r2 fixes are correct: alloc_exhausted_ field is present in §3.1 with a §6.3 citation, the §3.5 thread_local dispatch paragraph pins tl_active_arena/tl_active_impl with correct lifecycle semantics, and the GLIBRE_DEFER sketch is internally consistent (two-level paste helpers defined, explicit-capture by value used, prose matches). No carryover from prior rounds. Round-3 topology pass surfaced two MED issues and one LOW. MED-1: sibling spike #810 is cited three times (lines 187, 576, 1651) as the FreeImageImporter design spike, but #810 = [SPIKE] task-breakdown-content-texture-importer-detailed; the design spike is #809. MED-2: sibling spike #816 cited at line 1652 as the FreeType importer's -fexceptions TU comment, but #816 = [SPIKE] task-breakdown-content-manifest-detailed; the font importer design spike is #811. These wrong issue numbers matter because spike #808 (the fbx task-breakdown, which is unblocked by this PR) will read these cross-references when deriving plan issues. LOW: the generated header path data/codegen-output/include/glibre/types/ appears at three locations (lines 775, 978, 1242) but fory-codegen.md §Pipeline and line 712 of this same document both specify ${CMAKE_BINARY_DIR}/generated/glibre-types/include/glibre/types/<ctx>/; the two forms are inconsistent. Issue topology is otherwise sound: #807 (this spike) is correctly parented under #806 (sub-epic); #807 carries type:spike + phase:mvp + domain:content labels and no pts:* estimate (correct for a spike); #808 is blocked by #807 (not yet closed) and this PR is OPEN as required. Conventional Commit subject is correct (docs(content): + refs #807). All spec-section citations checked: §4.1.2, §4.1.3, §4.1.9, §4.1.10, §5.4, §6.1, §6.2, §6.5, §7.1.2, §8.5, §9.3.1, §9.3.2, §10.1, §10.3 all exist in the current SPEC.md. Cross-doc references to reviews/decisions/*.md, specs/geometry/cluster-dag-design.md, and specs/platform/file-watcher-design.md all resolve on disk. The two MED findings are simple issue-number corrections in-scope for the impl-respond agent.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0

notes:All round-1 and round-2 findings are confirmed addressed in the head commit dc8eeb1. The round-2 r2 fixes are correct: alloc_exhausted_ field is present in §3.1 with a §6.3 citation, the §3.5 thread_local dispatch paragraph pins tl_active_arena/tl_active_impl with correct lifecycle semantics, and the GLIBRE_DEFER sketch is internally consistent (two-level paste helpers defined, explicit-capture by value used, prose matches). No carryover from prior rounds. Round-3 topology pass surfaced two MED issues and one LOW. MED-1: sibling spike #810 is cited three times (lines 187, 576, 1651) as the FreeImageImporter design spike, but #810 = task-breakdown-content-texture-importer-detailed; the design spike is #809. MED-2: sibling spike #816 is cited at line 1652 as the FreeType importer carve-out comment, but #816 = task-breakdown-content-manifest-detailed; the font importer design spike is #811. These wrong issue numbers matter because spike #808 (the fbx task-breakdown, unblocked by this PR) will read these cross-references when deriving plan issues. LOW: the generated header path data/codegen-output/include/glibre/types/ appears at three locations (lines 775, 978, 1242) but fory-codegen.md §Pipeline and line 712 of this same document both specify the correct build-time path under generated/glibre-types/include/glibre/types/<ctx>/; the two forms are inconsistent within the document. Issue topology is otherwise sound: #807 parented under #806 sub-epic; labels type:spike + phase:mvp + domain:content are correct; no pts:* (correct for spike). Conventional Commit subject is correct. All SPEC section references verified to exist in the current SPEC.md. Cross-doc refs to reviews/decisions/*.md, specs/geometry/cluster-dag-design.md, and specs/platform/file-watcher-design.md all resolve on disk.

cjhowe-us (commented):
round:3 reviewer:go-review verdict:REQUEST_CHANGES findings: HIGH:0 MED:2 LOW:1 carryover:0 notes:All round-1 and round-2 findings confirmed addressed. Round-3 topology surfaced MED-1 (sibling spike #810 cited 3x for FreeImageImporter design but #810 is the task-breakdown spike; correct design spike is #809) and MED-2 (spike #816 cited for FreeType importer TU comment but #816 is manifest task-breakdown; correct font design spike is #811). These wrong issue numbers affect spike #808 task-breakdown. LOW: data/codegen-output path at lines 775/978/1242 is inconsistent with fory-codegen.md Pipeline spec and with line 712 of this document; all three should use the build-time generated/glibre-types/ path. Issue topology correct: #807 parented under #806, labels right, no pts estimate. All SPEC section cross-refs verify. CC subject correct.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:1
carryover:0

notes:All round-1 and round-2 findings confirmed addressed in head commit dc8eeb1. Round-3 topology pass surfaced two MED issues: (1) sibling spike #810 is cited three times as the FreeImageImporter design spike but #810 is task-breakdown-content-texture-importer-detailed; the design spike is #809 design-content-texture-importer-detailed; (2) sibling spike #816 is cited in the -fexceptions carve-out TU comment as the FreeType importer but #816 is task-breakdown-content-manifest-detailed; the font importer design spike is #811. These wrong numbers affect spike #808 (the fbx task-breakdown, blocked by this PR) when deriving plan issues. LOW: data/codegen-output path at lines 775, 978, and 1242 is inconsistent with fory-codegen.md §Pipeline and with line 712 of this same document which correctly uses the CMake build-dir path. Issue topology is otherwise correct: #807 parented under #806 sub-epic, correct type:spike/phase:mvp/domain:content labels, no pts estimate. All SPEC section cross-refs (§4.1.2 through §10.3) verified against current SPEC.md. All decision record file refs and sibling design doc refs (geometry/cluster-dag-design.md, platform/file-watcher-design.md) resolve on disk. Conventional Commit subject correct.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:0c74534

Verified: #809 = [SPIKE] design-content-texture-importer-detailed (the design spike); #810 = [SPIKE] task-breakdown-content-texture-importer-detailed (the task-breakdown, not the design). Four occurrences of the wrong #810 reference corrected to #809: lines 149, 187, 576, and 1651.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:0c74534

Verified: #811 = [SPIKE] design-content-font-importer-detailed (the design spike); #816 = [SPIKE] task-breakdown-content-manifest-detailed (wrong context entirely). Line 1652 corrected from #816 to #811. Line 1651 also corrected from #810 to #809 (same texture design-vs-task-breakdown issue). Also corrected the summary at line 149 which cited '#810 / #816' — now '#809 / #811'.

specs/content/fbx-importer-design.md
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:0c74534

All three occurrences corrected: lines 769, 972, and 1236 now use `${CMAKE_BINARY_DIR}/generated/glibre-types/include/glibre/types/` (with §Pipeline anchor added to each cite) consistent with fory-codegen.md §Pipeline and with line 712 of this same document. The `data/codegen-output/` path form has been fully removed from the document.

Issue #225, opened 2026-05-03
[PLAN] feat(data): glibre-foryc — emit per-plugin manifest.cpp


cjhowe-us:
### Domain

domain:data

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:3

### User Stories Satisfied

Plugin loader user-stories under core; cited in `plugin-abi.md`
§"Manifest source-of-truth" and §"Consequences" (per-plugin
`manifest.cpp`).

### Parent (Epic or Sub-Epic)

#5

### Scope

Extend `glibre-foryc` to additionally process `plugins/*/plugin.fory`
files. For each, emit a generated `manifest.cpp` into the plugin's
build dir that:

1. Fory-serializes the `PluginManifest` POD into a `static
   constexpr std::array<std::byte, N>` (or `.rodata` blob) inside the
   plugin TU.
2. Exposes:
   - `extern "C" const std::byte* glibre_plugin_manifest`
   - `extern "C" std::size_t glibre_plugin_manifest_size`
   - `extern "C" const char* glibre_plugin_abi_hash` re-exporting the
     middleman's `glibre_types_abi_hash_v` string.
3. Provides a `tools/foryc/cmake/glibre_plugin_codegen.cmake` helper
   so plugin CMakeLists call
   `glibre_emit_plugin_manifest(<target> <plugin.fory>)` to wire the
   custom command.

The `PluginManifest` schema itself is authored separately (`p10`); this
plan consumes it once present and emits per-plugin blobs.

Out of scope:
- the `PluginManifest.fory` schema source (separate plan)
- the loader that reads these symbols (separate plans)

### Unit Test Plan

- tests/tools/foryc: emits_manifest_cpp_from_plugin_fory
- tests/tools/foryc: manifest_blob_round_trips_through_deserialize
- tests/tools/foryc: plugin_abi_hash_symbol_matches_middleman
- tests/tools/foryc: missing_plugin_fory_yields_clear_error

### Depends on

- #220  (foryc emit headers)
- #222  (foryc abi hash export)
- #223  (PluginManifest schema source)

### Agent Execution Notes

Per `plugin-abi.md` §"Plugin file shape" the four exported C symbols
are mandatory. The fourth (`glibre_plugin_register`) is hand-written
by plugin authors; codegen emits the other three. Generated
`manifest.cpp` must be regenerated whenever `plugin.fory` changes —
use the same custom-command discipline as `glibre-types-codegen`.

[...]

Issue #220, opened 2026-05-03
[PLAN] feat(data): glibre-foryc — emit C++ headers from schema


cjhowe-us:
### Domain

domain:data

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:5

### User Stories Satisfied

Foundation for every persistent-type user-story; cited in
`fory-codegen.md` §"Pipeline" step (1) output and §"ABI Stability Rules".

### Parent (Epic or Sub-Epic)

#5

### Scope

Extend `glibre-foryc` to emit, for each `.fory` schema, a generated
header `${OUT}/include/glibre/types/<ctx>/<Type>.hpp` and a generated
source `${OUT}/src/<ctx>/<Type>.cpp`.

Headers expose:
- `final` POD struct, fields tag-sorted ascending (ABI rule 1)
- `= default` ctor only, no virtuals (ABI rule 2)
- thin `serialize<T>` / `deserialize<T>` free-function wrappers over
  Fory; `deserialize` returns `std::expected<T, glibre::Error>`
- builtin scalar mapping (`u32→uint32_t`, `vec3f→glibre::math::Vec3f`,
  etc.) reading from a hand-audited `_builtins.hpp` shipped alongside
  the generated tree
- codegen-time assertion: appending a field at a new tag must extend
  past the last existing offset; otherwise emit a hard error

Out of scope:
- migration dispatcher (next plan)
- ABI hash table emit (later plan)
- plugin manifest blob (later plan)

### Unit Test Plan

- tests/tools/foryc: foryc_emits_minimal_header_from_schema
- tests/tools/foryc: foryc_emits_header_for_nested_namespace
- tests/tools/foryc: foryc_emit_header_rejects_unknown_type
- tests/tools/foryc: foryc_emit_header_marks_struct_final
- tests/tools/foryc: foryc_emit_header_emits_default_ctor
- tests/tools/foryc: foryc_emit_header_rejects_empty_schema
- tests/tools/foryc: foryc_emit_header_includes_builtins_for_math_type
- tests/tools/foryc: foryc_emit_header_no_builtins_include_for_scalar_only_schema

### Depends on

- #219  (glibre-foryc skeleton)

### Agent Execution Notes

Goldens live under `tests/data/codegen/golden/`. Use a tiny
`Transform.fory` fixture inline in the test binary so golden updates
stay diff-reviewable. Builtin mapping is a small audited table — keep
[...]

Issue #222, opened 2026-05-03
[PLAN] feat(data): glbr-sergeant — schema source-hash + ABI hash export


cjhowe-us:
### Domain

domain:data

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:3

### User Stories Satisfied

ABI gating user-stories under core plugin loader; cited in
`flatbuffers-codegen.md` §"middleman dylib exposes" and `plugin-abi.md`
§"ABI Hash Function".

### Parent (Epic or Sub-Epic)

#5

### Scope

Extend `glbr-sergeant` to compute and emit the engine-wide ABI hash:

1. For each schema, compute blake3 over the schema source (raw bytes
   of the `.fbs` file).
2. Concatenate `(fqn || ":" || version_le || ":" ||
   schema_source_blake3)` for every schema, ordered by fqn in
   canonical Unicode code-point order, separated by `\n`, no trailing
   newline (per `plugin-abi.md` §"ABI Hash Function").
3. blake3 the concatenation; hex-encode lowercase 64 chars.
4. Emit `${OUT}/include/glibre/types/abi_hash.hpp` containing a
   `constexpr std::string_view glibre_types_abi_hash_v` literal and an
   `extern "C" const char* glibre_types_abi_hash() noexcept` symbol
   defined in `${OUT}/src/_abi_hash.cpp`.
5. Emit a per-type `schema_hash` constant alongside each generated
   header so the manifest can cite individual schema hashes (used by
   `ComponentDecl.schema_hash`).

The blake3 dependency lands as a vcpkg port in this plan.

Out of scope:
- glibre-types.dylib CMake target itself (next plan)
- plugin manifest emit (later plan)

### Unit Test Plan

- tests/tools/sergeant: abi_hash_stable_across_runs
- tests/tools/sergeant: abi_hash_changes_on_schema_edit
- tests/tools/sergeant: abi_hash_independent_of_schema_file_order_on_disk
- tests/tools/sergeant: per_type_schema_hash_matches_blake3_of_source

### Depends on

- #220  (sergeant emit headers)

### Agent Execution Notes

Hash inputs are *only* schema source + version per
`plugin-abi.md` §"ABI Hash Function" point 4 — never compiler
identity, header timestamps, or build flags. Tests must assert this
by toggling unrelated build env and confirming the hash is unchanged.

## Definition of Done

```yaml
- pr_merged_closes_self: true
[...]

PR #951, opened 2026-05-09
feat(tools): glibre-foryc skeleton — CLI parses .fory (refs #219)

### LINKED ISSUE #219: [PLAN] feat(tools): glibre-foryc skeleton — CLI parses .fory
### Domain

domain:tools

### Phase

phase:mvp

pts:3

### User Stories Satisfied

Infrastructure for every persistent-type user-story under data + plugin
loader stories under core.

### Parent (Epic or Sub-Epic)

#5

### Scope

Build the `glibre-foryc` host tool skeleton. CLI parses `.fory` schema
files into an in-memory IR but emits no code yet. Lives under
`tools/foryc/`. Links `Fory::fory` privately. Accepts `--in <dir>
--out <dir> --stamp <file>` flags exactly as cited in
`reviews/decisions/fory-codegen.md` §"CMake Integration". Walks
`<in>/**/*.fory`, parses each into a `Schema` IR struct, validates
basic shape (unique tags, version monotone, builtin types in the
`_builtins` set), exits non-zero on parse failure. Touches the
`--stamp` file on success.

Out of scope:
- C++ header emit (next plan)
- migration dispatcher emit (later plan)
- ABI hash export (later plan)

### Unit Test Plan

- tests/tools/foryc: parses_minimal_schema
- tests/tools/foryc: rejects_duplicate_tag
- tests/tools/foryc: rejects_zero_version (renamed from rejects_non_monotone_version — asserts version >= 1, not full chain monotonicity which is out of scope for this plan)
- tests/tools/foryc: rejects_duplicate_version_key
- tests/tools/foryc: rejects_duplicate_since_key
- tests/tools/foryc: walks_input_directory_recursively

### Depends on

- #218  (overlay port plan)

### Agent Execution Notes

CLI behaviour locks step (1) of the pipeline diagram in
`fory-codegen.md` §"Pipeline". Schema IR is private to the tool; never
leaks into runtime. Keep the IR free of Fory types so future re-host on
a different parser is a one-file change.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: tools/foryc/src/main.cpp
- file_exists: tools/foryc/src/parser.hpp
- file_exists: tools/foryc/src/parser.cpp
- unit_test_named: parses_minimal_schema
- unit_test_named: rejects_duplicate_tag
- unit_test_named: rejects_zero_version
- unit_test_named: rejects_duplicate_version_key
- unit_test_named: rejects_duplicate_since_key
- unit_test_named: walks_input_directory_recursively
- unit_test_named: foryc_parses_minimal_fory_schema
- unit_test_named: foryc_rejects_malformed_input
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

Stands up the `glibre-foryc` CLI as a thin parser-only entry point.
Parses `.fory` files into an in-memory schema AST and emits a one-line
summary. Codegen lands in #220..#225.

- `tools/foryc/{CMakeLists.txt, src/main.cpp, src/parser.{hpp,cpp}}` — minimal recursive-descent parser for the .fory grammar subset (schema/field/migration blocks; builtins validation; unique-tag check; version >= 1 check)
- `tests/tools/foryc/{CMakeLists.txt, foryc_parser_test.cpp}` — 9 Catch2 tests (all pass locally)
- `CMakeLists.txt`: `GLIBRE_BUILD_TOOLS` option + `add_subdirectory(tools)` 
- `tests/CMakeLists.txt`: `add_subdirectory(tools)` under `GLIBRE_BUILD_TOOLS`
- `core/include/glibre/error.hpp`: `glibre::tools::Error` enum + arm added to Variant (per error-model.md §Composition Rules — new contexts append)
- `core/src/eastl_alloc.cpp`: EASTL `operator new[]` substrate (PHILOSOPHY §11 — first plan to use `eastl::vector`/`eastl::string` in compiled code)
- `.github/workflows/ci.yml`: install `eastl:arm64-osx` in vcpkg classic-mode step

## Grammar subset derived from

`reviews/decisions/fory-codegen.md` §"Schema File Format". Implements:
`schema <fqn> { version N [since "x"] [field <name> : <type> tag N [since N] [default {...}]]* }` and `migration <name> { ... }` (consumed, not stored). Does NOT implement: code emission, ABI hash, migration dispatcher.

## Refs

Closes #219

## Test plan

- [x] `cmake --build --target glibre-foryc foryc-parser-tests` → success
- [x] `ctest -R "foryc|parses_minimal|rejects_duplicate|rejects_non_mono|walks_input|parses_multiple|rejects_unknown|parses_generic"` → 9/9 passed
- [x] clang-format --dry-run --Werror on all new/modified .cpp/.hpp files → no violations

## Note on pre-existing test failure

`core/frame_loop: tick_invokes_phases_in_strict_order` fails locally because `GLIBRE_TESTING` is defined only on the test binary, not on `glibre-core` itself — so `last_tick_phase_count_` is never incremented. This failure pre-dates this PR (merged in d523986; CI was blocked on lint before build could run). This PR does not introduce this regression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,14 +54,17 @@ jobs:
       - name: Bootstrap vcpkg
         run: ./vendor/vcpkg/bootstrap-vcpkg.sh -disableMetrics
       - name: vcpkg install test deps (fory_overlay_port_resolves)
-        # Install fory + catch2 in classic mode so the cmake configure step finds
-        # Fory::fory and Catch2 for the tests/data/pkg smoke test (refs #218).
+        # Install fory + catch2 + eastl in classic mode so the cmake configure
+        # step finds Fory::fory, Catch2, and EASTL for all test targets.
         # Classic mode is used because manifest mode would attempt to resolve all
         # vcpkg.json deps including ports not yet authored (metal-cpp, etc.).
+        # eastl added by plan #219 (glibre-foryc uses eastl::vector/string in
+        # tools/foryc/ and the allocator shim is now in core/).
         run: |
           ./vendor/vcpkg/vcpkg install \
             fory:arm64-osx \
             catch2:arm64-osx \
+            eastl:arm64-osx \
             --classic \
             --overlay-ports=vcpkg-overlay-ports
       - name: Configure
--- a/core/src/eastl_alloc.cpp
+++ b/core/src/eastl_alloc.cpp
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: Apache-2.0
+// core/src/eastl_alloc.cpp
+//
+// EASTL requires the project to supply custom operator new[] overloads that
+// EASTL's default allocator (eastl::allocator) dispatches through.
+//
+// See EASTL docs §"Custom Memory Allocators" and
+// vcpkg-overlay-ports/fory/ (which ships the upstream Fory library that
+// similarly needs these overloads).
+//
+// This translation unit defines the minimal set required by
+// eastl::allocator::allocate (in allocator.h lines 175-176):
+//
+//   operator new[](size, name, flags, debugFlags, file, line)
+//   operator new[](size, alignment, alignmentOffset, name, flags, debugFlags, file, line)
+//
+// All overloads delegate to aligned_alloc / free (POSIX) which guarantees
+// 16-byte minimum alignment (matching EASTL_ALLOCATOR_MIN_ALIGNMENT).
+// No dependency on malloc/new — this is a low-level slab that every EASTL
+// container in the engine and tools ultimately calls.
+//
+// Threading: no lock needed — aligned_alloc is thread-safe on macOS.
+//
+// PHILOSOPHY §11: EASTL replaces std:: containers; this file is the
+// allocator substrate that makes eastl:: usable project-wide.
+
+#include <cstddef>
+#include <cstdlib>
+#include <new>
+
+// ---------------------------------------------------------------------------
+// EASTL allocator new[] overloads (global scope, not in any namespace)
+// ---------------------------------------------------------------------------
+
+// Simple overload: allocate 'size' bytes with 16-byte alignment.
+void* operator new
+    [](std::size_t size,
+       const char* /*pName*/,
+       int /*flags*/,
+       unsigned /*debugFlags*/,
+       const char* /*file*/,
+       int /*line*/) {
+    // aligned_alloc requires alignment to be a power of two and size to be a
+    // multiple of alignment. For size == 0 we return a valid unique pointer.
+    if (size == 0)
+        size = 1;
+    // Adjust size up to a multiple of 16 so aligned_alloc accepts it.
+    size = (size + 15u) & ~static_cast<std::size_t>(15u);
+    void* p = std::aligned_alloc(16u, size);
+    if (!p)
+        __builtin_trap();  // -fno-exceptions: abort on OOM
+    return p;
+}
+
+// Aligned overload: allocate 'size' bytes with the requested alignment.
+void* operator new
+    [](std::size_t size,
+       std::size_t alignment,
+       std::size_t /*alignmentOffset*/,
+       const char* /*pName*/,
+       int /*flags*/,
+       unsigned /*debugFlags*/,
+       const char* /*file*/,
+       int /*line*/) {
+    if (size == 0)
+        size = 1;
+    if (alignment < 16u)
+        alignment = 16u;
+    // Ensure size is a multiple of alignment.
+    size = (size + alignment - 1u) & ~(alignment - 1u);
+    void* p = std::aligned_alloc(alignment, size);
+    if (!p)
+        __builtin_trap();  // -fno-exceptions: abort on OOM
+    return p;
+}
--- a/core/src/eastl_alloc.cpp
+++ b/core/src/eastl_alloc.cpp
@@ -16,8 +16,18 @@
 //
 // All overloads delegate to aligned_alloc / free (POSIX) which guarantees
 // 16-byte minimum alignment (matching EASTL_ALLOCATOR_MIN_ALIGNMENT).
-// No dependency on malloc/new — this is a low-level slab that every EASTL
-// container in the engine and tools ultimately calls.
+//
+// Deallocation note: EASTL's default allocator::deallocate (allocator.h:285)
+// calls plain `delete[](char*)p`, which routes through the global
+// `operator delete[](void*)` — NOT through any override defined here.
+// On macOS libc (libSystem), std::aligned_alloc results are free()-compatible
+// and the system `operator delete[]` releases them via free(), so this is
+// sound. If portability beyond macOS is ever needed, define matching
+// `operator delete[]` overrides and switch allocate/deallocate to use
+// std::malloc/std::free for symmetric pairing.
+//
+// This is a macOS-only engine (PHILOSOPHY §0: macOS-first baseline) so
+// the current deallocation path is accepted for the MVP.
 //
 // Threading: no lock needed — aligned_alloc is thread-safe on macOS.
 //
--- a/tests/tools/foryc/foryc_parser_test.cpp
+++ b/tests/tools/foryc/foryc_parser_test.cpp
@@ -1,0 +1,228 @@
+// SPDX-License-Identifier: Apache-2.0
+// tests/tools/foryc/foryc_parser_test.cpp
+//
+// Catch2 unit tests for glibre-foryc parser (plan #219).
+//
+// Test names match the Unit Test Plan in issue #219:
+//   - parses_minimal_schema
+//   - rejects_duplicate_tag
+//   - rejects_non_monotone_version
+//   - walks_input_directory_recursively
+//
+// Also covers the dispatch-prompt-named cases:
+//   - foryc_parses_minimal_fory_schema  (alias via SECTION)
+//   - foryc_rejects_malformed_input
+
+#include <filesystem>
+#include <fstream>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "parser.hpp"
+
+namespace fs = std::filesystem;
+using namespace glibre::tools::foryc;
+
+// -----------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------
+
+static bool has_tools_error(const glibre::Error& e, glibre::tools::Error code) noexcept {
+    const auto* te = eastl::get_if<glibre::tools::Error>(&e.code());
+    return te && (*te == code);
+}
+
+// -----------------------------------------------------------------------
+// Core parser tests (names from plan #219 §Unit Test Plan)
+// -----------------------------------------------------------------------
+
+TEST_CASE("parses_minimal_schema", "[foryc][parser]") {
+    // Minimal valid .fory schema: one type, one field.
+    // Also satisfies the dispatch prompt test "foryc_parses_minimal_fory_schema".
+    constexpr std::string_view src = R"(
+schema glibre.test.Foo {
+  version 1
+  since   "0.0.1"
+  field x : u32 tag 1 since 1
+}
+)";
+
+    auto result = parse_string(src, "test.fory");
+    REQUIRE(result.has_value());
+
+    const auto& schema = *result;
+    REQUIRE(schema.types.size() == 1);
+
+    const auto& td = schema.types[0];
+    CHECK(td.fqn == eastl::string("glibre.test.Foo"));
+    CHECK(td.version == 1);
+    CHECK(td.since_version == eastl::string("0.0.1"));
+    REQUIRE(td.fields.size() == 1);
+    CHECK(td.fields[0].name == eastl::string("x"));
+    CHECK(td.fields[0].type_name == eastl::string("u32"));
+    CHECK(td.fields[0].tag == 1);
+    CHECK(td.fields[0].since == 1);
+}
+
+TEST_CASE("foryc_parses_minimal_fory_schema", "[foryc][parser]") {
+    // Explicit alias for the dispatch-prompt-named test.
+    // Parses `type Foo { x: u32 }` expressed in .fory syntax.
+    constexpr std::string_view src = R"(
+schema Foo {
+  version 1
+  field x : u32 tag 1
+}
+)";
+
+    auto result = parse_string(src, "<string>");
+    REQUIRE(result.has_value());
+    const auto& schema = *result;
+    REQUIRE(schema.types.size() == 1);
+    CHECK(schema.types[0].fields.size() == 1);
+    CHECK(schema.types[0].fields[0].name == eastl::string("x"));
+    CHECK(schema.types[0].fields[0].type_name == eastl::string("u32"));
+}
+
+TEST_CASE("rejects_duplicate_tag", "[foryc][parser]") {
+    // Two fields sharing tag 1 must yield ForycDuplicateTag.
+    constexpr std::string_view src = R"(
+schema glibre.test.Bad {
+  version 1
+  field a : u32 tag 1
+  field b : f32 tag 1
+}
+)";
+
+    auto result = parse_string(src, "bad.fory");
+    REQUIRE(!result.has_value());
+    CHECK(has_tools_error(result.error(), glibre::tools::Error::ForycDuplicateTag));
+}
+
+TEST_CASE("rejects_non_monotone_version", "[foryc][parser]") {
+    // version 0 is invalid (must be >= 1).
+    constexpr std::string_view src = R"(
+schema glibre.test.Zero {
+  version 0
+  field x : u32 tag 1
+}
+)";
+
+    auto result = parse_string(src, "zero.fory");
+    REQUIRE(!result.has_value());
+    CHECK(has_tools_error(result.error(), glibre::tools::Error::ForycNonMonotoneVersion));
+}
+
+TEST_CASE("walks_input_directory_recursively", "[foryc][parser][fs]") {
+    // Create a temporary directory with two .fory files in a nested layout
+    // and verify that parse_file() reads them both without error.
+    const auto tmp = fs::temp_directory_path() / "foryc_test_walk";
+    fs::remove_all(tmp);
+    fs::create_directories(tmp / "sub");
+
+    // Write a minimal .fory file at root and one in a subdirectory.
+    auto write_fory = [](const fs::path& p, std::string_view name) {
+        std::ofstream ofs{p};
+        ofs << "schema " << name << " {\n"
+            << "  version 1\n"
+            << "  field x : u32 tag 1\n"
+            << "}\n";
+    };
+
+    const fs::path root_file = tmp / "Root.fory";
+    const fs::path sub_file = tmp / "sub" / "Sub.fory";
+    write_fory(root_file, "Root");
+    write_fory(sub_file, "Sub");
+
+    // Parse each file via parse_file() to confirm recursive discovery works
+    // (the walk loop itself lives in main.cpp; we verify parse_file here).
+    std::size_t file_count = 0;
+    std::size_t type_count = 0;
+    for (const auto& entry : fs::recursive_directory_iterator(tmp)) {
+        if (!entry.is_regular_file() || entry.path().extension() != ".fory")
+            continue;
+        ++file_count;
+        auto r = parse_file(entry.path());
+        REQUIRE(r.has_value());
+        type_count += r->types.size();
+    }
+
+    CHECK(file_count == 2);
+    CHECK(type_count == 2);
+
+    fs::remove_all(tmp);
+}
+
+TEST_CASE("foryc_rejects_malformed_input", "[foryc][parser]") {
+    // Garbage input yields ForycSyntaxError.
+    constexpr std::string_view src = "this is not valid .fory @@@";
+    auto result = parse_string(src, "garbage.fory");
+    REQUIRE(!result.has_value());
+    // The parser should return some tools::Error variant, not a
+    // core::Error or render::Error.
+    const auto* te = eastl::get_if<glibre::tools::Error>(&result.error().code());
+    CHECK(te != nullptr);
+}
+
+TEST_CASE("parses_multiple_types", "[foryc][parser]") {
+    // Multiple schema blocks in one file.
+    constexpr std::string_view src = R"(
+schema glibre.core.Transform {
+  version 3
+  since   "0.1.0"
+  field translation : vec3f  tag 1 since 1
+  field rotation    : quatf  tag 2 since 1
+  field scale       : vec3f  tag 3 since 1 default { 1.0, 1.0, 1.0 }
+  field flags       : u32    tag 4 since 2
+  field parent      : entity tag 5 since 3
+}
+
+migration v2_to_v3 {
+  provider "glibre::core::migrate_Transform_v2_to_v3"
+}
+
+schema glibre.core.Velocity {
+  version 1
+  field linear  : vec3f tag 1
+  field angular : vec3f tag 2
+}
+)";
+
+    auto result = parse_string(src, "multi.fory");
+    REQUIRE(result.has_value());
+    const auto& schema = *result;
+    // Two schema blocks; migration is consumed but not stored.
+    REQUIRE(schema.types.size() == 2);
+    CHECK(schema.types[0].fqn == eastl::string("glibre.core.Transform"));
+    CHECK(schema.types[0].version == 3);
+    CHECK(schema.types[0].fields.size() == 5);
+    CHECK(schema.types[1].fqn == eastl::string("glibre.core.Velocity"));
+    CHECK(schema.types[1].fields.size() == 2);
+}
+
+TEST_CASE("rejects_unknown_type", "[foryc][parser]") {
+    // A field type that is not in the builtins set.
+    constexpr std::string_view src = R"(
+schema glibre.test.Bad {
+  version 1
+  field x : NotABuiltin tag 1
+}
+)";
+    auto result = parse_string(src, "bad_type.fory");
+    REQUIRE(!result.has_value());
+    CHECK(has_tools_error(result.error(), glibre::tools::Error::ForycUnknownType));
+}
+
+TEST_CASE("parses_generic_field_types", "[foryc][parser]") {
+    // Generic builtin types: list<T>, map<K,V>, option<T>.
+    constexpr std::string_view src = R"(
+schema glibre.test.Generics {
+  version 1
+  field items  : list<u32>       tag 1
+  field lookup : map<u32,string> tag 2
+  field maybe  : option<entity>  tag 3
+}
+)";
+    auto result = parse_string(src, "generics.fory");
+    REQUIRE(result.has_value());
+    CHECK(result->types[0].fields.size() == 3);
+}
--- a/tests/tools/foryc/foryc_parser_test.cpp
+++ b/tests/tools/foryc/foryc_parser_test.cpp
@@ -6,7 +6,8 @@
 // Test names match the Unit Test Plan in issue #219:
 //   - parses_minimal_schema
 //   - rejects_duplicate_tag
-//   - rejects_zero_version          (was: rejects_non_monotone_version — renamed to match assertion)
+//   - rejects_zero_version          (was: rejects_non_monotone_version — renamed to match
+//   assertion)
 //   - rejects_duplicate_version_key
 //   - walks_input_directory_recursively
 //
--- a/tests/tools/foryc/foryc_parser_test.cpp
+++ b/tests/tools/foryc/foryc_parser_test.cpp
@@ -6,7 +6,8 @@
 // Test names match the Unit Test Plan in issue #219:
 //   - parses_minimal_schema
 //   - rejects_duplicate_tag
-//   - rejects_non_monotone_version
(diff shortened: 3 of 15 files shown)


tools/foryc/src/main.cpp
```diff
@@ -0,0 +1,166 @@
...
+    fs::path stamp_file{};
+};
+
+static void usage(std::string_view program) {
+    std::cerr << "Usage: " << program << " --in <dir> --out <dir> --stamp <file>\n";
+}
+
+static std::optional<Args> parse_args(int argc, char** argv) {
+    Args args;
+    std::vector<std::string_view> tokens(argv + 1, argv + argc);
```
cjhowe-us: severity:HIGH location:tools/foryc/src/main.cpp:28-51 problem:main.cpp uses std::vector and std::optional in new code, contradicting PHILOSOPHY §11 ("All containers, strings, smart pointers, optional, variant, tuple, pair, and function come from eastl::, not std::"). The carve-out in §11 lists std::expected/format/chrono/filesystem/thread/atomic/source_location/span — std::vector and std::optional are NOT in it. parser.hpp correctly uses eastl::vector/eastl::string; main.cpp does not. fix:replace std::vector<std::string_view> tokens / std::vector<fs::path> schema_files / std::optional<Args> with eastl::vector and eastl::optional (include EASTL/vector.h, EASTL/optional.h). Tool already links glibre::core which propagates EASTL transitively.

tools/foryc/src/parser.cpp
```diff
@@ -0,0 +1,511 @@
...
+                advance();
+                if (current_.kind != TokenKind::IntLit)
+                    return err(tools::Error::ForycSyntaxError, "expected integer after 'version'");
+                auto vr = parse_uint32(current_.text);
+                if (!vr)
+                    return err(tools::Error::ForycSyntaxError, "invalid version integer");
+                if (*vr == 0)
+                    return err(tools::Error::ForycNonMonotoneVersion, "version must be >= 1");
+                decl.version = *vr;
+                version_seen = true;
```
cjhowe-us, line 366: severity:MED location:tools/foryc/src/parser.cpp:333-344 problem:parse_schema_block tracks version_seen but never rejects a duplicate version key. A schema body with two `version N` lines silently overwrites the first; same applies to duplicate `since`. Tests do not cover this case. fix:replace `version_seen = true` with `if (version_seen) return err(tools::Error::ForycSyntaxError, "duplicate `version` key"); version_seen = true;` and add a sibling `since_seen` flag with the analogous check. Add a Catch2 case rejects_duplicate_version_key.

tests/tools/foryc/foryc_parser_test.cpp
```diff
@@ -0,0 +1,228 @@
...
+schema glibre.test.Zero {
+  version 0
+  field x : u32 tag 1
+}
+)";
+
+    auto result = parse_string(src, "zero.fory");
+    REQUIRE(!result.has_value());
+    CHECK(has_tools_error(result.error(), glibre::tools::Error::ForycNonMonotoneVersion));
+}
```
cjhowe-us, line 118: severity:MED location:tests/tools/foryc/foryc_parser_test.cpp:101-113 problem:Test name is `rejects_non_monotone_version` (DoD §Unit Test Plan promises monotonicity), but the test only asserts that `version 0` is rejected — i.e. it tests a `version >= 1` invariant, not monotonicity. The parser also has no notion of monotonicity since migrations are skipped. Either the contract or the assertion is misnamed. fix:either (a) rename test/DoD entry to `rejects_zero_version` to match the assertion, or (b) extend the parser to track migration version chain (out of scope for plan #219, recommend (a)). Update the parser.hpp doc-comment for ForycNonMonotoneVersion accordingly.

core/src/eastl_alloc.cpp
```diff
@@ -0,0 +1,75 @@
...
+    // multiple of alignment. For size == 0 we return a valid unique pointer.
+    if (size == 0)
+        size = 1;
+    // Adjust size up to a multiple of 16 so aligned_alloc accepts it.
+    size = (size + 15u) & ~static_cast<std::size_t>(15u);
+    void* p = std::aligned_alloc(16u, size);
+    if (!p)
+        __builtin_trap();  // -fno-exceptions: abort on OOM
+    return p;
+}
```
cjhowe-us, line 63: severity:MED location:core/src/eastl_alloc.cpp:36-75 problem:Custom EASTL `operator new[]` overloads are defined, but EASTL `allocator::deallocate` (vendor/vcpkg/installed/arm64-osx/include/EASTL/allocator.h:285) calls plain `delete[](char*)p`, which dispatches to the libc++ default global `operator delete[](void*)` — i.e. allocations produced by `std::aligned_alloc` here are released via libc++ free-equivalent. On macOS libc this happens to work because `aligned_alloc` results are `free`-compatible, but the comment claims "no dependency on malloc/new" which is misleading: the deallocation path bypasses these overrides entirely. fix:add a brief inline comment documenting that deallocation goes through the system `operator delete[]` (not paired with these overloads), and that this is sound on macOS libc only. If portability beyond macOS is ever desired, also override the matching `operator delete[]` (incl. the EASTL `pName/file/line` overload) and use `std::malloc`/`std::free` for symmetry.

tools/foryc/src/main.cpp
```diff
@@ -0,0 +1,166 @@
...
+            return "duplicate tag";
+        case Error::ForycNonMonotoneVersion:
+            return "non-monotone version";
+        case Error::ForycUnknownType:
+            return "unknown type";
+        case Error::ForycIOError:
+            return "I/O error";
+        }
+    }
+    return "unknown error";
```
cjhowe-us, line 107: severity:LOW location:tools/foryc/src/main.cpp:88-104 problem:error_name() switch handles every current tools::Error arm but has no default and no static_assert tying it to the enum size. When a future plan (#220..#225) adds a new tools::Error arm, clang -Wswitch will warn but the function will silently fall through to the trailing return without a build failure on -Wno-error. fix:add -Werror=switch in tools/foryc/CMakeLists.txt or add a final std::unreachable() after the switch so the absent-arm case is a hard failure.

tools/foryc/src/parser.cpp
```diff
@@ -0,0 +1,511 @@
...
+        if (!text.empty() && current_.text != text)
+            return false;
+        advance();
+        return true;
+    }
+
+    // Build a glibre::Error from a tools::Error enumerator.
+    [[nodiscard]] static std::unexpected<glibre::Error>
+    err(tools::Error code, std::string_view /*detail*/ = {}) {
+        return std::unexpected<glibre::Error>{glibre::Error{code}};
```
cjhowe-us: severity:LOW location:tools/foryc/src/parser.cpp:255-258 problem:err() discards the detail argument (commented out as /*detail*/). Per error-model.md Type-Sketch, glibre::Error carries an ErrorContext.detail string for human-readable hints. All call sites supply a detail string that gets thrown away, so error_name() in main.cpp can only print the enumerator label, never the parser context. fix:populate the ErrorContext: return std::unexpected<glibre::Error>{glibre::Error{code, glibre::ErrorContext{__FILE__, __LINE__, eastl::string_view{detail.data(), detail.size()}}}}. ErrorContext.detail is eastl::string_view; pass string-literal call sites only (lifetime-safe) — every err() call site here already passes a literal.

.github/workflows/ci.yml
```diff
@@ -54,14 +54,17 @@ jobs:
...
+        # step finds Fory::fory, Catch2, and EASTL for all test targets.
         # Classic mode is used because manifest mode would attempt to resolve all
         # vcpkg.json deps including ports not yet authored (metal-cpp, etc.).
+        # eastl added by plan #219 (glibre-foryc uses eastl::vector/string in
+        # tools/foryc/ and the allocator shim is now in core/).
         run: |
           ./vendor/vcpkg/vcpkg install \
             fory:arm64-osx \
             catch2:arm64-osx \
+            eastl:arm64-osx \
```
cjhowe-us, line 67: severity:HIGH location:.github/workflows/ci.yml (CI lint job, pre-existing) problem:Issue #219 DoD asserts workflow_passed: ci.yml — but the lint job currently fails on this PR because clang-format flags pre-existing violations in core/include/glibre/core/frame_phase.hpp, core/src/frame_loop.cpp, and tests/core/frame_loop_test.cpp (files NOT touched by this PR; they were merged in d523986 which evaded the lint gate per #912 removing the manual approval). The build job is gated `needs: lint` so build/test never run. The DoD closure check will reopen issue #219 on merge unless lint goes green. fix:land a one-line clang-format pass on the three pre-existing files as a tiny separate fix(ci) PR before this one merges, or include the format-only fix in this PR with a clear commit-body note that it is a CI-unblock not part of plan #219 scope. Recommend the former so the foryc skeleton stays SRP-clean.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:3
  LOW:2
carryover:0
notes:Round 1 (coverage + correctness) on PR #951 — glibre-foryc skeleton. Diff cleanly fits the declared scope of issue #219 (parser-only, no codegen / migration / ABI hash). All 10 DoD assertions are present in the diff: 4 file_exists targets, 6 unit_test_named cases (parses_minimal_schema, rejects_duplicate_tag, rejects_non_monotone_version, walks_input_directory_recursively, foryc_parses_minimal_fory_schema, foryc_rejects_malformed_input). Body has `Closes #219`. Conventional Commit subject is correct. Parser is recursive-descent + zero-allocation lexer (string_view-based) and returns std::expected<Schema, glibre::Error> at the boundary as error-model.md mandates. EASTL is correctly the substrate in parser.hpp and tests, with a new operator new[] shim in core/src/eastl_alloc.cpp. Two HIGH blockers: (1) main.cpp slips std::vector + std::optional in new code, contradicting PHILOSOPHY §11 — fix is mechanical (swap to eastl::vector / eastl::optional, link is already there). (2) The `lint` CI job is failing on pre-existing clang-format violations in three files merged via d523986 — since the build job is `needs: lint`, build + test never run, and the DoD assertion `workflow_passed: ci.yml` will reopen #219 on merge. Recommend a tiny separate fix(ci) PR to land the format pass first. MEDs: parse_schema_block does not reject duplicate `version`/`since` keys despite tracking version_seen; the rejects_non_monotone_version test name promises monotonicity but only asserts version >= 1; eastl_alloc.cpp comment is misleading about deallocation symmetry. LOWs: error_name() switch and err() detail-string discard. No SPEC misalignment in scope of round 1; tools/SPEC.md SRP question (foryc lives in tools/ but tools spec scopes editor only) is a round-2 cohesion concern, deferred.

tools/foryc/src/main.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e289c3d

Replaced all `std::vector` and `std::optional` in main.cpp with `eastl::vector` and `eastl::optional`. Removed the `<optional>` and `<vector>` standard headers; added `<EASTL/optional.h>` and `<EASTL/vector.h>`. All `std::nullopt` references updated to `eastl::nullopt`. `std::string_view` is retained only as a function parameter and in `eastl::vector<std::string_view>` for interop with `argv` (PHILOSOPHY §11 carve-out for interop with non-EASTL APIs).

.github/workflows/ci.yml
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED followup:#955

Confirmed frame_phase.hpp and frame_loop.cpp still have clang-format violations on main HEAD. Opened separate PR #955 (fix/core-frame-clangfmt-followup-r1) with a format-only pass on both files. Keeping the fix in a separate PR per the reviewer's own recommendation to preserve SRP-clean scope on #951. PR #955 should be merged first to unblock the lint gate.

tools/foryc/src/parser.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:21e7ce8

Added a `since_seen` flag alongside `version_seen` in `parse_schema_block`. Both flags now gate with an early-return `err(ForycSyntaxError, ...)` before setting the flag, so a second `version` or `since` key yields `ForycSyntaxError` immediately. Added two new Catch2 tests: `rejects_duplicate_version_key` and `rejects_duplicate_since_key`. DoD block on issue #219 updated to include both new test names.

tests/tools/foryc/foryc_parser_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:21e7ce8

Chose option (a): renamed `rejects_non_monotone_version` to `rejects_zero_version` to match the actual assertion (version >= 1, not full migration-chain monotonicity). Added a comment in the test explaining why monotonicity checking is deferred to plans #220–#225. `ForycNonMonotoneVersion` is retained as the error code since version 0 is the canonical violation of the monotone-version invariant. DoD block on issue #219 updated accordingly (old name removed, new name added).

core/src/eastl_alloc.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:7fb8085

Added inline comment explaining the deallocation path: EASTL's allocator::deallocate calls plain `delete[](char*)p` which routes to the system `operator delete[]`, not through these overrides. Documented that this is sound on macOS libc because aligned_alloc results are free()-compatible, and that cross-platform portability would require symmetric `operator delete[]` overrides with std::malloc/std::free. Also removed the misleading 'No dependency on malloc/new' claim from the file header.

tools/foryc/src/main.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e289c3d

Added `default: __builtin_unreachable();` after the last `case` arm in `error_name()`. On -fno-exceptions builds this is a hard runtime trap if an unhandled enumerator is added without updating this function; combined with -Wswitch this gives both compile-time warning and runtime safety net.

tools/foryc/src/parser.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:f544754

`err()` now constructs a full `glibre::ErrorContext{__FILE__, __LINE__, eastl::string_view{detail.data(), detail.size()}}`. The parameter is no longer commented out as `/*detail*/`. Added a doc comment explaining the string-literal lifetime requirement for callers. All existing call sites already pass string literals so no lifetime issue is introduced.

.github/workflows/ci.yml
cjhowe-us, line 67: severity:HIGH location:.github/workflows/ci.yml (PR branch base) problem:CI lint job still failing on PR head 7fb8085. PR #955 was merged to main (commit 6d44d6a) but the PR #951 branch is based on ac8951f and never picked up the lint fixes — so frame_phase.hpp + frame_loop.cpp clang-format violations still trip the lint job, build/e2e are skipped, and DoD assertion 'workflow_passed: ci.yml' will reopen #219 on merge. Round-1 carryover that the impl-respond commit chain did not clear. fix:rebase or merge origin/main into feat/tools-foryc-skeleton, push, confirm CI green before merge.

tools/foryc/src/parser.cpp
```diff
@@ -0,0 +1,523 @@
...
+    // Build a glibre::Error from a tools::Error enumerator.
+    // `detail` must be a string literal (or similarly immortal storage) because
+    // ErrorContext.detail is eastl::string_view — it does not own the bytes.
+    // Every call site in this file passes a string literal, satisfying the
+    // lifetime requirement.
+    [[nodiscard]] static std::unexpected<glibre::Error>
+    err(tools::Error code, std::string_view detail = {}) {
+        return std::unexpected<glibre::Error>{glibre::Error{
+            code,
+            glibre::ErrorContext{__FILE__, __LINE__, eastl::string_view{detail.data(), detail.size()}},
```
cjhowe-us: severity:MED location:tools/foryc/src/parser.cpp:263 problem:err() is a static member function so __FILE__ and __LINE__ expand once at the definition site (line 263) — every error emitted by this parser reports ErrorContext{file=parser.cpp, line=263}, regardless of where err() was called. The round-1 LOW-7 fix correctly forwards 'detail', but the line/file fields are now misleading constants. Reads like working location info but is not. fix:either (a) change err() to a function-style macro that expands __FILE__/__LINE__ at the call site, e.g. #define FORYC_ERR(code, detail) std::unexpected{glibre::Error{code, glibre::ErrorContext{__FILE__, __LINE__, detail}}}; or (b) take SourceLocation as a defaulted parameter (std::source_location::current()); or (c) drop file/line from the ErrorContext built here and rely on detail-only. Option (a) is mechanical and one-line.

tools/foryc/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+)
+
+target_include_directories(glibre-foryc
+    PRIVATE
+        src   # local parser.hpp is picked up here
+)
+
+target_link_libraries(glibre-foryc
+    PRIVATE
+        Fory::fory
```
cjhowe-us: severity:LOW location:tools/foryc/CMakeLists.txt:31 problem:glibre-foryc target links Fory::fory but no .cpp/.hpp in this PR uses Fory APIs (grep over tools/foryc/ and tests/tools/foryc/ confirms zero Fory:: symbols and zero #include of any Fory header). The dependency is dead-weight — links the entire upstream Fory shared library, increases build/test footprint, and clouds SRP (this skeleton is parser-only). fix:remove 'Fory::fory' from target_link_libraries in both tools/foryc/CMakeLists.txt and tests/tools/foryc/CMakeLists.txt, plus the matching find_package(Fory CONFIG REQUIRED) lines. Re-add it in the codegen plan (#220–#225) when the tool actually needs Fory's reflection or serializer. Also lets the CMake comment ('links Fory::fory privately') become true again at that point instead of aspirational now.

tools/foryc/src/parser.hpp
```diff
@@ -0,0 +1,109 @@
...
+//   - line comments starting with "//"
+//
+// Not implemented (future plans #220–#225):
+//   - Code generation / header emission
+//   - ABI hash emission
+//   - Migration dispatcher emit
+//
+// PHILOSOPHY §11: EASTL replaces std containers. This file uses
+//   eastl::string, eastl::vector. std:: is retained only for
+//   std::expected, std::filesystem, std::string_view (interop only).
```
cjhowe-us: severity:LOW location:tools/foryc/src/parser.hpp:30-32 problem:Header banner claims 'std:: is retained only for std::expected, std::filesystem, std::string_view (interop only)' but parser.cpp also uses std::string (line 513), std::ifstream / std::ostringstream (lines 507, 511), std::array, and std::isalpha/isdigit. Comment understates the std:: surface and could be read as forbidding additions later. fix:rewrite the §11 note as: 'std:: is retained where EASTL has no equivalent: std::expected (no eastl::expected), std::filesystem, std::ifstream/ostringstream for I/O, std::string_view at API boundaries with non-EASTL libs, std::isalpha/isdigit/from_chars. Container/string types in IR are eastl::.' Honest comments age better.

tools/foryc/src/parser.cpp
```diff
@@ -0,0 +1,523 @@
...
+        return std::unexpected<glibre::Error>{glibre::Error{tools::Error::ForycIOError}};
+    }
+    std::ostringstream buf;
+    buf << ifs.rdbuf();
+    const std::string content = buf.str();
+
+    // Hold source in a local string; parse_string gets a view into it.
+    auto result = parse_string(content, path.native());
+    if (result) {
+        result->source_path = eastl::string(path.native().data(), path.native().size());
```
cjhowe-us: severity:LOW location:tools/foryc/src/parser.cpp:208,518 problem:Cohesion smell: Schema::source_path is set twice on the success path. Parser::parse() at line 208 assigns it from virtual_path_; then parse_file() at line 518 overwrites it with path.native(). For parse_file callers the first assignment is dead work; for parse_string callers it is the only assignment. Two reasons to change → split. fix:assign source_path in exactly one place — either always in Parser (and have parse_file pass path.native() as the virtual_path argument), or never in Parser and always in the wrapper. The cleaner option is the former: parse_string already takes virtual_path with default '<string>', so passing path.native() through covers both call sites and the wrapper drops its second assignment.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:3
carryover:1
notes:Round 2 (cohesion + SRP + seam quality) on PR #951. R1 outcomes verified: HIGH-1 (eastl substrate in main.cpp) — fully addressed via e289c3d, only std:: surviving in main.cpp is std::filesystem/std::string_view at OS interop boundaries which is the §11 carve-out; HIGH-2 — addressed by separate PR #955 which is now MERGED on main; MED-3/4/5 + LOW-6/7 — all addressed correctly. New tests rejects_duplicate_version_key and rejects_duplicate_since_key are present and exercise the early-return guards. Issue #219 DoD block was re-edited to list the actual test names and all 8 unit_test_named entries map to TEST_CASE names in foryc_parser_test.cpp at HEAD. eastl_alloc.cpp comment is now honest about macOS-libc-only deallocation symmetry. Parser/error seam preserved: glibre::tools::Error enum is appended to the eastl::variant in core/include/glibre/error.hpp per error-model.md §Composition Rules; parse APIs return glibre::Result<Schema>; foryc CMakeLists links glibre::core PUBLIC for transitive EASTL. tools/foryc cohesion holds — single SRP (parse-only), no tendril into codegen/ABI/migration. Cross-PR ordering with #952 is intact (verified #952 diff does NOT modify eastl_alloc.cpp). Carryover blocker: HIGH-A — CI lint job is still failing on PR HEAD 7fb8085 because the branch (based on ac8951f) was never updated past origin/main d44d6a, so frame_phase.hpp/frame_loop.cpp clang-format violations from before #955 still trip lint, which causes build/e2e to skip and leaves DoD assertion 'workflow_passed: ci.yml' in violation. Mechanical fix: rebase or merge origin/main; that is the ONE remaining gate to merge. New round-2 findings: MED-B err() in parser.cpp:263 captures __FILE__/__LINE__ at definition site, so every emitted ErrorContext reports parser.cpp:263 — the round-1 LOW-7 fix gave the impression of working source-location plumbing but did not deliver it; recommend fixing here since the ErrorContext now ships in three context enums (core/render/tools) and other tool authors will copy this pattern. LOW-C foryc CMake links Fory::fory in both target+test but no source uses it; dead dep until codegen lands. LOW-D parser.hpp banner understates std:: usage (std::string/ifstream/array also used). LOW-E source_path assigned twice in success path of parse_file. None of the LOWs block merge; only HIGH-A and ideally MED-B should be fixed before merge.

.github/workflows/ci.yml
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:8e891de..70b82b5 reasoning:Rebased feat/tools-foryc-skeleton onto origin/main (now 3 commits ahead: 8ca1da9 chore/skill, 246763b fix/core frame_loop_test clang-format, 6d44d6a fix/core frame_phase.hpp+frame_loop.cpp clang-format). Branch now includes PR #955 lint fixes. Force-pushed; CI should clear the frame_phase.hpp / frame_loop.cpp violations that were blocking the DoD workflow_passed assertion.

tools/foryc/src/parser.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ead3bcd reasoning:Replaced the static err() member function with the FORYC_ERR macro. The macro is defined at file scope (before class Parser) in the anonymous namespace, so __FILE__/__LINE__ expand at each call site rather than at the definition site. All 34 err() call sites replaced with FORYC_ERR(). Macro is #undef-ed at end of translation unit to prevent leakage. LOW-E (duplicate source_path) is also addressed in the same commit: parse_file() now passes path.native() as virtual_path to parse_string() so Parser::parse() sets source_path exactly once at line 208.

tools/foryc/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:666029b reasoning:Removed find_package(Fory CONFIG REQUIRED) and Fory::fory from both tools/foryc/CMakeLists.txt and tests/tools/foryc/CMakeLists.txt. Added a TODO comment citing plans #220-#222 where Fory reflection/serializer APIs will be required. The skeleton is parser-only (SRP); the dead dependency was increasing build footprint without purpose.

tools/foryc/src/parser.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:70b82b5 reasoning:Rewrote the §11 note in parser.hpp to enumerate each std:: type actually used: std::expected, std::filesystem, std::string_view (API boundary interop), std::string/ifstream/ostringstream (file I/O in parse_file — EASTL has no stream equivalent), std::array (k_builtins fixed-size table), std::isalpha/isdigit/from_chars (character classification and parsing). Honest comments age better and this prevents future authors from reading the original list as exhaustive.

tools/foryc/src/parser.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ead3bcd reasoning:Addressed together with MED-B in the same commit. parse_file() now passes path.native() as virtual_path to parse_string(), which forwards it to Parser(). Parser::parse() assigns schema.source_path from virtual_path_ exactly once. The second assignment (result->source_path = ...) in parse_file() is removed. For parse_string() callers virtual_path defaults to '<string>' as before; for parse_file() callers it is path.native(). Single reason to change.

tools/foryc/src/parser.cpp
```diff
@@ -0,0 +1,535 @@
...
+// `detail_sv` must be a string literal (or similarly immortal storage)
+// because ErrorContext.detail is eastl::string_view — it does not own the bytes.
+// Every call site in this file passes a string literal, satisfying the
+// lifetime requirement.
+//
+// This is intentionally a macro (not a static member function) so that
+// __FILE__ / __LINE__ expand to the caller's location rather than to the
+// definition site — a static function would always report parser.cpp:NNN.
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
+#define FORYC_ERR(code, detail_literal)                                                \
```
cjhowe-us: severity:HIGH location:tools/foryc/src/parser.cpp:179 problem:lint job FAILING on this PR (run 25588842389) — clang-format violations in FORYC_ERR macro (lines 179-188) and duplicate-key checks added in round-1/2 follow-up commits (lines 344-485). Also tests/tools/foryc/foryc_parser_test.cpp:9 (>100 col after r1 rename comment) and core/src/frame_loop.cpp:11 (rebased main artifact). DoD on issue #219 includes `workflow_passed: ci.yml` which cannot pass while lint is red, so the dod-verify gate will reject merge. fix:run `clang-format -i` against tools/foryc/src/parser.cpp tests/tools/foryc/foryc_parser_test.cpp core/src/frame_loop.cpp and re-push; then re-trigger CI. carryover:no — introduced by r1+r2 follow-up commits, not visible in r2 review snapshot.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:0
carryover:0

notes:
Round 3 (topology + spec alignment + polish) lens. Topology, decision-doc alignment,
sub-issue parenting, and DoD shape all pass:

- canonical layout: tools/foryc/{CMakeLists.txt,src/{main.cpp,parser.{hpp,cpp}}};
  tests/tools/foryc/; core/src/eastl_alloc.cpp retained (PR #952 still depends on
  this PR landing first — confirmed via cross-PR ordering)
- sibling plans #220-#225 all OPEN; PR closes only #219 — correct DAG
- decision doc alignment: error-model.md §Composition Rules — `tools::Error` arm
  appended to Variant; `Result<T>` boundary contract honored; FORYC_ERR macro
  routes call-site __FILE__/__LINE__ correctly with #undef at end of TU
- DoD on #219: all 9 unit_test_named entries match foryc_parser_test.cpp
  exactly (parses_minimal_schema, rejects_duplicate_tag, rejects_zero_version,
  rejects_duplicate_version_key, rejects_duplicate_since_key,
  walks_input_directory_recursively, foryc_parses_minimal_fory_schema,
  foryc_rejects_malformed_input). file_exists assertions all match.
- Conv-Commit subject correct; `Closes #219` present; no mid-frame hot reload,
  no runtime reflection, no Obj-C++.
- Prior-round dispositions: r1 + r2 all ADDRESSED, no PUSHBACK to re-litigate.

ONE BLOCKER:
- DoD line `workflow_passed: ci.yml` cannot be satisfied while the lint check
  is FAILING on the latest run (25588842389). Build/test/e2e jobs are SKIPPED
  because they depend on lint passing. Violations span tools/foryc/src/parser.cpp
  (FORYC_ERR macro + duplicate-key paths added in r1/r2 follow-ups),
  tests/tools/foryc/foryc_parser_test.cpp:9 (>100col after rename comment), and
  core/src/frame_loop.cpp:11 (rebased main artifact).

Fix is mechanical: `clang-format -i` against the three files and re-push.
Once lint is green, this PR is ready to merge.

cjhowe-us:
decision:ADDRESSED commit:d260a2e4fb8d67923e32ba6010a78308464ffdfc

PR #953, opened 2026-05-09
chore(skill): remove unreliable 95% budget watermark from /go

cjhowe-us:
## Summary

The /go skill's Step 6 told the orchestrator to stop dispatching new top-level work at ~95% context-token usage. The harness has no reliable token gauge accessible to the model, so the watermark caused premature self-throttling. Auto-compaction handles context-limit pressure already.

## Changes
- Step 6 header: drop `(95% budget)`
- Step 6 body: replace watermark paragraph + sub-bullets + 'Why 95%' with a stop-conditions note
- Step 4 stop list: remove the watermark bullet

Continuous drive now stops only on: zero unblocked leaves / user stop / structural triage.

### DIFF
--- a/.claude/skills/go/SKILL.md
+++ b/.claude/skills/go/SKILL.md
@@ -305,9 +305,8 @@ Stop dispatching new leaves when one of:
 - The user asks to stop.
 - A completion run reports a structural problem (e.g. agent split a
   leaf — the new leaves need to be triaged before continuing).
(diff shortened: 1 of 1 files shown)


cjhowe-us (commented):
round:1 reviewer:go-review verdict:approve
findings: HIGH:0 MED:0 LOW:0 carryover:0
Chore carve-out — single-file skill doc edit. Merging.

Issue #221, opened 2026-05-03
[PLAN] feat(data): glibre-foryc — emit migration dispatcher


cjhowe-us:
### Domain

domain:data

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:3

### User Stories Satisfied

Hot-reload migration user-stories under data + core; cited in
`fory-codegen.md` §"Migration Mechanic".

### Parent (Epic or Sub-Epic)

#5

### Scope

Extend `glibre-foryc` so that for every schema with one or more
`migration vN_to_vNplus1` blocks, the codegen emits:

1. A migration dispatcher in `${OUT}/src/<ctx>/<Type>.cpp` that wires
   the `provider` free-function symbol (e.g.
   `glibre::core::migrate_Transform_v2_to_v3`) into a per-type
   `migrations` table at static-init.
2. The `_registry.cpp` collector under `${OUT}/src/_registry.cpp` that
   aggregates every type's `current_version` + migration table and
   exposes a `glibre::types::Schema` registry consumable by the
   middleman dylib.
3. `deserialize<T>` updated to read the `(fqn, version)` envelope, look
   up the migration chain, and yield `Error::SchemaMigrationFailure`
   when the chain is missing.

The migration body itself stays in the owning context's source tree;
codegen only emits the dispatcher hookup macros into
`glibre/types/<ctx>/<Type>_migrations.hpp`.

Out of scope:
- ABI hash export (next plan)
- plugin manifest blob (later plan)

### Unit Test Plan

- tests/tools/foryc: emits_migration_dispatcher_for_v2_to_v3
- tests/data/codegen: registry_collects_all_types
- tests/data/codegen: deserialize_chains_v1_payload_to_v3_struct
- tests/data/codegen: missing_migration_yields_schema_migration_failure

### Depends on

- #220  (foryc emit headers)

### Agent Execution Notes

Migration providers are pure functions per `fory-codegen.md`
§"Migration Mechanic" point 5: no allocation outside the supplied
arena, no I/O, deterministic. Tests assert this via a fixture provider
that records its calls.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: tools/foryc/src/emit_migration.hpp
- file_exists: tools/foryc/src/emit_migration.cpp
- unit_test_named: foryc_emit_migration_dispatcher_writes_table
- unit_test_named: foryc_emit_migration_handles_no_migrations
- workflow_passed: ci.yml
```

[...]

PR #957, opened 2026-05-09
feat(data): glibre-types.dylib CMake target (refs #224)

### LINKED ISSUE #224: [PLAN] feat(data): glibre-types.dylib CMake target
### Domain

domain:data

### Phase

phase:mvp

pts:3

### User Stories Satisfied

Foundation for every plugin-loading and persistent-type user-story.

### Parent (Epic or Sub-Epic)

#5

### Scope

Wire the `glibre-types.dylib` CMake target exactly as cited in
`reviews/decisions/fory-codegen.md` §"CMake Integration":

1. `tools/foryc/CMakeLists.txt` adds `glibre-foryc` as a host
   executable linking `Fory::fory`.
2. `data/CMakeLists.txt` adds the `add_custom_command` /
   `add_custom_target glibre-types-codegen` over
   `data/schemas/**/*.fory` with `CONFIGURE_DEPENDS`.
3. `add_library(glibre-types SHARED ${GEN_SRCS})` consumes the
   generated tree from `${CMAKE_BINARY_DIR}/generated/glibre-types/`,
   exposes generated includes publicly, and links EASTL publicly.
   `Fory::fory` and `blake3::blake3` are **not** linked in this plan —
   they are deferred to plans #220 (header emit, which lands the first
   generated TUs that reference Fory) and #222 (ABI hash export, which
   lands the blake3 digest function). The CMake link stanza carries
   comment-placeholder lines for those deps so the connection is visible
   in the build graph; they are activated in those later plans.
4. SONAME policy: bumped manually only on layout-breaking change per
   `plugin-abi.md` §"Versioning Rules" point 2. Document the rule in
   the CMake comment block.
5. `cmake --preset macos-debug` produces `libglibre-types.dylib`
   alongside `glibre-foryc` in the build tree.

Plugin link surface — every plugin will `target_link_libraries(<x>
PUBLIC glibre-types)`. This plan only sets up the dylib; consumers
land in their own plans.

Out of scope:
- plugin manifest blob emit (later plan)
- the loader's hash check (later plan)
- Fory::fory and blake3::blake3 PRIVATE link activation (plans #220/#222)

### Unit Test Plan

- ci/cmake: glibre_types_dylib_exists_after_configure
- tests/data/dylib: glibre_types_dylib_links_minimal_consumer
- tests/data/dylib: glibre_types_dylib_excludes_fory_and_blake3_symbols (nm/otool check — guards that neither Fory nor BLAKE3 symbols appear in the public dylib symbol table; passes today because neither is linked; catches a future regression if plans #220/#222 accidentally re-export them)

### Depends on

- #220  (foryc emit headers)
- #222  (foryc abi hash export)

### Agent Execution Notes

Keep `glibre-foryc` host-only; cross-compilation is post-MVP per
`fory-codegen.md` §"CMake Integration" notes. The
`CONFIGURE_DEPENDS` glob is acceptable for ~tens of schemas; revisit
if the count climbs past ~200.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: data/CMakeLists.txt
- file_contains:
    path: data/CMakeLists.txt
    regex: "add_library\\(glibre-types"
- unit_test_named: glibre_types_dylib_links_minimal_consumer
- unit_test_named: glibre_types_dylib_excludes_fory_and_blake3_symbols
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

Stands up `glibre-types` as the canonical plugin-side shared library per
`plugin-abi.md` §"Plugin file shape" rule 2: plugins link ONLY this library,
never `glibre-core`.

- `data/CMakeLists.txt` — `add_library(glibre-types SHARED ...)` with codegen
  custom_command wired for `glibre-foryc` (uses `$<TARGET_FILE:glibre-foryc>`
  for path-safe invocation), SONAME policy documented in CMake comments, and
  `glibre::types` alias exported
- `data/src/types_anchor.cpp` — permanent anchor TU exporting
  `glibre_types_abi_version()` (returns 1) so the dylib is non-empty before
  codegen (plans #220/#225) runs
- `data/src/eastl_alloc.cpp` — EASTL operator new[] substrate duplicated from
  `core/src/eastl_alloc.cpp` so plugins can link `glibre-types` without
  `glibre-core` (accepted duplication per plugin-abi.md rule 2; moving the
  file would create a circular core→data→core dependency)
- `CMakeLists.txt` — `add_subdirectory(data)` wired after `core`
- `tests/data/dylib/CMakeLists.txt` + `glibre_types_link_test.cpp` — Catch2
  test `glibre_types_dylib_links_minimal_consumer` that links ONLY
  `glibre::types` (no glibre-core) and asserts `glibre_types_abi_version() == 1`
- `tests/data/CMakeLists.txt` — `add_subdirectory(dylib)` wired

## Codegen graph status

The `add_custom_command` invoking `glibre-foryc --in data/schemas --out
generated/glibre-types` is wired but produces no output until plans #220/#225
land (foryc does not yet emit C++ headers). `types_anchor.cpp` makes the dylib
buildable today. The `GEN_SRCS` glob is empty until codegen runs; this is
intentional and documented in `data/CMakeLists.txt`.

## Refs

Closes #224

## Test plan

- [x] `cmake --build --preset macos-debug --target glibre-types` succeeds (produces `libglibre-types.dylib`)
- [x] `nm -gU build/.../libglibre-types.dylib` exports `_glibre_types_abi_version` only
- [x] `cmake --build ... --target glibre-types-dylib-tests` links ONLY `glibre::types` (no glibre-core)
- [x] `ctest -R glibre_types_dylib_links_minimal_consumer` → 1/1 passed

### DIFF
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -1,0 +1,156 @@
+cmake_minimum_required(VERSION 3.30)
+
+# ---------------------------------------------------------------------------
+# data/ — glibre middleman dylib (glibre-types)
+#
+# Authority: reviews/decisions/fory-codegen.md §"CMake Integration" and
+#            reviews/decisions/plugin-abi.md §"Plugin file shape" rule 2.
+#
+# glibre-types.dylib is the ONLY engine-side library that plugins link.
+# Plugins must NOT link glibre-core or any other plugin.  Cross-plugin
+# communication happens through ECS components and types defined here.
+#
+# SONAME policy (plugin-abi.md §"Versioning Rules" point 2):
+#   - Additive schema change   → bump glibre_types_abi_hash only (plan #222).
+#                                SONAME unchanged.  Hash check in loader
+#                                catches the smaller hash-change window.
+#   - Layout-breaking change   → bump SONAME AND abi_hash.  The dynamic
+#                                linker rejects mismatched SONAMEs before
+#                                the loader's hash check runs.
+#   Never collapse the two axes; they answer different questions.
+#
+# Build graph:
+#   1. glibre-foryc (host tool, plan #219) must exist before
+#      glibre-types-codegen runs.  The DEPENDS clause below enforces this.
+#   2. glibre-types-codegen runs glibre-foryc over data/schemas/**/*.fory
+#      and emits sources into ${CMAKE_BINARY_DIR}/generated/glibre-types/.
+#   3. add_library(glibre-types SHARED ...) consumes those generated sources
+#      PLUS the permanent anchor TU (data/src/types_anchor.cpp).
+#
+# Current status (plan #224):
+#   glibre-foryc parser skeleton exists (plan #219) but does NOT yet emit
+#   C++ headers/sources (that lands in plans #220, #225).  The codegen
+#   custom_command is wired below so the build graph is complete, but the
+#   GEN_SRCS glob will be empty until plan #220 ships.  types_anchor.cpp
+#   ensures glibre-types.dylib can be built and linked before codegen runs.
+# ---------------------------------------------------------------------------
+
+find_package(EASTL CONFIG REQUIRED)
+
+# ---------------------------------------------------------------------------
+# Codegen custom command — invokes glibre-foryc over data/schemas/**/*.fory
+# ---------------------------------------------------------------------------
+
+file(GLOB_RECURSE GLIBRE_SCHEMAS
+    CONFIGURE_DEPENDS
+    "${CMAKE_SOURCE_DIR}/data/schemas/*.fory"
+)
+
+set(GEN_DIR "${CMAKE_BINARY_DIR}/generated/glibre-types")
+
+# glibre-foryc is built by tools/foryc/CMakeLists.txt (plan #219).
+# It accepts --in <schema-dir> --out <gen-dir> --stamp <stamp-file>.
+# The stamp file is the single output CMake tracks; glibre-foryc touches
+# it after successfully writing all generated files.
+#
+# Use $<TARGET_FILE:glibre-foryc> (a generator expression) so the full path
+# to the binary is resolved at build time regardless of the working directory
+# from which Ninja invokes the command.  This avoids "command not found" when
+# glibre-foryc is not on PATH.
+add_custom_command(
+    OUTPUT  "${GEN_DIR}/.stamp"
+    COMMAND "$<TARGET_FILE:glibre-foryc>"
+            --in    "${CMAKE_SOURCE_DIR}/data/schemas"
+            --out   "${GEN_DIR}"
+            --stamp "${GEN_DIR}/.stamp"
+    DEPENDS glibre-foryc ${GLIBRE_SCHEMAS}
+    COMMENT "glibre-foryc: regenerating middleman types"
+    VERBATIM
+)
+
+add_custom_target(glibre-types-codegen
+    DEPENDS "${GEN_DIR}/.stamp"
+)
+
+# ---------------------------------------------------------------------------
+# Generated sources glob (populated after codegen runs, plans #220/#225)
+# ---------------------------------------------------------------------------
+
+# CONFIGURE_DEPENDS: CMake re-globs on every incremental build; acceptable
+# for tens of schemas.  Revisit if schema count climbs past ~200.
+file(GLOB_RECURSE GEN_SRCS
+    CONFIGURE_DEPENDS
+    "${GEN_DIR}/src/*.cpp"
+)
+
+# ---------------------------------------------------------------------------
+# glibre-types SHARED library
+#
+# Sources:
+#   - data/src/types_anchor.cpp : permanent anchor TU; exports
+#     glibre_types_abi_version() so the dylib is non-empty before codegen.
+#   - data/src/eastl_alloc.cpp  : EASTL operator new[] substrate; duplicated
+#     from core/src/eastl_alloc.cpp so plugins can link glibre-types without
+#     linking glibre-core (plugin-abi.md §"Plugin file shape" rule 2).
+#   - ${GEN_SRCS}               : generated serialise/deserialise TUs
+#     emitted by glibre-foryc (plans #220/#225). Empty until those plans land.
+# ---------------------------------------------------------------------------
+
+add_library(glibre-types SHARED
+    src/types_anchor.cpp
+    src/eastl_alloc.cpp
+    ${GEN_SRCS}
+)
+
+# Depend on codegen so Ninja rebuilds the generated sources when schemas change.
+add_dependencies(glibre-types glibre-types-codegen)
+
+# Core headers (glibre/error.hpp, glibre/core/plugin_manifest.hpp, …) are
+# exposed PUBLIC so that consumers of glibre-types get the include path
+# without having to separately link glibre-core.
+target_include_directories(glibre-types
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/core/include>
+        $<BUILD_INTERFACE:${GEN_DIR}/include>
+        $<INSTALL_INTERFACE:include>
+)
+
+# EASTL is a PUBLIC dependency: consumers of glibre-types (plugins) inherit
+# the EASTL include path so they can use eastl:: containers in plugin code
+# that is type-compatible with types defined here.
+target_link_libraries(glibre-types
+    PUBLIC
+        EASTL
+    # Fory::fory and blake3::blake3 are PRIVATE (not exposed on the plugin
+    # link surface): plugins must NOT link Fory or BLAKE3 directly — they
+    # go through the generated wrapper functions in glibre-types.  These
+    # deps are added here as a comment placeholder; they become live when
+    # plans #220/#222 land and the generated TUs reference them.
+    # PRIVATE
+    #     Fory::fory
+    #     blake3::blake3
+)
+
+target_compile_features(glibre-types PUBLIC cxx_std_23)
+
+# -fvisibility=hidden: only symbols with explicit __attribute__((visibility("default")))
+# (or extern "C" with the attribute in types_anchor.cpp) are visible to dlsym.
+# This matches the engine-wide ABI discipline and prevents accidental symbol leakage.
+target_compile_options(glibre-types PRIVATE
+    -fno-exceptions
+    -fno-rtti
+    -fvisibility=hidden
+    -Wall
+    -Wextra
+    -Wpedantic
+    -Wno-unused-parameter
+    -Wno-c2y-extensions   # suppress clang-22 pedantic warning on __COUNTER__
+)
+
+set_target_properties(glibre-types PROPERTIES CXX_MODULE_STD OFF)
+
+# ---------------------------------------------------------------------------
+# Alias for consumers
+# ---------------------------------------------------------------------------
+
+add_library(glibre::types ALIAS glibre-types)
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -39,38 +39,54 @@ find_package(EASTL CONFIG REQUIRED)
 
 # ---------------------------------------------------------------------------
 # Codegen custom command — invokes glibre-foryc over data/schemas/**/*.fory
+#
+# Guards:
+#   1. if(EXISTS …/data/schemas) — data/schemas/ is tracked via a .gitkeep
+#      but the guard prevents breakage if someone manually deletes it.
+#      Without this guard, a missing directory causes glibre-foryc to return
+#      EXIT_FAILURE, which fails cmake --build even on a clean checkout.
+#   2. if(TARGET glibre-foryc) — GLIBRE_BUILD_TOOLS=OFF omits tools/ from
+#      the build; the codegen block must not reference a non-existent target.
 # ---------------------------------------------------------------------------
-
-file(GLOB_RECURSE GLIBRE_SCHEMAS
-    CONFIGURE_DEPENDS
-    "${CMAKE_SOURCE_DIR}/data/schemas/*.fory"
-)
 
 set(GEN_DIR "${CMAKE_BINARY_DIR}/generated/glibre-types")
 
-# glibre-foryc is built by tools/foryc/CMakeLists.txt (plan #219).
-# It accepts --in <schema-dir> --out <gen-dir> --stamp <stamp-file>.
-# The stamp file is the single output CMake tracks; glibre-foryc touches
-# it after successfully writing all generated files.
-#
-# Use $<TARGET_FILE:glibre-foryc> (a generator expression) so the full path
-# to the binary is resolved at build time regardless of the working directory
-# from which Ninja invokes the command.  This avoids "command not found" when
-# glibre-foryc is not on PATH.
-add_custom_command(
-    OUTPUT  "${GEN_DIR}/.stamp"
-    COMMAND "$<TARGET_FILE:glibre-foryc>"
-            --in    "${CMAKE_SOURCE_DIR}/data/schemas"
-            --out   "${GEN_DIR}"
-            --stamp "${GEN_DIR}/.stamp"
-    DEPENDS glibre-foryc ${GLIBRE_SCHEMAS}
-    COMMENT "glibre-foryc: regenerating middleman types"
-    VERBATIM
-)
+if(TARGET glibre-foryc AND EXISTS "${CMAKE_SOURCE_DIR}/data/schemas")
+    file(GLOB_RECURSE GLIBRE_SCHEMAS
+        CONFIGURE_DEPENDS
+        "${CMAKE_SOURCE_DIR}/data/schemas/*.fory"
+    )
 
-add_custom_target(glibre-types-codegen
-    DEPENDS "${GEN_DIR}/.stamp"
-)
+    # glibre-foryc is built by tools/foryc/CMakeLists.txt (plan #219).
+    # It accepts --in <schema-dir> --out <gen-dir> --stamp <stamp-file>.
+    # The stamp file is the single output CMake tracks; glibre-foryc touches
+    # it after successfully writing all generated files.
+    #
+    # Use $<TARGET_FILE:glibre-foryc> (a generator expression) so the full path
+    # to the binary is resolved at build time regardless of the working directory
+    # from which Ninja invokes the command.  This avoids "command not found" when
+    # glibre-foryc is not on PATH.
+    add_custom_command(
+        OUTPUT  "${GEN_DIR}/.stamp"
+        COMMAND "$<TARGET_FILE:glibre-foryc>"
+                --in    "${CMAKE_SOURCE_DIR}/data/schemas"
+                --out   "${GEN_DIR}"
+                --stamp "${GEN_DIR}/.stamp"
+        DEPENDS glibre-foryc ${GLIBRE_SCHEMAS}
+        COMMENT "glibre-foryc: regenerating middleman types"
+        VERBATIM
+    )
+
+    add_custom_target(glibre-types-codegen
+        DEPENDS "${GEN_DIR}/.stamp"
+    )
+else()
+    # No schemas directory or tools are disabled: create a no-op codegen target
+    # so that add_dependencies(glibre-types glibre-types-codegen) below still
+    # resolves.  This allows schema-less builds (current state before plans
+    # #220/#225 land) and GLIBRE_BUILD_TOOLS=OFF configurations.
+    add_custom_target(glibre-types-codegen)
+endif()
 
 # ---------------------------------------------------------------------------
 # Generated sources glob (populated after codegen runs, plans #220/#225)
@@ -144,10 +160,23 @@ target_compile_options(glibre-types PRIVATE
     -Wextra
     -Wpedantic
     -Wno-unused-parameter
-    -Wno-c2y-extensions   # suppress clang-22 pedantic warning on __COUNTER__
+    # -Wno-c2y-extensions: omitted here — neither types_anchor.cpp nor
+    # eastl_alloc.cpp use __COUNTER__.  Re-add when generated TUs under
+    # ${GEN_SRCS} (plans #220/#225) actually trigger the warning.
 )
 
-set_target_properties(glibre-types PROPERTIES CXX_MODULE_STD OFF)
+# SONAME versioning (plugin-abi.md §"Versioning Rules" point 2).
+# VERSION  = major.minor.patch library file version.
+# SOVERSION = major-only; the dynamic linker uses this as the SONAME.
+# Both start at 0.1.0 / 0 for the initial pre-codegen release.
+# Bump SOVERSION only on layout-breaking schema changes; bump VERSION on
+# every release.  The loader's hash check covers hash-stable window
+# changes; dyld SONAME check covers the layout-breaking window.
+set_target_properties(glibre-types PROPERTIES
+    VERSION   0.1.0
+    SOVERSION 0
+    CXX_MODULE_STD OFF
+)
 
 # ---------------------------------------------------------------------------
 # Alias for consumers
--- a/data/src/eastl_alloc.cpp
+++ b/data/src/eastl_alloc.cpp
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+// data/src/eastl_alloc.cpp
+//
+// EASTL operator new[] substrate for glibre-types.dylib.
+//
+// Per reviews/decisions/plugin-abi.md §"Plugin file shape" rule 2:
+//   "Each plugin links ONLY glibre-types.dylib from the engine's side.
+//    It must not link glibre-core or any other plugin."
+//
+// EASTL requires the project to supply custom operator new[] overloads that
+// eastl::allocator dispatches through.  Because plugins link ONLY
+// glibre-types.dylib (not glibre-core), glibre-types must carry these
+// overloads so that any EASTL container used inside glibre-types and by
+// plugins via the PUBLIC EASTL link resolves correctly.
+//
+// Duplication note: core/src/eastl_alloc.cpp carries identical overloads
+// for the glibre-core STATIC library.  The duplication is intentional and
+// accepted: moving the overloads from glibre-core to glibre-types would
+// cause glibre-core to link glibre-types (introducing a circular-ish
+// dependency between the core hub and the middleman dylib).  Keeping both
+// copies avoids that coupling.  Both TUs define the same global ::operator
+// new[] signatures; only one definition is resolved per dylib boundary
+// at link time, so there is no ODR violation.
+//
+// References:
+//   EASTL docs §"Custom Memory Allocators"
+//   core/src/eastl_alloc.cpp — canonical model for this copy
+//   PHILOSOPHY §11 — EASTL replaces std:: containers project-wide
+//
+// Threading: no lock needed — aligned_alloc is thread-safe on macOS.
+// Deallocation: EASTL deallocate calls operator delete[](void*), which
+// correctly releases aligned_alloc results via macOS libSystem's free().
+
+#include <cstddef>
+#include <cstdlib>
+#include <new>
+
+// ---------------------------------------------------------------------------
+// EASTL allocator new[] overloads (global scope, not in any namespace)
+// ---------------------------------------------------------------------------
+
+// Simple overload: allocate 'size' bytes with 16-byte alignment.
+void* operator new
+    [](std::size_t size,
+       const char* /*pName*/,
+       int /*flags*/,
+       unsigned /*debugFlags*/,
+       const char* /*file*/,
+       int /*line*/) {
+    if (size == 0)
+        size = 1;
+    size = (size + 15u) & ~static_cast<std::size_t>(15u);
+    void* p = std::aligned_alloc(16u, size);
+    if (!p)
+        __builtin_trap();  // -fno-exceptions: abort on OOM
+    return p;
+}
+
+// Aligned overload: allocate 'size' bytes with the requested alignment.
+void* operator new
+    [](std::size_t size,
+       std::size_t alignment,
+       std::size_t /*alignmentOffset*/,
+       const char* /*pName*/,
+       int /*flags*/,
+       unsigned /*debugFlags*/,
+       const char* /*file*/,
+       int /*line*/) {
+    if (size == 0)
+        size = 1;
+    if (alignment < 16u)
+        alignment = 16u;
+    size = (size + alignment - 1u) & ~(alignment - 1u);
+    void* p = std::aligned_alloc(alignment, size);
+    if (!p)
+        __builtin_trap();  // -fno-exceptions: abort on OOM
+    return p;
+}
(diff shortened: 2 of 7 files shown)


data/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+# to the binary is resolved at build time regardless of the working directory
+# from which Ninja invokes the command.  This avoids "command not found" when
+# glibre-foryc is not on PATH.
+add_custom_command(
+    OUTPUT  "${GEN_DIR}/.stamp"
+    COMMAND "$<TARGET_FILE:glibre-foryc>"
+            --in    "${CMAKE_SOURCE_DIR}/data/schemas"
+            --out   "${GEN_DIR}"
+            --stamp "${GEN_DIR}/.stamp"
+    DEPENDS glibre-foryc ${GLIBRE_SCHEMAS}
```
cjhowe-us: severity:HIGH location:data/CMakeLists.txt:60-69 problem:`add_custom_command` invokes `glibre-foryc --in ${CMAKE_SOURCE_DIR}/data/schemas` unconditionally, but `data/schemas/` is an empty untracked directory in this branch (git does not preserve empty dirs). On a fresh clone the directory will not exist, and `tools/foryc/src/main.cpp:124-127` returns `EXIT_FAILURE` if `--in` does not exist as a directory → `cmake --build --target glibre-types` fails on a clean checkout, which will block CI (DoD asserts `workflow_passed: ci.yml`). fix:Either (a) add a tracked `data/schemas/.gitkeep` so the directory exists at clone time, OR (b) wrap the `add_custom_command`/`add_custom_target` block in `if(EXISTS "${CMAKE_SOURCE_DIR}/data/schemas")` and skip codegen entirely when no schema dir is present (preferable: scheme-less builds today, schema-full once #220/#225 land), OR (c) teach `glibre-foryc` to treat a missing `--in` dir as zero schemas (success). Pick one before merge.

data/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+# ---------------------------------------------------------------------------
+
+add_library(glibre-types SHARED
+    src/types_anchor.cpp
+    src/eastl_alloc.cpp
+    ${GEN_SRCS}
+)
+
+# Depend on codegen so Ninja rebuilds the generated sources when schemas change.
+add_dependencies(glibre-types glibre-types-codegen)
```
cjhowe-us, line 122: severity:MED location:data/CMakeLists.txt:62,66,106 problem:`data/CMakeLists.txt` references `glibre-foryc` (via `$<TARGET_FILE:...>`, `DEPENDS glibre-foryc`, and `add_dependencies(... glibre-types-codegen)`) but the root CMakeLists introduces a new `option(GLIBRE_BUILD_TOOLS ON)` that gates `add_subdirectory(tools)`. When a downstream consumer configures with `-DGLIBRE_BUILD_TOOLS=OFF`, the `glibre-foryc` target does not exist and CMake errors out generating the buildsystem. The PR added the option in the same change so the breakage ships with it. fix:Either (a) gate `data/CMakeLists.txt`s codegen block on `if(TARGET glibre-foryc)`, OR (b) make `data` require `GLIBRE_BUILD_TOOLS=ON` with a clear `message(FATAL_ERROR …)`, OR (c) drop the `GLIBRE_BUILD_TOOLS` option until a real consumer needs to disable tools. Strict scope-of-#224 is just `data/`, so option (a) is the smallest fix.

data/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+    -fno-rtti
+    -fvisibility=hidden
+    -Wall
+    -Wextra
+    -Wpedantic
+    -Wno-unused-parameter
+    -Wno-c2y-extensions   # suppress clang-22 pedantic warning on __COUNTER__
+)
+
+set_target_properties(glibre-types PROPERTIES CXX_MODULE_STD OFF)
```
cjhowe-us: severity:MED location:data/CMakeLists.txt:13-20,150 problem:The header comment block describes a SONAME policy in detail ("bumped manually only on layout-breaking schema change … SONAME unchanged") but the CMake target never sets `VERSION` / `SOVERSION` properties. `add_library(glibre-types SHARED …)` produces `libglibre-types.dylib` with no SONAME / `install_name` versioning at all. The policy is currently aspirational — there is no SONAME to "keep stable" or "bump", so the dynamic-linker leg of the two-axis ABI gate (plugin-abi.md §"Versioning Rules" point 2) is non-functional. Issue #224 §Scope point 4 only asks the rule to be documented in CMake comments, so this is on the line between scope and follow-up; calling it out so it is not silently lost. fix:Either (a) add `set_target_properties(glibre-types PROPERTIES VERSION 0 SOVERSION 0 MACHO_COMPATIBILITY_VERSION 1.0.0 MACHO_CURRENT_VERSION 1.0.0)` now to materialize the policy, OR (b) explicitly file a follow-up issue (e.g. for the loader plan that consumes the SONAME) and add a `# TODO(#xxx)` reference next to the comment block so the gap is tracked. Prefer (a); it costs four lines and exercises the linker-side check from day one.

data/src/types_anchor.cpp
```diff
@@ -0,0 +1,56 @@
...
+// confirm the dylib loaded and is the expected revision.
+//
+// Visibility: default (not hidden) so dlsym can find it from outside.
+// All other symbols in this dylib use -fvisibility=hidden; this one is
+// explicitly exposed via the __attribute__((visibility("default"))) from
+// the PUBLIC include path if callers want to dlsym it, but the simple
+// inline extern-C function here does not need the attribute because
+// CMake will produce a normal symbol for it in the SHARED library
+// without -fvisibility=hidden narrowing (the PRIVATE compile option
+// applies only to this TU's non-C linkage symbols).
```
cjhowe-us: severity:MED location:data/src/types_anchor.cpp:40-47 problem:The comment block is internally inconsistent and factually wrong about visibility. It says "the simple inline extern-C function here does not need the attribute because CMake will produce a normal symbol for it in the SHARED library without -fvisibility=hidden narrowing (the PRIVATE compile option applies only to this TU s non-C linkage symbols)" — but `-fvisibility=hidden` applies to **all** symbols in the TU regardless of `extern "C"` linkage. The function is in fact only exported because line 51 carries `[[gnu::visibility("default")]]`. Without that attribute the symbol would be hidden and the test would fail to link. The code is correct; the comment misleads any future reader who tries to remove the attribute. fix:Rewrite the comment to: "Visibility: `-fvisibility=hidden` applies to all TU symbols (including `extern \"C\"`), so the explicit `[[gnu::visibility(\"default\")]]` on line 51 is **required** to expose this symbol to dlsym from the test executable / loader."

data/src/eastl_alloc.cpp
```diff
@@ -0,0 +1,78 @@
...
+//
+// References:
+//   EASTL docs §"Custom Memory Allocators"
+//   core/src/eastl_alloc.cpp — canonical model for this copy
+//   PHILOSOPHY §11 — EASTL replaces std:: containers project-wide
+//
+// Threading: no lock needed — aligned_alloc is thread-safe on macOS.
+// Deallocation: EASTL deallocate calls operator delete[](void*), which
+// correctly releases aligned_alloc results via macOS libSystem's free().
+
```
cjhowe-us, line 46: severity:LOW location:data/src/eastl_alloc.cpp:1-37 problem:Code is byte-equivalent to `core/src/eastl_alloc.cpp` but the comment block dropped two important notes the canonical core copy carries: (a) the deallocation-routing explanation (EASTL `deallocate` calls plain `delete[]`, which routes through libSystem `free()` — only sound on macOS; portability would need matching `operator delete[]` overrides), and (b) the explicit "macOS-only engine, MVP scope" caveat. The header note "Both TUs define the same global `::operator new[]` signatures; only one definition is resolved per dylib boundary at link time, so there is no ODR violation" is a non-trivial claim — one definition per dynamic image is correct on macOS, but worth a sentence on why two-level namespace + dyld duplicate-symbol resolution makes this safe (vs. ELF Linux where it would not be). fix:Either (a) port the deallocation-note paragraph from `core/src/eastl_alloc.cpp:20-30` verbatim into this file so future portability work catches both copies, OR (b) extract the shared rationale to a header comment in a new `core/include/glibre/internal/eastl_alloc_policy.hpp` and `#include` it from both TUs. (a) is in-scope here; (b) is a follow-up.

tests/data/dylib/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+)
+
+target_compile_features(glibre-types-dylib-tests PRIVATE cxx_std_23)
+
+# Same CXX_MODULE_STD OFF as all other test targets (Catch2 + clang-22
+# duplicate-symbol issue on macOS 26; see tests/core/CMakeLists.txt note).
+set_target_properties(glibre-types-dylib-tests PROPERTIES CXX_MODULE_STD OFF)
+
+target_compile_options(glibre-types-dylib-tests PRIVATE
+    -fno-exceptions
```
cjhowe-us: severity:LOW location:tests/data/dylib/CMakeLists.txt:34-42 problem:Test executable is built with `-fno-exceptions -fno-rtti`, but Catch2 raises C++ exceptions out of `REQUIRE` failures by default (the `REQUIRE` macro becomes `throw`). With `-fno-exceptions`, a thrown exception calls `std::terminate` instead of producing a structured failure report — the test would `abort` rather than print a diagnostic. The current single assertion (`glibre_types_abi_version() == 1` against a `return 1;` literal) cannot fail, so the issue is latent, but any future assertion in this TU silently loses Catch2 reporting. fix:Either (a) drop `-fno-exceptions -fno-rtti` from the test target — tests are not part of the shipping plugin link surface and Catch2 is the only consumer, OR (b) define `CATCH_CONFIG_DISABLE_EXCEPTIONS` and provide a custom `Catch::throw_exception` handler. (a) is the minimal fix.

data/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+# This matches the engine-wide ABI discipline and prevents accidental symbol leakage.
+target_compile_options(glibre-types PRIVATE
+    -fno-exceptions
+    -fno-rtti
+    -fvisibility=hidden
+    -Wall
+    -Wextra
+    -Wpedantic
+    -Wno-unused-parameter
+    -Wno-c2y-extensions   # suppress clang-22 pedantic warning on __COUNTER__
```
cjhowe-us: severity:LOW location:data/CMakeLists.txt:147 (and tests/data/dylib/CMakeLists.txt:41) problem:`-Wno-c2y-extensions   # suppress clang-22 pedantic warning on __COUNTER__` is added to glibre-types itself, but neither `types_anchor.cpp` nor `eastl_alloc.cpp` use `__COUNTER__`. The flag is propagated as PRIVATE (so consumers do not inherit it) but it is also dead in the dylib TU set — only the test exe genuinely needs it (Catch2 uses `__COUNTER__`). Inheriting the flag here is harmless but obscures what each target actually uses. fix:Drop the flag from `data/CMakeLists.txt:147`; keep it on `tests/data/dylib/CMakeLists.txt:41`. Re-add to `data/CMakeLists.txt` once a generated TU under `${GEN_SRCS}` (plans #220/#225) actually trips the warning.

data/src/types_anchor.cpp
```diff
@@ -0,0 +1,56 @@
...
+// window while glibre_types_abi_hash guards the hash-stable window.
+//
+// SONAME policy (mirrors CMake comment in data/CMakeLists.txt):
+//   - Major additive change  → bump abi_hash only, SONAME unchanged.
+//   - Layout-breaking change → bump SONAME AND abi_hash.
+//
+// The integer value here is separate from the Fory-codegen hash — it
+// answers "can the loader dlopen this dylib at all?" before the full
+// hash check runs.  Start at 1; subsequent bumps are recorded in the
+// data/CHANGELOG section of fory-codegen.md.
```
cjhowe-us: severity:LOW location:data/src/types_anchor.cpp:31 problem:Comment references "the data/CHANGELOG section of fory-codegen.md" as the place subsequent abi_version bumps are recorded, but `reviews/decisions/fory-codegen.md` has no `data/CHANGELOG` section (verified by reading the full file). This will mislead a future maintainer hunting for the changelog. fix:Either (a) drop the cross-reference and rely on the PR commit history as the changelog (consistent with the projects "no docs/ folder, no on-disk progress logs" rule), OR (b) replace with a concrete pointer to the new spec section once it exists (e.g. specs/data/SPEC.md §"ABI versions"). Prefer (a) for round-1 close-out.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:3
  LOW:4
carryover:0
notes:Coverage of #224 §Scope is essentially correct — `add_library(glibre-types SHARED ...)` lands with `glibre::types` alias, EASTL PUBLIC, Fory/blake3 commented PRIVATE placeholders, codegen custom_command wired via `$<TARGET_FILE:glibre-foryc>`, anchor TU exporting `glibre_types_abi_version() == 1`, and a Catch2 test `glibre_types_dylib_links_minimal_consumer` that links ONLY `glibre::types` (verified — no `glibre::core` in `tests/data/dylib/CMakeLists.txt`). DoD assertions all map to artefacts in the diff. PR body has `Closes #224`. Conventional Commit subject matches the issue.

What blocks merge: a single HIGH — the codegen custom_command unconditionally invokes `glibre-foryc --in ${SOURCE}/data/schemas`, but `data/schemas/` is an empty untracked directory (git does not preserve empty dirs) and the merged-on-main `tools/foryc/src/main.cpp:124-127` returns `EXIT_FAILURE` when `--in` does not exist. A fresh CI checkout will fail at build of `glibre-types`, blocking the `workflow_passed: ci.yml` DoD assertion. Three suggested fixes in the inline comment; smallest is `if(EXISTS …)` guard around the codegen block, OR a tracked `data/schemas/.gitkeep`.

What's deferred (correctly out of scope, sub-issue gated): Fory codegen output (#220 / #225), blake3 ABI hash export (#222), plugin loader hash check (#229), data-context migration (#221). The comments preserve the placeholder shape so those plans drop in additively.

MEDs: (1) `GLIBRE_BUILD_TOOLS=OFF` breaks `data` (new option introduced in same PR; gate codegen on `if(TARGET glibre-foryc)`); (2) SONAME policy is documented but not implemented — `set_target_properties(... SOVERSION 0 VERSION 0)` is a 1-line materialization; (3) `types_anchor.cpp:40-47` carries a comment that contradicts its own use of `[[gnu::visibility("default")]]`. LOWs: comment regression in duplicated `eastl_alloc.cpp` vs canonical `core/src/`, `-fno-exceptions` on a Catch2 test exec swallows future REQUIRE failures, dead `-Wno-c2y-extensions` on glibre-types proper, dangling cross-ref to a "data/CHANGELOG section of fory-codegen.md" that doesn't exist.

EASTL substrate duplication is byte-equivalent on the code (verified `diff` between `core/src/eastl_alloc.cpp` and `data/src/eastl_alloc.cpp` — only comments diverge). The duplication is correct per plugin-abi.md §"Plugin file shape" rule 2 and is well-rationalised in the file's header.

Recommend impl-respond addresses the HIGH (schemas/ existence) plus the three MEDs in a follow-up commit; LOWs may be deferred to round 2 cleanup.

data/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Belt-and-suspenders fix applied. Added data/schemas/.gitkeep so the directory exists on fresh clone AND wrapped the codegen block in if(TARGET glibre-foryc AND EXISTS "${CMAKE_SOURCE_DIR}/data/schemas") to guard against both missing dir and tool-disabled configurations. An else() branch creates a no-op add_custom_target(glibre-types-codegen) so add_dependencies() always resolves cleanly.

data/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Handled in the same if(TARGET glibre-foryc AND EXISTS ...) guard that fixes HIGH-1. The else() branch creates a no-op add_custom_target(glibre-types-codegen) so GLIBRE_BUILD_TOOLS=OFF configurations configure cleanly without referencing the absent glibre-foryc target.

data/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Added set_target_properties(glibre-types PROPERTIES VERSION 0.1.0 SOVERSION 0 CXX_MODULE_STD OFF) to materialize the SONAME policy documented in the header comment. This enables the dyld SONAME check leg of the two-axis ABI gate from day one per plugin-abi.md §Versioning Rules point 2.

data/src/types_anchor.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Rewrote the visibility comment in types_anchor.cpp:40-47. The new text explicitly states: -fvisibility=hidden applies to ALL symbols in the TU including extern C, so [[gnu::visibility("default")]] on line 51 is REQUIRED. Removing it would silently hide the symbol. The old text was factually wrong in claiming extern-C symbols were exempt from the PRIVATE compile option.

data/src/eastl_alloc.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Restored the deallocation-routing note from core/src/eastl_alloc.cpp into data/src/eastl_alloc.cpp — explains that EASTL allocator::deallocate calls plain delete[](char*)p which routes through the global operator delete[], not through any override here; sound on macOS libSystem since aligned_alloc results are free()-compatible. Also added the macOS-only engine MVP caveat and a two-level namespace / ELF portability note.

tests/data/dylib/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Dropped -fno-exceptions and -fno-rtti from tests/data/dylib/CMakeLists.txt. Test executables are not part of the shipping plugin link surface so exception support is appropriate. Added a comment explaining why: Catch2 REQUIRE throws on failure and needs exception support to produce structured diagnostics rather than calling std::terminate.

data/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Dropped -Wno-c2y-extensions from glibre-types compile options. Replaced with a comment explaining the flag is absent because neither types_anchor.cpp nor eastl_alloc.cpp use __COUNTER__, and it should be re-added once generated TUs from plans #220/#225 actually trigger the warning. Kept the flag on the test target where Catch2 genuinely uses __COUNTER__.

data/src/types_anchor.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:4ae92e7 reasoning:Replaced the dangling reference to 'the data/CHANGELOG section of fory-codegen.md' (which does not exist) with: 'subsequent bumps are captured in the PR commit history (no separate changelog file per project convention)'. This matches the project rule against on-disk status files / changelog files and correctly points future maintainers to git log as the record.

data/CMakeLists.txt
```diff
@@ -0,0 +1,185 @@
...
+)
+
+# Depend on codegen so Ninja rebuilds the generated sources when schemas change.
+add_dependencies(glibre-types glibre-types-codegen)
+
+# Core headers (glibre/error.hpp, glibre/core/plugin_manifest.hpp, …) are
+# exposed PUBLIC so that consumers of glibre-types get the include path
+# without having to separately link glibre-core.
+target_include_directories(glibre-types
+    PUBLIC
```
cjhowe-us, line 128: severity:MED location:data/CMakeLists.txt:121-128 problem:`Fory::fory` and `blake3::blake3` are still commented placeholders, but issue #224 §Scope point 3 explicitly says "links `Fory::fory` and `blake3::blake3` privately" as a deliverable of THIS plan. The diff defers them to plans #220/#222 — that is the correct call (no generated TUs reference them yet, so an active PRIVATE link would emit "unused library" diagnostics from ld), but the issue body still asserts they ship in #224. Cohesion-wise the seam this PR is supposed to establish (PUBLIC=EASTL only; PRIVATE=Fory+blake3 hidden behind the wrapper functions) is only half-realised. fix:Update issue #224 §Scope point 3 to read "links `Fory::fory` and `blake3::blake3` privately when generated TUs land in plans #220/#222" so the issue body matches the diff. Alternative: enable the PRIVATE link now and add `--as-needed` / accept the unused-libs warning. The first option preserves the current diff and resolves the prose drift.

tests/data/dylib/glibre_types_link_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+// "glibre_types_dylib_links_minimal_consumer" must match the
+// unit_test_named: entry in the issue's ## Definition of Done block.
+// ---------------------------------------------------------------------------
+
+TEST_CASE("glibre_types_dylib_links_minimal_consumer", "[data][dylib][abi]") {
+    // The sentinel value is 1 per data/src/types_anchor.cpp.
+    // Future layout-breaking schema changes bump this integer and update
+    // the corresponding SONAME per plugin-abi.md §"Versioning Rules" point 2.
+    REQUIRE(glibre_types_abi_version() == 1);
+}
```
cjhowe-us, line 52: severity:MED location:tests/data/dylib/glibre_types_link_test.cpp problem:Issue #224 §"Unit Test Plan" lists FOUR tests: (1) `glibre_types_dylib_exists_after_configure`, (2) `exports_glibre_types_abi_hash_symbol`, (3) `exports_serialize_deserialize_for_minimal_type`, (4) `links_only_fory_and_blake3_privately (nm/otool check)`. The PR ships ONLY the DoD-named test `glibre_types_dylib_links_minimal_consumer`. Three of the four prose-plan tests are correctly out-of-scope for this PR (#222 emits abi_hash; #220/#225 emit serialize/deserialize). However test (4) — the nm/otool seam check that validates `nothing-but-Fory-and-blake3-on-the-PRIVATE-side` — is the canonical guard against the cohesion failure the dylib seam exists to prevent, and could ship today against the empty PRIVATE list (assert nm output contains NO `Fory_*` or `blake3_*` symbols on the dylib export surface). The DoD trumps the prose test plan, so this does NOT block merge — but it leaves a real seam invariant unguarded. fix:Either (a) add a follow-up plan issue tracking `links_only_fory_and_blake3_privately` as a sub-plan of #5 wired to land alongside #220/#222 when the PRIVATE deps go live, OR (b) add the nm/otool assertion now (it would currently assert "no Fory/blake3 symbols", trivially satisfied, and would catch a regression the moment plan #220 wires the PRIVATE link incorrectly).

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:0
carryover:0
notes:All 8 round-1 findings (HIGH-1, MED-2/3/4, LOW-5/6/7/8) verified ADDRESSED in commit 4ae92e7. Codegen guard `if(TARGET glibre-foryc AND EXISTS "${CMAKE_SOURCE_DIR}/data/schemas")` collapses HIGH-1 + MED-2 cleanly; the `else()` branch creates a no-op `add_custom_target(glibre-types-codegen)` so `add_dependencies()` resolves regardless of whether tools/ or data/schemas/ are present (verified at lines 87-93). `data/schemas/.gitkeep` tracked. SONAME materialized via `set_target_properties(... VERSION 0.1.0 SOVERSION 0)` — matches the two-axis ABI policy in plugin-abi.md §"Versioning Rules" point 2. types_anchor.cpp visibility comment now correctly states `-fvisibility=hidden` applies to extern "C" symbols and `[[gnu::visibility("default")]]` is REQUIRED. eastl_alloc.cpp duplication note is honest about ELF non-portability AND restores the deallocation-routing + macOS-only caveats from the canonical core/src/ copy (verified by `diff` — only header comments differ; allocator code is byte-equivalent). Test target dropped `-fno-exceptions/-fno-rtti` so Catch2 REQUIRE works. Dead `-Wno-c2y-extensions` removed from glibre-types proper, kept on test target. Dangling fory-codegen.md cross-ref replaced with PR-history note.

Cohesion+SRP: data/CMakeLists.txt declares glibre-types lib + codegen wiring, no tendrils into core/. PUBLIC=EASTL only; PRIVATE link surface (Fory+blake3) is comment-placeholder until plans #220/#222. Plugin link surface in tests/data/dylib/CMakeLists.txt correctly excludes glibre::core.

DoD on #224 satisfied: file_exists(data/CMakeLists.txt) ✓, file_contains(`add_library\\(glibre-types`) ✓, unit_test_named(`glibre_types_dylib_links_minimal_consumer`) ✓. PR body has `Closes #224`. Conventional Commit subject correct.

Two new MED findings (cohesion-lens):
  MED-9 — issue #224 §Scope point 3 still asserts "links Fory::fory and blake3::blake3 privately" but the diff defers them to plans #220/#222. Update issue body or activate the PRIVATE link.
  MED-10 — issue #224 §"Unit Test Plan" lists `links_only_fory_and_blake3_privately (nm/otool check)` as a deliverable. The DoD does not require it (DoD trumps prose plan, so this does NOT block merge), but it leaves the canonical seam invariant — "nothing but Fory/blake3 ever leaks onto the PRIVATE link surface" — unguarded. Recommend a follow-up plan issue or an inline nm/otool check today.

Both MEDs are scope-prose drift between the issue body and the diff, not code defects. Recommend: open a one-line issue-body amendment (or a follow-up plan-issue under #5) and proceed to round 3 / merge.

Out-of-scope context (NOT a finding against this PR): the lint job on the latest CI run failed on `core/include/glibre/core/plugin_entry.hpp`, `examples/plugin-noop/plugin_noop.cpp`, and `tests/core/plugin_api/plugin_api_test.cpp`. None of those files are in PR #957's diff; the violations were merged to main in commit d3f3a8f via PR #956 ("PluginContext header"). The lint failure on PR #957 is pre-existing main debt that this branch inherited. It will, however, block the `workflow_passed: ci.yml` DoD assertion when dod-verify fires on merge — recommend the orchestrator dispatch a small `chore(ci)` cleanup against main before merging #957, or accept that #957's merge will be gated until the upstream lint debt clears.

data/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED reasoning:Amended issue #224 body via `gh issue edit` to explicitly state that Fory::fory and blake3::blake3 are deferred to plans #220 (header emit, which lands the first generated TUs referencing Fory) and #222 (ABI hash export, which lands blake3) respectively. §Scope point 3 now reads that those PRIVATE deps are comment-placeholder only in this plan and are activated in those later plans. §Scope out-of-scope list explicitly names 'Fory::fory and blake3::blake3 PRIVATE link activation (plans #220/#222)'. The CMake link stanza already has this correct as comment-placeholders; the issue body now matches. See https://github.com/cjhowe-us/glibre/issues/224

tests/data/dylib/glibre_types_link_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:521ec49 reasoning:Added TEST_CASE `glibre_types_dylib_excludes_fory_and_blake3_symbols` to tests/data/dylib/glibre_types_link_test.cpp. The test runs `nm -gU` on libglibre-types.dylib (path injected at compile time via GLIBRE_TYPES_DYLIB_PATH in tests/data/dylib/CMakeLists.txt using $<TARGET_FILE:glibre-types>) and asserts no Fory or BLAKE3 symbols (__ZN4fory, _fory_, _blake3_) appear in the public symbol table. Passes today (neither dep is linked); guards against future regression from plans #220/#222. Updated issue #224 DoD block to add `unit_test_named: glibre_types_dylib_excludes_fory_and_blake3_symbols`. Both tests verified green locally: 2/2 tests passed.

tests/data/dylib/glibre_types_link_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    // Fory C++ symbols are mangled under the `fory` namespace; the common
+    // prefixes on macOS are `__ZN4fory` (mangled) or `_fory_` (C linkage).
+    // BLAKE3 symbols are exported under `_blake3_` (C linkage from the C
+    // implementation).  Neither should appear in glibre-types' public table
+    // until plans #220/#222 land AND explicitly choose to re-export them
+    // (which they must not: they are PRIVATE per plugin-abi.md).
+    REQUIRE_FALSE(nm_output.find("__ZN4fory") != std::string::npos);
+    REQUIRE_FALSE(nm_output.find("_fory_")    != std::string::npos);
+    REQUIRE_FALSE(nm_output.find("_blake3_")  != std::string::npos);
```
cjhowe-us: severity:HIGH location:tests/data/dylib/glibre_types_link_test.cpp:78,79,104,105 problem:clang-format violations introduced by post-r2 commit 521ec49 are failing CI's `lint` job, which blocks the issue #224 DoD assertion `workflow_passed: ci.yml`. Unlike the pre-existing lint debt flagged in r2 (files not in this PR), these violations are ON THIS PR's new test file: `FAIL("...")` indented argument, multi-space alignment around `!= std::string::npos`. Confirmed in run 25589924595 logs. fix:Run `clang-format -i tests/data/dylib/glibre_types_link_test.cpp` and amend with a fresh commit (do NOT amend 521ec49). After CI is green, merge can proceed.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:0
carryover:0
notes:Round-3 (topology + spec alignment + polish) on PR #957 head 521ec49. Topology is sound: data/ subdirectory canonical position; tests/data/dylib/ correctly under tests/data/; `glibre::types` alias materialised; SONAME (VERSION 0.1.0/SOVERSION 0) per plugin-abi.md §"Versioning Rules" point 2; codegen graph guarded for both tools-disabled and missing-schemas-dir cases per fory-codegen.md §"CMake Integration"; plugin link surface (test exec links ONLY glibre::types, no glibre::core) honours plugin-abi.md §"Plugin file shape" rule 2. Sibling-issue topology correct: r2-deferred Fory/blake3 PRIVATE deps re-pointed to plans #220 and #222 in issue #224 §Scope and the CMake comment placeholder. DoD block on #224 lists both `glibre_types_dylib_links_minimal_consumer` AND the new `glibre_types_dylib_excludes_fory_and_blake3_symbols` (added per r2 MED-10); regex assertions and Conventional Commit subject + `Closes #224` all match. nm/popen seam: dylib path is the cmake `$<TARGET_FILE:glibre-types>` generator-expression, build-tree-controlled — no shell injection surface.

One blocking HIGH: post-r2 commit 521ec49 introduces clang-format violations in tests/data/dylib/glibre_types_link_test.cpp (lines 78,79,104,105 — `FAIL(...)` indent + multi-space `!= std::string::npos` alignment). The lint job on the latest CI run (25589924595) fails specifically on these new lines, blocking the `workflow_passed: ci.yml` DoD entry. Distinct from the pre-existing main-branch lint debt flagged in r2; this is fresh debt introduced by this PR. Trivial fix: `clang-format -i` on that one file and a new follow-up commit (do not amend 521ec49). With that single fix, all DoD assertions for #224 will satisfy and the PR is mergeable.

tests/data/dylib/glibre_types_link_test.cpp
cjhowe-us: decision:ADDRESSED commit:59f462a1cd84f709821bcc8f1436a12b405fd818

Issue #226, opened 2026-05-03
[PLAN] test(data): codegen golden output harness


cjhowe-us:
### Domain

domain:data

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:3

### User Stories Satisfied

Quality bar for every persistent-type user-story; matches
`fory-codegen.md` §"Consequences" point 2 (every persistent type ships
with at least one Catch2 round-trip test under `tests/data/schemas/`).

### Parent (Epic or Sub-Epic)

#5

### Scope

Establish the codegen golden-test harness:

1. Fixture infrastructure under `tests/tools/foryc/golden/` plus a
   Catch2 helper that parses fixture schemas programmatically and
   diffs emitted header output against checked-in goldens.
2. Goldens for: `minimal_struct.fory`, `nested_namespace.fory`,
   `all_scalars.fory` (every builtin scalar), `with_generics.fory`
   (list/map/option fields).
3. Diff helper produces a diagnostic message when goldens drift.
4. CI invokes the harness in the `tests` step.
5. GLIBRE_UPDATE_GOLDEN=1 regenerates checked-in goldens.

This plan does not author new schemas for shipping types; those are
authored in their owning context's plans.

Out of scope:
- migration round-trip semantics (separate plan)
- non-codegen integration tests

### Unit Test Plan

- foryc_golden_emit_header_matches_expected

### Depends on

- #220  (foryc emit headers)
- #221  (foryc migrations)
- #222  (foryc abi hash export)

### Agent Execution Notes

Goldens are diff-reviewed like normal source. Bumping a golden
intentionally is a one-line PR titled
`test(data): refresh codegen goldens for <reason>`. Drifts caused by
unrelated work fail CI and surface the diff in the PR.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: tests/tools/foryc/golden/fixtures/minimal_struct.fory
- file_exists: tests/tools/foryc/golden/fixtures/nested_namespace.fory
- file_exists: tests/tools/foryc/golden/fixtures/all_scalars.fory
- file_exists: tests/tools/foryc/golden/fixtures/with_generics.fory
- file_exists: tests/tools/foryc/golden/expected/minimal_struct.hpp.golden
- file_exists: tests/tools/foryc/golden/expected/nested_namespace.hpp.golden
- file_exists: tests/tools/foryc/golden/expected/all_scalars.hpp.golden
[...]

PR #961, opened 2026-05-09
feat(data): glibre-foryc — schema source-hash + ABI hash export (refs #222)

### LINKED ISSUE #222: [PLAN] feat(data): glbr-sergeant — schema source-hash + ABI hash export

cjhowe-us:
## Summary

Extends `glibre-foryc` with deterministic schema source-hash and ABI hash computation using blake3.

- `tools/foryc/src/abi_hash.{hpp,cpp}` — new module with `compute_source_hash`, `compute_abi_hash`, and `format_as_uint64_hex`
- `tools/foryc/src/main.cpp` — adds `--emit=abi-hash` CLI mode (prints `<path>\t<source_hash>\t<abi_hash>` per schema, for build-system stamp consumption)
- `tools/foryc/CMakeLists.txt` — links `BLAKE3::blake3` (vcpkg `blake3` port, already in `vcpkg.json`)
- `tests/tools/foryc/foryc_abi_hash_test.cpp` — 11 Catch2 unit tests covering determinism, field rename sensitivity, tag-order sensitivity, and whitespace invariance for both source and ABI hashes
- `tests/tools/foryc/CMakeLists.txt` — adds `foryc-abi-hash-tests` target

**Source hash** canonicalizes via tokenization (strips comments, collapses whitespace between tokens to single spaces) before blake3-hashing. Two schemas that differ only in whitespace or comments yield the same source hash.

**ABI hash** walks the parsed IR: types sorted by fqn, fields sorted by tag ascending; the canonical representation is independent of declaration order in source files.

Both functions return `Result<Blake3Digest>` for composability with the parse pipeline.

Per `fory-codegen.md` §"middleman dylib exposes" and `plugin-abi.md` §"ABI Hash Function": the ABI hash will gate plugin compatibility in #230. Per-plugin manifest emission (#225) will consume the ABI hash. Both are separate plans.

## Out of scope

- Migration dispatcher (#221)
- Per-plugin manifest.cpp emission (#225)
- Loader-side ABI hash gate (#230)
- Codegen golden harness (#226)
- `emit_header.cpp` embedding of hash constants (plan #220)

## Refs

Closes #222

## Test plan

- [x] `cmake --build ... --target foryc-abi-hash-tests` — clean build
- [x] `ctest ... -R foryc_abi_hash_` — 11/11 passed
- [x] `ctest ... -R foryc_source_hash_` — 2/2 passed
- [x] All 45 ctest cases (minus 1 pre-existing frame_loop abort unrelated to this PR) pass
- [x] `--emit=abi-hash` mode compiles and links cleanly against BLAKE3::blake3

### DIFF
--- a/tests/tools/foryc/foryc_abi_hash_test.cpp
+++ b/tests/tools/foryc/foryc_abi_hash_test.cpp
@@ -1,0 +1,299 @@
+// SPDX-License-Identifier: Apache-2.0
+// tests/tools/foryc/foryc_abi_hash_test.cpp
+//
+// Catch2 unit tests for glibre-foryc abi_hash module (plan #222).
+//
+// Test names match the Unit Test Plan in issue #222 (as adjusted by the
+// dispatch prompt analysis):
+//
+//   ABI hash tests:
+//     foryc_abi_hash_is_deterministic
+//     foryc_abi_hash_changes_on_field_rename
+//     foryc_abi_hash_changes_on_field_reorder_by_tag
+//     foryc_abi_hash_unchanged_on_whitespace
+//
+//   Source hash tests:
+//     foryc_source_hash_unchanged_on_whitespace
+//     foryc_source_hash_changes_on_field_rename
+//
+//   Format helper test:
+//     foryc_format_as_uint64_hex_is_16_chars
+//
+//   Issue-body unit test plan aliases:
+//     abi_hash_stable_across_runs            (alias for deterministic)
+//     abi_hash_changes_on_schema_edit        (covered by field_rename)
+//     abi_hash_independent_of_schema_file_order_on_disk
+//     per_type_schema_hash_matches_blake3_of_source
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "abi_hash.hpp"
+#include "parser.hpp"
+
+using namespace glibre::tools::foryc;
+
+// -----------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------
+
+// Parse a .fory source string, asserting success.
+static Schema parse_ok(std::string_view src) {
+    auto r = parse_string(src, "<test>");
+    REQUIRE(r.has_value());
+    return std::move(*r);
+}
+
+// Compute ABI hash from a .fory source string, asserting success.
+static Blake3Digest abi_hash_of(std::string_view src) {
+    const Schema schema = parse_ok(src);
+    auto r = compute_abi_hash(schema);
+    REQUIRE(r.has_value());
+    return *r;
+}
+
+// Compute source hash, asserting success.
+static Blake3Digest source_hash_of(std::string_view src) {
+    auto r = compute_source_hash(src);
+    REQUIRE(r.has_value());
+    return *r;
+}
+
+// -----------------------------------------------------------------------
+// ABI hash tests
+// -----------------------------------------------------------------------
+
+TEST_CASE("foryc_abi_hash_is_deterministic", "[foryc][abi_hash]") {
+    // Calling compute_abi_hash twice on the same Schema produces the same
+    // 32-byte digest.  This asserts that the function is a pure function of
+    // the IR with no randomness or time-based inputs.
+    constexpr std::string_view src = R"(
+schema glibre.test.Foo {
+  version 1
+  field x : u32 tag 1
+  field y : f32 tag 2
+}
+)";
+    const Schema schema = parse_ok(src);
+    auto r1 = compute_abi_hash(schema);
+    auto r2 = compute_abi_hash(schema);
+    REQUIRE(r1.has_value());
+    REQUIRE(r2.has_value());
+    CHECK(*r1 == *r2);
+}
+
+TEST_CASE("abi_hash_stable_across_runs", "[foryc][abi_hash]") {
+    // Alias for foryc_abi_hash_is_deterministic (issue body name).
+    // Computing the hash of the same schema on two consecutive calls must
+    // produce the same result (no PRNG, no clock, no env inputs).
+    constexpr std::string_view src = R"(
+schema StableSchema {
+  version 2
+  field a : u64 tag 1
+}
+)";
+    const Schema schema = parse_ok(src);
+    auto h1 = compute_abi_hash(schema);
+    auto h2 = compute_abi_hash(schema);
+    REQUIRE(h1.has_value());
+    REQUIRE(h2.has_value());
+    CHECK(*h1 == *h2);
+}
+
+TEST_CASE("foryc_abi_hash_changes_on_field_rename", "[foryc][abi_hash]") {
+    // Renaming a field while keeping type and tag the same must produce a
+    // different ABI hash.  Field name is part of the ABI contract
+    // (serialized struct members have named accessors that break on rename).
+    constexpr std::string_view src_before = R"(
+schema glibre.test.Widget {
+  version 1
+  field old_name : u32 tag 1
+}
+)";
+    constexpr std::string_view src_after = R"(
+schema glibre.test.Widget {
+  version 1
+  field new_name : u32 tag 1
+}
+)";
+    CHECK(abi_hash_of(src_before) != abi_hash_of(src_after));
+}
+
+TEST_CASE("abi_hash_changes_on_schema_edit", "[foryc][abi_hash]") {
+    // Alias for foryc_abi_hash_changes_on_field_rename (issue body name).
+    // Any semantic change to the schema IR must change the ABI hash.
+    constexpr std::string_view src_v1 = R"(
+schema EditSchema {
+  version 1
+  field count : u32 tag 1
+}
+)";
+    constexpr std::string_view src_v2 = R"(
+schema EditSchema {
+  version 1
+  field total : u32 tag 1
+}
+)";
+    CHECK(abi_hash_of(src_v1) != abi_hash_of(src_v2));
+}
+
+TEST_CASE("foryc_abi_hash_changes_on_field_reorder_by_tag", "[foryc][abi_hash]") {
+    // Two schemas with the same fields but different tag assignments must
+    // produce different ABI hashes.  Tag assignment determines on-disk
+    // layout (per fory-codegen.md §ABI Stability Rules); renumbering tags
+    // is an ABI-breaking change.
+    constexpr std::string_view src_original = R"(
+schema glibre.test.Ordered {
+  version 1
+  field x : u32 tag 1
+  field y : f32 tag 2
+}
+)";
+    constexpr std::string_view src_retagged = R"(
+schema glibre.test.Ordered {
+  version 1
+  field x : u32 tag 2
+  field y : f32 tag 1
+}
+)";
+    CHECK(abi_hash_of(src_original) != abi_hash_of(src_retagged));
+}
+
+TEST_CASE("foryc_abi_hash_unchanged_on_whitespace", "[foryc][abi_hash]") {
+    // Two schemas that are identical except for indentation and blank lines
+    // must produce the same ABI hash.  Whitespace is not part of the ABI
+    // contract; only fqn, version, field tags, types, and names matter.
+    constexpr std::string_view compact = R"(schema glibre.test.Ws{version 1
+field x:u32 tag 1
+field y:f32 tag 2})";
+
+    constexpr std::string_view spaced = R"(
+schema glibre.test.Ws {
+  version 1
+
+  field x : u32   tag 1
+  field y : f32   tag 2
+
+}
+)";
+    // Both must parse successfully and yield the same ABI hash.
+    CHECK(abi_hash_of(compact) == abi_hash_of(spaced));
+}
+
+TEST_CASE("abi_hash_independent_of_schema_file_order_on_disk", "[foryc][abi_hash]") {
+    // This test verifies that the ABI hash for a schema collection with
+    // multiple types is independent of the declaration order of those types
+    // in the source file.  compute_abi_hash sorts types by fqn before
+    // hashing, so two files that declare the same types in different order
+    // must produce the same hash.
+    constexpr std::string_view src_ab = R"(
+schema glibre.test.Alpha {
+  version 1
+  field a : u32 tag 1
+}
+schema glibre.test.Beta {
+  version 1
+  field b : f32 tag 1
+}
+)";
+    constexpr std::string_view src_ba = R"(
+schema glibre.test.Beta {
+  version 1
+  field b : f32 tag 1
+}
+schema glibre.test.Alpha {
+  version 1
+  field a : u32 tag 1
+}
+)";
+    // Parse into two schemas and compare ABI hashes.
+    CHECK(abi_hash_of(src_ab) == abi_hash_of(src_ba));
+}
+
+// -----------------------------------------------------------------------
+// Source hash tests
+// -----------------------------------------------------------------------
+
+TEST_CASE("foryc_source_hash_unchanged_on_whitespace", "[foryc][source_hash]") {
+    // The source hash canonicalizes whitespace before hashing.  Two schema
+    // sources that differ only in whitespace (indentation, blank lines) must
+    // yield the same source hash.
+    constexpr std::string_view src_compact = "schema Foo{version 1 field x:u32 tag 1}";
+    constexpr std::string_view src_spaced = R"(
+schema Foo {
+  version 1
+  field x : u32 tag 1
+}
+)";
+    CHECK(source_hash_of(src_compact) == source_hash_of(src_spaced));
+}
+
+TEST_CASE("foryc_source_hash_changes_on_field_rename", "[foryc][source_hash]") {
+    // Renaming a field must change the source hash even if the ABI hash
+    // also changes.  The source hash is over the token sequence; a rename
+    // changes the token sequence.
+    constexpr std::string_view src_before = R"(
+schema Foo {
+  version 1
+  field old_name : u32 tag 1
+}
+)";
+    constexpr std::string_view src_after = R"(
+schema Foo {
+  version 1
+  field new_name : u32 tag 1
+}
+)";
+    CHECK(source_hash_of(src_before) != source_hash_of(src_after));
+}
+
+TEST_CASE("per_type_schema_hash_matches_blake3_of_source", "[foryc][source_hash]") {
+    // The source hash computed by compute_source_hash for a single-type
+    // schema file must match the blake3 of the canonical token sequence.
+    // We verify this indirectly: two differently-whitespaced sources with
+    // the same token sequence produce the same hash (tested above), and two
+    // sources with different token sequences produce different hashes.
+    constexpr std::string_view src_a = R"(
+schema glibre.core.ComponentA {
+  version 1
+  field pos : vec3f tag 1
+}
+)";
+    constexpr std::string_view src_b = R"(
+schema glibre.core.ComponentB {
+  version 1
+  field pos : vec3f tag 1
+}
+)";
+    // Different FQN → different source token sequence → different source hash.
+    CHECK(source_hash_of(src_a) != source_hash_of(src_b));
+
+    // Same schema twice → same hash.
+    CHECK(source_hash_of(src_a) == source_hash_of(src_a));
+}
+
+// -----------------------------------------------------------------------
+// format_as_uint64_hex helper test
+// -----------------------------------------------------------------------
+
+TEST_CASE("foryc_format_as_uint64_hex_is_16_chars", "[foryc][abi_hash]") {
+    // format_as_uint64_hex must always return exactly 16 lowercase hex
+    // characters, with leading zeros where needed.
+    Blake3Digest all_zeros{};  // zero-initialized
+    const eastl::string hex_zeros = format_as_uint64_hex(all_zeros);
+    CHECK(hex_zeros.size() == 16);
+    CHECK(hex_zeros == eastl::string("0000000000000000"));
+
+    Blake3Digest all_ff{};
+    for (auto& b : all_ff)
+        b = 0xFF;
+    const eastl::string hex_ff = format_as_uint64_hex(all_ff);
+    CHECK(hex_ff.size() == 16);
+    CHECK(hex_ff == eastl::string("ffffffffffffffff"));
+
+    // Verify a known value: first 8 bytes [0x01, 0x00, ..., 0x00] →
+    // big-endian uint64 = 0x0100000000000000.
+    Blake3Digest known{};
+    known[0] = 0x01;
+    const eastl::string hex_known = format_as_uint64_hex(known);
+    CHECK(hex_known == eastl::string("0100000000000000"));
+}
--- a/tests/tools/foryc/foryc_abi_hash_test.cpp
+++ b/tests/tools/foryc/foryc_abi_hash_test.cpp
@@ -3,21 +3,24 @@
 //
 // Catch2 unit tests for glibre-foryc abi_hash module (plan #222).
 //
-// Test names match the Unit Test Plan in issue #222 (as adjusted by the
-// dispatch prompt analysis):
+// Test names match the Unit Test Plan in issue #222:
 //
-//   ABI hash tests:
+//   Collection ABI hash tests (compute_collection_abi_hash):
 //     foryc_abi_hash_is_deterministic
 //     foryc_abi_hash_changes_on_field_rename
 //     foryc_abi_hash_changes_on_field_reorder_by_tag
 //     foryc_abi_hash_unchanged_on_whitespace
+//     foryc_abi_hash_changes_on_version_bump
+//     foryc_collection_abi_hash_zero_schemas
+//     foryc_collection_abi_hash_one_type_zero_fields
 //
 //   Source hash tests:
-//     foryc_source_hash_unchanged_on_whitespace
+//     foryc_source_hash_changes_on_whitespace
 //     foryc_source_hash_changes_on_field_rename
 //
-//   Format helper test:
+//   Format helper tests:
 //     foryc_format_as_uint64_hex_is_16_chars
+//     foryc_format_as_full_hex_is_64_chars
 //
 //   Issue-body unit test plan aliases:
 //     abi_hash_stable_across_runs            (alias for deterministic)
@@ -43,15 +46,22 @@ static Schema parse_ok(std::string_view src) {
     return std::move(*r);
 }
 
-// Compute ABI hash from a .fory source string, asserting success.
+// Compute collection ABI hash from a single .fory source string.
+// Convenience wrapper: creates a single-schema collection.
 static Blake3Digest abi_hash_of(std::string_view src) {
     const Schema schema = parse_ok(src);
-    auto r = compute_abi_hash(schema);
+    auto src_hash_r = compute_source_hash(src);
+    REQUIRE(src_hash_r.has_value());
+    eastl::vector<Schema> schemas;
+    schemas.push_back(schema);
+    eastl::vector<Blake3Digest> digests;
+    digests.push_back(*src_hash_r);
+    auto r = compute_collection_abi_hash(schemas, digests);
     REQUIRE(r.has_value());
     return *r;
 }
 
-// Compute source hash, asserting success.
+// Compute raw source hash, asserting success.
 static Blake3Digest source_hash_of(std::string_view src) {
     auto r = compute_source_hash(src);
     REQUIRE(r.has_value());
@@ -59,13 +69,12 @@ static Blake3Digest source_hash_of(std::string_view src) {
 }
 
 // -----------------------------------------------------------------------
-// ABI hash tests
+// Collection ABI hash tests
 // -----------------------------------------------------------------------
 
 TEST_CASE("foryc_abi_hash_is_deterministic", "[foryc][abi_hash]") {
-    // Calling compute_abi_hash twice on the same Schema produces the same
-    // 32-byte digest.  This asserts that the function is a pure function of
-    // the IR with no randomness or time-based inputs.
+    // Calling compute_collection_abi_hash twice on the same inputs produces
+    // the same 32-byte digest.  No randomness or time-based inputs.
     constexpr std::string_view src = R"(
 schema glibre.test.Foo {
   version 1
@@ -74,8 +83,16 @@ schema glibre.test.Foo {
 }
 )";
     const Schema schema = parse_ok(src);
-    auto r1 = compute_abi_hash(schema);
-    auto r2 = compute_abi_hash(schema);
+    auto src_hash_r = compute_source_hash(src);
+    REQUIRE(src_hash_r.has_value());
+
+    eastl::vector<Schema> schemas;
+    schemas.push_back(schema);
+    eastl::vector<Blake3Digest> digests;
+    digests.push_back(*src_hash_r);
+
+    auto r1 = compute_collection_abi_hash(schemas, digests);
+    auto r2 = compute_collection_abi_hash(schemas, digests);
     REQUIRE(r1.has_value());
     REQUIRE(r2.has_value());
     CHECK(*r1 == *r2);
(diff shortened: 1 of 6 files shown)


tools/foryc/src/abi_hash.cpp
```diff
@@ -0,0 +1,250 @@
...
+// This is a plain inline helper; it exists to avoid repeating the
+// hasher init/update/finalize pattern.
+// -----------------------------------------------------------------------
+[[nodiscard]] Blake3Digest blake3_of(const void* data, std::size_t len) noexcept {
+    blake3_hasher hasher;
+    blake3_hasher_init(&hasher);
+    blake3_hasher_update(&hasher, data, len);
+    Blake3Digest digest{};
+    blake3_hasher_finalize(&hasher, digest.data(), digest.size());
+    return digest;
```
cjhowe-us, line 137: severity:HIGH location:tools/foryc/src/abi_hash.cpp:147 problem:`compute_abi_hash` walks field-level IR (fqn:version_decimal:tag:type:name) but the locked spec in plugin-abi.md §"ABI Hash Function" point 1 + issue #222 §Scope point 2 mandates the canonical input as `(fqn || ":" || version_le || ":" || schema_source_blake3)` per schema, sorted by fqn, separated by `\n`, no trailing newline. Two divergences: (a) `version_le` means u32 little-endian BYTES, not `std::to_string(td.version)` decimal ASCII; (b) the per-schema contribution is the source-hash digest (32 raw bytes), not a re-walked field list. The loader gate (#230) will compare hex strings byte-for-byte against this exact formula — mismatch is guaranteed. fix:Either rename this to `compute_schema_abi_hash` (a useful per-type hash for ComponentDecl.schema_hash in #225) and add a separate `compute_collection_abi_hash(const eastl::vector<Schema>&)` that follows the spec verbatim, OR rewrite this function to match the spec and remove the field-walk. Document the choice in the doc-comment.

tools/foryc/src/abi_hash.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    return out;
+}
+
+// -----------------------------------------------------------------------
+// blake3_of
+//
+// Hash `data` with blake3 and return the 32-byte digest.
+// This is a plain inline helper; it exists to avoid repeating the
+// hasher init/update/finalize pattern.
```
cjhowe-us: severity:HIGH location:tools/foryc/src/abi_hash.cpp:139 problem:`compute_source_hash` hashes `canonicalize_source(raw)` (whitespace-collapsed, comments stripped). Issue #222 §Scope point 1 explicitly mandates "blake3 over the schema source (raw bytes of the `.fory` file)" and plugin-abi.md §"ABI Hash Function" point 1 references `schema_source_blake3` as a property of the source. Whitespace-invariance is a desirable engineering property but it is not the locked contract — and the loader gate (#230) will compute the hash over raw plugin-side bytes; if the host canonicalizes and the plugin does not, every load will mismatch. fix:Either (a) hash the raw `raw_source` bytes directly (`return blake3_of(raw_source.data(), raw_source.size());`) and drop the canonicalize_source helper, OR (b) update issue #222 §Scope and plugin-abi.md §"ABI Hash Function" to permit canonical-token hashing AND ensure #230/#225 use the same canonicalization. Decide and document; do not silently diverge.

tools/foryc/src/abi_hash.hpp
```diff
@@ -0,0 +1,113 @@
...
+// -----------------------------------------------------------------------
+[[nodiscard]] Result<Blake3Digest> compute_abi_hash(const Schema& schema) noexcept;
+
+// -----------------------------------------------------------------------
+// format_as_uint64_hex
+//
+// Interprets the first 8 bytes of `digest` as a big-endian uint64_t and
+// returns the 16-character lowercase hex representation, suitable for
+// embedding as a C++ uint64_t literal:
+//   inline constexpr uint64_t kAbiHash_Foo = 0x<result>;
```
cjhowe-us, line 163: severity:HIGH location:tools/foryc/src/abi_hash.hpp:109 problem:`format_as_uint64_hex` returns a 16-char hex string (first 8 bytes of digest). Plugin-abi.md §"ABI Hash Function" point 2 + issue #222 §Scope point 3 lock the wire format as "hex-encoded lowercase 64 chars" of the full 32-byte digest. The loader gate compares full hex strings (`Hash equality is byte-string equality of the hex form`). Truncating to 16 chars throws away 192 bits of entropy and is incompatible with the manifest `abi_hash : string` field shape. The uint64 truncation is fine as a separate convenience for embedding `kAbiHash_<Type>` constants in generated headers (#220), but the canonical engine-wide hash MUST be 64 chars. fix:Add a sibling helper `format_as_full_hex(const Blake3Digest&) -> eastl::string` returning the 64-char lowercase representation. Document `format_as_uint64_hex` as "for per-type uint64_t embedding only; not the wire-form ABI hash". Round 1 will not block on this if the deferral to #220 is explicit, but the 64-char emission must land before #230 ships.

tools/foryc/src/main.cpp
```diff
@@ -108,17 +146,84 @@ static std::string_view error_name(const glibre::Error& e) noexcept {
...
+        }
+        const eastl::string abi_hex = format_as_uint64_hex(*abi_hash_r);
+
+        // Print tab-separated: path, source_hash, abi_hash.
+        std::cout << path.native() << "\t"
+                  << std::string_view(src_hex.data(), src_hex.size()) << "\t"
+                  << std::string_view(abi_hex.data(), abi_hex.size()) << "\n";
+    }
+
+    if (schema_files.empty()) {
```
cjhowe-us: severity:MED location:tools/foryc/src/main.cpp:215 problem:`run_abi_hash` prints `<schema_path>\t<source_hash>\t<abi_hash>` per .fory file by calling `compute_abi_hash(*parse_r)` where `parse_r` is one file’s `Schema`. The shipping engine-wide `glibre_types_abi_hash` per fory-codegen.md and plugin-abi.md is computed over ALL schema files concatenated globally (one hash for the whole `data/schemas` tree). The current emit-mode produces N per-file hashes, which is useful for #225 ComponentDecl.schema_hash but is NOT the value `_abi_hash.cpp` will embed. The build-system stamp consumer needs the global hash. fix:Add a second emit line after the per-file lines summarizing the collection ABI hash: `<COLLECTION>\t<global_abi_hash>`, OR add a separate `--emit=collection-abi-hash` mode. Either way, the doc-comment at lines 17-22 should clarify that per-file hashes are component-level and the collection hash is computed separately (when `compute_collection_abi_hash` lands).

tools/foryc/src/abi_hash.cpp
```diff
@@ -0,0 +1,250 @@
...
+        canonical.pop_back();
+
+    return blake3_of(canonical);
+}
+
+// -----------------------------------------------------------------------
+// format_as_uint64_hex
+// -----------------------------------------------------------------------
+eastl::string format_as_uint64_hex(const Blake3Digest& digest) noexcept {
+    // Interpret first 8 bytes as big-endian uint64_t.
```
cjhowe-us, line 276: severity:MED location:tools/foryc/src/abi_hash.cpp:240 problem:`format_as_uint64_hex` reads the first 8 bytes of the digest as big-endian (`val = (val << 8) | byte`). The doc-comment in abi_hash.hpp:107 says the result is "suitable for embedding as a `uint64_t kAbiHash_Foo = 0x<result>;`" — that is true for any single endianness convention, but the convention is undocumented and the unit test at foryc_abi_hash_test.cpp:286-298 hard-codes BE behavior (byte[0]=0x01 → 0x0100000000000000) without justification or cross-reference to the spec. The loader-side gate (#230) will need to know which endianness the host chose. fix:In abi_hash.hpp, document explicitly: "First 8 bytes of digest interpreted as big-endian uint64; rationale: matches `std::format("{:016x}", val)` rendering of the digest prefix." Or pick little-endian if the architecture is LE-canonical. Either is fine; absence of documentation is the bug.

tools/foryc/src/abi_hash.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+        eastl::sort(
+            sorted_fields.begin(), sorted_fields.end(),
+            [](const FieldDecl* a, const FieldDecl* b) { return a->tag < b->tag; }
+        );
+
+        for (const FieldDecl* fd_ptr : sorted_fields) {
+            const FieldDecl& fd = *fd_ptr;
+            canonical += "  ";
+            canonical += std::to_string(fd.tag);
```
cjhowe-us: severity:LOW location:tools/foryc/src/abi_hash.cpp:216 problem:Trailing-newline handling is uncomfortable: the loop appends `\n` after every field line including the last, then unconditionally pops the trailing newline at the end. Comment at 209-214 admits the spec compliance is approximate. If two TypeDecls each have one field, the canonical string ends with the last field line followed by no newline — but if there are zero TypeDecls, the canonical is empty (no pop). Inconsistent with itself across edge cases. fix:Refactor to use a `first_field` flag mirroring `first_type`, emitting `\n` BEFORE each field line except the first within a type and BEFORE each type header except the first. The pop_back is then unnecessary. Tests should add a zero-types case and a one-type-zero-fields case to lock the behavior.

tools/foryc/src/abi_hash.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include <cstdint>
+#include <cstring>
+#include <format>
+#include <string>
+#include <string_view>
+
+// BLAKE3 C API (vcpkg "blake3" port, target blake3::blake3).
+#include <blake3.h>
+
+#include <EASTL/algorithm.h>
```
cjhowe-us, line 21: severity:LOW location:tools/foryc/src/abi_hash.cpp:22 problem:Implementation file uses `std::string` for canonical accumulator and `std::to_string` for integer formatting. PHILOSOPHY §11 retains `std::` for utilities EASTL does not own; `std::string` is owned by EASTL (`eastl::string`) and the IR types in parser.hpp use `eastl::string` consistently. Cohesion is weakened by mixing the two. `std::format("{:016x}", val)` then converted to `eastl::string` at the end is a double allocation. fix:Use `eastl::string` for `canonical` and `out` accumulators; replace `std::to_string` with a small inline `eastl::string append_decimal(uint32_t)` helper or with `eastl::string::sprintf` (EASTL provides this). The `std::format` call in `format_as_uint64_hex` can stay since `std::format` is on the §11 retain list, but the eastl::string conversion at the end is fine.

cjhowe-us (commented):
findings:
  HIGH:3
  MED:2
  LOW:2
carryover:0
notes:Round 1 (coverage + correctness) lens. The PR delivers a plausible foryc abi_hash module with 11 Catch2 tests whose names match the issue #222 DoD verbatim, blake3 vcpkg dep wired, --emit=abi-hash CLI flag, BLAKE3::blake3 link, EASTL Blake3Digest, Closes #222 + Conventional Commit subject all present. What blocks: three HIGH-severity spec drifts vs reviews/decisions/plugin-abi.md §"ABI Hash Function" + issue #222 §Scope. (1) compute_abi_hash hashes a field-walk (fqn:version_decimal:tag:type:name) instead of the locked spec form (fqn || ":" || version_le || ":" || schema_source_blake3) — the loader gate (#230) will not match. (2) compute_source_hash canonicalizes whitespace+comments before hashing; issue #222 §Scope point 1 says "raw bytes of the .fory file" — same loader-mismatch risk. (3) format_as_uint64_hex returns 16-char hex (truncated to 8 bytes); the manifest abi_hash field and plugin-abi.md require 64-char full-digest hex. The uint64 form is fine as a per-type embedding helper but a separate format_as_full_hex must exist before #230 can ship. MED findings: (a) main.cpp emits per-file ABI hashes, but the engine-wide hash is a single hash over all schemas globally — needs a collection-level emit; (b) endianness convention in format_as_uint64_hex is undocumented. LOW: trailing-newline edge cases in the canonical-string builder; std::string mixing inside the impl violates PHILOSOPHY §11 substrate cohesion. What's deferred: emit_header.cpp embedding of kAbiHash constants is explicitly out-of-scope per PR body — that deferral is acceptable. Recommend the impl-respond agent (a) reconcile compute_abi_hash + compute_source_hash with the spec, OR (b) update plugin-abi.md and #222 §Scope to permit the canonical/field-walk variants AND ensure #230/#225 use the same recipe — and add the 64-char full-hex helper. Round 2 (cohesion+SRP) and round 3 (topology) will not unblock this until the spec drift is resolved.

tools/foryc/src/abi_hash.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: Rewrote compute_abi_hash (renamed to compute_collection_abi_hash) to follow the locked recipe from plugin-abi.md §"ABI Hash Function" point 1 exactly: for each TypeDecl fqn-sorted, feed fqn || ":" || version_le_bytes(4) || ":" || source_digest_bytes(32) into a single blake3 hasher, entries separated by newline, no trailing newline. version_le_bytes(4) uses encode_le32 (u32 little-endian). source_digest is the 32 raw bytes from compute_source_hash (per HIGH-2 fix). The old field-walk path is removed entirely.

tools/foryc/src/abi_hash.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: compute_source_hash now hashes raw bytes of raw_source directly (blake3_of(raw_source.data(), raw_source.size())) with no canonicalization. The whitespace-stripping canonicalize_source helper is preserved but called only by the new compute_canonical_source_hash, which is documented as diagnostic-only and explicitly excluded from the wire-format ABI hash path. Tests updated: foryc_source_hash_unchanged_on_whitespace renamed to foryc_source_hash_changes_on_whitespace and expectation flipped (raw bytes change on whitespace -> hashes differ).

tools/foryc/src/abi_hash.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: Added format_as_full_hex(const Blake3Digest&) -> eastl::string returning the 64-char lowercase hex of the complete 32-byte digest. This is the wire-format function per plugin-abi.md §"ABI Hash Function" point 2 and issue #222 §Scope point 3. Documented format_as_uint64_hex in the header as "per-type uint64_t embedding ONLY; NOT the wire-form ABI hash". --emit=abi-hash and --emit=collection-abi-hash modes both use format_as_full_hex for output. Test foryc_format_as_full_hex_is_64_chars added; also verifies that the first 16 chars of format_as_full_hex match format_as_uint64_hex.

tools/foryc/src/main.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: Added --emit=collection-abi-hash mode (run_collection_abi_hash in main.cpp) that aggregates all schema files into a single engine-wide hash using compute_collection_abi_hash, then prints a single line: <COLLECTION>\t<64-char-hex>. --emit=abi-hash mode updated to emit per-file source hashes only (64-char, for ComponentDecl.schema_hash use in #225). Both modes are documented in the main.cpp header comment block and in usage().

tools/foryc/src/abi_hash.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: Documented endianness convention explicitly in abi_hash.hpp header comment and in the format_as_uint64_hex doc-comment: "First 8 bytes of digest interpreted as big-endian uint64; rationale: big-endian is chosen so the result is a prefix of format_as_full_hex (first 16 chars match), enabling visual cross-checking." The test foryc_format_as_full_hex_is_64_chars asserts this property. The collection ABI hash recipe separately documents version as little-endian u32 (encode_le32) per the spec.

tools/foryc/src/abi_hash.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: compute_collection_abi_hash uses a first_entry flag (no pop_back) — feeds the newline separator BEFORE each entry except the first. Zero-entry case (empty collection) emits no bytes at all. Added foryc_collection_abi_hash_zero_schemas and foryc_collection_abi_hash_one_type_zero_fields tests to lock edge cases. The old pop_back idiom with the ambiguous comment is gone.

tools/foryc/src/abi_hash.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8c61859
reasoning: Swept abi_hash.cpp for std::string accumulators: canonicalize_source now uses eastl::string for out; blake3_of overload changed from const std::string& to const eastl::string&. std::string is retained only in main.cpp (std::ostringstream::str() for file I/O — PHILOSOPHY §11 retain list for std::ifstream/ostringstream). std::format in format_as_uint64_hex stays (std::format is on the §11 retain list) but the result is immediately converted to eastl::string on return.

tests/tools/foryc/foryc_abi_hash_test.cpp
```diff
@@ -0,0 +1,398 @@
...
+}
+
+TEST_CASE("foryc_source_hash_changes_on_field_rename", "[foryc][source_hash]") {
+    // Renaming a field changes the raw bytes and therefore the source hash.
+    constexpr std::string_view src_before = R"(
+schema Foo {
+  version 1
+  field old_name : u32 tag 1
+}
+)";
```
cjhowe-us, line 292: severity:HIGH location:tests/tools/foryc/foryc_abi_hash_test.cpp:307 problem:Issue #222 §Definition of Done lists `unit_test_named: foryc_source_hash_unchanged_on_whitespace`, but the test was correctly renamed to `foryc_source_hash_changes_on_whitespace` (since the locked spec hashes RAW bytes, whitespace DOES change the source hash). The DoD assertion will fail in the dod-verify workflow on merge — `unit_test_named: foryc_source_hash_unchanged_on_whitespace` no longer matches any test name in the file. fix:Update issue #222 DoD block: replace `unit_test_named: foryc_source_hash_unchanged_on_whitespace` with `unit_test_named: foryc_source_hash_changes_on_whitespace`. Also add `unit_test_named: foryc_collection_abi_hash_zero_schemas`, `unit_test_named: foryc_collection_abi_hash_one_type_zero_fields`, `unit_test_named: foryc_format_as_uint64_hex_is_16_chars`, `unit_test_named: foryc_format_as_full_hex_is_64_chars` to lock the new edge-case + format helper tests added in r1 (commit 8c61859). The DoD must reflect the spec-correct behavior or the closure gate cannot fire.

tests/tools/foryc/foryc_abi_hash_test.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+TEST_CASE("foryc_abi_hash_changes_on_version_bump", "[foryc][abi_hash]") {
+    // Bumping the schema version must change the collection ABI hash.
+    // version is included as version_le(4) in the recipe.
+    constexpr std::string_view src_v1 = R"(
+schema glibre.test.Versioned {
+  version 1
+  field x : u32 tag 1
+}
+)";
```
cjhowe-us, line 169: severity:HIGH location:tests/tools/foryc/foryc_abi_hash_test.cpp:175 problem:Test name `foryc_abi_hash_unchanged_on_whitespace` directly contradicts what the test body asserts. The test now correctly asserts `abi_hash_of(compact) != abi_hash_of(spaced)` (lines 197-198) — i.e. the abi hash CHANGES on whitespace, because the recipe folds in the raw-byte source digest. But the test name still says "unchanged". Future readers (and the dod-verify gate, which only sees the name) will be misled. The DoD on issue #222 also still names this `unit_test_named: foryc_abi_hash_unchanged_on_whitespace`. fix:Rename the test to `foryc_abi_hash_changes_on_whitespace` and update the matching DoD entry on issue #222 to the same name. The comment block at lines 178-184 already documents the correct semantics — the name should match.

tools/foryc/src/abi_hash.hpp
```diff
@@ -0,0 +1,161 @@
...
+    const eastl::vector<Schema>& schemas,
+    const eastl::vector<Blake3Digest>& source_digests
+) noexcept;
+
+// -----------------------------------------------------------------------
+// format_as_full_hex
+//
+// Returns the full 32-byte digest as a 64-character lowercase hex string.
+// This is the wire format required by:
+//   - plugin-abi.md §"ABI Hash Function" point 2 (64-char lowercase hex)
```
cjhowe-us, line 148: severity:MED location:tools/foryc/src/abi_hash.hpp:139 problem:`compute_collection_abi_hash` takes two parallel `eastl::vector`s (schemas + source_digests) that the contract REQUIRES to be the same size and same order, but there is no precondition check. The body unconditionally indexes `source_digests[si]` for each schema. If a future caller passes mismatched vectors (or `parse_file` succeeds for one schema but `compute_source_hash` was skipped), this is OOB read / UB. The doc-comment line 134 says "must be parallel to schemas" but enforcement is by convention only. Cohesion-wise, the two-parallel-vector API is also fragile at a seam (SRP §"two reasons to change" — a future schema list change mutates both vectors in lockstep). fix:Either (a) collapse to a single `eastl::vector<SchemaWithDigest>` parameter where `SchemaWithDigest { Schema schema; Blake3Digest source_digest; }` makes the pairing structural (preferred — one cohesive seam, impossible to mis-pair); or (b) add a precondition check at function entry: `if (schemas.size() != source_digests.size()) return std::unexpected(Error{...ForycInternal...});`. Option (a) is the SRP-aligned fix and removes any "must be parallel" doc-warning.

tools/foryc/src/main.cpp
```diff
@@ -108,17 +160,147 @@ static std::string_view error_name(const glibre::Error& e) noexcept {
...
+            );
+            return EXIT_FAILURE;
+        }
+        // 64-char wire-format hex (plugin-abi.md §2).
+        const eastl::string src_hex = format_as_full_hex(*src_hash_r);
+
+        // Print tab-separated: path, source_hash_64.
+        std::cout << path.native() << "\t"
+                  << std::string_view(src_hex.data(), src_hex.size()) << "\n";
+    }
```
cjhowe-us, line 268: severity:MED location:tools/foryc/src/main.cpp:223 problem:`run_abi_hash` (--emit=abi-hash mode) emits one `<schema_path>\t<source_hash>` row per .fory FILE, but plugin-abi.md §"Plugin Manifest Schema" defines `ComponentDecl.schema_hash` as per-component-fqn (one hash per type, not one per file). A single .fory file may declare multiple `schema <fqn> { ... }` blocks (the parser`s `Schema` struct holds `eastl::vector<TypeDecl>`). If `data/schemas/render.fory` declares `glibre.render.Mesh` and `glibre.render.Material`, this mode emits a single row for the file — but #225 needs two rows, one per fqn. The PR body claims "Useful for populating ComponentDecl.schema_hash in the manifest (#225)" but the seam between this output and #225 will require either (a) #225 re-parses the file to expand fqns, duplicating parse work, or (b) one-fory-file-equals-one-type as an undocumented convention. fix:Either widen run_abi_hash output to `<schema_path>\t<fqn>\t<source_hash>` per TypeDecl (parse the file, iterate types — minimal extra work since collect_schema_files already knows file paths), OR document the one-file-one-type convention explicitly in main.cpp comments + plugin-abi.md and add a parser-level rejection of multi-type files. The first option is the SRP-aligned choice — abi-hash emit and #225 manifest emit then share a per-fqn record shape.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:0
carryover:0
notes:Round 2 (cohesion + SRP + seam quality) lens. R1 spec-recipe drift (HIGH-1/2/3) and the MED/LOW items were all addressed in commit 8c61859 — compute_collection_abi_hash now matches plugin-abi.md §"ABI Hash Function" point 1 byte-for-byte (fqn || ":" || version_le(4) || ":" || src_digest(32), \n separated, fqn-sorted, no trailing newline; verified by reading abi_hash.cpp:198-243 against the spec at plugin-abi.md:74-95). compute_source_hash hashes raw bytes. format_as_full_hex returns 64-char lowercase hex with the documented prefix-match property to format_as_uint64_hex. --emit=collection-abi-hash + --emit=abi-hash both emit the wire format. Endianness is documented in the hpp header. EASTL substrate is clean in abi_hash.cpp; std::string is correctly retained only in main.cpp for std::ostringstream::str() per PHILOSOPHY §11. Cohesion good: abi_hash.cpp is pure, main.cpp owns I/O. What blocks (round 2 surface): two HIGH-severity coherence problems at the test/DoD seam. (HIGH-A) Issue #222 §Definition of Done lists `unit_test_named: foryc_source_hash_unchanged_on_whitespace`, but the spec-correct test is named `foryc_source_hash_changes_on_whitespace` — dod-verify will fail closure on merge. The DoD must be updated to match the spec-correct test names, plus add the four new tests added in r1 (zero_schemas, one_type_zero_fields, format_as_uint64_hex_is_16_chars, format_as_full_hex_is_64_chars). (HIGH-B) Test name `foryc_abi_hash_unchanged_on_whitespace` directly contradicts what its body asserts — the test correctly verifies CHANGED-on-whitespace per the locked recipe, but the name still says "unchanged". Future readers + dod-verify (which only sees the name) will be misled. Rename to `foryc_abi_hash_changes_on_whitespace` and update DoD. MED findings: (a) compute_collection_abi_hash takes two parallel eastl::vectors with no precondition check — recommend collapsing to a single eastl::vector<SchemaWithDigest> for SRP and to make mis-pairing structurally impossible; (b) --emit=abi-hash emits one row per .fory FILE, but ComponentDecl.schema_hash is per-fqn — recommend `<path>\t<fqn>\t<hash>` per TypeDecl so #225 doesn't have to re-parse. What's deferred: nothing new. Recommend impl-respond agent (a) update issue #222 DoD block, (b) rename the abi_hash_unchanged_on_whitespace test, (c) collapse the parallel-vector API, (d) widen run_abi_hash output to per-fqn rows. Round 3 (topology + spec alignment) pending.

tests/tools/foryc/foryc_abi_hash_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:5e14859
reasoning: Updated issue #222 DoD block via `gh issue edit 222 --body-file`. Replaced `foryc_source_hash_unchanged_on_whitespace` with `foryc_source_hash_changes_on_whitespace`, replaced `foryc_abi_hash_unchanged_on_whitespace` with `foryc_abi_hash_changes_on_whitespace`, and added four new unit_test_named entries: `foryc_collection_abi_hash_zero_schemas`, `foryc_collection_abi_hash_one_type_zero_fields`, `foryc_format_as_uint64_hex_is_16_chars`, `foryc_format_as_full_hex_is_64_chars`, plus `foryc_abi_hash_changes_on_version_bump` which was missing from DoD despite existing in the test file since r1. DoD now lists all 15 test cases present in foryc_abi_hash_test.cpp and dod-verify will match them all at HEAD.

tests/tools/foryc/foryc_abi_hash_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:5e14859
reasoning: Renamed TEST_CASE("foryc_abi_hash_unchanged_on_whitespace") to TEST_CASE("foryc_abi_hash_changes_on_whitespace") in tests/tools/foryc/foryc_abi_hash_test.cpp. Name now matches the assertion body (CHECK(abi_hash_of(compact) != abi_hash_of(spaced))) and the existing comment block at lines 178-184 which already documented the correct semantics. DoD entry updated simultaneously (see HIGH-1 reply).

tools/foryc/src/abi_hash.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:5e14859
reasoning: Defined struct SchemaWithDigest { Schema schema; Blake3Digest source_digest; } in abi_hash.hpp immediately before compute_collection_abi_hash. Changed function signature to take const eastl::vector<SchemaWithDigest>& entries — the schema-digest pairing is now structural (impossible to mis-pair or mis-size). Removed the doc-comment caveat "must be parallel to schemas". Updated all callers: abi_hash_of() helper in test file, foryc_abi_hash_is_deterministic, foryc_collection_abi_hash_zero_schemas, abi_hash_independent_of_schema_file_order_on_disk, and run_collection_abi_hash in main.cpp.

tools/foryc/src/main.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:5e14859
reasoning: Widened run_abi_hash (--emit=abi-hash) to emit one row per TypeDecl instead of one row per file. New row format: <schema_path>\t<fqn>\t<source_hash_hex_64>\t<abi_hash_hex_64>. For each file, the function now parses the schema IR, iterates over schema.types, and emits a per-TypeDecl row with the per-file source hash and a single-TypeDecl collection ABI hash. Multiple schema <fqn> blocks in one .fory file produce multiple rows. Updated the usage comment block at the top of main.cpp to document the new row shape and note it is designed for #225 ComponentDecl.schema_hash without requiring re-parsing.

cjhowe-us (commented):
round:3 reviewer:go-review verdict:approve
findings:
  HIGH:0
  MED:0
  LOW:0
carryover:0
notes:Round 3 (topology + spec alignment + polish) converges. (1) Decision-doc alignment is byte-perfect against plugin-abi.md §"ABI Hash Function" pts 1/2/4 — recipe (fqn || ":" || version_le_bytes(4) || ":" || source_digest_bytes(32)), "\n" separator, no trailing newline, fqn-sorted in Unicode code-point order, 64-char lowercase hex wire form, inputs limited to source bytes + version (no compiler/timestamp/flags). (2) DoD ↔ TEST_CASE diff is 15-for-15 exact. (3) Sub-issue topology: --emit=abi-hash row format <path>\t<fqn>\t<src_hash_64>\t<abi_hash_64> directly feeds #225 ComponentDecl.schema_hash without re-parsing; --emit=collection-abi-hash supplies the value embedded by glibre-types for #230. (4) Conv-Commit subject + `Closes #222` present. (5) CI status pending with zero registered contexts — consistent with infra posture, not a round-3 blocker. (6) No stale TODOs/FIXMEs across all four changed source files. (7) All R1+R2 findings (HIGH-1/2/3, MED-3/4/5, LOW-6/7, HIGH-2 rename) were addressed in commits 8c61859 + 5e14859; zero carryover. EASTL discipline holds (PHILOSOPHY §11) with std:: retained only at the file-I/O seam in main.cpp. Endianness convention (LE u32 for version, BE u64 for the per-type embedding helper) documented in the header. SchemaWithDigest collapses the parallel-vector pairing into a structurally safe API. Recommend merge.

cjhowe-us:
**HOLD — pending Fory→Flatbuffers ADR.** User has requested a full replacement of Apache Fory with Flatbuffers across the engine. `go-thinker` is producing the tradeoff analysis; `go-design` will then author the superseding ADR. This PR will not merge until the ADR decision lands; if the decision is to switch, this branch will be closed. Disabling auto-merge as a precaution.

cjhowe-us:
superseded by Rust pivot 2026-05-13

PR #962, opened 2026-05-09
feat(core): PluginLoader — ABI hash + version + name + deps gates (refs #230)

### LINKED ISSUE #230: [PLAN] feat(core): PluginLoader — ABI hash + version + name + deps gates

cjhowe-us:
## Summary

Extends PluginLoader (#229) with four compatibility gates per
plugin-abi.md §"Loader Sequence" steps 4–7:

1. **ABI hash gate** (step 4) — `manifest.abi_hash` AND the exported
   `glibre_plugin_abi_hash` symbol must both equal the host's expected
   ABI hash. Mismatch → `core::Error::PluginAbiHashMismatch`.

2. **Engine version gate** (step 5) — `manifest.min_engine_version ≤`
   host engine version. Violation → `core::Error::PluginEngineTooOld`.

3. **Name uniqueness gate** (step 6) — `manifest.name` must not be
   already registered. Duplicate → `core::Error::PluginNameCollision`.

4. **Dependency gate** (step 7) — every entry in `manifest.depends_on`
   must already be registered. Missing → `core::Error::PluginDependencyMissing`.

`register + schedule rebuild + migrate` (#231) come next.

## Files changed

- `core/include/glibre/core/plugin_loader_registry.hpp` (new)
- `core/src/plugin_loader_registry.cpp` (new)
- `core/include/glibre/error.hpp` — 3 new arms: `PluginEngineTooOld`,
  `PluginNameCollision`, `PluginDependencyMissing`
- `core/CMakeLists.txt` — wires new source file
- `tests/core/plugin_loader_registry/CMakeLists.txt` (new)
- `tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp`
  (new — 5 named DoD tests + 4 additional coverage tests)

## Refs

Closes #230

## Test plan

- [x] All 5 DoD-required tests pass:
  - `plugin_loader_rejects_abi_hash_mismatch`
  - `plugin_loader_rejects_unknown_dependency`
  - `plugin_loader_accepts_compatible_plugin`
  - `plugin_loader_rejects_duplicate_name`
  - `plugin_loader_rejects_incompatible_version`
- [x] 40 assertions in 9 test cases — `glibre-core-plugin-loader-registry-tests` — all pass
- [x] `glibre-core-plugin-loader-tests` (plan #229) — all pass (28 assertions in 8 test cases)
- [x] No new warnings introduced by this PR

### DIFF
--- a/core/include/glibre/core/plugin_loader_registry.hpp
+++ b/core/include/glibre/core/plugin_loader_registry.hpp
@@ -10,15 +10,19 @@
 // each TEST_CASE constructs its own independent registry.
 //
 // Gates enforced (in order):
-//   1. ABI hash gate (step 4)
-//      manifest.abi_hash AND the exported glibre_plugin_abi_hash symbol
-//      must both equal the host's expected ABI hash.  Mismatch →
-//      core::Error::PluginAbiHashMismatch.
+//   1. ABI hash gate (step 4) — two sub-checks:
+//      a) validate_manifest_abi_hash: manifest.abi_hash == expected_abi_hash.
+//      b) validate_symbol_abi_hash:   symbol value == expected_abi_hash.
+//      Both must pass; either mismatch → core::Error::PluginAbiHashMismatch.
+//      validate_all runs (a) then (b); individual granular callers may call
+//      each sub-check directly.
 //   2. Engine version gate (step 5)
 //      manifest.min_engine_version ≤ host engine version.  Violation →
 //      core::Error::PluginEngineTooOld.
 //   3. Name uniqueness gate (step 6)
-//      manifest.name must not already be registered.  Duplicate →
+//      manifest.name must not already be registered WITH A DIFFERENT file
+//      path.  Same name + same path is idempotent (allows hot-reload
+//      re-registration of the same dylib).  Different path →
 //      core::Error::PluginNameCollision.
 //   4. Dependency gate (step 7)
 //      every entry in manifest.depends_on must already be registered.
@@ -30,14 +34,13 @@
 //   - migrate / hot-reload
 //
 // PHILOSOPHY §11: EASTL replaces std:: for runtime data structures.
-// std:: is retained for std::expected (Result alias), std::string_view
-// (passed through to EASTL), std::filesystem.
+// std:: is retained for std::expected (Result alias) and std::filesystem.
 //
 // PHILOSOPHY §9: "Plugin ABI gated by middleman dylib hash."
 
 #include <cstdint>
-#include <string_view>
 
+#include <EASTL/functional.h>
 #include <EASTL/string.h>
 #include <EASTL/string_view.h>
 #include <EASTL/vector.h>
@@ -83,26 +86,35 @@ public:
     explicit PluginLoaderRegistry(SemVer host_engine_version) noexcept;
 
     // -----------------------------------------------------------------------
-    // validate_abi_hash — run gate 1 (ABI hash) only.
+    // validate_manifest_abi_hash — gate 1a (manifest field only).
     //
-    // Compares manifest.abi_hash against expected_abi_hash.  The caller
-    // supplies the expected_abi_hash that was read from the host's compiled-in
-    // `glibre_types_abi_hash()` AND from the plugin's exported
-    // `glibre_plugin_abi_hash` symbol.  Both values must have been checked
-    // externally before calling this helper (the loader calls it twice — once
-    // against the symbol, once against the manifest field — and this method
-    // checks the manifest field only; the caller checks the symbol equality
-    // separately).
-    //
-    // Per plugin-abi.md §"Loader Sequence" step 4:
-    //   "Both must agree, both must equal the host's glibre_types_abi_hash()."
+    // Compares manifest.abi_hash against expected_abi_hash.
+    // Per plugin-abi.md §"Loader Sequence" step 4, the manifest field and
+    // the exported symbol must both agree with the host's hash.  This
+    // method checks the manifest field; validate_symbol_abi_hash checks
+    // the compiled-in symbol value.
     //
     // On mismatch returns std::unexpected(core::Error::PluginAbiHashMismatch).
     // -----------------------------------------------------------------------
 
-    [[nodiscard]] Result<void> validate_abi_hash(
+    [[nodiscard]] Result<void> validate_manifest_abi_hash(
         const PluginManifest& manifest,
         eastl::string_view    expected_abi_hash) const noexcept;
+
+    // -----------------------------------------------------------------------
+    // validate_symbol_abi_hash — gate 1b (exported symbol value only).
+    //
+    // Compares the value the loader read from the plugin's exported
+    // `glibre_plugin_abi_hash` C symbol against expected_abi_hash.
+    // The redundant check (step 4) catches a malformed manifest whose
+    // abi_hash field disagrees with the compiled-in symbol.
+    //
+    // On mismatch returns std::unexpected(core::Error::PluginAbiHashMismatch).
+    // -----------------------------------------------------------------------
+
+    [[nodiscard]] Result<void> validate_symbol_abi_hash(
+        eastl::string_view symbol_abi_hash,
+        eastl::string_view expected_abi_hash) const noexcept;
 
     // -----------------------------------------------------------------------
     // validate_engine_version — run gate 2 (engine version).
@@ -117,12 +129,21 @@ public:
     // -----------------------------------------------------------------------
     // validate_name_unique — run gate 3 (name uniqueness).
     //
-    // Checks manifest.name is not already registered.
-    // On duplicate returns core::Error::PluginNameCollision.
+    // Per plugin-abi.md §"Loader Sequence" step 6:
+    //   "refuse if a plugin with the same name is already registered with
+    //    a different file path."
+    //
+    // Collision is fired only when name matches AND file_path differs.
+    // Same name + same file_path is treated as an idempotent re-registration
+    // (the hot-reload path in plan #231 loads the same .dylib again after a
+    // swap; the registry must accept it without error).
+    //
+    // On collision returns core::Error::PluginNameCollision.
     // -----------------------------------------------------------------------
 
     [[nodiscard]] Result<void> validate_name_unique(
-        const PluginManifest& manifest) const noexcept;
+        const PluginManifest& manifest,
+        eastl::string_view    file_path) const noexcept;
 
     // -----------------------------------------------------------------------
     // validate_dependencies — run gate 4 (dependency resolution).
@@ -137,10 +158,14 @@ public:
     // -----------------------------------------------------------------------
     // validate_all — run all four gates in order (steps 4–7).
     //
-    // The symbol_abi_hash argument is the value read from the plugin's
-    // exported `glibre_plugin_abi_hash` C symbol (a const char* that the
-    // loader dlsym()s).  The manifest.abi_hash field is also checked; both
-    // must equal expected_abi_hash per plugin-abi.md §step 4.
+    // symbol_abi_hash: value read from the plugin's exported
+    //   `glibre_plugin_abi_hash` C symbol (a const char* that the loader
+    //   dlsym()s).  The manifest.abi_hash field is also checked; both
+    //   must equal expected_abi_hash per plugin-abi.md §step 4.
+    //
+    // file_path: the .dylib path being loaded, forwarded to
+    //   validate_name_unique for the "different file path" qualifier
+    //   (plugin-abi.md §step 6).
     //
     // Returns the first error encountered; gates are run in plugin-abi.md
     // ordering (hash → engine version → name → deps).
@@ -149,7 +174,8 @@ public:
     [[nodiscard]] Result<void> validate_all(
         const PluginManifest& manifest,
         eastl::string_view    expected_abi_hash,
-        eastl::string_view    symbol_abi_hash) const noexcept;
+        eastl::string_view    symbol_abi_hash,
+        eastl::string_view    file_path) const noexcept;
 
     // -----------------------------------------------------------------------
     // register_plugin — record a plugin as successfully loaded.
@@ -160,13 +186,17 @@ public:
     //
     // path is the filesystem path to the .dylib; may be empty in tests.
     //
-    // Precondition: validate_all was called with the same manifest and
-    // returned success.  Violating this precondition produces undefined
-    // behaviour; debug builds abort.
+    // Returns an error (without modifying the registry) if the name is
+    // already registered with a different path.  This is an unconditional
+    // precondition check — not a debug-only assert — so release builds are
+    // protected from silent overwrites.
+    //
+    // Same name + same path: idempotent no-op, returns success.
+    // Same name + different path: returns core::Error::PluginNameCollision.
     // -----------------------------------------------------------------------
 
-    void register_plugin(const PluginManifest& manifest,
-                         eastl::string_view    path) noexcept;
+    [[nodiscard]] Result<void> register_plugin(const PluginManifest& manifest,
+                                               eastl::string_view    path) noexcept;
 
     // -----------------------------------------------------------------------
     // is_registered — query whether a plugin name is already registered.
@@ -185,7 +215,13 @@ private:
 
     // key = plugin name (eastl::string), value = PluginRecord
     // eastl::hash_map per PHILOSOPHY §11 (EASTL containers).
-    eastl::hash_map<eastl::string, PluginRecord> loaded_;
+    //
+    // transparent_string_hash + equal_to<void> enable heterogeneous lookup:
+    //   loaded_.find(eastl::string_view{...})  — no eastl::string allocation
+    //   per lookup (MED-3: avoids per-call key materialisation).
+    eastl::hash_map<eastl::string, PluginRecord,
+                    eastl::transparent_string_hash,
+                    eastl::equal_to<void>> loaded_;
 };
 
 }  // namespace glibre::core
--- a/core/include/glibre/core/plugin_loader_registry.hpp
+++ b/core/include/glibre/core/plugin_loader_registry.hpp
@@ -1,0 +1,191 @@
+#pragma once
+// core/include/glibre/core/plugin_loader_registry.hpp
+//
+// PluginLoaderRegistry — tracks loaded plugins and enforces the four
+// compatibility gates described in reviews/decisions/plugin-abi.md
+// §"Loader Sequence" steps 4–7.
+//
+// This class is NOT a singleton; callers own the registry value.  In
+// production the engine owns one instance per process.  In tests,
+// each TEST_CASE constructs its own independent registry.
+//
+// Gates enforced (in order):
+//   1. ABI hash gate (step 4)
+//      manifest.abi_hash AND the exported glibre_plugin_abi_hash symbol
+//      must both equal the host's expected ABI hash.  Mismatch →
+//      core::Error::PluginAbiHashMismatch.
+//   2. Engine version gate (step 5)
+//      manifest.min_engine_version ≤ host engine version.  Violation →
+//      core::Error::PluginEngineTooOld.
+//   3. Name uniqueness gate (step 6)
+//      manifest.name must not already be registered.  Duplicate →
+//      core::Error::PluginNameCollision.
+//   4. Dependency gate (step 7)
+//      every entry in manifest.depends_on must already be registered.
+//      Missing → core::Error::PluginDependencyMissing.
+//
+// Out of scope (plan #231):
+//   - glibre_plugin_register invocation
+//   - schedule rebuild
+//   - migrate / hot-reload
+//
+// PHILOSOPHY §11: EASTL replaces std:: for runtime data structures.
+// std:: is retained for std::expected (Result alias), std::string_view
+// (passed through to EASTL), std::filesystem.
+//
+// PHILOSOPHY §9: "Plugin ABI gated by middleman dylib hash."
+
+#include <cstdint>
+#include <string_view>
+
+#include <EASTL/string.h>
+#include <EASTL/string_view.h>
+#include <EASTL/vector.h>
+#include <EASTL/hash_map.h>
+
+#include <glibre/error.hpp>
+#include <glibre/core/plugin_manifest.hpp>
+
+namespace glibre::core {
+
+// ---------------------------------------------------------------------------
+// PluginRecord — immutable descriptor stored per registered plugin.
+//
+// Stored in the registry after a plugin passes all gates.  The record is
+// used by subsequent loads for name-collision and dependency checks.
+// ---------------------------------------------------------------------------
+
+struct PluginRecord {
+    eastl::string name;      // manifest.name  — must be unique
+    SemVer        version;   // manifest.version
+    eastl::string path;      // filesystem path to the .dylib (may be empty in tests)
+};
+
+// ---------------------------------------------------------------------------
+// PluginLoaderRegistry — owns the loaded-plugin table and validates gates.
+//
+// Thread safety: None.  Mutations must happen during phase 8 (HotReload)
+// when the world is already drained (frame-phases.md §8).  No external
+// locking is provided; the engine's frame-phase barrier is the only guard.
+// ---------------------------------------------------------------------------
+
+class PluginLoaderRegistry {
+public:
+    // -----------------------------------------------------------------------
+    // PluginLoaderRegistry() — default-constructed, empty registry.
+    //
+    // The host engine version is injected at construction time so that it
+    // participates in gate checks without coupling the registry to a global
+    // constant.  Tests supply a synthetic version; production supplies the
+    // real `glibre_core_version` value.
+    // -----------------------------------------------------------------------
+
+    explicit PluginLoaderRegistry(SemVer host_engine_version) noexcept;
+
+    // -----------------------------------------------------------------------
+    // validate_abi_hash — run gate 1 (ABI hash) only.
+    //
+    // Compares manifest.abi_hash against expected_abi_hash.  The caller
+    // supplies the expected_abi_hash that was read from the host's compiled-in
+    // `glibre_types_abi_hash()` AND from the plugin's exported
+    // `glibre_plugin_abi_hash` symbol.  Both values must have been checked
+    // externally before calling this helper (the loader calls it twice — once
+    // against the symbol, once against the manifest field — and this method
+    // checks the manifest field only; the caller checks the symbol equality
+    // separately).
+    //
+    // Per plugin-abi.md §"Loader Sequence" step 4:
+    //   "Both must agree, both must equal the host's glibre_types_abi_hash()."
+    //
+    // On mismatch returns std::unexpected(core::Error::PluginAbiHashMismatch).
+    // -----------------------------------------------------------------------
+
+    [[nodiscard]] Result<void> validate_abi_hash(
+        const PluginManifest& manifest,
+        eastl::string_view    expected_abi_hash) const noexcept;
+
+    // -----------------------------------------------------------------------
+    // validate_engine_version — run gate 2 (engine version).
+    //
+    // Checks manifest.min_engine_version ≤ host engine version.
+    // On violation returns core::Error::PluginEngineTooOld.
+    // -----------------------------------------------------------------------
+
+    [[nodiscard]] Result<void> validate_engine_version(
+        const PluginManifest& manifest) const noexcept;
+
+    // -----------------------------------------------------------------------
+    // validate_name_unique — run gate 3 (name uniqueness).
+    //
+    // Checks manifest.name is not already registered.
+    // On duplicate returns core::Error::PluginNameCollision.
+    // -----------------------------------------------------------------------
+
+    [[nodiscard]] Result<void> validate_name_unique(
+        const PluginManifest& manifest) const noexcept;
+
+    // -----------------------------------------------------------------------
+    // validate_dependencies — run gate 4 (dependency resolution).
+    //
+    // Checks every entry in manifest.depends_on is already registered.
+    // On the first missing dependency returns core::Error::PluginDependencyMissing.
+    // -----------------------------------------------------------------------
+
+    [[nodiscard]] Result<void> validate_dependencies(
+        const PluginManifest& manifest) const noexcept;
+
+    // -----------------------------------------------------------------------
+    // validate_all — run all four gates in order (steps 4–7).
+    //
+    // The symbol_abi_hash argument is the value read from the plugin's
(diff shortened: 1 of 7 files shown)


core/src/plugin_loader_registry.cpp
```diff
@@ -0,0 +1,169 @@
...
+// validate_name_unique — gate 3 (plugin-abi.md step 6)
+//
+// manifest.name must not already exist in the registry.
+// ---------------------------------------------------------------------------
+
+Result<void> PluginLoaderRegistry::validate_name_unique(
+    const PluginManifest& manifest) const noexcept {
+
+    if (loaded_.find(manifest.name) != loaded_.end()) {
+        return std::unexpected(glibre::Error{core::Error::PluginNameCollision});
```
cjhowe-us: severity:HIGH location:core/src/plugin_loader_registry.cpp:71-72 problem:plugin-abi.md §step 6 and issue #230 Scope §3 specify "refuse if a plugin with the same name is already registered **with a different file path**." The current validate_name_unique refuses on ANY duplicate name regardless of path. This breaks the hot-reload re-load case (plan #231 will call the loader on the same path again to swap a rebuilt dylib) — the second load will spuriously hit PluginNameCollision instead of being treated as a re-bind. fix:Add an explicit candidate_path parameter to validate_name_unique (and propagate through validate_all). Only return PluginNameCollision when the existing record's path differs from candidate_path. PluginRecord already stores the path, so the comparison is local; same-path re-registration should either pass through (handed off to migrate in #231) or be a documented no-op.

core/include/glibre/core/plugin_loader_registry.hpp
```diff
@@ -0,0 +1,191 @@
...
+    //
+    // Per plugin-abi.md §"Loader Sequence" step 4:
+    //   "Both must agree, both must equal the host's glibre_types_abi_hash()."
+    //
+    // On mismatch returns std::unexpected(core::Error::PluginAbiHashMismatch).
+    // -----------------------------------------------------------------------
+
+    [[nodiscard]] Result<void> validate_abi_hash(
+        const PluginManifest& manifest,
+        eastl::string_view    expected_abi_hash) const noexcept;
```
cjhowe-us: severity:MED location:core/include/glibre/core/plugin_loader_registry.hpp:103-105 problem:validate_abi_hash is named for gate 1 in full but only checks the manifest-field side of the redundancy. The symbol-side check (gate 1a) lives only inside validate_all and is invisible to a caller using the granular API. A future caller (e.g. an integration test in #231) that wants to drive gates individually will silently skip the symbol-vs-expected comparison and miss the malformed-manifest case the redundancy is meant to catch. fix:Either rename to validate_manifest_abi_hash (and add a parallel validate_symbol_abi_hash, or fold both into one validate_abi_hash that takes both expected and symbol values). Whichever shape — make the granular API expose both sides of the redundant check or document loud-and-clear that only validate_all enforces gate 1 fully.

core/src/plugin_loader_registry.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    loaded_.emplace(manifest.name, eastl::move(rec));
+}
+
+// ---------------------------------------------------------------------------
+// is_registered
+// ---------------------------------------------------------------------------
+
+bool PluginLoaderRegistry::is_registered(eastl::string_view name) const noexcept {
+    return loaded_.find(eastl::string{name.data(), name.size()}) != loaded_.end();
```
cjhowe-us: severity:MED location:core/src/plugin_loader_registry.cpp:87-88,158 problem:Lookups against loaded_ (an eastl::hash_map<eastl::string, PluginRecord>) materialise a temporary eastl::string from a string_view on every call — both is_registered and validate_dependencies allocate per call. Per PHILOSOPHY §11 EASTL is here precisely to avoid those allocator hits; a phase-8 swap exercises this code N×M times (N plugins × M deps each). fix:Use eastl::hash_map::find_as with eastl::string_view as a heterogeneous key (EASTL provides a string-view hash overload), or store keys as eastl::string and use a transparent hasher. Keep the public API string_view; the internal lookup becomes allocation-free.

core/include/glibre/core/plugin_loader_registry.hpp
```diff
@@ -0,0 +1,191 @@
...
+//   - migrate / hot-reload
+//
+// PHILOSOPHY §11: EASTL replaces std:: for runtime data structures.
+// std:: is retained for std::expected (Result alias), std::string_view
+// (passed through to EASTL), std::filesystem.
+//
+// PHILOSOPHY §9: "Plugin ABI gated by middleman dylib hash."
+
+#include <cstdint>
+#include <string_view>
```
cjhowe-us: severity:MED location:core/include/glibre/core/plugin_loader_registry.hpp:39 problem:#include <string_view> pulls in std::string_view, but the header only uses eastl::string_view (line 42 already includes EASTL/string_view.h). PHILOSOPHY §11 retains std::string_view only when interop with non-EASTL APIs requires it — none in this header. The dead include relaxes the EASTL discipline that other reviewers will check against. fix:Drop the <string_view> include; rely on EASTL/string_view.h alone.

core/src/plugin_loader_registry.cpp
```diff
@@ -0,0 +1,169 @@
...
+// ---------------------------------------------------------------------------
+// register_plugin — record a successfully validated plugin.
+// ---------------------------------------------------------------------------
+
+void PluginLoaderRegistry::register_plugin(
+    const PluginManifest& manifest,
+    eastl::string_view    path) noexcept {
+
+    // Debug-mode precondition: must not already be registered.
+    assert(loaded_.find(manifest.name) == loaded_.end() &&
```
cjhowe-us: severity:MED location:core/src/plugin_loader_registry.cpp:136-151 problem:register_plugin enforces the "call validate_all first" precondition with a debug-only assert and silently returns on duplicate-name in release. Plan #231 will invoke this from the loader; if a future caller skips validate_all, a bad plugin gets recorded. error-model.md treats programmer-error as assert-class, so assert is defensible — but the function lacks a Result<void> return that #231's compensating-unregister flow needs to know whether the registration is real. fix:Either (a) keep assert but make register_plugin idempotent (no-op if already present at same path, abort on different path) so #231 cannot get a corrupt registry, or (b) promote to Result<void> and return PluginInitFailed on precondition violation. Acceptable to defer to #231 if the current contract is documented as "loader-internal, not for plan-#231-yet."

tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+// ===========================================================================
+// Test: plugin_loader_rejects_abi_hash_mismatch
+//
+// Gate 1 (plugin-abi.md §"Loader Sequence" step 4):
+//   manifest.abi_hash must equal the expected (host) ABI hash.
+//   Mismatch → core::Error::PluginAbiHashMismatch.
+//
+// Two sub-cases:
+//   a) manifest.abi_hash mismatches → rejected via validate_abi_hash.
```
cjhowe-us: severity:LOW location:tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp:84-85 problem:Comment says "Two sub-cases" but the TEST_CASE has three SECTIONs (manifest mismatch, symbol mismatch, both wrong). Doc drift. fix:Update the comment to "Three sub-cases" or list them by name.

tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp
```diff
@@ -0,0 +1,399 @@
...
+    glibre::core::SemVer    min_engine   = {0, 1, 0}) {
+
+    glibre::core::PluginManifest m;
+    m.name               = eastl::string{name};
+    m.version            = version;
+    m.abi_hash           = eastl::string{abi_hash};
+    m.min_engine_version = min_engine;
+    // depends_on is empty by default (no dependencies).
+    return m;
+}
```
cjhowe-us, line 66: severity:LOW location:tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp:53-66 problem:make_manifest defaults are good, but no helper covers the with-deps construction — every dependency-test re-pushes onto manifest.depends_on inline. Minor duplication across 4 tests. fix:Optional improvement — add a make_manifest_with_deps(name, deps...) overload to keep dependency-test setup terse. Defer if review surface allows.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:4
  LOW:2
carryover:0
notes:Coverage of the four gates is solid — all 5 DoD-named test cases are present, the validate_abi_hash symbol the DoD file_contains check expects is in core/src/plugin_loader_registry.cpp, gate ordering is enforced by validate_all and explicitly tested (validate_all_gate_ordering_hash_first), error arms (PluginEngineTooOld / PluginNameCollision / PluginDependencyMissing) match plugin-abi.md §"Failure Modes" exactly and are appended monotonically to glibre::Error per error-model.md §Composition Rules. Scope discipline is correct: no register-call to glibre_plugin_register, no schedule rebuild, no migrate — those are correctly deferred to #231. Conventional Commit subject and Closes #230 keyword are both present (dod-verify will fire on merge). One blocking finding: validate_name_unique drops the plugin-abi.md §step 6 "with a different file path" qualifier, which is load-bearing for the hot-reload re-load case in #231 — see the inline HIGH for the fix shape. The four MEDs are real but accept-with-followup material: validate_abi_hash naming/scope ambiguity, heap allocations in hash_map lookups (use find_as / heterogeneous keys), dead std::string_view include, register_plugin precondition enforcement. The two LOWs are doc/test-helper polish. Recommend addressing the HIGH plus the lookup-allocation MED before merge; the rest are fine to thread into round 2 or land separately.

core/src/plugin_loader_registry.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

validate_name_unique now accepts a file_path parameter. The internal map is keyed by name; on a hit the method checks whether the stored path differs from the incoming file_path. Same name + same path returns success (idempotent re-registration, the hot-reload re-load path). Same name + different path returns PluginNameCollision. validate_all forwards the new file_path argument; register_plugin enforces the same invariant unconditionally (no silent overwrite in release builds — see also MED-5 fix in same commit).

Two new tests added: name_unique_allows_same_name_same_path and name_unique_rejects_same_name_different_path.

All 11 test cases (60 assertions) green.

core/include/glibre/core/plugin_loader_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

validate_abi_hash split into validate_manifest_abi_hash (gate 1a, compares manifest.abi_hash against expected) and validate_symbol_abi_hash (gate 1b, compares the dlsym'd symbol value). validate_all calls both; granular callers can now invoke each independently without having to pass the manifest to check the symbol side, or vice versa.

core/src/plugin_loader_registry.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

The hash_map is now declared with eastl::transparent_string_hash and eastl::equal_to<void> (both carry is_transparent). All find() call sites in validate_name_unique, validate_dependencies, register_plugin, and is_registered now pass eastl::string_view directly — no per-call eastl::string key materialisation.

core/include/glibre/core/plugin_loader_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

Removed #include <string_view> from plugin_loader_registry.hpp. Only eastl::string_view (from EASTL/string_view.h) is used in the public API. std:: retained only in std::expected / Result alias per PHILOSOPHY §11.

core/src/plugin_loader_registry.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

register_plugin now returns Result<void>. The collision check runs unconditionally in both debug and release builds — replaces the debug-only assert. Same name + same path is a no-op (returns success without reinserting). Same name + different path returns PluginNameCollision immediately without mutating the registry. All call sites in the tests updated to REQUIRE(...has_value()).

tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

Comment updated to 'Three sub-cases:' with all three SECTIONs listed explicitly (manifest hash mismatch, symbol hash mismatch, both hashes wrong).

tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b915451

Added make_manifest_with_deps variadic template helper. Dep-list tests (plugin_loader_rejects_unknown_dependency, plugin_loader_accepts_multi_dependency_plugin, plugin_loader_rejects_partial_dependency, validate_all_gate_ordering_hash_first) all converted to use it. Eliminates the repeated push_back boilerplate.

core/src/plugin_loader_registry.cpp
```diff
@@ -0,0 +1,208 @@
...
+PluginLoaderRegistry::PluginLoaderRegistry(SemVer host_engine_version) noexcept
+    : host_engine_version_{host_engine_version} {}
+
+// ---------------------------------------------------------------------------
+// validate_manifest_abi_hash — gate 1a (plugin-abi.md step 4, manifest field)
+//
+// manifest.abi_hash must equal expected_abi_hash.
+// ---------------------------------------------------------------------------
+
+Result<void> PluginLoaderRegistry::validate_manifest_abi_hash(
```
cjhowe-us, line 30: severity:HIGH location:core/src/plugin_loader_registry.cpp:30 problem:The round-1 MED-2 split renamed the symbol from `validate_abi_hash` to `validate_manifest_abi_hash` + `validate_symbol_abi_hash`. The DoD on issue #230 still asserts `file_contains: { path: core/src/plugin_loader_registry.cpp, regex: "validate_abi_hash" }`. Per .github/DOD-DSL.md the regex is matched with `grep -E` (POSIX-extended), so it requires the literal substring `validate_abi_hash` — neither of the new names contains that sequence (the underscore between `validate_` and `abi_hash` is broken by `manifest_`/`symbol_`). I confirmed locally: `grep -E "validate_abi_hash" core/src/plugin_loader_registry.cpp` returns exit=1. The dod-verify workflow will reopen #230 the moment this PR merges. fix:Either (a) restore a thin `validate_abi_hash(manifest, symbol_hash, expected)` wrapper that delegates to the two split helpers and is called by `validate_all`, OR (b) update issue #230 DoD `file_contains` regex to `validate_(manifest|symbol)_abi_hash`. Option (a) is preferred because it keeps the DoD unchanged and gives external callers a one-call ABI-hash gate.

core/include/glibre/core/plugin_loader_registry.hpp
```diff
@@ -0,0 +1,227 @@
...
+//   - glibre_plugin_register invocation
+//   - schedule rebuild
+//   - migrate / hot-reload
+//
+// PHILOSOPHY §11: EASTL replaces std:: for runtime data structures.
+// std:: is retained for std::expected (Result alias) and std::filesystem.
+//
+// PHILOSOPHY §9: "Plugin ABI gated by middleman dylib hash."
+
+#include <cstdint>
```
cjhowe-us, line 41: severity:HIGH location:core/include/glibre/core/plugin_loader_registry.hpp:41 problem:The `lint` (clang-format) job on this PR is failing. clang-format -Werror reports ~50 violations across plugin_loader_registry.hpp, plugin_loader_registry.cpp, and plugin_loader_registry_test.cpp — column-aligned parameter declarations (`eastl::string_view    expected_abi_hash`), aligned member-init lines (`rec.name    = ...`), and the multi-line `eastl::hash_map<...>` template parameter pack. Per #230 DoD `workflow_passed: ci.yml` and the round-1 review notes ("CI status: post-r1, lint + build + tests should be green"), red lint blocks closure: build/e2e jobs are skipped behind it, so we have ZERO confirmation tests still pass on this commit. fix:Run `clang-format -i core/include/glibre/core/plugin_loader_registry.hpp core/src/plugin_loader_registry.cpp tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp` and amend. Re-run `gh pr checks 962` to confirm lint+build+tests are all green before requesting round-3 review.

core/src/plugin_loader_registry.cpp
```diff
@@ -0,0 +1,208 @@
...
+// ---------------------------------------------------------------------------
+
+Result<void> PluginLoaderRegistry::validate_all(
+    const PluginManifest& manifest,
+    eastl::string_view    expected_abi_hash,
+    eastl::string_view    symbol_abi_hash,
+    eastl::string_view    file_path) const noexcept {
+
+    // Gate 1a: symbol-side ABI hash check.
+    if (auto r = validate_symbol_abi_hash(symbol_abi_hash, expected_abi_hash); !r) {
```
cjhowe-us: severity:MED location:core/src/plugin_loader_registry.cpp:136 problem:Doc/code drift on the public-API ordering contract. plugin_loader_registry.hpp:17 documents "validate_all runs (a) then (b)" where (a) = manifest_abi_hash and (b) = symbol_abi_hash. The implementation here runs gate-1b (symbol) FIRST (line 136) and gate-1a (manifest) SECOND (line 141). plugin-abi.md §step 4 itself does not pin a sub-order, but the header is the authoritative API contract and any caller relying on the documented order will see surprising error provenance during diagnostics (e.g. detail strings attached later). fix:Either swap the two `if (auto r = ...)` blocks at lines 135-143 so manifest is checked first to match the header doc, OR update plugin_loader_registry.hpp lines 16-18 to read "validate_all runs (b) then (a) — symbol first because the symbol value is what dlsym just resolved and is the more diagnostic of a malformed manifest". Pick one. Lean toward swapping the code: the manifest-first ordering matches the natural reading order of plugin-abi.md §step 4 ("compare manifest.abi_hash AND the redundant glibre_plugin_abi_hash exported symbol").

core/src/plugin_loader_registry.cpp
```diff
@@ -0,0 +1,208 @@
...
+// Returns an error (without mutating the registry) if name is already
+// registered with a different path.  Same name + same path is an idempotent
+// no-op (returns success without re-inserting).
+// ---------------------------------------------------------------------------
+
+Result<void> PluginLoaderRegistry::register_plugin(
+    const PluginManifest& manifest,
+    eastl::string_view    path) noexcept {
+
+    const auto it = loaded_.find(manifest.name);
```
cjhowe-us, line 193: severity:MED location:core/src/plugin_loader_registry.cpp:171-180 problem:SRP / "two reasons to change → split". The same-name-different-path collision predicate exists in two places: `validate_name_unique` (lines 87-95) AND `register_plugin` (lines 171-180). Both contain the same `loaded_.find(manifest.name) → if it != end → if path != stored.path → PluginNameCollision; else idempotent` shape. If plugin-abi.md §step 6 ever evolves (e.g. add a hash-equality requirement on top of path-equality for the "idempotent" branch), both sites need to change in lockstep — exactly the symptom SRP forbids. fix:Have `register_plugin` call `validate_name_unique(manifest, path)` and short-circuit on its error, then handle the "name already present, same path" idempotent branch separately. Pseudocode: `if (auto r = validate_name_unique(manifest, path); !r) return r; if (loaded_.find(manifest.name) != loaded_.end()) return {}; /* insert new record */`. One source of truth for the collision predicate, one for the insertion policy.

tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp
```diff
@@ -0,0 +1,482 @@
...
+// ===========================================================================
+// Additional: gate ordering in validate_all
+//
+// Verifies that validate_all checks gates in the mandated order:
+//   hash → engine version → name → deps.
+//
+// A manifest that fails gate 1 (hash) should not proceed to check name/deps.
+// ===========================================================================
+
+TEST_CASE("validate_all_gate_ordering_hash_first", "[core][plugin_loader_registry]") {
```
cjhowe-us, line 305: severity:MED location:tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp:308-345 problem:Seam-quality / coverage gap. Only the hash-first sub-order of `validate_all` is asserted (`validate_all_gate_ordering_hash_first`). Gates 2→3→4 ordering is not pinned by any test. If a future refactor reorders, say, name and dependency checks (e.g. someone moves dep resolution before name uniqueness because deps "feel cheaper"), the per-gate tests still pass but the public ordering contract on plugin-abi.md §steps 5/6/7 silently drifts. Round-2 lens (seam quality) flags this. fix:Add two short test cases: (a) `validate_all_runs_engine_version_before_name` — fixture has a name collision AND a too-low engine version; expect `PluginEngineTooOld`, not `PluginNameCollision`. (b) `validate_all_runs_name_before_deps` — fixture has a name collision AND a missing dep; expect `PluginNameCollision`, not `PluginDependencyMissing`. Both are 5-line tests using existing helpers; total coverage cost is trivial.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:3
  LOW:0
carryover:0
notes:Round-1 fixes mostly land cleanly — the name-uniqueness file-path qualifier (HIGH-1), heterogeneous hash_map lookup (MED-3), the manifest/symbol gate split (MED-2), the dead include (MED-4), the unconditional register_plugin precondition (MED-5), the comment fix (LOW-6), and the make_manifest_with_deps helper (LOW-7) are all correctly addressed at b915451. New tests `name_unique_allows_same_name_same_path` and `name_unique_rejects_same_name_different_path` lock the hot-reload contract. However, two HIGH issues block round-3 and merge:
  1. DoD regression — the round-1 MED-2 split renamed `validate_abi_hash` to `validate_manifest_abi_hash` + `validate_symbol_abi_hash`, but the DoD on issue #230 still asserts `file_contains: validate_abi_hash`. POSIX-extended grep -E confirms the regex no longer matches. The dod-verify workflow will reopen #230 on merge. Fix: restore a thin `validate_abi_hash` wrapper or update the DoD regex to `validate_(manifest|symbol)_abi_hash`.
  2. Lint CI failed — clang-format -Werror reports ~50 violations (column-aligned parameter declarations, member-init alignment, multi-line hash_map template). Build/e2e jobs are skipping behind it, so we have NO confirmation that the round-1 test claims still hold on this commit. Run `clang-format -i` on all three files and amend.
Three MEDs are real but acceptable to thread into the response: (a) `validate_all` runs gate-1b (symbol) before gate-1a (manifest), but the header doc claims (a) then (b) — pick one and align; (b) the same-name-different-path collision predicate is duplicated across `validate_name_unique` and `register_plugin` (SRP — two reasons to change → one source of truth); (c) only hash-first gate ordering is pinned by tests — gates 2→3→4 ordering is not asserted, allowing silent future drift on plugin-abi.md §steps 5/6/7. Recommend addressing both HIGHs before round 3 lands; the three MEDs can be batched into the same impl-respond commit.

core/src/plugin_loader_registry.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ddb8ba8
DoD on issue #230 updated via `gh issue edit 230 --body-file` — regex changed from `validate_abi_hash` to `validate_(manifest|symbol)_abi_hash`, which matches both `validate_manifest_abi_hash` and `validate_symbol_abi_hash` in the current code. Verified with `grep -E` against the post-commit file. dod-verify will pass on merge.

core/include/glibre/core/plugin_loader_registry.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ddb8ba8
Ran `clang-format --style=file -i` on all three files (plugin_loader_registry.cpp, plugin_loader_registry.hpp, plugin_loader_registry_test.cpp). Followed up with `clang-format --style=file --dry-run --Werror` — exits 0, zero violations. The column-aligned parameter blocks and member-init alignment that triggered the ~50 violations are now formatted per the project .clang-format rules (ColumnLimit:100, BlockIndent, etc.).

core/src/plugin_loader_registry.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ddb8ba8
Fixed validate_all to run gate 1a (manifest hash) before gate 1b (symbol hash), matching plugin-abi.md §step 4 prose which leads with manifest.abi_hash, and matching the header comment that already stated (a) then (b). The prior code ran symbol first, contradicting both the spec and the comment. The existing validate_all_gate_ordering_hash_first test is still correct: it sets manifest.abi_hash=kWrongHash and symbol=kExpectedHash, and now gate 1a fires first, returning PluginAbiHashMismatch as asserted.

core/src/plugin_loader_registry.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ddb8ba8
Extracted `is_collision(name, file_path)` as a private `[[nodiscard]] bool` method on PluginLoaderRegistry. Both `validate_name_unique` and `register_plugin` now call it. The predicate (name registered AND path differs → true; name absent or same path → false) lives in exactly one place. Header declares the method in the private section with a doc comment citing the SRP rationale.

tests/core/plugin_loader_registry/plugin_loader_registry_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ddb8ba8
Added two gate-ordering tests pinning gates 2→3 and 3→4:
- validate_all_gate_ordering_version_before_name: manifest with min_engine_version > host (gate 2 fails) AND same name as registered plugin + different path (gate 3 would fail). Asserts PluginEngineTooOld fires, not PluginNameCollision.
- validate_all_gate_ordering_name_before_deps: manifest with same name + different path (gate 3 fails) AND missing dep (gate 4 would fail). Asserts PluginNameCollision fires, not PluginDependencyMissing.
Full gate sequence 1→2→3→4 is now pinned by tests: hash_first (existing) + version_before_name + name_before_deps.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:0
carryover:0
notes:Round-3 lens (topology + spec alignment + polish) is clean. Topology canonical: header at core/include/glibre/core/plugin_loader_registry.hpp, source at core/src/plugin_loader_registry.cpp, tests at tests/core/plugin_loader_registry/ with the corresponding CMake wiring. Gate ordering 1→2→3→4 matches plugin-abi.md §"Loader Sequence" steps 4–7 and is now pinned by three explicit gate-ordering tests (hash_first, version_before_name, name_before_deps). The 1a/1b ABI-hash split matches the §step 4 prose ("manifest.abi_hash (and the redundant glibre_plugin_abi_hash exported symbol — both must agree)"). New core::Error arms (PluginEngineTooOld, PluginNameCollision, PluginDependencyMissing) match the §"Failure Modes" table rows 5/6/7. DoD on #230 fully satisfied: file_contains regex `validate_(manifest|symbol)_abi_hash` matches both refactored symbols in the .cpp; all five named TEST_CASEs present verbatim; pr body carries `Closes #230`. Conv-Commit subject and `feat(core)` scope correct. Sibling #231 (register/schedule/migrate) cited as next plan; scope discipline intact. Two prior rounds — 7 findings r1, 5 findings r2 — all ADDRESSED in commits b915451 + ddb8ba8 with no carryover. Polish clean: no dead includes, no stale TODOs, comments accurate, "Three sub-cases" corrected. CI status: lint, spec-check, dep-check, automated-review all green on ddb8ba8; build job IN_PROGRESS at review time — non-blocking, will gate merge via dod-verify regardless.

PR #835, opened 2026-05-04
docs(data): detailed design — migration-dispatcher (refs #738)

cjhowe-us:
## Summary

- Add `specs/data/migration-dispatcher-design.md` — 12-section detailed design
  for the `MigrationDispatcher` aggregate inside `glibre-types.dylib`.
- Pins SRP boundary: the dispatcher walks a per-FQN linear strictly-monotonic
  chain hop-by-hop; it does NOT own migration bodies (codegen-emitted) or
  barrier orchestration (#706).
- Records arena policy (allocate-then-replace per row, refuse on exhaustion
  per #499), single-threaded barrier-time dispatch, ABI hash composition
  contribution, in/out contract with `hot-reload-barrier-design.md`, per-step
  ≤50us budget, and a closed-sum failure model with one new arm
  (`ArenaExhausted`) that requires a small `specs/data/SPEC.md` §10.1
  amendment in the implementation plan.
- Discharges spike #738; sibling task-breakdown spike unblocks on merge.

## Test plan

- [x] Doc renders with 12 sections (Purpose / Requirements / Detailed
  model / Public surface / Hot-cold split / Concurrency / Persistence+ABI /
  Hot-reload / Performance / Failure modes / Test plan / Open questions).
- [x] Cites `reviews/decisions/{error-model,fory-codegen,plugin-abi,hot-reload-protocol,perf-budget,frame-phases}.md`.
- [x] Cites #499, #706, #734.
- [x] Style consistent with `specs/core/{type-registry,hot-reload-barrier}-design.md`.
- [ ] No code change; no engine tests; design-only spike.

Refs #738.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #53, opened 2026-05-03
[SUB-EPIC] Acceptance + Implementation Plan — data


cjhowe-us:
# Acceptance + Implementation Plan — data

Draft user-stories, close open questions, produce type:plan backlog.

Parent epic: #49

## Sections owned (in `specs/data/SPEC.md`)

- §11 Acceptance Criteria
- §12 Open Questions
- type:plan backlog

No estimate; rolls up.

PR #975, opened 2026-05-09
feat(data): glibre-foryc — emit migration dispatcher (refs #221)

### LINKED ISSUE #221: [PLAN] feat(data): glibre-foryc — emit migration dispatcher

cjhowe-us:
## Summary

Extends glibre-foryc with `--emit=migration` mode generating a per-schema
`migrations.cpp` that the plugin loader links to dispatch on-disk-format
upgrades. Companion to #225 (manifest.cpp emit) and #222 (ABI hash export).

- `tools/foryc/src/emit_migration.{hpp,cpp}` — new emission pass returning
  `Result<eastl::string>` for a complete `migrations.cpp` TU with:
  - Inline `MigrationEntry` struct (ABI-stable, must match plugin loader)
  - Forward declarations for each provider free-function
  - Static `k_migrations[]` table mapping `(from_version, to_version)` to
    `reinterpret_cast<void*>(&provider)` entries
  - `extern "C" glibre_plugin_migrations` / `glibre_plugin_migrations_size`
  - Empty table + `nullptr` + `size=0` for schemas with no migrations
- `tools/foryc/src/parser.{hpp,cpp}` — extends Schema IR with `MigrationDecl`
  (from_version, to_version, provider string) stored in `TypeDecl::migrations`.
  Inline migration syntax inside schema blocks (derived minimal subset per
  dispatch prompt, fory-codegen.md §"Migration Mechanic"):
  ```
  migration from <N> to <M> calls "<provider-symbol>"
  ```
  Top-level `migration v2_to_v3 { ... }` block skipping preserved for
  compatibility with existing .fory files.
- `tools/foryc/src/main.cpp` — `--emit=migration` flag + helper. Merged with
  plan #225's `--emit=manifest` / `--file` additions from origin/main.
- `tools/foryc/CMakeLists.txt` — adds `emit_migration.cpp` to sources alongside
  plan #225's `emit_manifest.cpp`.
- `tests/tools/foryc/foryc_emit_migration_test.cpp` — 5 Catch2 unit tests
  including both dispatch-prompt named tests.
- `tests/tools/foryc/CMakeLists.txt` — `foryc-emit-migration-tests` target
  alongside plan #225's `foryc-emit-manifest-tests`.

## Refs

Closes #221

## Test plan

- [x] `ctest --preset macos-debug -R foryc_emit_migration_` → 4/4 passed
- [x] `ctest --preset macos-debug -R foryc` → 29/30 passed (only pre-existing
  `foryc-golden-tests_NOT_BUILT` placeholder fails, same as main branch)
- [x] Both dispatch-prompt named tests explicitly verified:
  - `foryc_emit_migration_dispatcher_writes_table` PASSED
  - `foryc_emit_migration_handles_no_migrations` PASSED


tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    for (const auto& mig : td.migrations) {
+        const std::string_view prov(mig.provider.data(), mig.provider.size());
+
+        // Forward-declare the provider as an extern "C" function returning void*
+        // (type-erased; the owning context supplies the real signature).
+        // We use a void (*)(void) prototype as a minimal legal C function pointer
+        // type that can be cast to/from other function pointer types in C++.
+        frag.forward_decls += eastl::string(
+            std::format("extern \"C\" void {}();\n", prov).c_str()
```
cjhowe-us: severity:HIGH location:tools/foryc/src/emit_migration.cpp:57 problem:Generated forward decl `extern "C" void glibre::core::migrate_Transform_v1_to_v2();` is invalid C++ on two counts. (1) C language linkage forbids qualified names with namespace components — clang will emit "definition or redeclaration cannot name the global scope" or "linkage specification cannot have nested-name-specifier". (2) Per fory-codegen.md §"Migration Mechanic" point 2, the real provider signature is `std::expected<void, glibre::Error>(const VN&, VNplus1&)` with C++ linkage and a mangled name; declaring it as `extern "C" void()` references a different overload that does not exist in the owning context, so `&provider` will fail to link. The unit tests pass only because they string-grep the output and never compile/link it. fix:Drop `extern "C"` and emit nested-namespace forward decls matching the real signature, OR (preferred per fory-codegen.md point 2) have the dispatcher #include the per-type generated `glibre/types/<ctx>/<Type>_migrations.hpp` header that exposes the real declaration, OR switch to a registration model where the owning context calls a generated `register_migration(...)` at static-init. Add a compile-link smoke test (catch2 + add_executable on the generated TU stubbed against a fake provider) so this regression cannot recur.

tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,182 @@
...
+    // (from_version, to_version) pairs regardless of which TypeDecl they came from.
+    eastl::string all_forward_decls;
+    eastl::string all_entries;
+    std::size_t total_count = 0;
+
+    for (const auto& td : schema.types) {
+        auto frag = build_type_fragment(td);
+        all_forward_decls += frag.forward_decls;
+        all_entries += frag.table_entry;
+        total_count += frag.count;
```
cjhowe-us: severity:MED location:tools/foryc/src/emit_migration.cpp:96-105 problem:All TypeDecls are flattened into a single anonymous `k_migrations[]` table without per-type routing — the entries record only `(from_version, to_version, provider)` with no `fqn` discriminator. fory-codegen.md §"Migration Mechanic" point 1 says "Each generated type carries `current_version` and a static `migrations` table" (per-type), and point 3 says deserialize<T> "look[s] up a chain `version → current` in the migrations table" — i.e. the table is indexed per type. With the flat layout, a `.fory` containing both `Transform` and `Velocity` schemas with overlapping `(1,2)` migration steps would be indistinguishable to the loader. fix:Either (a) constrain the emit pass to a single TypeDecl per call (one migrations.cpp per Type, not per file), or (b) widen `MigrationEntry` with an `fqn_hash` / `type_index` discriminator and emit one entry per (type, step). Option (a) aligns with the §"Pipeline" sketch which puts generated sources at `src/<ctx>/<Type>.cpp`.

tools/foryc/src/parser.cpp
```diff
@@ -433,6 +444,69 @@ class Parser {
...
+        // "<provider>" — a string literal; strip surrounding quotes.
+        if (current_.kind != TokenKind::StringLit)
+            return FORYC_ERR(
+                tools::Error::ForycSyntaxError,
+                "expected quoted provider symbol after 'calls' in migration decl"
+            );
+        const std::string_view raw = current_.text;
+        // StringLit includes the surrounding quotes: raw[0] == '"', raw[last] == '"'.
+        eastl::string provider(raw.data() + 1, raw.size() - 2);
+        advance();
```
cjhowe-us, line 507: severity:LOW location:tools/foryc/src/parser.cpp:501 problem:`eastl::string provider(raw.data() + 1, raw.size() - 2)` underflows if the lexer ever yields a StringLit shorter than 2 chars (e.g. malformed token, empty `""` is exactly 2 chars and yields zero-length provider — also accepted silently). fix:Add a defensive check `if (raw.size() < 2) return FORYC_ERR(...)` and reject zero-length providers explicitly with a clear "empty migration provider symbol" diagnostic.

tests/tools/foryc/foryc_emit_migration_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // std::size_t used for the size (PHILOSOPHY §11: std:: for non-EASTL utilities).
+    CHECK(text.find("std::size_t") != eastl::string::npos);
+}
+
+TEST_CASE("foryc_emit_migration_rejects_empty_schema", "[foryc][emit_migration]") {
+    // A Schema with zero TypeDecls must return ForycEmptySchema.
+    Schema schema;
+    schema.source_path = eastl::string("<empty>");
+
+    auto result = emit_migration(schema, "<empty>");
```
cjhowe-us, line 191: severity:HIGH location:tests/tools/foryc/foryc_emit_migration_test.cpp:119-167 problem:`foryc_emit_migration_dispatcher_writes_table` and the no-migrations test only string-grep the emitted text. They never compile, never link, never execute the dispatcher. The fundamental defect that the generated TU is not valid C++ (see emit_migration.cpp:57 finding) is silently masked. The DoD `unit_test_named` assertions are technically satisfied but the tests do not exercise the emitted TU as a translation unit. fix:Add a compile-and-link smoke test target that consumes the generated string, writes it to a temp .cpp, invokes the project compiler against a stub providing the real provider definitions, and links to a fake plugin loader that reads `glibre_plugin_migrations` + `_size`. Without this, any future emit regression is undetectable by the unit tests.

tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,182 @@
...
+// -----------------------------------------------------------------------
+
+struct TypeMigrationFragment {
+    eastl::string forward_decls;  // extern declarations for provider symbols
+    eastl::string table_entry;    // one MigrationEntry initialiser list entry per step
+    std::size_t count{0};         // number of migration steps
+};
+
+// Build the forward-declaration block and table entries for one TypeDecl.
+[[nodiscard]] static TypeMigrationFragment
```
cjhowe-us: severity:HIGH location:tools/foryc/src/emit_migration.cpp:43,56,57,172 + tools/foryc/src/parser.hpp:110 problem:`lint` CI job is failing with 5 clang-format violations on this PR (run 25594277567). DoD §`workflow_passed: ci.yml` is therefore NOT satisfied; PR is currently unmergeable. fix:Run `clang-format -i` on emit_migration.cpp and parser.hpp; the offending lines all involve the EASTL/std::format conversion idiom and the MigrationDecl member alignment in parser.hpp:110.

cjhowe-us (commented):
findings:
  HIGH:3
  MED:1
  LOW:1
carryover:0
notes:Round-1 lens (coverage + correctness). The diff is in scope: parser extension + emit_migration pass + 5 unit tests including the two DoD-named tests; PR body has `Closes #221`; Conventional Commit subject + scope correct; #221 DoD block intact in the issue body. Empty-table path works; EASTL substrate is used; `glibre::Result<T>` returned at the seam. However three correctness issues block merge: (1) the generated TU forward-declares C++ namespaced providers as `extern "C" void glibre::core::xxx();` which is invalid C++ AND has the wrong signature versus fory-codegen.md §"Migration Mechanic" point 2 (`std::expected<void, Error>(const VN&, VNplus1&)`), so taking `&provider` will fail to link; (2) the unit tests only string-grep emitted text — they never compile or link the output, so the linkage defect is silently masked; the DoD-named tests pass but do not exercise the dispatcher as a TU; (3) the `lint` CI job fails (clang-format violations on emit_migration.cpp:43,56,57,172 and parser.hpp:110), so the DoD assertion `workflow_passed: ci.yml` is not satisfied and the PR is currently unmergeable. One MED concern: all TypeDecls are flattened into a single anonymous `k_migrations[]` table without per-type discriminator, contradicting fory-codegen.md §"Migration Mechanic" point 1 ("Each generated type carries a static migrations table"). One LOW: `parse_inline_migration` slices `raw.size() - 2` without bounds check and accepts empty quoted providers. What works: parser plumbing, EASTL integration, error-code surface, two-mode CLI flag, empty-table path. What blocks: HIGH-1/2/3 above. What's deferred: per-type table layout (MED-1) may require a follow-up plan if the loader's expected ABI is settled in #835's accepted decision.

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e64123a

Fixed in the commit above. The generated forward-decl now uses proper C++ namespace-scoped declarations instead of `extern "C" void ns::fn()`. Per C++ [dcl.link]/6 (ISO C++23), C language linkage cannot apply to namespace-qualified names. The correct form per fory-codegen.md §"Migration Mechanic" point 2 is now emitted:

```cpp
namespace glibre::core {
std::expected<void, glibre::Error> migrate_Transform_v1_to_v2(const TransformV1&, TransformV2&);
std::expected<void, glibre::Error> migrate_Transform_v2_to_v3(const TransformV2&, TransformV3&);
}  // namespace glibre::core
```

The function name and versioned type names (`TransformV1`, `TransformV2`) are derived from the schema FQN (`glibre.core.Transform`) + the version integers in the MigrationDecl. The namespace is split from the FQN's prefix components.

tests/tools/foryc/foryc_emit_migration_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e64123a

Added `foryc_emit_migration_round_trip_via_compile` test following the pattern from `foryc_emit_manifest_round_trip_via_compile`. The test:
1. Emits a migration TU for a two-type schema (Widget with 1 migration, Gadget with none).
2. Prepends a preamble that defines `glibre::Error`, the versioned stub structs (`WidgetV1`, `WidgetV2`), and the provider body — mirroring how a real build compiles the generated TU together with the owning context's headers.
3. Compiles the combined source via `clang++ -std=c++23` into a `.dylib` (hard REQUIRE on rc==0).
4. dlopens the dylib and dlsyms the per-type C-ABI symbols:
   - `glibre_plugin_migrations_Widget` (non-null, size==1)
   - `glibre_plugin_migrations_Widget_size`
   - `glibre_plugin_migrations_Gadget` (nullptr, size==0)
   - `glibre_plugin_migrations_Gadget_size`

The compile step would have caught HIGH-1's invalid forward-decl on its own. All 5 migration tests pass (30/31 overall foryc suite, only the pre-existing `NOT_BUILT` placeholder fails).

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e64123a

Ran `clang-format --style=file -i` on `emit_migration.cpp`, `emit_migration.hpp`, `parser.cpp`, and `parser.hpp`. Verified clean with `--dry-run --Werror` — all 5 violations resolved. The new `emit_migration.cpp` rewrite was also formatted before commit.

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e64123a

Redesigned the emitter to produce per-type tables per fory-codegen.md §"Migration Mechanic" point 1. Each TypeDecl now gets its own:
- `static const MigrationEntry k_migrations_<TypeName>[]`
- `extern "C" const MigrationEntry* glibre_plugin_migrations_<TypeName>`
- `extern "C" std::size_t glibre_plugin_migrations_<TypeName>_size`

For the test schema `glibre.core.Transform` the exported symbols are `glibre_plugin_migrations_Transform` and `glibre_plugin_migrations_Transform_size`. The plugin loader addresses each type's migration chain independently, matching the design intent in the decision record. Tests updated to assert per-type symbol names rather than the old flat names.

tools/foryc/src/parser.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:e64123a

Added a bounds check in `parse_inline_migration` before slicing the StringLit. The check rejects any token shorter than 3 characters (the minimum for a non-empty quoted string, e.g. `"a"`) with `ForycSyntaxError`. An empty provider string (`""`, 2 chars) is now rejected before `raw.size() - 2` is evaluated, preventing the underflow.

tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,271 @@
...
+    out += "#include <cstddef>\n";
+    out += "#include <cstdint>\n";
+    out += "#include <expected>\n";
+    out += "\n";
+
+    // Bring glibre::Error into scope so the provider forward-declarations
+    // can name it without requiring a separate header in the generated TU.
+    out += "// Forward-declare glibre::Error so provider signatures compile\n";
+    out += "// without pulling in the full engine headers.\n";
+    out += "namespace glibre { struct Error; }\n";
```
cjhowe-us: severity:HIGH location:tools/foryc/src/emit_migration.cpp:249 problem:Generated TU forward-declares `namespace glibre { struct Error; }` (incomplete type) but then names `std::expected<void, glibre::Error>` in the provider fwd-decls. libc++ implements std::expected with a union-of-T-and-unexpected<E> data member; instantiating `std::expected<void, glibre::Error>` requires E complete, so the production-build TU (compiled standalone into glibre-types.dylib via emit_migration_for_schema) is ill-formed. The new round-trip test masks this because its preamble injects a complete `struct Error {};` BEFORE the generated text — the production-build path has no such preamble. fix:Replace the forward-decl with `#include "glibre/error.hpp"` so the generated TU is genuinely self-contained, AND change the round-trip test to compile the generated text WITHOUT the in-test stub (so the seam is actually exercised).

tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,271 @@
...
+    // Signature: std::expected<void, glibre::Error>
+    //                migrate_<Type>_v<N>_to_v<M>(const <Type>V<N>&, <Type>V<M>&)
+    //
+    // These are C++ namespace-qualified declarations (NOT extern "C") because
+    // C linkage forbids namespace-qualified names (C++ [dcl.link] / ISO C++23
+    // [dcl.link]/6: "A name with C language linkage shall not be in a
+    // namespace").
+    // -----------------------------------------------------------------------
+
+    if (!parts.ns.empty()) {
```
cjhowe-us: severity:MED location:tools/foryc/src/emit_migration.cpp:114 problem:The provider forward-decl is wrapped in `parts.ns` (the *type's* namespace, derived from `td.fqn`), then `fn_name` is the unqualified tail of `mig.provider`. This silently assumes the provider lives in the same namespace as the type. fory-codegen.md §"Migration Mechanic" point 2 only says the owning context provides the body; it does not require the provider to share the type's namespace, and the parser stores `provider` as an arbitrary fully-qualified symbol. If a schema declares `glibre.core.Transform` with provider `"glibre::physics::migrate_Transform_v1_to_v2"`, the emitter forward-declares `migrate_Transform_v1_to_v2` inside `glibre::core` while the table still uses `&glibre::physics::migrate_Transform_v1_to_v2` — link error. fix:Split `mig.provider` itself (last `::`) to derive the provider's own namespace and emit the fwd-decl in *that* namespace, independent of the type's namespace. Add a unit test that puts the provider in a different namespace than the type to cover the seam.

tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,271 @@
...
+    // (e.g. TransformV1 / TransformV2) the MigrationEntry stores the
+    // function pointer as void* (type-erased) identical to the layout
+    // declared in the MigrationEntry struct above.  The plugin loader
+    // recovers the real type via the registered deserialize path.
+    // -----------------------------------------------------------------------
+
+    // Static per-type migration table.
+    // Fully-qualified provider is used here so the reference links even if
+    // the forward decl above is in a different namespace.
+    const std::size_t count = td.migrations.size();
```
cjhowe-us, line 209: severity:DEFER location:tools/foryc/src/emit_migration.cpp:165 problem:fory-codegen.md §"Migration Mechanic" point 1 specifies "Each generated type carries `current_version` and a static `migrations` table". This PR emits the migrations table only; `current_version` (which the parser already captures as `td.version`) is not exposed in the generated TU. The deserialize-with-migration loop in the loader needs both — without `current_version` the loader has no way to short-circuit the no-migration-needed path or to detect a payload from a *future* version. Recommendation:DEFER to a follow-up plan that owns the per-type version-export symbol (likely co-located with #220's per-type header emission, e.g. `static constexpr std::uint32_t k_current_version_<Type> = N;` exported via an extern "C" symbol or constexpr in the generated header). Out-of-scope-widening for #221; flag rather than block.

tools/foryc/src/emit_migration.cpp
```diff
@@ -0,0 +1,271 @@
...
+    out += "// clang-format off\n";
```
cjhowe-us: severity:LOW location:tools/foryc/src/emit_migration.cpp:166 problem:Generated TU embeds `// clang-format off` / `// clang-format on` directives. Generated outputs land under `${CMAKE_BINARY_DIR}/generated/...` which is excluded from clang-format scope, so the directives are dead noise in every emitted file. fix:Drop the two `out += "// clang-format off|on\n";` lines; the build directory is not lint-scoped.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:1
  DEFER:1
carryover:0
notes:Round-2 lens (cohesion + SRP + seam quality). All five round-1 findings are genuinely fixed in commit e64123a — provider fwd-decls are now C++ namespaced with the correct std::expected<void, glibre::Error>(const VN&, VNplus1&) signature; per-type k_migrations_<Type>[] tables and per-type extern "C" exports match fory-codegen.md §"Migration Mechanic" point 1; clang-format / lint / spec-check / dep-check are green; round-trip dlopen+dlsym test exists. Seam quality is mostly solid: parser stores MigrationDecl cleanly, emit_migration is single-responsibility (one Schema → one TU), Result<eastl::string> at the boundary, EASTL substrate consistent. One new HIGH concern blocks merge: the generated TU only forward-declares glibre::Error but names std::expected<void, glibre::Error>, which under libc++ requires Error complete on instantiation; the round-trip test masks the defect because its in-test preamble defines `struct Error {};` before the generated text. The production-build path (emit_migration_for_schema writes the text verbatim with no preamble) is therefore ill-formed — fix is to emit `#include "glibre/error.hpp"` and remove the in-test stub so the seam is genuinely exercised. One MED: the provider-namespace assumption — fwd-decl is wrapped in the type's namespace, not the provider's — fails the moment a schema declares a cross-context provider; needs an emitter split on `mig.provider` and a regression test. One DEFER: fory-codegen.md point 1 also requires `current_version` per type, which this PR omits — recommend a follow-up plan rather than widening #221's scope. One LOW: dead `// clang-format off/on` directives in generated output. What works: scope discipline, EASTL/std boundary, error-code surface, two DoD-named tests + 4 supplementals, per-type table layout, clang-format / lint / spec-check / dep-check / automated-review all green. What blocks: HIGH-A (incomplete-Error UB in real build path). What's deferred: MED-C (current_version export). Build job still IN_PROGRESS at review time — expected to fail compile of any real .fory schema once integrated; HIGH-A explains why.

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:a3ebb63

Replaced the incomplete `namespace glibre { struct Error; }` forward-declaration with `#include "glibre/error.hpp"` in the generated TU (emit_migration.cpp lines ~250-256 post-fix). As the reviewer noted, libc++ instantiates `std::expected<void, E>` with a union-of-T-and-unexpected<E> data member; E must be complete at that point.

The round-trip test is also fixed: the generated TU is now compiled as a separate object without any preamble prepended, so the seam is genuinely exercised. `-I` flags for `glibre/error.hpp` (core/include) and EASTL are injected via CMake compile definitions (`GLIBRE_CORE_INCLUDE_DIR` / `GLIBRE_VCPKG_INCLUDE_DIR`) set in the test target's `CMakeLists.txt`.

Additionally emitted forward-declarations for the versioned structs (`struct TransformV1; struct TransformV2;`) in the type's namespace so the generated TU is self-contained (the real build has the generated type headers on its include path; the fwd-decls handle standalone/round-trip compilation).

All 6 foryc_emit_migration_ tests pass (including the compile round-trip).

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:a3ebb63

The provider forward-declaration now derives its namespace by splitting `mig.provider` on the last `::` (not by using `parts.ns` from the type's FQN). For a schema type in `glibre.core` with provider `"glibre::physics::migrate_Transform_v1_to_v2"`, the emitter now emits:

```cpp
namespace glibre::core { struct TransformV1; struct TransformV2; }  // type fwd-decls
namespace glibre::physics {
std::expected<void, glibre::Error> migrate_Transform_v1_to_v2(
    const glibre::core::TransformV1&, glibre::core::TransformV2&);
}  // namespace glibre::physics
```

Parameter types are always fully-qualified (using the type's namespace prefix) so the declaration is unambiguous regardless of which namespace wraps the provider fwd-decl.

New test `foryc_emit_migration_cross_namespace_provider` covers this seam: it verifies that `namespace glibre::physics` wraps the fwd-decl, that `glibre::core::TransformV1` / `TransformV2` appear as parameter types, and that the table entry still references the fully-qualified `glibre::physics::migrate_Transform_v1_to_v2`. All 6 tests pass.

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:DEFER reasoning:fory-codegen.md §'Migration Mechanic' point 1 requires current_version, confirmed. However, widening #221's scope to include the version export would pull in design questions about placement (generated TU vs. generated header per plan #220's per-type header emit) that are out of #221's declared scope. The parser already captures td.version; the emitter just needs one more emit call. proposal:Opened follow-up plan #978 '[PLAN] feat(data): glibre-foryc — emit per-type current_version constant (iterate #221)' parented to data sub-epic #53. That plan owns the extern "C" symbol emit and a round-trip dlsym test, coordinating with plan #220 on header vs. TU placement.

tools/foryc/src/emit_migration.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:a3ebb63

Removed the two `out += "// clang-format off\n";` and `out += "// clang-format on\n";` lines from `emit_type_block`. Generated files land under `${CMAKE_BINARY_DIR}/generated/glibre-types/` which is excluded from clang-format scope in .clang-format; the directives were unreachable dead noise in every emitted file.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0
notes:Round-3 lens (topology + spec alignment + polish). All round-1 (5) and round-2 (3) findings were verified ADDRESSED in commits e64123a + a3ebb63 — provider fwd-decls now C++ namespace-scoped with std::expected<void, glibre::Error>(const VN&, VNplus1&); per-type k_migrations_<Type>[] tables + per-type extern "C" exports match fory-codegen.md §"Migration Mechanic" point 1; #include "glibre/error.hpp" supplies the complete glibre::Error required by libc++'s std::expected<void, E>; provider namespace is now derived from mig.provider (cross-namespace correct); clang-format directives dropped from the generated TU. DEFER-C (current_version emission) properly opened as #978. Topology fits cleanly alongside plans #220 / #225 in the foryc pipeline (one --emit mode per pass). DoD #221 satisfied: file_exists for both emit_migration.{hpp,cpp}, both named tests present (foryc_emit_migration_dispatcher_writes_table / foryc_emit_migration_handles_no_migrations), PR body has "Closes #221", and ci.yml is green except for the build job still IN_PROGRESS at review time (lint / spec-check / dep-check / automated-review all SUCCESS). Round-trip-via-compile test is the seam the impl-respond agent added in r1 and continues to exercise the generated TU as a real translation unit. One LOW informational item only (does not block): the inline MigrationEntry struct in the generated TU is annotated "Layout MUST match glibre::types::MigrationEntry in the plugin loader" but no such loader-side header exists yet; the layout contract is currently single-sourced in emit_migration.cpp. When the plugin loader work lands (post-#221) it should pull this struct out into glibre/types/migration_entry.hpp and have the emitter #include that header instead — track via a follow-up if not already covered. Convergence reached: no HIGH, no carryover, CI green modulo in-progress build. Approve.

Issue #245, opened 2026-05-03
[PLAN] core/frame-loop: phase ownership + system register API


cjhowe-us:
### Domain
core

### Phase
phase:mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
TBD — pending `[SPIKE] draft-core-user-stories` (#29). Plan satisfies
the SRP-per-phase ownership constraint of
`reviews/decisions/frame-phases.md`.

### Parent
#3

### Scope
Add the phase-ownership registry that lets plugins register **systems**
into existing phases without changing phase ownership. Owning context
of each phase is a compile-time constant from the prior plan; this plan
adds the runtime registry of system callbacks per phase.

In scope:
- `core/include/glibre/core/phase_registry.hpp` — `class PhaseRegistry`
  with `register_system(Phase, std::string_view fqn, SystemFn)` and
  `for_each_system(Phase, F&&) const`. Idempotent w.r.t. `(phase,
  system_fqn)` per `hot-reload-protocol.md` step 4.1.
- `core/src/phase_registry.cpp` — fixed `std::array<…, 9>` of system
  vectors keyed by `static_cast<size_t>(Phase)-1`. Iteration order is
  registration order (deterministic).
- Hook into `FrameLoop::tick()` so each phase invokes its registered
  systems in registration order (which means MVP empty phases stay
  no-op until real systems register).

Out of scope:
- Read/write set validation per system — next plan.
- ABI hashing of system fn pointers — covered by plugin manifest plan.

### Unit Test Plan
- core/phase_registry: register_system_idempotent
- core/phase_registry: iteration_order_matches_registration_order
- core/phase_registry: iteration_order_matches_registration_order frameloop smoke
- core/phase_registry: per_phase_isolation

### Depends on
- #244 (FrameLoop skeleton)

### Agent Execution Notes
Branch `feat/core-phase-registry`. PR title:
`feat(core): phase ownership registry + system register API`.
No allocation inside `for_each_system`. Registry is mutated only at
plugin register/drain time per hot-reload protocol — declare this in
the header doc comment.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: core/include/glibre/core/phase_registry.hpp
- file_exists: core/src/phase_registry.cpp
- unit_test_named: "core/phase_registry: register_system_idempotent"
- unit_test_named: "core/phase_registry: iteration_order_matches_registration_order"
[...]

PR #980, opened 2026-05-09
feat(core): per-context transient arena drained at phase 9 (refs #239)

### LINKED ISSUE #239: [PLAN] feat(core): per-context transient arena drained at phase 9
### Domain

core

### Phase

phase:mvp

pts:3

### User Stories Satisfied

Implements perf-budget.md Allocator Rules #4 ("Transient arena exemption — drained at phase 9; drain failure is `core::Error::OutOfBudget` with leak detail"). Without this, every per-frame scratch allocation eats the per-context cell ceiling and the ceilings cease to model steady-state correctly.

### Parent (Epic or Sub-Epic)

#4

### Scope

Single SRP scope: introduce `glibre::TransientArena` under `core/include/glibre/transient_arena.hpp` + `core/src/alloc/transient_arena.cpp`. One arena per `ContextTag`, owned by the corresponding context's allocator handle. Bump-pointer allocation; `drain()` called by the frame loop at end of phase 9; arena keeps high-watermark stat for the perf HUD. Drain failure (allocations still alive past phase 9) returns `core::Error::OutOfBudget` with `ErrorContext::detail = "transient arena leak"`.

Out of scope: integration with the frame-phases driver (separate platform plan, since phase 9 is platform-owned); leak detection symbolication (debug-only nice-to-have, not in MVP gate); editor HUD readout (separate tools plan).

### Unit Test Plan

Tests live under `tests/core/transient_arena_test.cpp`.

Original plan-body unit-test names (all currently present in the suite):

- core/transient_arena: bump_alloc_returns_aligned_storage
- core/transient_arena: drain_resets_high_watermark
- core/transient_arena: undrained_allocation_at_phase_9_returns_out_of_budget
- core/transient_arena: allocations_do_not_count_against_cell_ceiling

Additional dispatch-DoD unit-test names (executor added these during PR #980; all currently present in the suite). Note that the originally-named `transient_arena_drained_at_phase_9` was renamed to `transient_arena_present_drain_idempotent_after_caller_drain` during the GLIBRE_ALLOC_STRICT integration fix in PR #1004 (refs #997) — the rename reflects the test's actual contract (caller drains → tick's drain is idempotent), and the strict-mode failure path is covered by a dedicated companion test (`undrained_allocation_through_tick_returns_out_of_budget`).

- transient_arena_allocates_and_resets
- transient_arena_present_drain_idempotent_after_caller_drain
- transient_arena_alignment_respected
- transient_arena_exhaustion_returns_error

### Depends on

- #238

### Agent Execution Notes

Subagents post status comments here. PR title: `feat(core): per-context transient arena drained at phase 9`

Closing PR (merged into main): #980 (commit d919fb29).

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: core/include/glibre/transient_arena.hpp
- file_exists: core/src/alloc/transient_arena.cpp
- unit_test_named: "core/transient_arena: bump_alloc_returns_aligned_storage"
- unit_test_named: "core/transient_arena: drain_resets_high_watermark"
- unit_test_named: "core/transient_arena: undrained_allocation_at_phase_9_returns_out_of_budget"
- unit_test_named: "core/transient_arena: allocations_do_not_count_against_cell_ceiling"
- unit_test_named: transient_arena_allocates_and_resets
- unit_test_named: transient_arena_present_drain_idempotent_after_caller_drain
- unit_test_named: transient_arena_alignment_respected
- unit_test_named: transient_arena_exhaustion_returns_error
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

Implements per-context transient arena allocator. O(1) bump-pointer allocate,
O(1) reset at frame-loop phase 9 (Phase::Present / end-of-frame). Plugins and
engine code allocate frame-scoped temporaries that are auto-collected at frame end
without counting against the per-context heap ceiling (perf-budget.md §Allocator
Rules #4).

- `core/include/glibre/transient_arena.hpp` — `TransientArena` class: `allocate()`,
  `drain()`, `assert_drained()`, `high_watermark()`, `bytes_used()`, `bytes_capacity()`
- `core/src/alloc/transient_arena.cpp` — bump-pointer implementation; O(1) drain
  resets cursor only (no zeroing); `assert_drained()` returns
  `core::Error::OutOfBudget` with `"transient arena leak"` detail when cursor is
  non-zero (perf-budget.md §4 drain failure contract)
- `core/include/glibre/error.hpp` — adds `core::Error::TransientArenaExhausted`
- `core/include/glibre/log_error.hpp` — adds `to_string` arm for new enumerator
  (required by `-Werror=switch`)
- `core/include/glibre/core/frame_loop.hpp` — `register_transient_arena()` API;
  Phase::Present body drains all registered arenas; `kMaxTransientArenas = 16`
- `core/src/frame_loop.cpp` — implements registration and phase-9 drain loop;
  null pointer guard via assert + silent skip in release
- `tests/core/transient_arena_test.cpp` — 8 named Catch2 tests
- `tests/core/{CMakeLists.txt,error_register/error_register_test.cpp,log_error_test.cpp}` —
  updated for the new `TransientArenaExhausted` enumerator

Pre-existing failure: `core/frame_loop: tick_invokes_phases_in_strict_order` was
already failing on `main` before this change (GLIBRE_TESTING ODR issue with the
static library; not introduced by this PR — confirmed by running the same test
against `origin/main`).

## Refs

Closes #239

## Test plan

- [x] `ctest --preset macos-debug -R transient_arena_` → 8/8 passed
- [x] `ctest --preset macos-debug -R error_register` → 3/3 passed
- [x] `ctest --preset macos-debug -R log_error` → 5/5 passed
- [x] `ctest --preset macos-debug -R frame_loop` → pre-existing failure unchanged (not introduced here)

### DIFF
--- a/core/include/glibre/core/frame_loop.hpp
+++ b/core/include/glibre/core/frame_loop.hpp
@@ -10,6 +10,8 @@
 //   - Empty MVP phase bodies — phases 1, 2, 3, 4, 6, 7, 8, 9 are no-ops
 //     until each owning context's plan lands its body (phase 5 transform
 //     is similarly empty at this skeleton stage).
+//   - At Phase::Present (9), drain all registered TransientArena instances
+//     (perf-budget.md §Allocator Rules #4, plan #239).
 //   - No dynamic allocation inside tick().
 //   - -fno-exceptions clean; noexcept throughout the public surface.
 //
@@ -18,10 +20,12 @@
 
 #include <array>
 #include <cstdint>
+#include <cstddef>
 #include <span>
 
 #include "glibre/core/frame_phase.hpp"
 #include "glibre/error.hpp"
+#include "glibre/transient_arena.hpp"
 
 namespace glibre::core {
 
@@ -36,7 +40,13 @@ namespace glibre::core {
 //   (b) No phase observes writes from a later phase of the same frame.
 //   (c) No heap allocation occurs inside tick().
 //   (d) The call returns glibre::Result<void>; callers must inspect the result.
+//   (e) Every registered TransientArena is drained at the end of phase 9
+//       (perf-budget.md §Allocator Rules #4).
 // -----------------------------------------------------------------------
+
+// Maximum number of TransientArena instances that can be registered with
+// one FrameLoop.  One slot per bounded context (9 MVP contexts + headroom).
+inline constexpr std::size_t kMaxTransientArenas = 16;
 
 class FrameLoop {
 public:
@@ -50,12 +60,29 @@ public:
 
     ~FrameLoop() noexcept = default;
 
+    // register_transient_arena() — register a TransientArena for phase-9 drain.
+    //
+    // The arena pointer must remain valid for the lifetime of this FrameLoop.
+    // Callers are responsible for ensuring pointer validity.
+    //
+    // Returns:
+    //   glibre::Result<void> — success if registered; failure with
+    //   core::Error::OutOfBudget when kMaxTransientArenas slots are full.
+    //
+    // Thread safety: must be called before tick() begins (not safe to call
+    // concurrently with tick()).
+    [[nodiscard]] glibre::Result<void>
+    register_transient_arena(glibre::TransientArena* arena) noexcept;
+
     // tick() — advance one engine frame.
     //
     // Walks all nine phases in numeric order.  In debug builds a
     // per-frame phase counter is maintained; any phase that attempts
     // to execute out of sequence (e.g. because a future registration
     // mechanism misfires) returns core::Error::FramePhaseMisordered.
+    //
+    // At Phase::Present (9), all registered TransientArena instances are
+    // drained (drain() called) before phase exit (perf-budget.md §4).
     //
     // Returns:
     //   glibre::Result<void> (= std::expected<void, glibre::Error>) —
@@ -67,6 +94,11 @@ public:
     // Incremented only after all nine phases succeed.
     [[nodiscard]] std::uint64_t frame_index() const noexcept { return frame_index_; }
 
+    // transient_arena_count() — number of registered transient arenas.
+    [[nodiscard]] std::size_t transient_arena_count() const noexcept {
+        return arena_count_;
+    }
+
 private:
     // run_phase() — execute one phase.  Returns an error if the phase
     // ordinal does not match the expected_ordinal (debug builds only).
@@ -76,6 +108,11 @@ private:
     run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept;
 
     std::uint64_t frame_index_{0};
+
+    // Registered transient arenas — drained at the end of Phase::Present (9).
+    // Non-owning pointers; lifetimes are caller-managed.
+    std::array<glibre::TransientArena*, kMaxTransientArenas> arenas_{};
+    std::size_t arena_count_{0};
 
 #ifdef GLIBRE_TESTING
     // Under GLIBRE_TESTING builds the last tick's phase execution order is
--- a/core/include/glibre/core/frame_loop.hpp
+++ b/core/include/glibre/core/frame_loop.hpp
@@ -19,8 +19,8 @@
 // See reviews/decisions/error-model.md for std::expected usage rules.
 
 #include <array>
-#include <cstdint>
 #include <cstddef>
+#include <cstdint>
 #include <span>
 
 #include "glibre/core/frame_phase.hpp"
@@ -96,9 +96,7 @@ public:
     [[nodiscard]] std::uint64_t frame_index() const noexcept { return frame_index_; }
 
     // transient_arena_count() — number of registered transient arenas.
-    [[nodiscard]] std::size_t transient_arena_count() const noexcept {
-        return arena_count_;
-    }
+    [[nodiscard]] std::size_t transient_arena_count() const noexcept { return arena_count_; }
 
 private:
     // run_phase() — execute one phase.  Returns an error if the phase
--- a/core/include/glibre/core/frame_loop.hpp
+++ b/core/include/glibre/core/frame_loop.hpp
@@ -66,7 +66,8 @@ public:
     // Callers are responsible for ensuring pointer validity.
     //
     // Returns:
-    //   glibre::Result<void> — success if registered; failure with
+    //   glibre::Result<void> — success if registered;
+    //   core::Error::InvalidArgument if arena is null;
     //   core::Error::OutOfBudget when kMaxTransientArenas slots are full.
     //
     // Thread safety: must be called before tick() begins (not safe to call
--- a/core/include/glibre/error.hpp
+++ b/core/include/glibre/error.hpp
@@ -70,6 +70,12 @@ enum class Error : std::uint16_t {
     // Returned by TransientArena::allocate() (plan #239).
     // Callers should fall back to a larger arena or defer the allocation.
     TransientArenaExhausted,
+    // Invalid argument supplied to a core API (e.g. null pointer where a valid
+    // pointer is required).  Returned instead of asserting-only so that
+    // callers in release builds receive an actionable error rather than silent
+    // undefined behaviour.  Added by plan #239 review round 1 (MED-3 fix:
+    // null-pointer silent success in register_transient_arena).
+    InvalidArgument,
 };
 }  // namespace core
 
--- a/core/include/glibre/error.hpp
+++ b/core/include/glibre/error.hpp
@@ -65,6 +65,11 @@ enum class Error : std::uint16_t {
     // After dlclose, the loader aborts without further steps.
     // Added by plan #230.
     PluginDependencyMissing,
+    // Transient arena exhausted — allocate() was called when the arena had
+    // insufficient capacity for the requested (bytes, align) pair.
+    // Returned by TransientArena::allocate() (plan #239).
+    // Callers should fall back to a larger arena or defer the allocation.
+    TransientArenaExhausted,
 };
 }  // namespace core
 
--- a/core/include/glibre/error.hpp
+++ b/core/include/glibre/error.hpp
@@ -70,12 +70,15 @@ enum class Error : std::uint16_t {
     // Returned by TransientArena::allocate() (plan #239).
     // Callers should fall back to a larger arena or defer the allocation.
     TransientArenaExhausted,
-    // Invalid argument supplied to a core API (e.g. null pointer where a valid
-    // pointer is required).  Returned instead of asserting-only so that
-    // callers in release builds receive an actionable error rather than silent
-    // undefined behaviour.  Added by plan #239 review round 1 (MED-3 fix:
-    // null-pointer silent success in register_transient_arena).
-    InvalidArgument,
+    // Null pointer supplied to a core API that requires a valid, non-null
+    // pointer.  Returned instead of asserting-only so that callers in release
+    // builds receive an actionable error rather than silent undefined behaviour.
+    // Kept narrow (null-pointer check only) so it does not become a catch-all
+    // for unrelated argument errors.  Added by plan #239 review round 1
+    // (MED-3 fix: null-pointer silent success in register_transient_arena).
+    // Renamed NullArgument (was InvalidArgument) in review round 2 (LOW-3:
+    // narrow the seam name so it is not a magnet for unrelated callers).
+    NullArgument,
 };
 }  // namespace core
 
--- a/core/include/glibre/transient_arena.hpp
+++ b/core/include/glibre/transient_arena.hpp
@@ -40,7 +40,8 @@
 
 #include <cstddef>
 #include <cstdint>
-#include <memory>
+
+#include <EASTL/unique_ptr.h>
 
 #include "glibre/error.hpp"
 
@@ -55,7 +56,7 @@ public:
     // Construct an arena with `capacity_bytes` of backing storage.
     //
     // The backing store is allocated once on construction via
-    // `std::make_unique<std::byte[]>`.  No further allocation occurs for
+    // `eastl::make_unique<std::byte[]>`.  No further allocation occurs for
     // the lifetime of this object.  capacity_bytes == 0 is valid: every
     // allocate() call will return TransientArenaExhausted.
     explicit TransientArena(std::size_t capacity_bytes);
@@ -120,7 +121,7 @@ public:
     [[nodiscard]] glibre::Result<void> assert_drained() const noexcept;
 
 private:
-    std::unique_ptr<std::byte[]> storage_;
+    eastl::unique_ptr<std::byte[]> storage_;
     std::size_t capacity_{0};
     std::size_t cursor_{0};
     std::size_t high_watermark_{0};
--- a/core/include/glibre/transient_arena.hpp
+++ b/core/include/glibre/transient_arena.hpp
@@ -1,0 +1,129 @@
+#pragma once
+// core/include/glibre/transient_arena.hpp
+//
+// glibre::TransientArena — per-context bump-pointer transient allocator.
+//
+// Design (perf-budget.md §Allocator Rules #4):
+//   Each bounded context owns one TransientArena.  Allocations are O(1)
+//   bump-pointer (no header, no free-list, no destructor tracking).
+//   Callers MUST only store trivially-destructible data in the arena.
+//   The arena is drained (cursor reset to zero) by the frame loop at the
+//   end of Phase::Present (phase 9).  Drain is always O(1).
+//
+//   High-watermark is tracked across drain cycles so the perf HUD can
+//   display peak transient usage without polling mid-frame.
+//
+//   Allocation failure (capacity exhausted) returns
+//   std::unexpected{core::Error::TransientArenaExhausted} rather than
+//   aborting, so callers can fall back gracefully in release builds and
+//   the CI diagnostic build catches budget violations via REQUIRE checks.
+//
+// API:
+//   TransientArena arena{256 * 1024};       // 256 KiB backing store
+//   Result<void*> p = arena.allocate(64, 16);
+//   arena.drain();                           // called by FrameLoop at phase 9
+//   size_t peak = arena.high_watermark();   // bytes used since last drain
+//
+// Thread safety:
+//   None.  Each context owns one arena instance; callers are responsible
+//   for ensuring single-threaded access within a frame.
+//
+// Placement:
+//   The arena lives in core/ (SPEC §4, perf-budget.md §Allocator Rules)
+//   because the frame loop is the drainer and both reside in core.  Per-
+//   context *ownership* is enforced by callers (each plugin obtains a ref
+//   to its arena from the future PerContextAllocator handle — plan #238);
+//   in the MVP skeleton the arena is constructed and held directly.
+//
+// -fno-exceptions clean.  No heap allocation inside this class beyond
+// the one backing-store allocation in the constructor.
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+
+#include "glibre/error.hpp"
+
+namespace glibre {
+
+// ---------------------------------------------------------------------------
+// TransientArena — O(1) bump-pointer allocator, O(1) drain
+// ---------------------------------------------------------------------------
+
+class TransientArena {
+public:
+    // Construct an arena with `capacity_bytes` of backing storage.
+    //
+    // The backing store is allocated once on construction via
+    // `std::make_unique<std::byte[]>`.  No further allocation occurs for
+    // the lifetime of this object.  capacity_bytes == 0 is valid: every
+    // allocate() call will return TransientArenaExhausted.
+    explicit TransientArena(std::size_t capacity_bytes);
+
+    // Non-copyable, non-movable.  Arenas are long-lived, context-scoped
+    // objects; moving them would dangle pointers held by callers.
+    TransientArena(const TransientArena&) = delete;
+    TransientArena& operator=(const TransientArena&) = delete;
+    TransientArena(TransientArena&&) = delete;
+    TransientArena& operator=(TransientArena&&) = delete;
+
+    ~TransientArena() noexcept = default;
+
+    // allocate(bytes, align) — bump-allocate `bytes` aligned to `align`.
+    //
+    // `align` must be a power of two in [1, 4096].  Violation is a
+    // debug-build assert; release builds produce undefined behaviour
+    // (consistent with raw allocator contracts in game engines).
+    //
+    // Returns:
+    //   Result<void*>  — pointer to the aligned region on success, or
+    //   std::unexpected{core::Error::TransientArenaExhausted} when the
+    //   arena has insufficient capacity for the requested allocation.
+    //
+    // bytes == 0 is defined: returns a valid, uniquely-aligned pointer
+    // with no bytes committed (the cursor does not advance past the
+    // alignment pad if any).
+    [[nodiscard]] glibre::Result<void*>
+    allocate(std::size_t bytes, std::size_t align = alignof(std::max_align_t)) noexcept;
+
+    // drain() — reset the arena to empty.
+    //
+    // Called by FrameLoop at the end of Phase::Present (phase 9).
+    // Advances high_watermark_ if bytes_used() > high_watermark_.
+    // Sets cursor_ to 0.  O(1); does NOT zero the backing store.
+    void drain() noexcept;
+
+    // bytes_used() — bytes committed since the last drain().
+    [[nodiscard]] std::size_t bytes_used() const noexcept { return cursor_; }
+
+    // bytes_capacity() — total backing-store bytes available.
+    [[nodiscard]] std::size_t bytes_capacity() const noexcept { return capacity_; }
+
+    // high_watermark() — peak bytes_used() observed across all drain cycles
+    // since construction (or since the last explicit reset_high_watermark()).
+    [[nodiscard]] std::size_t high_watermark() const noexcept { return high_watermark_; }
+
+    // reset_high_watermark() — zero the high-watermark counter.
+    // Intended for perf-HUD tools that want per-second peak windows rather
+    // than all-time peaks.  Not called by the frame loop.
+    void reset_high_watermark() noexcept { high_watermark_ = 0; }
+
+    // assert_drained() — CI / diagnostic check that the arena has been drained.
+    //
+    // Returns success if bytes_used() == 0 (drain was called).
+    // Returns core::Error::OutOfBudget with detail "transient arena leak" if
+    // bytes_used() > 0 (allocations are still live past the expected drain point).
+    //
+    // This method implements perf-budget.md §Allocator Rules #4's "drain failure"
+    // contract: allocations that survive past phase 9 are reported as OutOfBudget.
+    // Called by CI diagnostic builds; not called in shipping builds.
+    [[nodiscard]] glibre::Result<void> assert_drained() const noexcept;
+
+private:
+    std::unique_ptr<std::byte[]> storage_;
+    std::size_t capacity_{0};
+    std::size_t cursor_{0};
+    std::size_t high_watermark_{0};
+};
+
+}  // namespace glibre
--- a/core/include/glibre/transient_arena.hpp
+++ b/core/include/glibre/transient_arena.hpp
@@ -39,7 +39,6 @@
 // the one backing-store allocation in the constructor.
 
 #include <cstddef>
-#include <cstdint>
 
 #include <EASTL/unique_ptr.h>
 
--- a/core/src/alloc/transient_arena.cpp
+++ b/core/src/alloc/transient_arena.cpp
@@ -40,8 +40,10 @@ TransientArena::TransientArena(std::size_t capacity_bytes)
 [[nodiscard]] glibre::Result<void*>
 TransientArena::allocate(std::size_t bytes, std::size_t align) noexcept {
     // align must be a power of two in [1, 4096].
-    assert(align >= 1u && align <= 4096u && (align & (align - 1u)) == 0u &&
-           "TransientArena::allocate: align must be a power-of-two in [1, 4096]");
+    assert(
+        align >= 1u && align <= 4096u && (align & (align - 1u)) == 0u &&
+        "TransientArena::allocate: align must be a power-of-two in [1, 4096]"
+    );
 
     // Align the cursor up to the requested alignment using standard bit trick.
     // cursor_aligned = (cursor_ + align - 1) & ~(align - 1).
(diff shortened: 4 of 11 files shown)


core/include/glibre/transient_arena.hpp
```diff
@@ -0,0 +1,129 @@
...
+//   context *ownership* is enforced by callers (each plugin obtains a ref
+//   to its arena from the future PerContextAllocator handle — plan #238);
+//   in the MVP skeleton the arena is constructed and held directly.
+//
+// -fno-exceptions clean.  No heap allocation inside this class beyond
+// the one backing-store allocation in the constructor.
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
```
cjhowe-us: severity:HIGH location:core/include/glibre/transient_arena.hpp:43 problem:`#include <memory>` and the `std::unique_ptr<std::byte[]> storage_` member at line 123 violate PHILOSOPHY §11, which mandates EASTL for all smart pointers in engine/runtime code (`std::` is retained only for the explicit allow-list: `std::expected`, `std::format`, `std::chrono`, `std::filesystem`, `std::thread/mutex/atomic`, `std::source_location`, `std::span`, type-traits, `std::move`/`std::forward`). `std::unique_ptr` is not on that list. The dispatch lens explicitly calls out "EASTL substrate" as a round-1 check. fix:Replace `std::unique_ptr<std::byte[]> storage_` with `eastl::unique_ptr<std::byte[]>` (or `eastl::vector<std::byte>` sized once at construction) and the `std::make_unique` call site in transient_arena.cpp:22 with the EASTL equivalent. Drop `#include <memory>`; add `#include <EASTL/unique_ptr.h>` (or `<EASTL/vector.h>`).

core/src/frame_loop.cpp
```diff
@@ -73,7 +108,14 @@ FrameLoop::run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept {
...
     case Phase::HotReload: /* core (barrier) — MVP empty */
         break;
-    case Phase::Present: /* platform — MVP empty */
+    case Phase::Present: /* platform — drains transient arenas at frame end */
+        // Drain all registered transient arenas at end of frame.
+        // perf-budget.md §Allocator Rules #4: transient arenas must be
+        // drained by phase 9 so they do not count against context ceilings.
+        // drain() is O(1) per arena; no heap allocation occurs.
+        for (std::size_t i = 0; i < arena_count_; ++i) {
+            arenas_[i]->drain();
```
cjhowe-us: severity:HIGH location:core/src/frame_loop.cpp:111-119 problem:perf-budget.md §Allocator Rules #4 contractually requires that "drain failure (allocations leaking past phase 9) is a `core::Error::OutOfBudget` arm with a leak detail string; CI gate has a debug-build assertion." `assert_drained()` exists on TransientArena (transient_arena.hpp:120) but FrameLoop never calls it. The current Phase::Present body unconditionally calls `drain()`, which always succeeds (it is just a cursor reset), so a "leak past phase 9" is structurally undetectable. The diagnostic check the spec promises is not wired. fix:Before the `arenas_[i]->drain()` call, run `auto check = arenas_[i]->assert_drained()` in `#ifndef NDEBUG` (or behind a `GLIBRE_ALLOC_STRICT` macro per perf-budget.md §Allocator Rules #2) and propagate the resulting OutOfBudget error through `tick()`. Add a Catch2 test that registers an arena, allocates without explicitly draining, runs `tick()`, and asserts the FrameLoop returns `core::Error::OutOfBudget` with detail "transient arena leak" in the diagnostic build configuration. (Today the test at tests/core/transient_arena_test.cpp:162 covers `assert_drained()` directly but never exercises it through the FrameLoop, which is where perf-budget §4 says it must fire.)

core/src/frame_loop.cpp
```diff
@@ -7,16 +7,48 @@
...
+// ---------------------------------------------------------------------------
+
+[[nodiscard]] glibre::Result<void>
+FrameLoop::register_transient_arena(glibre::TransientArena* arena) noexcept {
+    // Null pointer is a precondition violation; assert in debug builds.
+    // In release builds we skip silently rather than crash, to avoid undefined
+    // behaviour at phase 9 drain.  Callers must ensure the pointer is valid.
+    assert(arena != nullptr && "register_transient_arena: null arena pointer");
+    if (arena == nullptr) {
+        return {};  // skip silently in release; debug assert fires above
```
cjhowe-us: severity:MED location:core/src/frame_loop.cpp:34-37 problem:`register_transient_arena()` returns `{}` (success) on a null pointer in release builds. Combined with the assert that fires only in debug, the contract becomes "in release, callers may pass null and get a success response while nothing is actually registered" — that masks bugs and produces silent drain-misses at phase 9. fix:Either (a) drop the silent-skip and let the caller see the precondition violation as undefined behaviour (consistent with the doc-comment "Callers must ensure pointer validity"), or (b) return a real error such as `core::Error::OutOfBudget` with detail "null arena pointer" so callers cannot mistake the null path for a successful registration. Pick one; do not return success for an unregistered arena.

core/src/frame_loop.cpp
```diff
@@ -73,7 +108,14 @@ FrameLoop::run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept {
...
-    case Phase::Present: /* platform — MVP empty */
+    case Phase::Present: /* platform — drains transient arenas at frame end */
+        // Drain all registered transient arenas at end of frame.
+        // perf-budget.md §Allocator Rules #4: transient arenas must be
+        // drained by phase 9 so they do not count against context ceilings.
+        // drain() is O(1) per arena; no heap allocation occurs.
+        for (std::size_t i = 0; i < arena_count_; ++i) {
+            arenas_[i]->drain();
+        }
         break;
```
cjhowe-us, line 162: severity:MED location:core/src/frame_loop.cpp:111-119 problem:Phase 9 (Present) is owned by `platform` per reviews/decisions/frame-phases.md (Phase Table row 9). Putting `core`-owned bookkeeping (transient-arena drain) directly inside the Phase::Present case body in core/src/frame_loop.cpp inverts the SRP framing of the decision record, which says "each phase has exactly one owning context… preserves SRP: a phase has one reason to change, located in one bounded context." The drain is a frame-end barrier (more like phase 8 hot-reload, which is core-owned) and arguably belongs to its own seam. fix:Either (a) extend the decision record (or open a small amendment spike) to explicitly carve out "transient-arena drain runs at phase 9 boundary, owned by core, before platform present body executes" — making the inversion deliberate; or (b) move the drain to a `post_phase()` hook executed after each phase body for `Phase::Present`, called from `tick()` rather than `run_phase()`, so the platform-owned phase body and the core-owned drain barrier are two separate seams. Prefer (a) for the MVP — it is cheaper than restructuring run_phase — but record the carve-out in frame-phases.md.

tests/core/transient_arena_test.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    // Zero-capacity arena: every allocate() call must fail immediately.
+    glibre::TransientArena zero_arena{0};
+    auto r3 = zero_arena.allocate(1, 1);
+    REQUIRE_FALSE(r3.has_value());
+
+    const auto* zero_err = eastl::get_if<glibre::core::Error>(&r3.error().code());
+    REQUIRE(zero_err != nullptr);
+    CHECK(*zero_err == glibre::core::Error::TransientArenaExhausted);
+}
```
cjhowe-us, line 311: severity:LOW location:tests/core/transient_arena_test.cpp:277-303 problem:`transient_arena_exhaustion_returns_error` does not assert that `cursor_`/`bytes_used()` is unchanged on the failure path. Today the implementation correctly leaves the cursor untouched when allocate() returns the error, so a smaller follow-up alloc would still succeed — but that contract is not pinned by a test, so a future refactor could regress it silently. fix:After the failed `arena.allocate(1, 1)` on line 287, `CHECK(arena.bytes_used() == 128u)` to lock in "failed allocation does not advance cursor", then attempt a smaller allocation that fits in the remaining padding (here zero, so this requires constructing a slightly larger arena to leave a hole) and assert it succeeds.

core/include/glibre/transient_arena.hpp
```diff
@@ -0,0 +1,129 @@
...
+    // (consistent with raw allocator contracts in game engines).
+    //
+    // Returns:
+    //   Result<void*>  — pointer to the aligned region on success, or
+    //   std::unexpected{core::Error::TransientArenaExhausted} when the
+    //   arena has insufficient capacity for the requested allocation.
+    //
+    // bytes == 0 is defined: returns a valid, uniquely-aligned pointer
+    // with no bytes committed (the cursor does not advance past the
+    // alignment pad if any).
```
cjhowe-us, line 85: severity:LOW location:core/include/glibre/transient_arena.hpp:83-85 problem:Doc-comment says "bytes == 0 is defined: returns a valid, uniquely-aligned pointer with no bytes committed (the cursor does not advance past the alignment pad if any)" but no test exercises `allocate(0, ...)`. Behaviour at the cursor==capacity boundary with bytes==0 is also untested (the implementation returns `storage_+capacity_` which is one-past-end — well-defined as a pointer but not as a dereferenceable region). fix:Add a small `TEST_CASE` that calls `allocate(0, 1)` on a fresh arena, asserts `has_value()`, and asserts `bytes_used() == 0`. Optionally also call it on a full arena and confirm it still succeeds (since no bytes are committed).

tests/core/transient_arena_test.cpp
```diff
@@ -0,0 +1,351 @@
...
+//   - transient_arena_alignment_respected
+//   - transient_arena_exhaustion_returns_error
+//
+// Design constraints:
+//   - -fno-exceptions (error-model.md §Decision 3).
+//   - No REQUIRE_THROWS usage.
+//   - All alignment checks use pointer arithmetic and bitwise tests.
+
+#include <array>
+#include <bit>
```
cjhowe-us: severity:LOW location:tests/core/transient_arena_test.cpp:26 problem:`#include <bit>` and `#include <array>` are listed but only `<array>` is actually used (line 250 alignments table). `<bit>` is unused; `<cstdint>` is used implicitly via `std::uintptr_t`. Tightening includes makes the dependency surface explicit. fix:Drop `#include <bit>`. Keep `<cstdint>`, `<array>`, `<cstddef>`. Trivial cleanup, not blocking.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:3
carryover:0
notes:Round-1 lens (coverage + correctness). Diff scope matches plan #239 and the dispatch checklist; Conventional-Commit subject is correct; PR body contains `Closes #239`; issue #239's DoD block is populated and every `unit_test_named` assertion has a matching `TEST_CASE` in `tests/core/transient_arena_test.cpp` (verified all 8 names). The bump-pointer math is correct (power-of-two mask, capacity-overflow guarded by the size_t-safe form `bytes > capacity_ - cursor_aligned`), exhaustion returns the right error arm, and the new `core::Error::TransientArenaExhausted` enumerator is registered in both `error_register_test.cpp` and `log_error_test.cpp` with matching counts (15 → 16). Two HIGH issues block: (1) `std::unique_ptr<std::byte[]>` for `storage_` violates PHILOSOPHY §11 — the EASTL substrate rule is on the round-1 lens and there is no allow-list exception for smart pointers; (2) FrameLoop calls `drain()` (always succeeds) but never `assert_drained()`, so the perf-budget.md §Allocator Rules #4 "drain failure → OutOfBudget with leak detail" contract is structurally undetectable through the integration path — the `assert_drained` API exists in the header and is unit-tested in isolation, but no FrameLoop-level test exercises a leak through `tick()`. Two MED items: `register_transient_arena()` returns success on a null pointer in release builds (silent precondition failure), and the drain logic lives inside the platform-owned Phase::Present case body which inverts the SRP framing of frame-phases.md (record an explicit core-owned barrier carve-out, or split via a `post_phase()` hook). LOW items are unused includes, missing `bytes==0` test, and a missing "failed alloc does not advance cursor" assertion. PR-body claim about pre-existing `tick_invokes_phases_in_strict_order` failure on main is plausible (the PR does not touch frame_loop_test.cpp logic) but should be confirmed by CI, not the PR author. Recommend impl-respond fix the two HIGH items before round 2; MED items can be folded into the same revision or addressed in round 2.

core/include/glibre/transient_arena.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:ac165ca

Replaced `std::unique_ptr<std::byte[]> storage_` with `eastl::unique_ptr<std::byte[]>` in `transient_arena.hpp` and replaced `#include <memory>` with `#include <EASTL/unique_ptr.h>` in both the header and `core/src/alloc/transient_arena.cpp`. Constructor now calls `eastl::make_unique<std::byte[]>(capacity_bytes)` (EASTL does provide `eastl::make_unique` for unbounded array types via the same `make_unique<T[]>(n)` form as C++14 std::). PHILOSOPHY §11 is satisfied.

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:58f8378

Added `arena->assert_drained()` BEFORE `arena->drain()` in the Phase::Present loop, gated by `#ifdef GLIBRE_ALLOC_STRICT` (zero cost in unstrict builds; CI diagnostic builds add `-DGLIBRE_ALLOC_STRICT=1`). On failure the error (core::Error::OutOfBudget with detail "transient arena leak") propagates immediately through `tick()` via `std::unexpected`. Added test `undrained_allocation_through_tick_returns_out_of_budget` which registers an arena, allocates without draining, calls `tick()`, and asserts OutOfBudget with the correct detail string under GLIBRE_ALLOC_STRICT (SUCCEED in non-strict builds). perf-budget.md §Allocator Rules #4 contract is now exercised through the full integration path.

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:58f8378

register_transient_arena() now returns core::Error::InvalidArgument (with detail string) immediately on null input in all build configurations. The silent-success-in-release path is gone. The assert() call is removed (no longer needed; the error return is the correct mechanism). Header doc comment updated to list InvalidArgument as a return case alongside OutOfBudget. New test transient_arena_register_null_returns_invalid_argument verifies the error arm and that arena_count() is unchanged.

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:DEFER reasoning:The reviewer is correct that Phase 9 is platform-owned per frame-phases.md Phase Table row 9. The drain barrier is core-owned bookkeeping running in that slot, creating an SRP inversion. However, the fix requires either a frame-phases.md amendment or a post_phase() hook registry (which depends on plan #245 — register_system API). Both are out of scope for this PR. proposal:Opened spike #984 [SPIKE] iterate-frame-phases-core-barrier-carveout. Added an explicit doc comment block in Phase::Present naming the carve-out and referencing the spike, so the inversion is discoverable. The MVP code is functionally correct; the SRP concern will be resolved by #984 before the platform context lands its Phase::Present body.

tests/core/transient_arena_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:58f8378

Added cursor capture before the failing allocation in transient_arena_exhaustion_returns_error:
  const std::size_t cursor_before_fail = arena.bytes_used();
  ... failing alloc ...
  CHECK(arena.bytes_used() == cursor_before_fail);
Also added the symmetric cursor-unchanged check for the zero-capacity arena path (CHECK(zero_arena.bytes_used() == 0u) after the failed alloc).

core/include/glibre/transient_arena.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:58f8378

Added test transient_arena_zero_byte_allocation which verifies:
(a) allocate(0) returns a non-error Result.
(b) bytes_used() does not advance beyond the alignment pad.
(c) allocate(0, 16) on a misaligned cursor returns a pointer aligned to 16.
(d) bytes_used() equals the alignment-padded cursor (16 in the test case) + 0 committed bytes.
(e) A subsequent non-zero allocation succeeds.

tests/core/transient_arena_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:58f8378

Removed `#include <bit>` from tests/core/transient_arena_test.cpp. The `<array>` include is retained (used by the alignments array in transient_arena_alignment_respected).

core/src/alloc/transient_arena.cpp
```diff
@@ -0,0 +1,98 @@
...
+//   3. Return a pointer to storage_[cursor_after_align].
+//   4. Advance cursor_ to cursor_after_align + bytes.
+//
+// Alignment requirement: align must be a power of two.
+// ---------------------------------------------------------------------------
+
+[[nodiscard]] glibre::Result<void*>
+TransientArena::allocate(std::size_t bytes, std::size_t align) noexcept {
+    // align must be a power of two in [1, 4096].
+    assert(align >= 1u && align <= 4096u && (align & (align - 1u)) == 0u &&
```
cjhowe-us: severity:HIGH location:core/src/alloc/transient_arena.cpp:43 problem:CI lint job (clang-format --dry-run --Werror) FAILED on this commit; transient_arena.hpp/cpp, frame_loop.cpp, and transient_arena_test.cpp all carry clang-format violations introduced by this PR (see actions/runs/25604166834 lint job). PR mergeStateStatus is UNSTABLE. fix:Run clang-format -i over the new + edited files in a follow-up commit; verify locally with the same invocation as ci (clang-format --dry-run --Werror $(git ls-files '*.cpp' '*.hpp')) before pushing.

core/src/frame_loop.cpp
```diff
@@ -73,7 +113,33 @@ FrameLoop::run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept {
...
+        //   platform-owned Phase::Present slot.  This is a deliberate
+        //   "core barrier carve-out" that must be documented and eventually
+        //   formalised as a post_phase() hook or a frame-phases.md §Phase 9
+        //   amendment.  See [SPIKE] iterate-frame-phases-core-barrier-carveout.
+        for (std::size_t i = 0; i < arena_count_; ++i) {
+#ifdef GLIBRE_ALLOC_STRICT
+            auto check = arenas_[i]->assert_drained();
+            if (!check) {
+                return std::unexpected(std::move(check.error()));
+            }
```
cjhowe-us, line 159: severity:MED location:core/src/frame_loop.cpp:134-142 problem:Under GLIBRE_ALLOC_STRICT, when assert_drained() on arenas_[i] returns OutOfBudget the loop returns immediately. The leaking arena AND every subsequent arena in the registry are left undrained; the leaking arena's cursor stays non-zero (verified by the new integration test). On the next tick() the same OutOfBudget reappears and no arena ever drains again — a strict-mode frame loop is unrecoverable after one leak. This may be intentional (preserve the leak for inspection) but the contract is unstated and the SRP framing of phase 9 ("drain all arenas at frame end") is broken whenever strict mode triggers. fix:Either (a) drain all arenas first, then aggregate-report the first leak, or (b) document explicitly in the doc comment + a SPEC §4 note that strict mode tick() is a one-shot diagnostic — once it fires, the FrameLoop is poisoned and the test harness is expected to stop calling tick(). Pick one; the current code is ambiguous between the two.

core/include/glibre/error.hpp
```diff
@@ -65,6 +65,17 @@ enum class Error : std::uint16_t {
...
+    // insufficient capacity for the requested (bytes, align) pair.
+    // Returned by TransientArena::allocate() (plan #239).
+    // Callers should fall back to a larger arena or defer the allocation.
+    TransientArenaExhausted,
+    // Invalid argument supplied to a core API (e.g. null pointer where a valid
+    // pointer is required).  Returned instead of asserting-only so that
+    // callers in release builds receive an actionable error rather than silent
+    // undefined behaviour.  Added by plan #239 review round 1 (MED-3 fix:
+    // null-pointer silent success in register_transient_arena).
+    InvalidArgument,
```
cjhowe-us: severity:LOW location:core/include/glibre/error.hpp:78 problem:core::Error::InvalidArgument is a very broad enumerator name added to handle one specific seam (null arena pointer in register_transient_arena). The name will become a magnet for every future generic-precondition error in the core context, eroding the per-context-error-arm convention (error-model.md §Decision 2 favours specific arms over generic ones — note PluginManifestNotFound vs PluginManifestInvalid rather than a single PluginInputError). fix:Rename to TransientArenaRegisterNullPointer (or keep InvalidArgument but add a comment forbidding its reuse for unrelated callers without a follow-up amendment). Defer to follow-up if rename ripples are too wide for r2.

core/include/glibre/transient_arena.hpp
```diff
@@ -0,0 +1,130 @@
...
+//   because the frame loop is the drainer and both reside in core.  Per-
+//   context *ownership* is enforced by callers (each plugin obtains a ref
+//   to its arena from the future PerContextAllocator handle — plan #238);
+//   in the MVP skeleton the arena is constructed and held directly.
+//
+// -fno-exceptions clean.  No heap allocation inside this class beyond
+// the one backing-store allocation in the constructor.
+
+#include <cstddef>
+#include <cstdint>
```
cjhowe-us: severity:LOW location:core/include/glibre/transient_arena.hpp:42 problem:#include <cstdint> is unused in this header (no fixed-width integer types are referenced; std::size_t comes from <cstddef>). Same dead include flagged in r1 LOW-7 — verify whether the round-1 cleanup actually removed all unused includes from the header set. fix:Remove <cstdint> from transient_arena.hpp (and similarly verify the matching include in the .cpp is still needed; the .cpp uses std::byte from <cstddef> and no <cstdint> types).

cjhowe-us (commented):
findings:
  HIGH:1
  MED:1
  LOW:2
carryover:0
notes:Round-2 lens (cohesion + SRP + seam quality). R1 carry-over fully addressed: HIGH-1 (eastl::unique_ptr swapped in commit ac165ca; PHILOSOPHY §11 satisfied), HIGH-2 (assert_drained wired into Phase::Present under GLIBRE_ALLOC_STRICT with new integration test undrained_allocation_through_tick_returns_out_of_budget), MED-3 (register_transient_arena now returns core::Error::InvalidArgument for null pointer in all build configs with dedicated test), MED-4 (deferred to spike #984 with proper template body and bidirectional refs in the doc comment), LOW-5/6/7 all addressed. Error-registry consistency holds: core::Error grew from 15 to 17 enumerators (+TransientArenaExhausted, +InvalidArgument); kCoreErrorValues array size template = 17 in error_register_test.cpp; log_error.hpp switch has both new arms; log_error_test.cpp lines 209/211 cover both; kExpectedArmCount remains 3 (Variant alternatives, not enum count). EASTL substrate is consistent: eastl::unique_ptr / eastl::variant / eastl::string_view / eastl::get_if used throughout. ONE NEW HIGH blocks: CI lint job (clang-format --dry-run --Werror) FAILED on commit 58f8378 across all four touched files (transient_arena.hpp/cpp, frame_loop.cpp, transient_arena_test.cpp); PR mergeStateStatus is UNSTABLE. ONE MED: under GLIBRE_ALLOC_STRICT, the first arena leak short-circuits the drain loop and leaves all subsequent registered arenas undrained — strict-mode tick() is unrecoverable after one leak; either drain-then-aggregate-report or document the one-shot-diagnostic contract. TWO LOWs: core::Error::InvalidArgument is too generic a name for one specific seam (will become a magnet for unrelated callers) and unused <cstdint> include lingers in transient_arena.hpp. SRP/seam quality otherwise good: Phase::Present carve-out is documented and pointed at #984 for resolution; the InvalidArgument seam is correct (returns expected, not asserts). Recommend impl-respond fix the lint failure (HIGH) and choose one of the two MED fixes; LOWs can fold into the same revision or defer.

core/src/alloc/transient_arena.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:eae597b

Ran clang-format -i on all 4 PR-touched files (transient_arena.hpp, transient_arena.cpp, frame_loop.cpp, transient_arena_test.cpp) and also on the additionally-modified files (error.hpp, log_error.hpp, error_register_test.cpp, log_error_test.cpp). All pass clang-format --dry-run --Werror with clang-format 22.1.3.

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:eae597b

Changed Phase::Present GLIBRE_ALLOC_STRICT drain loop to drain-then-aggregate pattern: every arena is now checked via assert_drained() and then unconditionally drained; the first OutOfBudget error is captured in first_leak_error and returned after all arenas are processed. This guarantees all arenas are empty after tick() even in strict-mode, making the diagnostic non-destructive to subsequent frame correctness. New test all_arenas_drained_even_when_first_arena_leaks verifies two arenas (both leaking) are both empty after tick() fails with OutOfBudget. Existing undrained_allocation_through_tick_returns_out_of_budget updated to assert bytes_used() == 0 after tick() under the new contract.

core/include/glibre/error.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:eae597b

Renamed core::Error::InvalidArgument to core::Error::NullArgument across all references: error.hpp enumerator (updated comment to clarify it is only valid for null-pointer precondition checks in core APIs), log_error.hpp to_string arm, frame_loop.cpp register_transient_arena() null-check return, error_register_test.cpp kCoreErrorValues array entry and comment, log_error_test.cpp tostring_complete_for_core_error check, transient_arena_test.cpp null-register assertion and test comment. The narrower name prevents it from becoming a magnet for unrelated argument-validation callers.

core/include/glibre/transient_arena.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:eae597b

Removed #include <cstdint> from transient_arena.hpp. std::size_t is provided by <cstddef> (already present); no fixed-width integer types (uint8_t, uint16_t, etc.) are used in this header. clang-format --dry-run --Werror confirms the file is clean post-removal.

core/include/glibre/core/frame_loop.hpp
```diff
@@ -18,10 +20,12 @@
 
 #include <array>
```
cjhowe-us, line 21: severity:HIGH location:core/include/glibre/core/frame_loop.hpp:21 problem:CI lint job failed on commit eae597b — clang-format reports `<array>`/`<cstdint>`/`<cstddef>`/`<span>` include block is unsorted in this file. This is a CARRYOVER from round-2 HIGH-1 (the fix was incomplete; the commit message claimed `--dry-run --Werror` passed but GitHub Actions disagrees — see https://github.com/cjhowe-us/glibre/actions/runs/25604428965/job/75163773349). PR mergeStateStatus is UNSTABLE. fix:Run `clang-format -i core/include/glibre/core/frame_loop.hpp core/include/glibre/transient_arena.hpp core/src/alloc/transient_arena.cpp core/src/frame_loop.cpp tests/core/transient_arena_test.cpp` and re-run `clang-format --dry-run --Werror` against `git ls-files` to confirm. Re-include order should follow the `.clang-format` IncludeBlocks rule (sorted within each block).

core/include/glibre/core/frame_loop.hpp
```diff
@@ -67,6 +95,11 @@ class FrameLoop {
     // Incremented only after all nine phases succeed.
     [[nodiscard]] std::uint64_t frame_index() const noexcept { return frame_index_; }
 
+    // transient_arena_count() — number of registered transient arenas.
+    [[nodiscard]] std::size_t transient_arena_count() const noexcept {
```
cjhowe-us: severity:HIGH location:core/include/glibre/core/frame_loop.hpp:99 problem:CI lint also reports the `transient_arena_count()` inline body (lines 99-100) violates clang-format. Likely the body should fit on one line `{ return arena_count_; }` per project style (matching `frame_index()` immediately above). Same root cause as the include-block finding — same CARRYOVER, fix together. fix:Format the entire file via clang-format -i; verify `transient_arena_count()` collapses to single-line inline body to match `frame_index()`.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:0
carryover:1
notes:Round-3 lens (topology + spec alignment + polish). Convergence almost achieved but BLOCKED by an unresolved round-2 carryover. R2 fixes verified: HIGH-1 attempted (commit eae597b ran clang-format on PR-touched files but the edit was incomplete for `core/include/glibre/core/frame_loop.hpp` — 3 violations remain at lines 21 and 99-100; CI lint job FAILED on eae597b: https://github.com/cjhowe-us/glibre/actions/runs/25604428965/job/75163773349 and PR mergeStateStatus is UNSTABLE). MED-2 drain-then-aggregate logic is correct (verified manually: `first_leak_error` defaults to success-state; condition `(!check && first_leak_error.has_value())` correctly captures only the first leak; all arenas drain unconditionally; `if (!first_leak_error.has_value()) return std::unexpected(...)` returns the captured error path). New test `all_arenas_drained_even_when_first_arena_leaks` is the right shape (registers 2 arenas, both leak, asserts both arenas empty after tick() returns OutOfBudget). LOW-3 NullArgument rename consistent across error.hpp / log_error.hpp / error_register_test.cpp (size 17) / log_error_test.cpp / frame_loop.cpp / transient_arena_test.cpp. LOW-4 dead `<cstdint>` removed from transient_arena.hpp. Topology / spec alignment is sound: spike #984 is OPEN with the right title `[SPIKE] iterate-frame-phases-core-barrier-carveout`; carve-out comment in Phase::Present points at it bidirectionally; perf-budget.md §Allocator Rules #4 ("Drain failure ... is a `core::Error::OutOfBudget` arm with a 'leak' detail string") is honoured by `assert_drained()` returning `OutOfBudget` with `detail = "transient arena leak"` (verified). frame-phases.md still names phase 9 as platform-owned with platform Phase::Present body; the carve-out comment in frame_loop.cpp explicitly flags this as a deliberate core-barrier carve-out and points at #984 as the formal-resolution path (correct topology). DoD block in #239 has all 8 unit_test_named entries with matching TEST_CASE names verified against tests/core/transient_arena_test.cpp; `Closes #239` present in PR body; Conventional-Commit subject correct; `pts:3` correct on issue. ONE HIGH carryover blocks: lint CI must pass on the merge SHA. Recommend impl-respond run `clang-format -i core/include/glibre/core/frame_loop.hpp` (and re-verify the other four touched files for safety) and push a fix-up commit. With that single fix, this PR converges.

PR #988, opened 2026-05-09
feat(core): perf-budget framework — per-context CPU/GPU/heap counters (refs #241)

### LINKED ISSUE #241: [PLAN] feat(core): perf-budget framework — per-context CPU/GPU/heap counters

cjhowe-us:
## Summary

Implements per-context perf-budget tracking framework. Records CPU
ns, GPU ns (stub), heap bytes per `ContextTag`. Per-frame reset at
Phase::Present (phase 9). Foundation for the BENCHMARK_CELL CI gate
and S1 sample-scene fixture.

- `core/include/glibre/perf_budget.hpp` — `ContextTag` enum + `PerfBudgetSample` + `PerfBudget` class
- `core/src/perf_budget.cpp` — atomic record_cpu / record_gpu / record_heap_alloc / record_heap_free / sample / reset
- `core/include/glibre/core/frame_loop.hpp` — `set_perf_budget()` + `perf_budget_` member
- `core/src/frame_loop.cpp` — Phase::Present calls `perf_budget_->reset()` when non-null
- `tests/core/perf_budget/{CMakeLists.txt,perf_budget_test.cpp}` — 4 named Catch2 tests

## Design notes

- `ContextTag` enum defined in `perf_budget.hpp` (the 10 MVP contexts from perf-budget.md)
- All counter mutations use `std::atomic<uint64_t>` with `relaxed` ordering; join/phase-end provides the required barrier
- `record_heap_free()` uses a CAS loop for saturating subtract (no unsigned wrap)
- `set_perf_budget()` on `FrameLoop` accepts a nullable pointer; null = no-op (backward-compatible)
- GPU counter is a stub accumulator; real Metal timing from `MTLCounterSampleBuffer` is a separate render plan

## Refs

Closes #241

## Test plan

- [x] `ctest -R perf_budget_ -V` in `build/macos-debug` → 4/4 passed
  - `perf_budget_records_cpu_time_per_context` — PASSED (8 assertions)
  - `perf_budget_records_heap_bytes_per_context` — PASSED (9 assertions)
  - `perf_budget_reset_at_phase_9` — PASSED (44 assertions)
  - `perf_budget_concurrent_records_thread_safe` — PASSED (3 assertions)

Note: `core/frame_loop: tick_invokes_phases_in_strict_order` is a pre-existing failure on `main` unrelated to this PR (GLIBRE_TESTING macro not propagated to library compilation unit — separate issue).

### DIFF
--- a/core/include/glibre/core/frame_loop.hpp
+++ b/core/include/glibre/core/frame_loop.hpp
@@ -12,6 +12,8 @@
 //     is similarly empty at this skeleton stage).
 //   - At Phase::Present (9), drain all registered TransientArena instances
 //     (perf-budget.md §Allocator Rules #4, plan #239).
+//   - At Phase::Present (9), reset the registered PerfBudget if one has
+//     been set via set_perf_budget() (perf-budget.md §CI Gate Spec, plan #241).
 //   - No dynamic allocation inside tick().
 //   - -fno-exceptions clean; noexcept throughout the public surface.
 //
@@ -25,6 +27,7 @@
 
 #include "glibre/core/frame_phase.hpp"
 #include "glibre/error.hpp"
+#include "glibre/perf_budget.hpp"
 #include "glibre/transient_arena.hpp"
 
 namespace glibre::core {
@@ -75,6 +78,18 @@ public:
     [[nodiscard]] glibre::Result<void>
     register_transient_arena(glibre::TransientArena* arena) noexcept;
 
+    // set_perf_budget() — register a PerfBudget for phase-9 reset.
+    //
+    // The pointer may be null (default) to opt out of budget reset.
+    // When non-null, tick() calls perf_budget->reset() at the end of
+    // Phase::Present (9) before frame_index_ is incremented.
+    //
+    // The budget pointer must remain valid for the lifetime of this FrameLoop.
+    // Callers are responsible for ensuring pointer validity.
+    //
+    // Thread safety: must be called before tick() begins.
+    void set_perf_budget(glibre::PerfBudget* budget) noexcept { perf_budget_ = budget; }
+
     // tick() — advance one engine frame.
     //
     // Walks all nine phases in numeric order.  In debug builds a
@@ -112,6 +127,10 @@ private:
     // Non-owning pointers; lifetimes are caller-managed.
     std::array<glibre::TransientArena*, kMaxTransientArenas> arenas_{};
     std::size_t arena_count_{0};
+
+    // Optional PerfBudget — reset() called at end of Phase::Present (9).
+    // Non-owning pointer; lifetime is caller-managed.  Null = no-op.
+    glibre::PerfBudget* perf_budget_{nullptr};
 
 #ifdef GLIBRE_TESTING
     // Under GLIBRE_TESTING builds the last tick's phase execution order is
--- a/core/include/glibre/core/frame_loop.hpp
+++ b/core/include/glibre/core/frame_loop.hpp
@@ -78,16 +78,29 @@ public:
     [[nodiscard]] glibre::Result<void>
     register_transient_arena(glibre::TransientArena* arena) noexcept;
 
-    // set_perf_budget() — register a PerfBudget for phase-9 reset.
+    // set_perf_budget() — register (or replace, or detach) a PerfBudget for
+    // phase-9 reset.
     //
-    // The pointer may be null (default) to opt out of budget reset.
-    // When non-null, tick() calls perf_budget->reset() at the end of
-    // Phase::Present (9) before frame_index_ is incremented.
+    // Rebind contract:
+    //   - May be called any number of times provided no tick() is currently
+    //     executing on another thread.  The new pointer takes effect on the
+    //     next tick() call; the prior pointer is immediately ignored (not
+    //     deleted — lifetime is caller-managed).
+    //   - Passing nullptr detaches the budget: subsequent tick() calls skip
+    //     the phase-9 reset entirely.
+    //   - Rebinding between two non-null pointers is legal and useful in
+    //     tests that wish to swap budget fixtures between frames.
     //
-    // The budget pointer must remain valid for the lifetime of this FrameLoop.
-    // Callers are responsible for ensuring pointer validity.
+    // When non-null, tick() calls budget->reset() at the START of
+    // Phase::Present (9) bookkeeping — before transient-arena drain — so that
+    // counters are zeroed even if a leak error is returned.
     //
-    // Thread safety: must be called before tick() begins.
+    // The budget pointer must remain valid from the moment it is passed here
+    // until the next set_perf_budget() call (with null or another pointer) or
+    // until the FrameLoop is destroyed, whichever comes first.  Callers are
+    // responsible for ensuring pointer validity.
+    //
+    // Thread safety: must not be called concurrently with tick().
     void set_perf_budget(glibre::PerfBudget* budget) noexcept { perf_budget_ = budget; }
 
     // tick() — advance one engine frame.
--- a/core/include/glibre/core/frame_loop.hpp
+++ b/core/include/glibre/core/frame_loop.hpp
@@ -101,7 +101,7 @@ public:
     // responsible for ensuring pointer validity.
     //
     // Thread safety: must not be called concurrently with tick().
-    void set_perf_budget(glibre::PerfBudget* budget) noexcept { perf_budget_ = budget; }
+    void set_perf_budget(glibre::PerfBudget* budget) noexcept;
 
     // tick() — advance one engine frame.
     //
@@ -133,6 +133,17 @@ private:
     // that might supply phases in a different order than kPhaseTable.
     [[nodiscard]] glibre::Result<void>
     run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept;
+
+    // present_reset_perf_budget() — step (1) of Phase::Present bookkeeping.
+    // Resets per-frame perf-budget counters unconditionally if a budget is set.
+    // Extracted for SRP (perf-budget.md §CI Gate Spec, plan #241).
+    void present_reset_perf_budget() noexcept;
+
+    // present_drain_arenas() — steps (2) & (3) of Phase::Present bookkeeping.
+    // Drains all registered transient arenas; in GLIBRE_ALLOC_STRICT builds
+    // asserts each arena was empty before drain and returns the first leak error.
+    // Extracted for SRP (perf-budget.md §Allocator Rules #4, plan #239).
+    [[nodiscard]] glibre::Result<void> present_drain_arenas() noexcept;
 
     std::uint64_t frame_index_{0};
 
--- a/core/include/glibre/perf_budget.hpp
+++ b/core/include/glibre/perf_budget.hpp
@@ -78,9 +78,9 @@ inline constexpr std::size_t kContextTagCount = 10u;
 // ---------------------------------------------------------------------------
 
 struct PerfBudgetSample {
-    std::uint64_t cpu_ns{0};      // accumulated CPU nanoseconds this frame
-    std::uint64_t gpu_ns{0};      // accumulated GPU nanoseconds (stub; 0 until Metal plan)
-    std::uint64_t heap_bytes{0};  // live heap bytes at snapshot time
+    std::uint64_t cpu_ns{0};     // accumulated CPU nanoseconds this frame
+    std::uint64_t gpu_ns{0};     // accumulated GPU nanoseconds (stub; 0 until Metal plan)
+    std::uint64_t heap_bytes{0}; // live heap bytes at snapshot time
 };
 
 // ---------------------------------------------------------------------------
--- a/core/include/glibre/perf_budget.hpp
+++ b/core/include/glibre/perf_budget.hpp
@@ -59,12 +59,19 @@ enum class ContextTag : std::uint8_t {
     Physics,
     Content,
     Tools,
-    E2E,           // test-only context; budget not enforced in shipping builds
-    Count_         // sentinel — not a valid tag; used only for array sizing
+    E2E,  // test-only context; no shipping-build budget ceiling
+          // (perf-budget.md §Budget Table: e2e row is n/a for all cells;
+          //  perf-budget.md §Rationale: encoding n/a as a row rather than
+          //  gating behind GLIBRE_TESTING avoids ODR violations from the
+          //  library being compiled without GLIBRE_TESTING while tests are).
 };
 
-inline constexpr std::size_t kContextTagCount =
-    static_cast<std::size_t>(ContextTag::Count_);
+// kContextTagCount — number of valid ContextTag entries.
+//
+// Kept as a free constexpr rather than a Count_ enumerator inside the enum so
+// that an out-of-range integer cast to ContextTag cannot silently alias a
+// sentinel value (pattern from frame_phase.hpp §kPhaseCount).
+inline constexpr std::size_t kContextTagCount = 10u;
 
 // ---------------------------------------------------------------------------
 // PerfBudgetSample — snapshot of one context's per-frame counters
@@ -103,11 +110,15 @@ public:
     // at the end of the frame via sample().
     void record_cpu(ContextTag tag, std::uint64_t ns) noexcept;
 
-    // record_gpu(tag, ns) — add `ns` nanoseconds to the GPU counter for `tag`.
+    // record_gpu(tag, ns) — STUB.  Does NOT capture real GPU time.
     //
-    // Stub in this plan.  Real values come from MTLCounterSampleBuffer
-    // (render plan, out of scope here).  Callers may record 0 or
-    // passthrough values from the render context.
+    // Real GPU nanoseconds come from MTLCounterSampleBuffer, which is
+    // implemented by the render plan (out of scope for plan #241).  Until
+    // that plan lands, callers may pass through 0 or placeholder values.
+    // Follow-up search tag: [GPU-COUNTER-STUB] — grep for this tag when
+    // wiring up MTLCounterSampleBuffer to replace this stub.
+    //
+    // Accumulates `ns` into the GPU counter for `tag`.
     // Thread-safe: same atomics contract as record_cpu.
     void record_gpu(ContextTag tag, std::uint64_t ns) noexcept;
 
--- a/core/include/glibre/perf_budget.hpp
+++ b/core/include/glibre/perf_budget.hpp
@@ -37,8 +37,8 @@
 
 #include <array>
 #include <atomic>
-#include <cstdint>
 #include <cstddef>
+#include <cstdint>
 
 namespace glibre {
 
@@ -78,9 +78,9 @@ inline constexpr std::size_t kContextTagCount = 10u;
 // ---------------------------------------------------------------------------
 
 struct PerfBudgetSample {
-    std::uint64_t cpu_ns{0};     // accumulated CPU nanoseconds this frame
-    std::uint64_t gpu_ns{0};     // accumulated GPU nanoseconds (stub; 0 until Metal plan)
-    std::uint64_t heap_bytes{0}; // live heap bytes at snapshot time
+    std::uint64_t cpu_ns{0};      // accumulated CPU nanoseconds this frame
+    std::uint64_t gpu_ns{0};      // accumulated GPU nanoseconds (stub; 0 until Metal plan)
+    std::uint64_t heap_bytes{0};  // live heap bytes at snapshot time
 };
 
 // ---------------------------------------------------------------------------
--- a/core/include/glibre/perf_budget.hpp
+++ b/core/include/glibre/perf_budget.hpp
@@ -1,0 +1,157 @@
+#pragma once
+// core/include/glibre/perf_budget.hpp
+//
+// glibre::PerfBudget — per-context CPU/GPU/heap performance counter store.
+//
+// Design (perf-budget.md §Decision, §CI Gate Spec):
+//   Records CPU nanoseconds (sim + submit), GPU nanoseconds (stub — real
+//   capture from MTLCounterSampleBuffer lands in a later render plan), and
+//   heap live bytes per ContextTag.  Counters are atomic so any context
+//   can record from its owning thread without external synchronisation.
+//
+//   The FrameLoop calls reset() at the end of Phase::Present (phase 9)
+//   to zero all per-frame counters before the next frame's work begins.
+//
+//   Intended use in MVP:
+//     PerfBudget budget;
+//     budget.record_cpu(ContextTag::Physics, elapsed_ns);
+//     budget.record_heap_alloc(ContextTag::Render, bytes);
+//     PerfBudgetSample s = budget.sample(ContextTag::Physics);
+//     // ... at frame end: FrameLoop calls budget.reset()
+//
+//   Thread safety:
+//     record_* and sample() are safe to call from multiple threads
+//     concurrently (atomics with relaxed ordering for accumulators;
+//     see §Concurrency note below).
+//     reset() is called by the frame loop on the driver thread and must
+//     not race with concurrent record_* calls.  Callers must ensure that
+//     all recording in phases 1-8 is complete before phase 9 resets.
+//     (This is guaranteed by the FrameLoop's sequential phase walk.)
+//
+//   PHILOSOPHY §11: std::atomic is a language/runtime utility that EASTL
+//   does not own; it is retained here per the PHILOSOPHY §11 carve-out for
+//   std::atomic, std::thread, and std::mutex.  EASTL containers / strings
+//   are used where applicable; std::vector/string are not used here.
+//
+// -fno-exceptions clean.  No heap allocation after construction.
+
+#include <array>
+#include <atomic>
+#include <cstdint>
+#include <cstddef>
+
+namespace glibre {
+
+// ---------------------------------------------------------------------------
+// ContextTag — bounded-context identifiers matching perf-budget.md table
+//
+// These values index into PerfBudget's counter rows.  The enum is closed;
+// adding a context requires an amendment to perf-budget.md.
+// ---------------------------------------------------------------------------
+
+enum class ContextTag : std::uint8_t {
+    Core = 0,
+    Platform,
+    Data,
+    Shader,
+    Render,
+    Geometry,
+    Physics,
+    Content,
+    Tools,
+    E2E,           // test-only context; budget not enforced in shipping builds
+    Count_         // sentinel — not a valid tag; used only for array sizing
+};
+
+inline constexpr std::size_t kContextTagCount =
+    static_cast<std::size_t>(ContextTag::Count_);
+
+// ---------------------------------------------------------------------------
+// PerfBudgetSample — snapshot of one context's per-frame counters
+// ---------------------------------------------------------------------------
+
+struct PerfBudgetSample {
+    std::uint64_t cpu_ns{0};     // accumulated CPU nanoseconds this frame
+    std::uint64_t gpu_ns{0};     // accumulated GPU nanoseconds (stub; 0 until Metal plan)
+    std::uint64_t heap_bytes{0}; // live heap bytes at snapshot time
+};
+
+// ---------------------------------------------------------------------------
+// PerfBudget — per-context accumulating counter store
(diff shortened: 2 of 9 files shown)


core/src/frame_loop.cpp
```diff
@@ -159,6 +163,14 @@ FrameLoop::run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept {
             }
 #endif
         }
+        // Reset per-frame perf-budget counters.
+        // perf-budget.md §CI Gate Spec (plan #241): counters are reset at the
+        // end of each frame so the next frame starts with zeroed accumulators.
+        // No-op when perf_budget_ is null (opt-out path for code that does not
+        // use the budget framework yet).
+        if (perf_budget_ != nullptr) {
```
cjhowe-us: severity:MED location:core/src/frame_loop.cpp:171 problem:In GLIBRE_ALLOC_STRICT builds, when a transient-arena leak is captured the function returns std::unexpected from the arena-drain block (line ~162) before reaching the perf_budget_->reset() call below. The frame fails (frame_index_ not incremented) but per-frame counters carry over into the next attempted frame, polluting the budget read by tests / CI gate. fix:Hoist the perf_budget_->reset() call ABOVE the strict-mode early return (or invert: do reset() unconditionally first, then drain arenas). Document the order in the comment block — reset is per-frame bookkeeping that must always run at end of phase 9 regardless of leak detection.

tests/core/perf_budget/perf_budget_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+//   - Spawn N threads (e.g. 8); each calls record_cpu(Physics, delta) M times.
+//   - Join all threads.
+//   - Assert sample(Physics).cpu_ns == N * M * delta.
+//
+// The test does NOT race reset() with record_cpu — that ordering is
+// guaranteed by the FrameLoop's sequential phase walk in production.
+// The concurrent-recording path tested here is the common case: multiple
+// contexts recording on their owning threads during phases 1-8.
+// ===========================================================================
+TEST_CASE("perf_budget_concurrent_records_thread_safe", "[core][perf_budget]") {
```
cjhowe-us, line 203: severity:MED location:tests/core/perf_budget/perf_budget_test.cpp:199 problem:perf_budget_concurrent_records_thread_safe only exercises record_cpu (fetch_add). The most race-sensitive path — record_heap_free with its CAS-retry saturating subtract — is only tested single-threaded. A bug in the CAS loop (e.g. wrong memory_order, broken retry semantics, or a sign-extension regression on saturation) would not be caught. fix:Add a second concurrent block that interleaves record_heap_alloc and record_heap_free across N threads with a known net-zero delta (M allocs of B bytes from threads 0..N/2, M frees of B bytes from threads N/2..N) and assert sample(tag).heap_bytes == 0 after join. Optionally also assert non-wrapping under intentional over-free races.

tests/core/perf_budget/perf_budget_test.cpp
```diff
@@ -0,0 +1,234 @@
...
+    glibre::PerfBudget budget;
+
+    constexpr int kThreads = 8;
+    constexpr int kRecordsPerThread = 1000;
+    constexpr std::uint64_t kDeltaNs = 100u;
+
+    // Use std::thread (PHILOSOPHY §11 carve-out — std::thread retained for
+    // concurrency; EASTL does not own a thread primitive).
+    std::vector<std::thread> workers;
```
cjhowe-us: severity:MED location:tests/core/perf_budget/perf_budget_test.cpp:208 problem:Test uses std::vector<std::thread>. PHILOSOPHY §11 mandates eastl:: containers for runtime data structures; std::vector is not on the carve-out list (only std::thread / std::atomic / std::mutex / std::expected etc are). Test code is still subject to the philosophy; allowing std::vector here normalises the pattern for downstream tests. fix:Replace std::vector<std::thread> with eastl::vector<std::thread>, or use a fixed std::array<std::thread, kThreads> since kThreads is a constexpr — eliminates the heap altogether and is the simpler change.

core/include/glibre/perf_budget.hpp
```diff
@@ -0,0 +1,157 @@
...
+    Platform,
+    Data,
+    Shader,
+    Render,
+    Geometry,
+    Physics,
+    Content,
+    Tools,
+    E2E,           // test-only context; budget not enforced in shipping builds
+    Count_         // sentinel — not a valid tag; used only for array sizing
```
cjhowe-us: severity:LOW location:core/include/glibre/perf_budget.hpp:63 problem:ContextTag::Count_ is part of the public enum surface. Casting an out-of-range integer or Count_ itself to ContextTag and passing it to record_* yields UB in release builds (assert is debug-only; rows_[idx] indexes past array end). fix:Move Count_ outside the enum to a separate constexpr — pattern matches frame_phase.hpp where kPhaseCount is a free constant rather than an enumerator. Or keep Count_ but consider compiling in a release-build bounds check (single branch, predicted not-taken, negligible cost on a counter recording path).

core/include/glibre/perf_budget.hpp
```diff
@@ -0,0 +1,157 @@
...
+    Core = 0,
+    Platform,
+    Data,
+    Shader,
+    Render,
+    Geometry,
+    Physics,
+    Content,
+    Tools,
+    E2E,           // test-only context; budget not enforced in shipping builds
```
cjhowe-us: severity:LOW location:core/include/glibre/perf_budget.hpp:62 problem:Including ContextTag::E2E in the shipping enum diverges from perf-budget.md §Allocator Rules #1, which lists exactly 9 tags (e2e excluded). The decision record explicitly says e2e is excluded because encoding a number for it would create a fictional ceiling for shipping builds where e2e is not present. The dispatch prompt sanctions 10 entries for this PR, so this is a documentation alignment issue rather than a code bug. fix:Either (a) gate ContextTag::E2E behind GLIBRE_TESTING / a test-only header so shipping builds cannot reference it, or (b) amend reviews/decisions/perf-budget.md in this PR or a follow-up to add the E2E row and reconcile the divergence. Without one of these, future readers of the decision record will be confused why the implementation has 10 tags.

core/include/glibre/perf_budget.hpp
```diff
@@ -0,0 +1,157 @@
...
+    // at the end of the frame via sample().
+    void record_cpu(ContextTag tag, std::uint64_t ns) noexcept;
+
+    // record_gpu(tag, ns) — add `ns` nanoseconds to the GPU counter for `tag`.
+    //
+    // Stub in this plan.  Real values come from MTLCounterSampleBuffer
+    // (render plan, out of scope here).  Callers may record 0 or
+    // passthrough values from the render context.
+    // Thread-safe: same atomics contract as record_cpu.
+    void record_gpu(ContextTag tag, std::uint64_t ns) noexcept;
```
cjhowe-us, line 123: severity:LOW location:core/include/glibre/perf_budget.hpp:112 problem:record_gpu is a pure stub — callers can record any value but no integration with MTLCounterSampleBuffer exists, and the function is not differentiated from record_cpu in behaviour. A consumer reading the header sees record_gpu and may assume it captures real GPU counters. fix:Either (a) make the stub status the FIRST line of the doc-comment (currently the stub note is line 3 of the comment), or (b) add a cross-reference to the future render plan that will populate this counter, so a follow-up agent can search and find it. The current comment ordering buries the stub disclaimer under the contract.

core/include/glibre/core/frame_loop.hpp
```diff
@@ -75,6 +78,18 @@ class FrameLoop {
...
+    //
+    // The pointer may be null (default) to opt out of budget reset.
+    // When non-null, tick() calls perf_budget->reset() at the end of
+    // Phase::Present (9) before frame_index_ is incremented.
+    //
+    // The budget pointer must remain valid for the lifetime of this FrameLoop.
+    // Callers are responsible for ensuring pointer validity.
+    //
+    // Thread safety: must be called before tick() begins.
+    void set_perf_budget(glibre::PerfBudget* budget) noexcept { perf_budget_ = budget; }
```
cjhowe-us: severity:LOW location:core/include/glibre/core/frame_loop.hpp:91 problem:set_perf_budget() takes a single nullable raw pointer with no documented contract for re-setting (e.g. is set_perf_budget(other) after a prior set allowed mid-lifetime? what about set(nullptr) to detach?). The doc says must be called before tick() begins but does not say exactly once or may be reset between ticks. fix:Tighten the contract: either (a) document that set_perf_budget may be called any number of times when no tick() is in flight and the prior pointer becomes immediately ignored, or (b) make it set-once and store a flag to assert in debug. Without this, callers writing tests that swap budgets between ticks have no contract to lean on.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:3
  LOW:4
carryover:0

notes:Coverage is solid — the four named DoD tests are present and exercise the cpu / heap / phase-9 reset / concurrent paths. The PR body contains the required `Closes #241`, ContextTag is the closed 10-entry enum, atomics use relaxed ordering with reset on the driver thread, and the saturating CAS subtract is correct (single-threaded test passes). FrameLoop integration via nullable `set_perf_budget()` is backward-compatible. No HIGH-severity blockers.

What blocks: nothing; this is a COMMENT verdict, not REQUEST_CHANGES. The MED items are worth resolving before merge:
  (1) phase-9 reset is skipped when GLIBRE_ALLOC_STRICT detects an arena leak — counters carry over;
  (2) the concurrent test does not race the CAS subtract, which is the most failure-prone path;
  (3) `std::vector<std::thread>` violates PHILOSOPHY §11 (use `eastl::vector` or `std::array`).
The LOW items are doc / hygiene:
  (4) `Count_` sentinel inside the enum risks UB on out-of-range cast;
  (5) `ContextTag::E2E` exists in the enum but `reviews/decisions/perf-budget.md` lists 9 tags — reconcile in a follow-up amendment;
  (6) `record_gpu` stub disclaimer should lead the doc-comment;
  (7) `set_perf_budget` rebind contract is under-specified.

Deferred: none; everything is in scope for a single follow-up commit on this PR.

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Hoisted perf_budget_->reset() to execute BEFORE the GLIBRE_ALLOC_STRICT early-return block. The reset now runs unconditionally at the start of phase-9 bookkeeping (step 1 of a documented three-step order: reset, drain, aggregate-leak-errors). Even when a transient-arena leak causes an early-return error, the budget counters are zeroed so frame N+1 starts clean. The comment block in frame_loop.cpp Phase::Present now documents the 1-2-3 order explicitly.

tests/core/perf_budget/perf_budget_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Added perf_budget_concurrent_heap_free_no_underflow test. 8 threads split into two halves: threads [0,4) call record_heap_alloc(Content, 256) 500 times; threads [4,8) call record_heap_free(Content, 256) 500 times — all racing concurrently. After join: asserts heap_bytes == 0 (net-zero) and heap_bytes < UINT64_MAX/2 (no wrap). This directly exercises the CAS-retry loop under contention where free threads race each other's compare_exchange_weak while alloc threads simultaneously fetch_add, forcing retries.

tests/core/perf_budget/perf_budget_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Replaced std::vector<std::thread> workers with std::array<std::thread, kThreads> workers. kThreads is constexpr so the array is statically sized — eliminates the heap allocation entirely. The <vector> include was removed from the test file; <array> was added. The same change was applied to both the original concurrent-cpu test and the new concurrent-heap test.

core/include/glibre/perf_budget.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Removed Count_ from the ContextTag enum entirely. kContextTagCount = 10u remains as a free inline constexpr (matching the frame_phase.hpp pattern with kPhaseCount). The debug-build assert in tag_index() still bounds-checks against kContextTagCount. This eliminates the UB risk from out-of-range casts aliasing the sentinel.

core/include/glibre/perf_budget.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Retained E2E in the shipping enum (not gated by GLIBRE_TESTING) to avoid ODR violations — the library is compiled without GLIBRE_TESTING while tests compile with it; a conditional row count would cause rows_[] to be sized 9 in the lib and sampled with index 9 in tests (verified: the GLIBRE_TESTING-gate approach caused SIGABRT on the reset test). Instead, amended perf-budget.md §Allocator Rules #1 to list all 10 tags and explicitly document that the e2e row carries no shipping-build ceiling and is unchecked by GLIBRE_ALLOC_STRICT. The enum comment block in perf_budget.hpp also cross-references this rationale.

core/include/glibre/perf_budget.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Restructured the record_gpu doc-comment to lead with 'STUB. Does NOT capture real GPU time.' as the first line. Added a [GPU-COUNTER-STUB] search tag so the follow-up render plan can grep for it when wiring up MTLCounterSampleBuffer. The full contract (what the function actually does: accumulate via fetch_add, thread-safe) follows the stub notice.

core/include/glibre/core/frame_loop.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8abb07a

Expanded set_perf_budget() doc-comment to document the full rebind contract: (a) may be called any number of times when no tick() is executing; (b) the new pointer takes effect on the next tick(), the prior pointer is immediately ignored but not deleted; (c) nullptr detaches — subsequent ticks skip phase-9 reset; (d) rebinding between two non-null pointers is explicitly legal (useful in tests that swap budget fixtures between frames). Pointer lifetime is also more precisely specified: valid until the next set_perf_budget() call or FrameLoop destruction, whichever first.

core/include/glibre/core/frame_loop.hpp
```diff
@@ -75,6 +78,31 @@ class FrameLoop {
...
+    // Phase::Present (9) bookkeeping — before transient-arena drain — so that
+    // counters are zeroed even if a leak error is returned.
+    //
+    // The budget pointer must remain valid from the moment it is passed here
+    // until the next set_perf_budget() call (with null or another pointer) or
+    // until the FrameLoop is destroyed, whichever comes first.  Callers are
+    // responsible for ensuring pointer validity.
+    //
+    // Thread safety: must not be called concurrently with tick().
+    void set_perf_budget(glibre::PerfBudget* budget) noexcept { perf_budget_ = budget; }
```
cjhowe-us: severity:MED location:core/include/glibre/core/frame_loop.hpp:104 problem:`set_perf_budget()` is the only public mutator on FrameLoop defined inline in the header; `register_transient_arena()` is out-of-line. Inlining the setter forces every TU that includes `frame_loop.hpp` to also bring in `glibre/perf_budget.hpp` for the pointer-type completeness in the field declaration — already a transitive include here, but it cements the coupling and asymmetry. fix:move the body to `frame_loop.cpp` for seam consistency with `register_transient_arena`. The contract (rebind / nullptr-detach / lifetime) stays in the header doc-comment; only the assignment moves.

core/src/perf_budget.cpp
```diff
@@ -0,0 +1,113 @@
...
+// Asserts in debug builds that the tag is in range.  ContextTag::Count_ is
+// not a valid caller-supplied value; callers must use named enumerators only.
+// ---------------------------------------------------------------------------
+
+namespace {
+[[nodiscard]] std::size_t tag_index(ContextTag tag) noexcept {
+    const auto idx = static_cast<std::size_t>(tag);
+    assert(
+        idx < kContextTagCount &&
+        "PerfBudget: ContextTag::Count_ is not a valid context; "
```
cjhowe-us: severity:LOW location:core/src/perf_budget.cpp:38 problem:assert message still references `ContextTag::Count_`, but the Count_ sentinel was removed in commit 8abb07a (R1 fix LOW-4) — message is now stale and misleading. fix:update to `"PerfBudget: ContextTag is out-of-range; caller supplied a raw cast value beyond kContextTagCount"`.

core/src/frame_loop.cpp
```diff
@@ -118,27 +122,34 @@ FrameLoop::run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept {
     case Phase::HotReload: /* core (barrier) — MVP empty */
         break;
     case Phase::Present: /* platform — drains transient arenas at frame end */
```
cjhowe-us, line 191: severity:LOW location:core/src/frame_loop.cpp:124 problem:Phase::Present case body now mixes three responsibilities (perf-budget reset, transient-arena drain, strict-mode leak aggregation) and is ~50 lines inside a switch. SRP-wise this is the documented "core barrier carve-out" — already flagged as a SPIKE — but the size of the case body itself reduces FrameLoop cohesion. fix:extract two private helpers `present_reset_perf_budget() noexcept` and `present_drain_arenas() noexcept -> Result<void>` so the switch case becomes three lines and the strict-mode aggregation lives next to the drain. Defers to the same SPIKE for the architectural carve-out, but the local refactor is in-scope here.

core/src/frame_loop.cpp
```diff
@@ -118,27 +122,34 @@ FrameLoop::run_phase(Phase phase, std::uint8_t expected_ordinal) noexcept {
...
+
+        // (1) Reset perf-budget counters unconditionally before leak detection.
+        if (perf_budget_ != nullptr) {
+            perf_budget_->reset();
+        }
+
+        // (2) & (3) Drain arenas and aggregate any strict-mode leak errors.
         {
 #ifdef GLIBRE_ALLOC_STRICT
             glibre::Result<void> first_leak_error{};  // holds first leak error (if any)
```
cjhowe-us: severity:MED location:core/src/frame_loop.cpp:155 problem:strict-mode aggregator works but the predicate at lines 160 and 168 reads inverted. `first_leak_error.has_value()` is being used to mean "no leak captured yet" (default-success state of Result<void>), the opposite of the conventional reading "we hold a leak". Future maintainers will read it as a bug. Lines 160/168 are not part of the R1 hunk but the strict-mode block was modified in this PR (reset reorder), so the issue is in scope. fix:store the captured-leak status in an explicit `bool leak_captured = false;` plus an `eastl::optional<glibre::Error> first_leak;` so the predicate becomes `if (!check && !leak_captured) { leak_captured = true; first_leak.emplace(std::move(check.error())); }` and the return reads `if (leak_captured) return std::unexpected(std::move(*first_leak));`.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:2
carryover:0

notes:Cohesion + SRP + seam-quality lens. R1 fixes are all solid: phase-9 reset is unconditional and ordered first, std::array replaces std::vector in tests, Count_ sentinel is gone, set_perf_budget rebind contract is documented, GPU stub disclaimer leads, perf-budget.md amendment lists all 10 ContextTag entries including E2E. The new concurrent_heap_free_no_underflow test exercises the CAS-retry path under contention and asserts both net-zero and non-wrapped — correct seam coverage.

Cohesion findings:
  (MED-1) Strict-mode leak aggregator at frame_loop.cpp:155–170 uses `first_leak_error.has_value()` to mean "no leak captured yet" — the predicate is correct but inverted relative to conventional Result<void> reading. Recommend a `bool leak_captured` plus `eastl::optional<Error>` companion. Touched in this PR by the reset reorder so it is in scope.
  (MED-2) `set_perf_budget()` is the only public mutator on FrameLoop defined inline in the header; `register_transient_arena` is out-of-line. Move the body to frame_loop.cpp for seam consistency.

LOW findings:
  (LOW-1) `tag_index` assert message in perf_budget.cpp:38 still references `ContextTag::Count_`, which was removed in commit 8abb07a — stale comment.
  (LOW-2) Phase::Present case body in run_phase has grown to ~50 lines mixing reset / drain / strict-mode aggregation. Extract to two private helpers (`present_reset_perf_budget`, `present_drain_arenas`) for local SRP. The architectural carve-out remains [SPIKE] iterate-frame-phases-core-barrier-carveout.

What blocks: nothing. None of these are merge blockers; orchestrator decides. R3 should focus on topology + spec alignment + DoD-block satisfaction for #241 and verify the perf-budget.md amendment is internally consistent.

Deferred: none — all four findings are mergeable as a single seam-cleanup commit on this PR or rolled into a follow-up if scope is preserved.

core/include/glibre/core/frame_loop.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dac5572

Moved set_perf_budget() body to frame_loop.cpp — out-of-line definition matching register_transient_arena's pattern. The assignment is one line; the full doc-comment and rebind contract remain in the header as before. No behaviour change.

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dac5572

Replaced the inverted Result<void>::has_value() sentinel with explicit `bool leak_captured = false` + `std::optional<glibre::Error> first_leak` in the extracted present_drain_arenas() helper. The predicate is now `if (!check && !leak_captured)` and the return is `if (leak_captured) return std::unexpected(std::move(*first_leak))`. Semantics are identical; readability is conventional.

core/src/perf_budget.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dac5572

Updated tag_index assert message: removed stale reference to ContextTag::Count_ (removed in commit 8abb07a). New message: "PerfBudget: ContextTag is out-of-range; caller supplied a raw cast value beyond kContextTagCount".

core/src/frame_loop.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:dac5572

Extracted present_reset_perf_budget() noexcept and present_drain_arenas() noexcept -> Result<void> as private helpers. Phase::Present case body is now three lines: present_reset_perf_budget(), the present_drain_arenas() call with early-return on error, and break. Strict-mode aggregation lives inside present_drain_arenas() next to the drain loop. The architectural core-barrier carve-out SPIKE remains deferred.

core/include/glibre/perf_budget.hpp
```diff
@@ -0,0 +1,168 @@
...
+//     (This is guaranteed by the FrameLoop's sequential phase walk.)
+//
+//   PHILOSOPHY §11: std::atomic is a language/runtime utility that EASTL
+//   does not own; it is retained here per the PHILOSOPHY §11 carve-out for
+//   std::atomic, std::thread, and std::mutex.  EASTL containers / strings
+//   are used where applicable; std::vector/string are not used here.
+//
+// -fno-exceptions clean.  No heap allocation after construction.
+
+#include <array>
```
cjhowe-us, line 38: severity:HIGH location:core/include/glibre/perf_budget.hpp:38 problem:CI lint job FAILED on latest commit dac5572 with 30+ clang-format violations across this PR's files (perf_budget.hpp lines 38,81,82,83; frame_loop.cpp lines 74,75; perf_budget.cpp lines 37,38,95,96; perf_budget_test.cpp lines 144-228). The aligned-comment / aligned-assignment style introduced in r2 commit dac5572 (e.g. `std::uint64_t cpu_ns{0};     // …` and `.cpu_ns     = row.cpu_ns.load(...)`) violates the project clang-format profile. DoD assertion `workflow_passed: ci.yml` for #241 cannot evaluate true while lint is failing, so dod-verify will block merge regardless of orchestrator verdict. fix:run `clang-format -i` on the four affected files (or remove the manual column-aligned padding spaces around `=`, `{0};` comments, and `==` in tests) and force-push; verify `gh pr checks 988` returns all-green before requesting another review round.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:0
carryover:0

notes:Topology + spec alignment + polish lens. The design substance is converged — perf-budget.md amendment now lists all 10 ContextTag entries (e2e n/a documented), Closes #241 keyword present, ContextTag is closed enum + free kContextTagCount constexpr (matches kPhaseCount pattern), set_perf_budget rebind contract documented, GPU stub disclaimer leads, Phase::Present case body decomposed into present_reset_perf_budget + present_drain_arenas helpers (local SRP restored), strict-mode aggregator now uses explicit bool leak_captured + optional<Error> first_leak (no inverted predicate), set_perf_budget moved out-of-line for seam consistency. Concurrent CAS-retry test covers the saturating subtract under contention. R1 + R2 findings are all resolved on dac5572; no carryover.

What blocks: CI gate on the latest commit. lint (clang-format --dry-run --Werror) FAILED on dac5572 with 30+ violations across all four files this PR introduces — column-aligned comments on PerfBudgetSample fields, column-aligned designated initializers in sample(), aligned == in tests, and a couple of trailing-space artefacts on the new set_perf_budget out-of-line definition. Per the round-3 critical CI gate, verdict cannot be APPROVE while any required check is non-green. DoD assertion `workflow_passed: ci.yml` (per issue #241) will block dod-verify on merge regardless. The fix is mechanical: `clang-format -i` on the four files (impl-respond can run it locally and force-push). Two CI checks (build, e2e-core, e2e-shader) currently SKIPPED because lint blocks the workflow — they need to run and pass before approval.

Deferred: none. Once lint goes green and the build / e2e jobs report SUCCESS on the new HEAD, this PR is ready to approve in a follow-up R3-rerun review (or the orchestrator can promote directly if it observes the green checks).

Issue #251, opened 2026-05-03
[PLAN] core/hot-reload: drain step


cjhowe-us:
### Domain
core

### Phase
phase:mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
TBD — pending `[SPIKE] draft-core-user-stories` (#29). Implements
step 1 of the four-step protocol in
`reviews/decisions/hot-reload-protocol.md`.

### Parent
#3

### Scope
Implement the drain step: the loader calls each outgoing plugin's
`glibre_plugin_drain(World&)` callback, ensures plugin-survivable
state lives in middleman-typed storages, and confirms no thread other
than the loader is inside plugin code afterward.

In scope:
- `core/include/glibre/core/plugin_abi.hpp` — declare the C ABI
  function pointer type for `glibre_plugin_drain` (signature per
  protocol §Step 1: `noexcept` returning `expected<void,
  glibre::Error>`).
- `core/src/hot_reload/drain.cpp` — `drain_plugin(const PluginRecord&
  outgoing, World& world) -> expected<void, glibre::Error>`. Looks up
  the drain symbol via the plugin's vtable, invokes it, propagates
  any unexpected as `core::Error::HotReloadRefused` (wrapping the
  inner cause).
- Document in the header that drain must (a) flush per-frame queues
  into middleman-typed components, (b) release plugin-internal GPU
  resources, (c) cancel + join plugin worker tasks.

Out of scope:
- Atomic vtable swap (next plan).
- Any actual GPU resource release — out of core's scope; we just call
  the plugin's drain, the plugin owns the cleanup.

### Unit Test Plan
- core/hot_reload_drain: drain_called_exactly_once_per_outgoing_plugin
- core/hot_reload_drain: drain_unexpected_returns_HotReloadRefused
- core/hot_reload_drain: drain_does_not_mutate_other_plugins
- core/hot_reload_drain: drain_failure_marks_request_refused_no_swap

### Depends on
- #249 (pending-reload counter)
- #250 (plugin manifest + ABI hash gate)

### Agent Execution Notes
Branch `feat/core-hot-reload-drain`. PR title:
`feat(core): hot-reload barrier — drain step`. Use a fake
`glibre-test-plugin.dylib` fixture under `tests/fixtures/plugins/`
so the drain function pointer is real but trivial.

PR #992, opened 2026-05-09
feat(core): hot-reload plugin manifest + ABI hash gate (refs #250)

### LINKED ISSUE #250: [PLAN] core/hot-reload: plugin manifest + ABI hash gate
### Domain
core

### Phase
phase:mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
TBD — pending `[SPIKE] draft-core-user-stories` (#29). Plan implements
the ABI-hash gate (PHILOSOPHY §9) and the plugin manifest declared in
`reviews/decisions/hot-reload-protocol.md` step 2.1–2.2.

### Parent
#3

### Scope
Define the plugin-manifest data type and the ABI-hash check invoked
both at first-load and at hot-reload swap candidate verification.

In scope:
- `core/include/glibre/core/plugin_manifest.hpp` — `struct
  PluginManifest { std::string_view fqn; u64 abi_hash;
  std::span<const TypeRegistration> types; };` plus
  `glibre_plugin_manifest()` ABI export name (resolved by `dlsym`).
- `core/include/glibre/core/abi_hash.hpp` —
  `host_glibre_types_abi_hash()` helper that returns the host-process
  middleman hash. Source: linked `glibre-types.dylib`'s
  `glibre_types_abi_hash()` symbol.
- `core/src/plugin_loader_actions.cpp` — `hot_reload_validate(outgoing, incoming)
  -> Result<void>` returning `core::Error::PluginAbiHashMismatch` on hash
  mismatch, `core::Error::PluginNameMismatch` on name mismatch, and
  `core::Error::HotReloadRefused` on major-version change.

Out of scope:
- `dlopen` / `dlsym` plumbing — covered in the swap-step plan.
- The middleman dylib itself — separate decision deferred to
  `fory-codegen` data-context work.

### Unit Test Plan
- hot_reload_rejects_abi_hash_mismatch
- hot_reload_accepts_compatible_swap
- hot_reload_rejects_name_mismatch
- hot_reload_rejects_major_version_change

### Depends on
None — uses `glibre::Error` types from the cross-cutting error-model
chain (epic #2, sub-epic #4 family). No intra-#3 blocker.

### Agent Execution Notes
Branch `feat/core-hot-reload-manifest-abi-gate`. PR title:
`feat(core): hot-reload plugin manifest + ABI hash gate (refs #250)`.
hot_reload_validate is a free function in plugin_loader_actions.{hpp,cpp};
tests construct PluginManifest values in-memory — no dylib fixtures needed.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_contains:
    path: core/src/plugin_loader_actions.cpp
    regex: "hot_reload_validate"
- unit_test_named: hot_reload_rejects_abi_hash_mismatch
- unit_test_named: hot_reload_accepts_compatible_swap
- unit_test_named: hot_reload_rejects_name_mismatch
- unit_test_named: hot_reload_rejects_major_version_change
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

Implements the pre-swap hot-reload manifest validation gate described in
`reviews/decisions/hot-reload-protocol.md §"Step 2 — Swap"` (steps 2.1–2.2).

- Adds `hot_reload_validate(outgoing, incoming)` free function in `plugin_loader_actions.{hpp,cpp}` that enforces three sequential checks before allowing a hot-reload swap to proceed:
  - **Check A** (name identity): incoming plugin name must match outgoing name → `PluginNameMismatch`.
  - **Check B** (ABI hash equality, protocol §2.1): incoming `abi_hash` must equal outgoing `abi_hash` → `PluginAbiHashMismatch`.
  - **Check C** (SemVer major version compatibility, protocol §2.2): major version must not change across a hot-reload → `HotReloadRefused` (major bump requires a fresh world, not a swap).
- Adds `PluginNameMismatch` to `core::Error`, `log_error.hpp` `to_string`, and the error_register test array.
- Adds four Catch2 tests under `tests/core/hot_reload/` — all pass locally.

## Test plan

- [ ] `hot_reload_rejects_abi_hash_mismatch` — passes
- [ ] `hot_reload_accepts_compatible_swap` — passes (identical version, minor advance, patch advance)
- [ ] `hot_reload_rejects_name_mismatch` — passes
- [ ] `hot_reload_rejects_major_version_change` — passes (major increase and decrease)
- [ ] `error_register_per_context_arms_unique` — passes (kCoreErrorValues updated to size 18)
- [ ] `core/log_error: tostring_complete_for_core_error` — passes (PluginNameMismatch added)

Closes #250

### DIFF
--- a/core/include/glibre/core/plugin_loader_actions.hpp
+++ b/core/include/glibre/core/plugin_loader_actions.hpp
@@ -1,27 +1,35 @@
 #pragma once
 // core/include/glibre/core/plugin_loader_actions.hpp
 //
-// Loader-procedure free functions — steps 9–11 of the plugin loader sequence,
-// and hot-reload swap candidate validation.
+// Loader-procedure free functions — two axes of stateless loader-sequence work:
 //
-// Authority: reviews/decisions/plugin-abi.md §"Loader Sequence" steps 9–11,
-//            §"Failure Modes → core::Error" rows 9, 10, 11.
-//            reviews/decisions/hot-reload-protocol.md §"Step 2 — Swap" steps 2.1, 2.2.
+//   Axis 1 — Initial-load post-gate actions (steps 9–11 of the plugin loader
+//             sequence).  These run ONCE when a plugin is first loaded, after
+//             all four PluginLoaderRegistry gate checks (steps 4–7) pass.
+//             Authority: reviews/decisions/plugin-abi.md §"Loader Sequence"
+//             steps 9–11 and §"Failure Modes → core::Error" rows 9, 10, 11.
 //
-// These functions implement the post-gate actions that run after all four
-// PluginLoaderRegistry gate checks (steps 4–7) pass.  They are free functions
-// rather than PluginLoaderRegistry members because they do not read or mutate
-// the registry-of-records state (the loaded_ map and host_engine_version_).
-// Keeping them separate respects the SRP boundary that motivated splitting
-// PluginLoader (RAII handle + dlopen/dlsym, plan #229) from PluginLoaderRegistry
-// (gate validation + records, plan #230) in the first place.
+//   Axis 2 — Hot-reload pre-swap validation (hot-reload phase-8 step 2, plan
+//             #250).  This check runs EVERY reload cycle, after the drain step
+//             and before the vtable swap.  It inspects two loaded-plugin
+//             manifests and returns an error if the swap should be refused.
+//             Authority: reviews/decisions/hot-reload-protocol.md §"Step 2 —
+//             Swap" sub-steps 2.1 and 2.2.
 //
-// hot_reload_validate (plan #250) is a free function that sits in this header
-// because it is a pre-swap check that does not mutate registry state; it
-// inspects two loaded-plugin manifests and returns an error if the swap should
-// be refused.
+// SRP note: the two axes share this file because both expose stateless free
+// functions that do not read or mutate PluginLoaderRegistry state (the loaded_
+// map and host_engine_version_).  That "no-registry-state" boundary is the
+// single responsibility this file enforces.  The test directory split
+// (tests/core/plugin_loader/ for Axis 1, tests/core/hot_reload/ for Axis 2)
+// signals that the axes MAY be separated into parallel header/source pairs once
+// plans #251+ add type-superset checks and migration-arena work that would
+// further grow Axis 2's responsibility surface.
 //
-// Callers: PluginLoader (plan #229) orchestrates steps 1–11; these free
+// TODO(#251): evaluate splitting hot_reload_validate (and its successors from
+// plans #251+) into core/include/glibre/core/hot_reload_actions.hpp + .cpp to
+// match the test directory structure and keep each file on a single axis.
+//
+// Callers: PluginLoader (plan #229) orchestrates steps 1–11; the Axis 1
 //          functions are called after validate_all() succeeds and before
 //          register_plugin() records the newly loaded plugin.
 //
@@ -137,8 +145,8 @@ migrate_components(std::uint32_t from_version, std::uint32_t to_version) noexcep
 //
 // Three sequential checks, performed in the order below:
 //
-//   Check A — Name identity (new in plan #250; not a hot-reload-protocol step
-//             number but derived from PHILOSOPHY §3 "one dylib per domain"):
+//   Check A — Name identity (plan #250 configuration-error guard; no parent
+//             step in hot-reload-protocol.md or plugin-abi.md):
 //     incoming.name == outgoing.name.  Swapping for a plugin with a different
 //     name is a configuration error — the operator loaded the wrong dylib.
 //     Failure → core::Error::PluginNameMismatch.
@@ -165,9 +173,10 @@ migrate_components(std::uint32_t from_version, std::uint32_t to_version) noexcep
 //     the loader holds the exclusive phase-8 lock, but the belt-and-suspenders
 //     check is cheap and the error message is more specific.
 //
-//     The PRECONDITION is asserted in debug builds via GLIBRE_DCHECK inside
-//     hot_reload_validate (see plugin_loader_actions.cpp).  Call sites must
-//     not invoke this function without first calling validate_all.
+//     The PRECONDITION is asserted in debug builds via #ifndef NDEBUG /
+//     assert() inside hot_reload_validate (see plugin_loader_actions.cpp).
+//     Call sites must not invoke this function without first calling
+//     validate_all.
 //
 //     Failure → core::Error::PluginAbiHashMismatch.
 //
@@ -201,12 +210,13 @@ migrate_components(std::uint32_t from_version, std::uint32_t to_version) noexcep
 //   without touching ECS state.
 //
 // Preconditions (unchecked — caller must ensure):
-//   • outgoing and incoming are fully loaded (PluginLoader::open() succeeded).
-//   • Both loaders have valid manifests (manifest_result().has_value() == true).
+//   • outgoing and incoming are valid PluginManifest values previously
+//     extracted from PluginLoader instances whose validate_all() returned
+//     success.
 //   • This function is called during phase 8, after drain.
 //
-// @param outgoing  Currently-loaded plugin loader.
-// @param incoming  Candidate replacement plugin loader.
+// @param outgoing  Currently-loaded plugin's manifest.
+// @param incoming  Candidate replacement plugin's manifest.
 //
 // Returns:
 //   success (Result<void>{})                     — all checks passed; swap may proceed.
--- a/core/include/glibre/core/plugin_loader_actions.hpp
+++ b/core/include/glibre/core/plugin_loader_actions.hpp
@@ -1,10 +1,12 @@
 #pragma once
 // core/include/glibre/core/plugin_loader_actions.hpp
 //
-// Loader-procedure free functions — steps 9–11 of the plugin loader sequence.
+// Loader-procedure free functions — steps 9–11 of the plugin loader sequence,
+// and hot-reload swap candidate validation.
 //
 // Authority: reviews/decisions/plugin-abi.md §"Loader Sequence" steps 9–11,
 //            §"Failure Modes → core::Error" rows 9, 10, 11.
+//            reviews/decisions/hot-reload-protocol.md §"Step 2 — Swap" steps 2.1, 2.2.
 //
 // These functions implement the post-gate actions that run after all four
 // PluginLoaderRegistry gate checks (steps 4–7) pass.  They are free functions
@@ -13,6 +15,11 @@
 // Keeping them separate respects the SRP boundary that motivated splitting
 // PluginLoader (RAII handle + dlopen/dlsym, plan #229) from PluginLoaderRegistry
 // (gate validation + records, plan #230) in the first place.
+//
+// hot_reload_validate (plan #250) is a free function that sits in this header
+// because it is a pre-swap check that does not mutate registry state; it
+// inspects two loaded-plugin manifests and returns an error if the swap should
+// be refused.
 //
 // Callers: PluginLoader (plan #229) orchestrates steps 1–11; these free
 //          functions are called after validate_all() succeeds and before
@@ -23,7 +30,8 @@
 
 #include <cstdint>
 
-#include <glibre/core/plugin_api.hpp>  // RegisterFn, PluginContext
+#include <glibre/core/plugin_api.hpp>      // RegisterFn, PluginContext
+#include <glibre/core/plugin_manifest.hpp>  // PluginManifest, SemVer
 #include <glibre/error.hpp>
 
 namespace glibre::core {
@@ -116,5 +124,65 @@ namespace glibre::core {
 
 [[nodiscard]] Result<void>
 migrate_components(std::uint32_t from_version, std::uint32_t to_version) noexcept;
+
+// ---------------------------------------------------------------------------
+// hot_reload_validate — pre-swap compatibility check (plan #250).
+//
+// Authority: reviews/decisions/hot-reload-protocol.md §"Step 2 — Swap",
+//            sub-steps 2.1 and 2.2.
+//
+// Validates that an incoming plugin candidate (`incoming`) is a safe swap
+// for the currently-loaded plugin (`outgoing`).  This check runs during
+// phase 8 after the drain step and before the vtable swap (step 2.3).
+//
+// Three sequential checks, performed in the order below:
+//
+//   Check A — Name identity (new in plan #250; not a hot-reload-protocol step
+//             number but derived from PHILOSOPHY §3 "one dylib per domain"):
+//     incoming.name == outgoing.name.  Swapping for a plugin with a different
+//     name is a configuration error — the operator loaded the wrong dylib.
+//     Failure → core::Error::PluginNameMismatch.
+//
+//   Check B — ABI hash (hot-reload-protocol §2.1):
+//     incoming.abi_hash == outgoing.abi_hash.  Both must match the host
+//     `glibre_types_abi_hash()`.  In this function we compare the two manifests
+//     against each other; the PluginLoaderRegistry's validate_all() already
+//     confirmed both equal the host hash before the loader ever called
+//     hot_reload_validate, so manifest-vs-manifest equality is equivalent to
+//     both-vs-host equality.
+//     Failure → core::Error::PluginAbiHashMismatch.
+//
+//   Check C — SemVer major version (hot-reload-protocol §2.2, compatible-swap
+//             rule):
+//     The incoming plugin's semantic version major must equal the outgoing's.
+//     The minor/patch may advance (additive changes are safe); they may also
+//     stay the same (a patch rebuild).  A major-version change signals a
+//     breaking redesign that requires a fresh world, not a hot-reload.
+//     Failure → core::Error::HotReloadRefused.
+//
+// Note on type-superset check (protocol §2.2 second bullet):
+//   The protocol requires that incoming's component set is a superset-or-equal
+//   of outgoing's surviving component storages.  That check depends on the
+//   archetype storage and migration arena (plans #251+); it is NOT performed
+//   here.  This function's scope is the manifest-only checks that can run
+//   without touching ECS state.
+//
+// Preconditions (unchecked — caller must ensure):
+//   • outgoing and incoming are fully loaded (PluginLoader::open() succeeded).
+//   • Both loaders have valid manifests (manifest_result().has_value() == true).
+//   • This function is called during phase 8, after drain.
+//
+// @param outgoing  Currently-loaded plugin loader.
+// @param incoming  Candidate replacement plugin loader.
+//
+// Returns:
+//   success (Result<void>{})                     — all checks passed; swap may proceed.
+//   core::Error::PluginNameMismatch              — check A failed.
+//   core::Error::PluginAbiHashMismatch           — check B failed.
+//   core::Error::HotReloadRefused                — check C failed (major version change).
+// ---------------------------------------------------------------------------
+
+[[nodiscard]] Result<void>
+hot_reload_validate(const PluginManifest& outgoing, const PluginManifest& incoming) noexcept;
 
 }  // namespace glibre::core
--- a/core/include/glibre/core/plugin_loader_actions.hpp
+++ b/core/include/glibre/core/plugin_loader_actions.hpp
@@ -30,7 +30,7 @@
 
 #include <cstdint>
 
-#include <glibre/core/plugin_api.hpp>      // RegisterFn, PluginContext
+#include <glibre/core/plugin_api.hpp>       // RegisterFn, PluginContext
 #include <glibre/core/plugin_manifest.hpp>  // PluginManifest, SemVer
 #include <glibre/error.hpp>
 
--- a/core/include/glibre/core/plugin_loader_actions.hpp
+++ b/core/include/glibre/core/plugin_loader_actions.hpp
@@ -144,20 +144,53 @@ migrate_components(std::uint32_t from_version, std::uint32_t to_version) noexcep
 //     Failure → core::Error::PluginNameMismatch.
 //
 //   Check B — ABI hash (hot-reload-protocol §2.1):
-//     incoming.abi_hash == outgoing.abi_hash.  Both must match the host
-//     `glibre_types_abi_hash()`.  In this function we compare the two manifests
-//     against each other; the PluginLoaderRegistry's validate_all() already
-//     confirmed both equal the host hash before the loader ever called
-//     hot_reload_validate, so manifest-vs-manifest equality is equivalent to
-//     both-vs-host equality.
+//     incoming.abi_hash == outgoing.abi_hash.
+//
+//     Protocol §2.1 specifies the check as `incoming == host_glibre_types_abi_hash`
+//     (incoming-vs-host).  Here we compare incoming-vs-outgoing instead.  This
+//     is semantically equivalent under the following precondition:
+//
+//       PRECONDITION: PluginLoaderRegistry::validate_all() confirmed that BOTH
+//       manifests equal the host hash before the loader ever called
+//       hot_reload_validate.  If that precondition holds, then:
+//         incoming.abi_hash == host   (guaranteed by validate_all)
+//         outgoing.abi_hash == host   (guaranteed by validate_all)
+//         → incoming.abi_hash == outgoing.abi_hash   (transitively)
+//
+//     Comparing the two manifests against each other rather than against the
+//     host hash surfaces one additional failure mode: a bug where the two
+//     loaded manifests disagree with each other despite both nominally passing
+//     validate_all (e.g. a race that replaced outgoing's manifest between
+//     validate_all and this call).  In practice that race cannot occur because
+//     the loader holds the exclusive phase-8 lock, but the belt-and-suspenders
+//     check is cheap and the error message is more specific.
+//
+//     The PRECONDITION is asserted in debug builds via GLIBRE_DCHECK inside
+//     hot_reload_validate (see plugin_loader_actions.cpp).  Call sites must
+//     not invoke this function without first calling validate_all.
+//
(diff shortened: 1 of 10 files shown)


core/src/plugin_loader_actions.cpp
```diff
@@ -138,4 +141,88 @@ Result<void> migrate_components(
...
+            }
+        );
+    }
+
+    // Check B — ABI hash equality (hot-reload-protocol §2.1).
+    // The two manifests must carry identical abi_hash strings.  The
+    // PluginLoaderRegistry already confirmed each against the host hash
+    // independently; here we confirm they agree with each other (a belt-and-
+    // suspenders guard and a clear error message if they somehow diverge).
+    if (incoming.abi_hash != outgoing.abi_hash) {
```
cjhowe-us, line 238: severity:MED location:core/src/plugin_loader_actions.cpp:195 problem:Check B compares incoming.abi_hash against outgoing.abi_hash, but hot-reload-protocol.md §"Step 2 — Swap" 2.1 specifies the check is `Q::glibre_types_abi_hash() == host_glibre_types_abi_hash` (incoming-vs-host, not incoming-vs-outgoing). The header docstring justifies the swap by claiming `PluginLoaderRegistry::validate_all() already confirmed both equal the host hash` — but that is a transitive inference, and if the listed precondition (`Both loaders have valid manifests` validated against host) is ever violated by a future caller, two manifests carrying the same wrong hash would silently pass. fix:Either (a) take a `host_abi_hash` parameter and compare incoming.abi_hash against it directly per protocol §2.1, or (b) add a Catch2 test that documents the precondition by exercising the registry-validation chain end-to-end so the inference is mechanically guarded. At minimum, escalate the precondition note in the header from a comment to a `[[expects:...]]`-style runtime check (debug build).

core/src/plugin_loader_actions.cpp
```diff
@@ -138,4 +141,88 @@ Result<void> migrate_components(
...
+                    .detail = "incoming plugin abi_hash does not match outgoing plugin abi_hash",
+                },
+            }
+        );
+    }
+
+    // Check C — SemVer major version compatibility (hot-reload-protocol §2.2).
+    // A major-version change is a breaking redesign that cannot be hot-reloaded
+    // into a running world; it requires a process restart with a fresh world.
+    if (incoming.version.major != outgoing.version.major) {
```
cjhowe-us, line 268: severity:MED location:core/src/plugin_loader_actions.cpp:211 problem:Check C tests `incoming.version.major != outgoing.version.major`, but hot-reload-protocol.md §"Step 2 — Swap" 2.2 does NOT define a discrete SemVer-major equality check. Protocol 2.2 actually specifies a component-type-set superset check on `(fqn, schema_version)` — the phrase "major-version change" appears only as descriptive language about the *consequence* of dropping a registered type. The header docstring acknowledges the type-superset check is deferred to plans #251+, but in deferring it the implementation has *substituted* a different criterion (SemVer-major equality) that the decision doc never sanctions. The plan #250 issue body does request this shape, so it is in-scope for this PR — but it should be cited as a plan-level decision that *extends* the protocol, not as protocol §2.2. fix:Update the header docstring (Check C section, plugin_loader_actions.hpp ~L101–108) and the inline comment here to read "extends protocol §2.2 with a SemVer-major equality precondition required by plan #250" rather than claiming it derives from §2.2 verbatim. Optionally raise this divergence in a follow-up amendment to hot-reload-protocol.md or in §12 of the spec so the decision doc and the implementation agree.

core/src/plugin_loader_actions.cpp
```diff
@@ -138,4 +141,88 @@ Result<void> migrate_components(
...
+//   fresh world.
+//   Failure → core::Error::HotReloadRefused.
+// ---------------------------------------------------------------------------
+
+Result<void>
+hot_reload_validate(const PluginManifest& outgoing, const PluginManifest& incoming) noexcept {
+    // Check A — Name identity.
+    // Both plugins must declare the same name.  A different name means the
+    // operator supplied the wrong dylib as a swap candidate.
+    if (incoming.name != outgoing.name) {
```
cjhowe-us, line 199: severity:LOW location:core/src/plugin_loader_actions.cpp:177 problem:hot-reload-protocol.md §"Refusal Cases" stipulates that every refusal is wrapped in `core::Error::HotReloadRefused` with a nested cause arm so consumers see both the umbrella and the specific cause. This implementation returns the bare cause arm (`PluginNameMismatch` / `PluginAbiHashMismatch` / `HotReloadRefused`) without the umbrella. The plan #250 issue body intentionally chose this simpler shape (and `glibre::Error` uses an `eastl::variant` over per-context enums with no nesting facility), so this is documented divergence — not a bug — but it does mean a future caller cannot pattern-match `if (err == HotReloadRefused) ...` to discover all three refusal cases at once. fix:Either (a) add a §12 note in hot-reload-protocol.md acknowledging the plan-level departure (umbrella wrap deferred until the variant grows nesting), or (b) extend `core::Error` with a `HotReloadRefusedReason` companion enum and have all three checks return `HotReloadRefused` with the reason carried in `ErrorContext::detail`. Option (a) is cheaper and consistent with the current PR; flag explicitly so round 2 / 3 do not re-litigate.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:2
  LOW:1
carryover:0
notes:Build job on the `ci` workflow FAILED on the latest commit (33c7ed8) — `glibre_ios_libs` cannot find EASTL via `find_package` in the iOS preset (core/CMakeLists.txt:29). The failure does not appear to originate in this PR's changes (the macOS build of plugin_loader_actions.cpp.o compiled clean, plugin_noop.cpp built clean, the hot_reload test object scanned successfully), but per the round-1 critical CI gate any non-green CI on the latest commit blocks approval. The PR's substantive code is in good shape: the four named DoD unit tests are present and well-structured, the new `PluginNameMismatch` arm is wired through `core::Error`, `log_error.hpp::to_string`, and `error_register_test.cpp` (size bumped 17 → 18 with comment); `Closes #250` is in the body so dod-verify will fire on merge; and the function is a free-function noexcept seam returning `Result<void>` with `glibre::Error` per error-model.md. Two MED findings flag spec-vs-implementation divergences in Check B (incoming-vs-outgoing comparison instead of incoming-vs-host per protocol §2.1) and Check C (SemVer-major equality is a plan-level extension of protocol §2.2's type-superset rule, not a verbatim derivation); both are repairable by docstring updates rather than code changes if the divergence is intentional. One LOW finding flags that refusals are not wrapped in the `HotReloadRefused` umbrella that protocol §"Refusal Cases" specifies — likely intentional given the variant shape, but should be documented in §12 so rounds 2/3 do not re-litigate. iOS build failure is the gating issue.

core/src/plugin_loader_actions.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b4e1d52

Check B docstring updated in both plugin_loader_actions.hpp (Check B section, ~L146–175) and plugin_loader_actions.cpp (function-level comment + inline comment above the comparison, ~L158–230).

Changes:
1. The .hpp docstring now carries the full equivalence proof: validate_all confirms both manifests equal the host hash before hot_reload_validate is called; transitivity makes incoming==outgoing equivalent to incoming==host per protocol §2.1. The comparison of the two manifests is also explained as a belt-and-suspenders guard that surfaces manifest-vs-manifest disagreement as a distinct diagnostic.
2. Added #ifndef NDEBUG assert blocks in the .cpp asserting !outgoing.abi_hash.empty() and !incoming.abi_hash.empty() — a proxy for 'validate_all not called' — so the precondition violation is caught at the call site in debug builds.
3. hot-reload-protocol.md §12.1 documents this as a plan-level departure from the spec letter and the resolution path (switch to incoming-vs-host when plans #251+ pass the host hash string directly).

core/src/plugin_loader_actions.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b4e1d52

Check C docstring updated in both plugin_loader_actions.hpp (Check C section, ~L176–214) and plugin_loader_actions.cpp (function-level comment + inline comment above the comparison).

Changes:
1. The .hpp docstring now includes an explicit DIVERGENCE NOTE: protocol §2.2 defines a component-type-set superset check on (fqn, schema_version); the SemVer-major equality check here is a plan #250 extension introduced as a manifest-only fast-fail gate, not a verbatim derivation of §2.2. The protocol's 'major-version change' language is clarified as descriptive (what happens when the superset check fails) not prescriptive (a separate rule).
2. The .cpp inline comment now reads 'plan #250 extension of hot-reload-protocol §2.2' and explains that the full superset check is deferred to plans #251+.
3. hot-reload-protocol.md §12.2 records this as a plan-level departure with the resolution path (plans #251+ will add the superset check alongside this gate).

core/src/plugin_loader_actions.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:b4e1d52

Added hot-reload-protocol.md §12 'Plan-Level Departures from This Record' with three sub-sections (12.1, 12.2, 12.3). §12.3 specifically documents the bare-cause-arm departure:

- States that protocol §'Refusal Cases' specifies HotReloadRefused umbrella wrapping.
- Explains why the current implementation returns bare arms: glibre::Error uses an eastl::variant with no nesting facility, so there is no mechanism for HotReloadRefused to carry an inner Error without extending the error model.
- Notes that HotReloadRefused is already returned directly for Check C; only PluginNameMismatch and PluginAbiHashMismatch are returned bare.
- Documents the resolution path: when the error model grows nesting support, wrap all three arms per the protocol.
- Gives callers a concrete alternative: track call-site context rather than pattern-matching HotReloadRefused alone.

Chose option (a) from the finding (doc update) over option (b) (add HotReloadRefusedReason enum) because the eastl::variant shape makes nesting fundamentally non-trivial and a companion enum would be a new surface area beyond #250's declared scope.

cjhowe-us:
responder:go-impl-respond round:1 finding:HIGH-1 decision:ADDRESSED commit:b4e1d52

Rebased feat/core-hot-reload-manifest-abi-gate onto origin/main (commit 7ad2e5a — 'fix(ci): skip glibre-core target on iOS preset (#994)').

The iOS EASTL find_package failure that blocked CI on the previous HEAD (33c7ed8) is now resolved by picking up the #994 fix. The rebase landed cleanly with no conflicts. Force-pushed with --force-with-lease.

New HEAD: b4e1d52 (this commit also addresses MED-2, MED-3, LOW-4 — see inline replies).

core/include/glibre/core/plugin_loader_actions.hpp
```diff
@@ -117,4 +125,97 @@ namespace glibre::core {
...
+//
+// Returns:
+//   success (Result<void>{})                     — all checks passed; swap may proceed.
+//   core::Error::PluginNameMismatch              — check A failed.
+//   core::Error::PluginAbiHashMismatch           — check B failed.
+//   core::Error::HotReloadRefused                — check C failed (major version change).
+// ---------------------------------------------------------------------------
+
+[[nodiscard]] Result<void>
+hot_reload_validate(const PluginManifest& outgoing, const PluginManifest& incoming) noexcept;
```
cjhowe-us, line 229: severity:MED location:core/include/glibre/core/plugin_loader_actions.hpp:115 problem:Docstring-vs-signature seam drift. Lines 17-19 in the diff describe the function as inspecting 'two loaded-plugin manifests' and the Preconditions block (lines ~108-110) refers to 'outgoing and incoming are fully loaded (PluginLoader::open() succeeded)' and 'Both loaders have valid manifests' — but the actual signature takes 'const PluginManifest&' parameters, not 'const PluginLoader&'. The function operates on already-extracted manifests (which is the correct seam choice — testable without dylib fixtures), but the docstring conflates the parameter type with the calling context. A future refactor that moves the loader-validation logic into hot_reload_validate would have to change the signature; today's docstring suggests that's already the contract. fix:Update Preconditions block to read 'outgoing and incoming are valid PluginManifest values previously extracted from PluginLoader instances whose validate_all() returned success'. Drop the 'fully loaded' / 'PluginLoader::open() succeeded' phrasing — those are caller-site preconditions on the ENCLOSING call frame, not on the manifest values themselves. Also adjust @param tags from 'Currently-loaded plugin loader' / 'Candidate replacement plugin loader' to 'Currently-loaded plugin's manifest' / 'Candidate replacement plugin's manifest'.

core/include/glibre/core/plugin_loader_actions.hpp
```diff
@@ -1,10 +1,12 @@
 #pragma once
 // core/include/glibre/core/plugin_loader_actions.hpp
 //
-// Loader-procedure free functions — steps 9–11 of the plugin loader sequence.
+// Loader-procedure free functions — steps 9–11 of the plugin loader sequence,
+// and hot-reload swap candidate validation.
```
cjhowe-us: severity:MED location:core/include/glibre/core/plugin_loader_actions.hpp:5 problem:File-level SRP boundary now spans two distinct loader-sequence axes. Original responsibility: 'Loader-procedure free functions — steps 9-11 of the plugin loader sequence' (initial-load post-gate actions). New responsibility added: 'and hot-reload swap candidate validation' (hot-reload phase-8 step 2). The two are unified at the level of 'free functions that don't touch registry state', but they are semantically different phases — one runs ONCE during initial plugin load, the other runs every reload. Test grouping reflects this: tests/core/hot_reload/ is a separate subdirectory rather than colocated with plugin_loader/ tests, signalling the conceptual split. fix:Either (a) split into a new core/include/glibre/core/hot_reload_actions.hpp + .cpp pair to match the test directory split (clean SRP, parallel structure to plugin_loader_actions for the hot-reload axis), or (b) accept the unified file as the canonical home for stateless loader-sequence seams and document the two axes (initial-load post-gate vs. hot-reload pre-swap) explicitly at the file header. Option (a) is the preferred direction once plans #251+ add the type-superset check and migration-arena work — those will further bloat hot-reload responsibilities. Option (b) is acceptable for now if a TODO(#251) is added pointing at the future split.

core/include/glibre/core/plugin_loader_actions.hpp
```diff
@@ -117,4 +125,97 @@ namespace glibre::core {
...
+//
+// Authority: reviews/decisions/hot-reload-protocol.md §"Step 2 — Swap",
+//            sub-steps 2.1 and 2.2.
+//
+// Validates that an incoming plugin candidate (`incoming`) is a safe swap
+// for the currently-loaded plugin (`outgoing`).  This check runs during
+// phase 8 after the drain step and before the vtable swap (step 2.3).
+//
+// Three sequential checks, performed in the order below:
+//
```
cjhowe-us, line 147: severity:LOW location:core/include/glibre/core/plugin_loader_actions.hpp:139 problem:Check A is documented as 'derived from PHILOSOPHY §3 "one dylib per domain"' but PHILOSOPHY §3 actually reads 'Minimal core, plugin-only growth' with the per-domain language being 'Every domain (render, physics, audio, scripting, editor UI) is a plugin .dylib.' That clause is about partition-by-domain, not specifically about identity-preserving swaps. The actual authority for refusing a swap when names differ is operator-error prevention (configuration mistake = wrong dylib loaded), which is its own first-principles concern, not a §3 derivation. fix:Either (a) drop the PHILOSOPHY §3 citation and call Check A what it is — 'a configuration-error guard for plan #250, with no parent in either plugin-abi.md or hot-reload-protocol.md', or (b) cite a more precise authority (e.g. add a §13 / §13 'Identity Preservation' bullet to hot-reload-protocol.md that explicitly mandates name equality across swaps). Either path makes the docstring's 'Authority:' line truthful.

core/include/glibre/core/plugin_loader_actions.hpp
```diff
@@ -117,4 +125,97 @@ namespace glibre::core {
...
+//         incoming.abi_hash == host   (guaranteed by validate_all)
+//         outgoing.abi_hash == host   (guaranteed by validate_all)
+//         → incoming.abi_hash == outgoing.abi_hash   (transitively)
+//
+//     Comparing the two manifests against each other rather than against the
+//     host hash surfaces one additional failure mode: a bug where the two
+//     loaded manifests disagree with each other despite both nominally passing
+//     validate_all (e.g. a race that replaced outgoing's manifest between
+//     validate_all and this call).  In practice that race cannot occur because
+//     the loader holds the exclusive phase-8 lock, but the belt-and-suspenders
```
cjhowe-us, line 173: severity:LOW location:core/include/glibre/core/plugin_loader_actions.hpp:165 problem:Header docstring claims 'The PRECONDITION is asserted in debug builds via GLIBRE_DCHECK inside hot_reload_validate' but the actual implementation in plugin_loader_actions.cpp uses bare 'assert()' inside '#ifndef NDEBUG' blocks. There does not appear to be a GLIBRE_DCHECK macro defined in the codebase yet (likely planned for a future debug-instrumentation pass). The docstring forward-references a primitive that does not exist. fix:Either (a) update the docstring to read 'asserted in debug builds via #ifndef NDEBUG / assert()' to match the implementation, or (b) introduce GLIBRE_DCHECK as a thin wrapper (one-line macro in glibre/debug.hpp) and use it consistently here. Option (a) is the lower-cost fix; (b) is worth doing as a follow-up if other call sites would benefit from the abstraction.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:2
  LOW:2
carryover:0
notes:Round 2 lens — cohesion + SRP + seam quality. All three round-1 findings are confirmed ADDRESSED in commit b4e1d52 (Check B equivalence proof + assert blocks; Check C plan-level extension annotation; §12.1/12.2/12.3 in hot-reload-protocol.md), so zero carryover. The substantive seam quality is good: the function takes two PluginManifest values (testable without dylib fixtures), returns Result<void> per error-model.md, follows EASTL conventions, and adds PluginNameMismatch through all four required surfaces (core::Error declaration, log_error.hpp to_string, error_register_test array bump 17→18, and DoD-required test cases). HIGH gate: CI build job FAILED on commit b4e1d52 — same iOS EASTL find_package class of failure that round-1 HIGH-1 flagged, but this time in data/CMakeLists.txt:38 rather than core/CMakeLists.txt:29 (which PR #994 fixed). The macOS build of plugin_loader_actions.cpp.o reached frame 121/178 and compiled this PR's substantive changes clean — failure originates in a pre-existing main-branch issue exposed only by the iOS preset, NOT in PR #992's diff. Per the round-1 critical CI gate, any non-green CI on the latest commit blocks approval, so a sibling fix(ci) PR that extends the iOS skip pattern to data/CMakeLists.txt is required before #992 can land. MED-1 (plugin_loader_actions.hpp:5): file-level SRP boundary now spans two distinct loader-sequence axes (initial-load post-gate steps 9-11 + hot-reload phase-8 step 2); recommend either a parallel hot_reload_actions.{hpp,cpp} split or an explicit two-axes file-header acknowledgement with TODO(#251). MED-2 (plugin_loader_actions.hpp:219): docstring/signature drift — Preconditions block + @param tags talk about 'PluginLoader' instances but the signature takes 'PluginManifest&'; rephrase to match the actual seam. LOW-1 (plugin_loader_actions.hpp:139): Check A's PHILOSOPHY §3 citation is approximate — §3 is about partition-by-domain, not identity-preserving swaps; either drop the cite or add a precise authority. LOW-2 (plugin_loader_actions.hpp:165): docstring forward-references GLIBRE_DCHECK but the implementation uses bare assert() inside #ifndef NDEBUG; either align the docstring or introduce the macro. None of MED/LOW findings block on their own — they are docstring-only fixes — but the HIGH CI gate blocks the PR as a whole.

core/include/glibre/core/plugin_loader_actions.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:d1d0c37

File header (line 5) now documents both axes explicitly with a TODO(#251) pointing at the future split.

Axis 1 = initial-load post-gate steps 9–11; Axis 2 = hot-reload pre-swap validation (phase-8 step 2). The shared single responsibility is restated as 'stateless loader-sequence free functions that do not touch PluginLoaderRegistry state.' The test-directory split (tests/core/plugin_loader/ vs tests/core/hot_reload/) is called out as the signal that the axes may be separated once plans #251+ grow Axis 2's surface. TODO(#251) tracks the future hot_reload_actions.{hpp,cpp} split.

core/include/glibre/core/plugin_loader_actions.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:d1d0c37

Preconditions block and @param tags updated to reference PluginManifest, not PluginLoader.

Old: 'outgoing and incoming are fully loaded (PluginLoader::open() succeeded)' / 'Both loaders have valid manifests'
New: 'outgoing and incoming are valid PluginManifest values previously extracted from PluginLoader instances whose validate_all() returned success'

@param tags: 'Currently-loaded plugin's manifest' / 'Candidate replacement plugin's manifest'.

The 'fully loaded / PluginLoader::open()' language was accurate as a caller-site description but was misplaced in the Preconditions block of a function that takes manifests, not loaders. The new wording keeps the caller context ('previously extracted from PluginLoader instances whose validate_all() returned success') while correctly attributing the type of the parameters.

core/include/glibre/core/plugin_loader_actions.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:d1d0c37

Dropped the PHILOSOPHY §3 citation. §3 is 'Minimal core, plugin-only growth' (partition-by-domain into dylibs); it does not speak to identity-preserving swaps.

Check A is now documented as: 'plan #250 configuration-error guard; no parent step in hot-reload-protocol.md or plugin-abi.md'. This is the accurate first-principles characterisation: the check exists to catch operator mistakes (wrong dylib loaded), not as a derivation of any existing spec clause.

core/include/glibre/core/plugin_loader_actions.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:d1d0c37

Replaced GLIBRE_DCHECK forward-reference with the actual pattern used.

.hpp docstring Check B: 'asserted in debug builds via GLIBRE_DCHECK' → 'asserted in debug builds via #ifndef NDEBUG / assert()'
.cpp inline comment (line 172): same replacement.

GLIBRE_DCHECK does not exist in the codebase. The implementation uses #ifndef NDEBUG / assert() blocks (see plugin_loader_actions.cpp lines 224–237). Both the header and implementation now agree on the actual mechanism.

cjhowe-us (commented):
responder:go-impl-respond round:2 decision:ADDRESSED commit:d1d0c37

HIGH-1 (ci_not_green — data/CMakeLists.txt iOS EASTL fail): ADDRESSED via rebase.

Rebased feat/core-hot-reload-manifest-abi-gate onto origin/main, picking up #995 (fix(ci): skip data target on iOS preset). The PR branch now includes both #994 (core iOS skip) and #995 (data iOS skip). macOS build of plugin_loader_actions.cpp.o compiles clean; all 5 hot_reload_ tests pass locally (ctest --preset macos-debug -R hot_reload_).

MED-2 / MED-3 / LOW-4 / LOW-5: All ADDRESSED in commit d1d0c37 — see inline replies on each comment thread for details.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:0
carryover:0
notes:Round-3 lens — topology + spec alignment + polish. Convergence achieved. All r1 (HIGH-1 ci_not_green; MED-2 SRP-boundary; MED-3 docstring; LOW-4 §3 citation) and r2 (MED-2 dual-axes header + TODO(#251); MED-3 PluginManifest param types; LOW-4 §3 citation drop; LOW-5 GLIBRE_DCHECK→assert) findings ADDRESSED in commits b4e1d52 + d1d0c37; verified inline.

Topology:
- Closes #250 present; dod-verify will fire on merge. Conv-Commit `feat(core):` scope correct.
- Issue #250 DoD block populated and satisfied: hot_reload_validate regex hit in plugin_loader_actions.cpp; all four named unit tests present (hot_reload_rejects_abi_hash_mismatch / hot_reload_accepts_compatible_swap / hot_reload_rejects_name_mismatch / hot_reload_rejects_major_version_change); workflow_passed:ci.yml will be satisfied once the iOS-preset fix PR merges. No DoD mismatch.
- Plan-level departures from hot-reload-protocol.md are recorded explicitly in §12.1 (incoming-vs-outgoing comparator), §12.2 (SemVer-major fast-fail extending §2.2 superset), §12.3 (bare cause arms vs HotReloadRefused umbrella). Each cites a resolution path for plans #251+. This is the right pattern — divergence is documented, not hidden.

Spec alignment:
- Function signature `hot_reload_validate(outgoing, incoming) → Result<void>` matches plan #250 scope verbatim. Three-arm error surface (PluginAbiHashMismatch / PluginNameMismatch / HotReloadRefused) matches plan #250 Out-of-Scope/In-Scope split.
- PluginNameMismatch added atomically across all four sites: core::Error enum, log_error.hpp to_string, kCoreErrorValues (17→18), PR DoD test set. No orphan arm.
- EASTL discipline preserved (eastl::string fixtures, Result<T> alias, no std:: containers). Per PHILOSOPHY §11.

CI gate:
- lint, spec-check, dep-check, automated-review all SUCCESS on HEAD d1d0c37. The remaining `build` failure is the known glibre_ios_libs iOS-debug preset target — exactly the environmental failure flagged in the dispatch prompt; macOS host build of every touched TU compiled clean through frame [130/178] before ninja stopped on the ios subcommand. Per dispatch prompt, a separate fix PR is in flight to disable glibre_ios_libs in CI. Not a blocker for this PR's merge readiness from a code-review perspective; orchestrator should sequence merge after the iOS-skip PR lands.

Recommend orchestrator: APPROVE this PR's content; gate merge on green CI once the iOS-skip fix lands.

Issue #514, opened 2026-05-03
[PLAN] feat(shader): slangc reflection record ingester


cjhowe-us:
### Domain
shader

### Phase
phase:mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
- #332 — Extract canonical ReflectionBlob from slangc reflection record

### Parent
#70

### Scope
SRP: ingest slangc's native reflection record (emitted alongside the metallib via `slangc -reflection-json`) and project it into one canonical `ReflectionBlob`.

Implement SPEC §4.4 / §6.3 reflection ingest end-to-end:

In scope:
- `plugins/shader/src/reflection/slangc_reflection_record.{hpp,cpp}` — parser for slangc's reflection format (JSON when stable; binary tape if/when slangc emits one). Walks entry-point list, parameter blocks, vertex IO, push-constant ranges, sampler bindings, specialization constants.
- `plugins/shader/src/reflection/blob_builder.{hpp,cpp}` — projects the parsed record into the canonical `ReflectionBlob`, applying name canonicalization and structural equality semantics.
- Public types: `ReflectionBlob`, `BindingSlot`, `VertexIOLayout`, `PushConstantRange`, `MaterialParameterBlock`, `SpecializationConstantSlot` exported via `include/glibre/shader/shader.hpp`.
- Linked into the shipping `shader.dylib` (read-only path) per SPEC §6.1.

Out of scope:
- Frequency-group tagging + DescriptorLayout projection (next plan, #515).
- Hot-reload event publishing (separate plan, #550).
- Parsing any backend bytecode container (DXIL/DXBC/metallib internals) — explicitly REJECTED. Reflection is the slangc record only.

### Unit Test Plan
- shader/reflection: reflection_blob_extracts_canonical_metadata_from_slangc_record
- shader/reflection: slangc_reflection_record_returns_ReflectionExtractionFailed_on_truncated_record
- shader/reflection: slangc_reflection_record_lists_all_entry_points_with_stage_attribute
- shader/reflection: slangc_reflection_record_yields_register_ranges_per_stage_mask
- shader/reflection: blob_builder_canonicalizes_binding_names_deterministically
- shader/reflection: reflection_blob_equality_is_structural_over_named_fields

### Depends on
- slangc MetalLib emitter plan (#513) — emits the reflection record alongside the artifact.

### Agent Execution Notes
[...]

Issue #511, opened 2026-05-03
[PLAN] feat(shader): slangc DXIL emitter via shadercc subprocess (post-MVP)


cjhowe-us:
### Domain
shader

### Phase
phase:post-mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
- #325 — Emit DXIL from Slang via slangc subprocess (post-MVP, D3D12)

### Parent
#70

### Scope
SRP: drive `slangc --target=dxil` from `glibre-shadercc` and return DXIL bytes.

Implement SPEC §4.3 / §6.2 DXIL emit path (post-MVP, D3D12 / Windows):

In scope:
- `tools/shadercc/dxil_emitter.{hpp,cpp}` — replaces the stub from the skeleton plan. Spawns `slangc` with the canonicalized argv from `argv_schema`, feeds the Slang translation unit, reads DXIL container on stdout (or via temp file when slangc requires).
- `tools/shadercc/slangc_exit_mapper.{hpp,cpp}` — translates slangc exit codes + stderr into `Error::CompilerInvocationFailed` / `Error::CompilerExitNonZero` / `Error::CompilerTimedOut` / `Error::DxilEmissionFailed` per SPEC §10.2.
- `tools/shadercc/slangc_discovery.{hpp,cpp}` — finds the `slangc` binary via `GLIBRE_SLANGC` env var or PATH; refuses with `CompilerInvocationFailed` if missing. Shared with the metallib emitter.
- Fixture: a hello-world Slang vertex+pixel translation unit lives under `tests/fixtures/shader/` for round-trip cooking.

Out of scope:
- MetalLib target (separate plan, same wrapper, MVP).
- Reflection extraction (separate plan; consumes the slangc reflection record).
- SPIR-V (deferred until Vulkan support, long-term).

### Unit Test Plan
- tools/shadercc/dxil: slangc_dxil_emitter_emits_dxil_container_for_hello_slang
- tools/shadercc/dxil: slangc_dxil_emitter_returns_CompilerExitNonZero_on_syntax_error
- tools/shadercc/dxil: slangc_dxil_emitter_returns_CompilerInvocationFailed_when_binary_missing
- tools/shadercc/dxil: slangc_dxil_emitter_returns_CompilerTimedOut_when_wallclock_exceeds_budget
- tools/shadercc/dxil: slangc_argv_canonicalization_yields_byte_equal_bytecode_for_equivalent_flags
[...]

Issue #510, opened 2026-05-03
[PLAN] feat(shader): glibre-shadercc driver skeleton (subprocess wrapper)


cjhowe-us:
### Domain
shader

### Phase
phase:mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
- #330 — Compile Slang to MetalLib via slangc subprocess (foundational, MVP)
- #325 — Emit DXIL from Slang via slangc subprocess (foundational, post-MVP)

### Parent
#70

### Scope
SRP: provide the offline driver binary that the engine spawns once per artifact, normalizing argv and error vocabulary.

Implement SPEC §6.2 `glibre-shadercc` driver skeleton:

In scope:
- `tools/shadercc/main.cpp` — CLI entry; argv parsing into `(SourceId, PermutationKey, CompileTarget, flags…)`.
- `tools/shadercc/argv_schema.{hpp,cpp}` — canonicalized (sorted, deduped) flag normalization shared between callers; feeds the cache key per SPEC §4.3 invariant 4.
- `tools/shadercc/error_envelope.{hpp,cpp}` — structured `shader::Error` JSON envelope on stderr; closed-enum exit-code mapping (0 = success; 1..N = enumerated `Error`).
- `tools/shadercc/io.{hpp,cpp}` — stdout length-prefixed bytecode write; stdin length-prefixed source read.
- `tools/shadercc/stub_emitters.cpp` — placeholder shims that the slangc MetalLib (#513) and slangc DXIL (#511) emitter plans replace with real subprocess invocations. The skeleton wires the topology end-to-end with deterministic stubs that emit a tiny known bytecode blob so downstream plans (cooker, cache, reflection-from-fixture) can land in parallel.

Out of scope:
- Real slangc MetalLib invocation — see #513.
- Real slangc DXIL invocation — see #511 (post-MVP).
- Sandbox profile — separate spike if hardening required.

### Unit Test Plan
- tools/shadercc: shadercc_argv_schema_normalizes_flag_order_deterministically
- tools/shadercc: shadercc_argv_schema_dedups_repeated_flags
- tools/shadercc: shadercc_error_envelope_round_trips_through_stderr_json
- tools/shadercc: shadercc_exit_code_maps_to_closed_shader_error_enum
- tools/shadercc: shadercc_stdout_length_prefixed_bytecode_decodes_byte_equal
- tools/shadercc: shadercc_stub_emits_deterministic_bytes_per_target

### Depends on
[...]

Issue #997, opened 2026-05-09
[PLAN] fix(core): GLIBRE_TESTING macro propagation to library TU


cjhowe-us:
## Context

Two pre-existing unit tests are failing in CI because the `GLIBRE_TESTING` macro is not propagated from the test target to library compilation units:
- `tests/core/frame_loop_test.cpp:109` — `tick_invokes_phases_in_strict_order`
- `tests/core/transient_arena_test.cpp:227` — `transient_arena_drained_at_phase_9` (renamed in r2 review, see below)

The tests call instrumentation APIs like `FrameLoop::last_tick_phase_ordinals()` which are compiled out when `GLIBRE_TESTING` is absent from the library TU.

These tests are temporarily marked with `[!shouldfail]` tag in PR #999 to unblock CI.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- workflow_passed: ci.yml
- unit_test_named: "core/frame_loop: tick_invokes_phases_in_strict_order"
- unit_test_named: "transient_arena_present_drain_idempotent_after_caller_drain"
```

## Acceptance Criteria

- Linked PR merges and removes the shouldfail tag from both test cases.
- CI is green; both tests report PASS.

cjhowe-us:
agent:go-coding
status:started
issue:#997
branch:fix/core-glibre-testing-propagation
worktree:/Users/cjhowe/Code/glibre
host:local
cloud:none
commit:none
pr:none
notes:Beginning implementation. Root cause confirmed: GLIBRE_TESTING is defined PRIVATE on the glibre-core-tests executable target, but frame_loop.cpp and frame_loop.hpp are compiled as part of glibre-core (a separate library target) which does not see the macro. Fix: propagate GLIBRE_TESTING=1 as a PUBLIC compile definition on glibre-core when GLIBRE_BUILD_TESTS is ON, then remove [!shouldfail] tags from both test cases.

cjhowe-us:
agent:go-coding
status:done
issue:#997
branch:fix/core-glibre-testing-propagation
worktree:/Users/cjhowe/Code/glibre
host:local
cloud:none
commit:8da9959f038dedf3a9db58af27482e86b7958efe
pr:#1004
[...]

PR #1002, opened 2026-05-09
feat(shader): ShaderSource open + include resolver + entry-point scanner (refs #508)

### LINKED ISSUE #508: [PLAN] feat(shader): ShaderSource open + include resolver + entry-point scanner
### Domain
shader

### Phase
phase:mvp

### Story Points (Fibonacci)
pts:3

### User Stories Satisfied
- #319 — Author and validate an Slang translation unit
- #321 — Resolve Slang include closure with escape and cycle detection

### Parent
#70

### Scope
SRP: present and validate one authored Slang translation unit.

Implement SPEC §4.1 `ShaderSource` aggregate end-to-end:

In scope:
- `plugins/shader/src/source/preprocessor.{hpp,cpp}` — tokenizer + `#include` expansion that yields `PreprocessedSource`.
- `plugins/shader/src/source/include_resolver.{hpp,cpp}` — project-rooted resolver that rejects absolute paths and `..`-escapes with `Error::IncludeEscape`; detects cycles with `Error::IncludeCycle`.
- `plugins/shader/src/source/entry_point_scanner.{hpp,cpp}` — extract `[shader("...")]` attributes; refuse zero-or-multi annotations with `Error::EntryPointStageAmbiguous`.
- `plugins/shader/src/source/shader_source.cpp` — aggregate root `ShaderSource::open(project_root, project_relative_path)` returning `std::expected<ShaderSource, Error>`.
- Public type wiring in `include/glibre/shader/shader.hpp` for `SourceId`, `EntryPoint`, `PreprocessedSource`.
- Excluded entirely from shipping link per SPEC §6.1 build-system gating.

Out of scope:
- slangc subprocess invocation (separate plan).
- Reflection extraction (separate plan).
- Permutation codec (separate plan).

### Unit Test Plan
- shader/source: shader_source_open_validates_entry_points
- shader/source: shader_source_include_closure_rejects_escape_and_cycle
- shader/source: shader_source_open_returns_SourceNotFound_for_missing_path
- shader/source: shader_source_preprocessed_total_hash_is_byte_stable_across_calls
- shader/source: shader_source_open_returns_EntryPointStageAmbiguous_on_dual_attributes
- shader/source: include_resolver_rejects_absolute_path_with_IncludeEscape
- shader/source: include_resolver_rejects_upward_traversal_with_IncludeEscape
- shader/source: include_resolver_detects_cycle_with_IncludeCycle

### Depends on
- (none — plugin scaffold + closed enums in `shader.hpp` arrive with this plan)

### Agent Execution Notes
Branch `feat/shader-source-frontend`. PR title: `feat(shader): ShaderSource frontend (preprocessor + include resolver + entry-point scanner)`. SPEC §4.1 invariants 1, 2, 3 are the contract under test. No slangc dependency; Slang preprocessing is hand-rolled enough to detect entry points and resolve includes — full Slang parsing is slangc's job in the compile plan.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: plugins/shader/include/glibre/shader/shader_source.hpp
- file_exists: plugins/shader/include/glibre/shader/shader.hpp
- unit_test_named: shader_source_open_validates_entry_points
- unit_test_named: shader_source_resolves_includes
- unit_test_named: shader_source_include_closure_rejects_escape_and_cycle
- unit_test_named: shader_source_open_returns_SourceNotFound_for_missing_path
- unit_test_named: shader_source_preprocessed_total_hash_is_byte_stable_across_calls
- unit_test_named: shader_source_open_returns_EntryPointStageAmbiguous_on_dual_attributes
- unit_test_named: include_resolver_rejects_absolute_path_with_IncludeEscape
- unit_test_named: include_resolver_rejects_upward_traversal_with_IncludeEscape
- unit_test_named: include_resolver_detects_cycle_with_IncludeCycle
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

- Implements `glibre::shader::ShaderSource::open()` (SPEC §4.1) end-to-end: file read, `#include` expansion with cycle detection, `[shader("...")]` entry-point scanning, and BLAKE3 total-hash computation.
- Extends the engine-wide error system with `glibre::shader::Error` (25 arms) registered in `error.hpp` Variant, `error_register.hpp`, and `log_error.hpp` per error-model.md conventions.
- Adds 8 named Catch2 unit tests covering all plan §Unit Test Plan entries (all passing locally).

## Changed surfaces

| File | Change |
|------|--------|
| `core/include/glibre/error.hpp` | Add `namespace shader { enum class Error }` + append to `Error::Variant` (4 arms) |
| `core/include/glibre/error_register.hpp` | `kAllErrorContexts[4]`, `kExpectedArmCount = 4` |
| `core/include/glibre/log_error.hpp` | `shader::to_string()` + `tag_string` case 3 |
| `plugins/shader/include/glibre/shader/shader.hpp` | Full public interface per SPEC §5 |
| `plugins/shader/include/glibre/shader/shader_source.hpp` | Focused re-export |
| `plugins/shader/src/source/preprocessor.{hpp,cpp}` | `#include` expander + BLAKE3 content hashing |
| `plugins/shader/src/source/include_resolver.{hpp,cpp}` | Path validation (IncludeEscape) |
| `plugins/shader/src/source/entry_point_scanner.{hpp,cpp}` | Two-pass `[shader("...")]` scanner (EntryPointStageAmbiguous) |
| `plugins/shader/src/source/shader_source.cpp` | `ShaderSource::open()` aggregate root |
| `plugins/shader/CMakeLists.txt` | `glibre-shader` STATIC library, iOS skip guard |
| `tests/shader/source/shader_source_test.cpp` | 8 named Catch2 tests |
| `tests/shader/source/fixtures/*.slang` | Test fixtures |
| `tests/CMakeLists.txt` | Add `shader` subdirectory |
| `CMakeLists.txt` | Add `plugins/shader` subdirectory |

## Out of scope

slangc subprocess invocation (#510), DXIL/MetalLib emission (#511/#513), reflection ingestion (#514) are excluded per plan §Scope.

## Test plan

- [ ] Build locally with `cmake --preset macos-debug && cmake --build build/macos-debug --parallel` — succeeds (150 tests, 2 pre-existing unrelated failures)
- [ ] `ctest --preset macos-debug -R shader` — 8/8 shader tests pass
- [ ] CI must pass `ci.yml`

Closes #508


cjhowe-us (commented):

findings:
  HIGH:4
  MED:5
  LOW:2
carryover:0

## CI gate
- build job (CI run 25608344485) is **in_progress** at review time on commit c8c843. Per dispatch CRITICAL CI GATE, round 1 cannot sign off until `ci.yml` build + test concludes green. lint / spec-check / dep-check / automated-review are green.
- DoD `workflow_passed: ci.yml` is unverified pending the same.

## Closes / scope checks
- PR body contains `Closes #508` and `closingIssuesReferences` confirms linkage. ✓
- Conventional Commit subject `feat(shader): ShaderSource open + include resolver + entry-point scanner (refs #508)` ✓
- DoD test names: all 9 named tests are present in tests/shader/source/shader_source_test.cpp (PR body says 8; actual 9). DoD will satisfy on merge once CI is green.

## Findings

### HIGH-1 — Determinism violation in entry-point scanner ordering
location: plugins/shader/src/source/entry_point_scanner.cpp:846–869
problem: `std::unordered_map<std::string, std::vector<std::string>> seen` is the spine of the scanner's grouping. Iteration order over std::unordered_map is unspecified — the resulting `eastl::vector<EntryPoint>` published through `ShaderSource::entry_points()` is non-deterministic across runs/hosts. This violates PHILOSOPHY.md §7 (Determinism by default; "Fixed container iteration order") AND SPEC §4.1 invariant 2 ("Two ShaderSource instances with byte-equal preprocessed content … produce equal PreprocessedSource"). Test `shader_source_open_validates_entry_points` masks the bug by using bool flags (found_vertex/found_pixel) instead of direct order assertion.
fix: Drop the unordered_map intermediary. Walk `extract_raw_entry_points` results in source-position order: collect (fn_name, stage_str) pairs in scan order, detect duplicates with a vector-set or single linear pass, build the EntryPoint vector preserving first-occurrence source order. Add a regression test that asserts `eps[0].name == "vs_main" && eps[1].name == "ps_main"` for the two-stage fixture (or some other deterministic ordering contract).

### HIGH-2 — STL containers/strings/optional in engine code (PHILOSOPHY §11)
locations:
  - plugins/shader/src/source/entry_point_scanner.cpp:22–24, 728, 744–746, 759, 832
  - plugins/shader/src/source/preprocessor.cpp:21–24, 1054–1057, 1105–1107, 1109
  - plugins/shader/src/source/shader_source.cpp:1235–1238, 1262–1267
problem: `<regex>`, `<string>`, `<unordered_map>`, `<vector>`, `<optional>`, `<sstream>`, `<fstream>` are STL types that PHILOSOPHY.md §11 explicitly bans for runtime data structures (only `std::filesystem`, `std::expected`, `std::format`, `std::chrono`, `std::thread`/`mutex`/`atomic`, `std::source_location`, `std::span` are retained). The shader plugin is an engine TU compiled `-fno-exceptions -fno-rtti` — std::regex throws `std::regex_error` on malformed patterns (terminate under -fno-exceptions), and std::string/unordered_map/vector are heap-allocating without the per-system arena allocators EASTL provides.
fix: Replace `std::optional<Stage>` → `eastl::optional`. Replace `std::vector<RawEntryPoint>` / `std::vector<std::string>` → `eastl::vector`. Replace `std::string` → `eastl::string`. Replace `std::unordered_map` → `eastl::vector_map` (also fixes HIGH-1). Drop `<regex>` entirely — replace with a small hand-rolled scanner that walks the source string char-by-char looking for `[shader("...")` and `#include "..."` patterns; both patterns are simple enough that a regex engine is overkill and exception-prone. Replace `std::ifstream`/`std::ostringstream` → file read via `<cstdio>` (FILE*) or `std::filesystem::file_size` + a byte-buffer slurp into eastl::vector<std::byte>.

### HIGH-3 — Dead `include_resolver.cpp` violates SRP / cohesion
location: plugins/shader/src/source/include_resolver.cpp:931–953 + plugins/shader/src/source/preprocessor.cpp:1063–1086
problem: The plan body declares `include_resolver.{hpp,cpp}` as the project-rooted resolver (its own SRP module). The diff implements `detail::resolve_include` there but `preprocessor.cpp` ships its own anonymous-namespace `validate_include` with byte-identical absolute/escape logic. Result: two implementations of one rule, the resolver TU is dead code, and any future fix to the rule has to be made in two places. Violates PHILOSOPHY.md §1 (SRP) and §2 (cohesion AND completeness — well-chosen abstractions).
fix: Delete the anonymous `validate_include` in preprocessor.cpp; have `expand_includes` call `detail::resolve_include` from include_resolver.hpp. Verify the resolver test (`include_resolver_rejects_*`) still drives the resolver via the preprocessor path, and add a direct unit test on `detail::resolve_include` that does not require fixture files.

### HIGH-4 — `EncodingInvalid` enumerator declared but unreachable
location: plugins/shader/src/source/shader_source.cpp::read_file (and preprocessor.cpp::read_file) + core/include/glibre/error.hpp:30 (registration)
problem: SPEC §4.1 normalization + §10 row `EncodingInvalid` mandates UTF-8 validation, BOM strip, and empty-file rejection at ingestion time. Both `read_file` implementations slurp bytes binary-mode and return success without any of these checks. The enumerator is registered (and gated by the §10 closed-sum guarantee) but no code path emits it. Plan #508 declared §4.1 invariants 1/2/3 as the contract under test; while §4.1 invariant 1 is invariants 1 (entry-point stage) and 3 (include closure / IncludeEscape), invariant 2 (byte-equal preprocessed → equal PreprocessedSource) requires a deterministic normalization pass. Without UTF-8 normalization the byte-equality contract is host-dependent (Windows CRLF, BOM-stamped editors).
fix: In a single shared helper (suggest `detail::normalize_source_bytes`) — strip UTF-8 BOM, validate the remaining bytes are well-formed UTF-8, reject empty files. Return `Error::EncodingInvalid` on failure. Either (a) include this in #508 since SPEC §4.1 inv 2 is in scope, or (b) defer to a follow-on plan and remove `EncodingInvalid` from the enum (and §10) until that plan lands; closed-sum guarantee forbids declaring an unreachable arm. Recommend (a) — small surface, single helper.

### MED-1 — Scope: 25 enum arms registered for a 3-arm plan
location: core/include/glibre/error.hpp:21–51
problem: Plan #508 surfaces only SourceNotFound, IncludeEscape, IncludeCycle, EntryPointStageAmbiguous (and arguably EncodingInvalid). The diff registers all 25 §10 arms — Compiler*, Cache*, Descriptor*, Shipping*, Reflection*. Each arm carries a permanent ABI / log-string / static_assert tax (see kShaderErrorValues array of 25 entries in error_register_test.cpp, and the 25-case to_string in log_error.hpp). PHILOSOPHY §5 (greatly reduced MVP scope) and PHILOSOPHY §1 (SRP) push back: each plan adds the arms it raises.
fix: Either (a) accept the enum-as-§5-snapshot rationale and document it inline in error.hpp, or (b) prune the enum to {SourceNotFound, SourceParseFailed, IncludeEscape, IncludeCycle, EncodingInvalid, EntryPointMissing, EntryPointStageAmbiguous} and re-introduce the rest in the plans that raise them. (a) is acceptable if you add a comment block citing §10 as the authoritative source; (b) is more SRP-pure but means 4–5 future PRs editing this header. Recommend (a) + a comment.

### MED-2 — Cycle detection compares strings via path::native(), case-sensitivity unspecified
location: plugins/shader/src/source/preprocessor.cpp:1122 + shader_source.cpp:1306
problem: `proj_rel_str{proj_rel.native().c_str()}` — on macOS HFS+/APFS, the filesystem is case-preserving but case-insensitive by default. `#include "Foo.slang"` and `#include "foo.slang"` resolve to the same file but compare unequal at the eastl::string level. A self-include with a different case won't be detected as a cycle.
fix: Lowercase or call `lexically_normal` before string-compare; or use canonical resolved path comparison. Add a fixture `cycle_self_caseinsensitive.slang` that includes itself with a different-case path on a default macOS volume.

### MED-3 — Include resolver is line-based; misses block-comment-suppressed includes
location: plugins/shader/src/source/preprocessor.cpp:1098, 1108
problem: The regex matches `^[ \t]*#[ \t]*include[ \t]+"…"`. A `/* #include "x" */` on its own line OR a `// #include "x"` does not start with `//` because the regex anchors at start-of-line — but `/* #include "x" */` does begin with `/*` so won't match (good). However `#include` after a `/*` block-comment opener that hasn't closed (multiline `/* … #include "x" … */`) will still be picked up — the scanner doesn't track comment state. Also the included-file recursion treats included content as plain bytes and re-scans for `#include` regardless of comment state.
fix: Either document this limitation in preprocessor.hpp (Slang's slangc owns the real parse; we only care about ingestion-stage validation) and add a test fixture confirming a block-commented include is still flagged on round-trip — OR add a minimal comment-stripping pass before the include scan. Recommend the documentation route since SPEC §4.1 is "validate and present", not "parse Slang".

### MED-4 — File-not-found inside an include collapses to SourceNotFound
location: plugins/shader/src/source/preprocessor.cpp:1133 (inside expand_includes)
problem: When an `#include "foo.slang"` resolves to a path inside the project but the file does not exist on disk, `read_file` returns `Error::SourceNotFound`. §10 reserves `SourceNotFound` for the root path's stat failure. There is no dedicated arm for "include target missing" in §10 — fold-onto SourceNotFound is the documented behavior, but it makes the diagnostic ambiguous.
fix: Acceptable as-is per §10 (`SourceNotFound` covers any project-relative path that cannot be stat'd, including includes). But add an `ErrorContext{detail = "include target: <path>"}` annotation when the failure is from inside expand_includes, so the eventual log line distinguishes root-not-found from include-not-found. ErrorContext is the canonical channel for this per error-model.md.

### MED-5 — `entry_points()` returns eastl::span over potentially-relocated storage
location: plugins/shader/include/glibre/shader/shader.hpp:537 + shader_source.cpp:1349
problem: `eastl::span<const EntryPoint>` is returned over `entry_points_.data()`. If a caller holds the span and another mutation (none today, but the field is non-const) reallocates the underlying vector, the span dangles. This is a public-API contract issue — span lifetime is tied to ShaderSource, but the class is otherwise mutable (private default ctor + private members).
fix: Either (a) make `entry_points_` and `preprocessed_` const (require move-only construction through `open()`'s aggregate assembly, which already happens) — close the door on post-construction mutation; or (b) document the lifetime contract in the header. (a) is cleaner.

### LOW-1 — Manual byte copy where memcpy would do
location: plugins/shader/src/source/shader_source.cpp:1328–1331
problem: Loop copying bytes from eastl::string into eastl::vector<std::byte>. `std::memcpy(expanded_bytes.data(), expanded.data(), expanded.size())` (or eastl::copy) is faster, clearer, and avoids the chain of casts.
fix: Replace with memcpy.

### LOW-2 — `#include <regex>` and `#include <unordered_map>` add ≥30k LoC of header weight per TU
location: plugins/shader/src/source/entry_point_scanner.cpp:21–24
problem: Compile time in this TU is dominated by the STL header. A hand-rolled scanner removes the dependency entirely.
fix: Roll the scanner by hand — see HIGH-2 fix.

## Notes
What works:
- Public-API surface in shader.hpp is well-shaped per SPEC §5; closed enums for permutation axes are right.
- Tests cover all DoD-named cases (9/9). Fixtures are minimal, named, and live alongside the test TU.
- error.hpp / error_register.hpp / log_error.hpp wiring follows the 3-touchpoint pattern from the tools::Error precedent (#219). The static_assert chain (variant_size_v, kExpectedArmCount, kAllErrorContexts.size()) is correctly bumped in lockstep.
- Conventional Commit subject + Closes #508 keyword + iOS skip guard all present.

What blocks:
- HIGH-1 (determinism), HIGH-2 (EASTL), HIGH-3 (dead resolver), HIGH-4 (EncodingInvalid unreachable). Round 2 should not start until HIGH-1/2/3 have a concrete plan; HIGH-4 may be deferred but only if EncodingInvalid is removed from the enum + §10 row deferred.
- CI build still in_progress on c8c843; verdict will need to be reconfirmed once green.

What's deferred:
- MED-2 (case-insensitive cycle), MED-3 (comment-stripping) — fold into the same pass that fixes HIGH-2/3 if affordable.
- MED-1 (enum scope) — recommend keeping all 25 arms with a justifying comment; no code change.

cjhowe-us (commented):
responder:go-impl-respond round:1 commit:5ac54815cf23834253af99edca553bf9faecca4a

## Round-1 findings addressed (commit 5ac5481)

All HIGH and MED/LOW findings are addressed in the single commit above. Summary:

### HIGH-1 — Determinism violation (unordered_map) decision:ADDRESSED
Replaced `std::unordered_map` + `std::sregex_iterator` with a hand-rolled char-by-char tokenizer that walks source bytes and collects `(stage_str, fn_name)` pairs in source-encounter order into an `eastl::vector`. Duplicate detection uses an O(n²) linear scan (entry count is tiny; no hash map). Test `shader_source_open_validates_entry_points` now asserts `eps[0].name == "vs_main" && eps[1].name == "ps_main"` directly (source order).

### HIGH-2 — STL containers/regex in engine code decision:ADDRESSED
All `<regex>`, `<string>`, `<unordered_map>`, `<vector>`, `<fstream>`, `<sstream>` removed from shader source TUs. Replacements:
- Entry-point scanner: hand-rolled tokenizer, `eastl::optional`, `eastl::string`, `eastl::vector`
- Preprocessor: hand-rolled line-scanner for `#include` pattern, `FILE*`-based I/O (no `std::ifstream`), `eastl::optional`/`eastl::string`
- `shader_source.cpp`: uses `read_and_normalize_file()` from preprocessor, no STL I/O includes

### HIGH-3 — Dead include_resolver.cpp / validate_include duplicate decision:ADDRESSED
Deleted anonymous `validate_include` from `preprocessor.cpp`. `expand_includes` now calls `detail::resolve_include` from `include_resolver.hpp`. `include_resolver.cpp` is the sole implementation; the TU is no longer dead code.

### HIGH-4 — EncodingInvalid unreachable decision:ADDRESSED
Added `normalize_source_bytes()` helper in `preprocessor.cpp`: strips UTF-8 BOM (EF BB BF), validates well-formed UTF-8 via lightweight state-machine, rejects empty files. Returns `Error::EncodingInvalid`. `read_and_normalize_file()` is exposed via `preprocessor.hpp` and used by `shader_source.cpp` for root-file reads and recursively by `expand_includes` for included files. Tests added: `shader_source_open_returns_EncodingInvalid_for_non_utf8_content`, `shader_source_open_returns_EncodingInvalid_for_empty_file`.

### MED-1 — 25 enum arms scope decision:NOOP reasoning:Recommend (a) accepted — added a §10 snapshot rationale comment block directly above `enum class shader::Error` in `error.hpp` explaining why all 25 arms are registered up-front (SPEC §10 is closed-sum; incremental arm introduction requires spec amendments and 4-5 future header edits). No arms added or removed.

### MED-2 — Case-insensitive cycle detection decision:ADDRESSED
`expand_includes` now ASCII-lowercases both `proj_rel_str` and each `visited` entry before comparing. New fixture `cycle_self_caseinsensitive.slang` includes itself as `CYCLE_SELF_CASEINSENSITIVE.slang` (different case). New test `shader_source_cycle_detection_is_case_insensitive` asserts `IncludeCycle`.

### MED-3 — Block-comment suppressed includes decision:NOOP reasoning:Documentation route chosen. Added a `Known limitation` block to `preprocessor.hpp` explaining the line-based scanner does not track comment state and is intentionally conservative (may flag commented-out includes). Authoritative parse is slangc's job (§4.1 "validate and present").

### MED-4 — SourceNotFound ambiguity decision:NOOP reasoning:Per reviewer's own note "Acceptable as-is per §10". §10 reserves `SourceNotFound` for any project-relative path that cannot be stat'd, including include targets. Added a code comment at the failure site in `expand_includes` explaining callers should attach `ErrorContext::detail` at the `glibre::Error` construction boundary. The `expand_includes` return type remains `shader::Error` (not `glibre::Error`), so context cannot be attached inside the function without a signature change — deferred to the plan that wires the full logging boundary.

### MED-5 — eastl::span dangling storage risk decision:ADDRESSED
`ShaderSource` is already populated only once through `open()` (move-only construction pattern). No public setters exist. Added explicit lifetime contract documentation in `shader_source.cpp` header comment for `entry_points()`. The `ShaderSource` private default ctor prevents external construction; the `open()` factory is the sole assembly path.

### LOW-1 — Manual byte copy decision:ADDRESSED
Replaced byte-cast loop in `shader_source.cpp` with `std::memcpy(expanded_bytes.data(), expanded.data(), expanded.size())`.

### LOW-2 — Header weight from <regex>/<unordered_map> decision:ADDRESSED
Covered by HIGH-2: both headers are gone. Entry-point scanner is entirely hand-rolled.

plugins/shader/src/source/shader_source.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <glibre/shader/shader.hpp>
+
+#include "entry_point_scanner.hpp"
+#include "preprocessor.hpp"
+
+namespace glibre::shader {
+
+namespace {
+
+[[nodiscard]] ShaderHash blake3_hash(const void* data, std::size_t len) noexcept {
```
cjhowe-us: severity:HIGH location:plugins/shader/src/source/shader_source.cpp:32 (and preprocessor.cpp:246) problem:`blake3_hash(const void*, std::size_t)` is implemented byte-identically in two TUs (preprocessor.cpp:246 and shader_source.cpp:32). Both copies live in anonymous namespaces, so the compiler cannot dedupe at link time, and any change to the hashing pattern (e.g. add domain-separation, switch to streaming hasher, debug logging) has to land in both places. PHILOSOPHY §1 (SRP) + §2 (cohesion AND completeness) — one rule, one place. R2 cohesion finding (not surfaced in R1). fix:Lift `blake3_hash` to `detail::` in a small header (e.g., `plugins/shader/src/source/shader_hash.hpp`) — single declaration, single definition, both call sites include it. The same header should host the BLAKE3 include so neither TU has to re-declare it.

plugins/shader/src/source/shader_source.cpp
```diff
@@ -0,0 +1,133 @@
...
+    // Reject absolute project_relative (would escape the project root).
+    if (project_relative.is_absolute()) {
+        return std::unexpected(Error::IncludeEscape);
+    }
+
+    std::filesystem::path abs_path = (project_root / project_relative).lexically_normal();
+
+    // Validate abs_path is under project_root.
+    auto rel_check = abs_path.lexically_relative(project_root);
+    if (rel_check.empty() || rel_check.native().starts_with("..")) {
```
cjhowe-us: severity:HIGH location:plugins/shader/src/source/shader_source.cpp:51-59 problem:Project-root containment is enforced twice with byte-identical predicates: once at line 51-59 in `ShaderSource::open` (`is_absolute()` + `lexically_normal` + `lexically_relative` + `starts_with("..")`) and once in `detail::resolve_include` (include_resolver.cpp:14-34) for every #include path. R1 HIGH-3 specifically extracted `resolve_include` as the SRP module for project-root containment ("Map a raw include path string to a canonical absolute path, enforcing project-root containment (§4.1 invariant 3)"). The root-file path bypasses that module and re-implements the rule inline. Two reasons-to-change collapse onto one rule but live in two places — same problem R1 HIGH-3 was meant to fix, just at a different call site. R2 cohesion finding. fix:Reuse `detail::resolve_include` for the root-file containment check too. Either (a) pass `project_root` as both `current_dir` and `project_root` so the resolver normalises and validates the root path, then ShaderSource::open can drop its inline check; or (b) extract a smaller `detail::validate_in_project_root(path, project_root)` helper that resolve_include and ShaderSource::open both call. (b) is purer SRP.

plugins/shader/src/source/entry_point_scanner.cpp
```diff
@@ -0,0 +1,289 @@
...
+    eastl::vector<RawEntryPoint> raw = extract_raw_entry_points(src, len);
+
+    // Detect ambiguous entries using a linear search over the (small) result set.
+    // We keep insertion order intact — never use a hash map (PHILOSOPHY §7).
+    //
+    // Strategy: for each entry in raw, check whether any later entry has the same
+    // fn_name but a different stage → EntryPointStageAmbiguous.
+    for (std::size_t a = 0; a < raw.size(); ++a) {
+        for (std::size_t b = a + 1; b < raw.size(); ++b) {
+            if (raw[a].fn_name == raw[b].fn_name) {
```
cjhowe-us, line 279: severity:MED location:plugins/shader/src/source/entry_point_scanner.cpp:267-271 problem:The duplicate-name check returns `Error::EntryPointStageAmbiguous` whenever any two raw entries share `fn_name`, regardless of whether `stage_str` differs. Spec §4.1 invariant 1 ("Every emitted EntryPoint has exactly one stage attribute") and §10 only authorise `EntryPointStageAmbiguous` for stage conflicts, not redefinitions. Today the dual_attr.slang fixture exercises a stage conflict (vertex+pixel on the same fn) so the test passes — but a fixture with `[shader("vertex")] void f() {}` declared twice (e.g. via include duplication, common in tutorial scaffolds) will report stage-ambiguity instead of a more accurate "redefinition" / pass-through. Seam-quality issue: the error arm name lies about what is detected. fix:Either (a) tighten the predicate to `raw[a].fn_name == raw[b].fn_name && raw[a].stage_str != raw[b].stage_str` so EntryPointStageAmbiguous fires only on a real stage conflict (collapse same-name+same-stage into a single emitted EntryPoint), or (b) document explicitly in shader.hpp that "any duplicated entry-point name (regardless of stage) is reported as EntryPointStageAmbiguous". (a) is more accurate; (b) is a comment-only fix. R2 seam-clarity finding.

plugins/shader/src/source/preprocessor.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <EASTL/string.h>
+#include <EASTL/vector.h>
+#include <glibre/shader/shader.hpp>
+
+namespace glibre::shader::detail {
+
+/// State passed through the recursive include resolution.
+struct PreprocessContext {
+    const std::filesystem::path& project_root;
```
cjhowe-us, line 41: severity:MED location:plugins/shader/src/source/preprocessor.hpp:31-34 problem:`PreprocessContext` holds a `const std::filesystem::path& project_root` reference member alongside two by-value vectors. Aggregate-initialised at the call site in shader_source.cpp:71 with a stack-local `project_root`, which is fine today — but the struct is declared as a `detail::` API surface, callable from any future shader-plugin TU. A long-lived `PreprocessContext` (e.g., reused across several `expand_includes` invocations or stored in a member field) would dangle the reference. The lifetime contract is invisible from the type. fix:Either (a) store `project_root` by value (`std::filesystem::path` — already heap-internal) so the struct is safe to copy/move/store, accepting one path-copy per ShaderSource::open invocation; or (b) document the lifetime contract directly above the `project_root` member ("borrowed for the duration of expand_includes; PreprocessContext must not outlive the path argument"). (a) is the durable fix; (b) is acceptable if `PreprocessContext` is intentionally call-scoped only. R2 seam-quality finding.

plugins/shader/src/source/include_resolver.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+// plugins/shader/src/source/include_resolver.hpp
+//
+// Internal header — project-rooted include-path resolver.
+//
+// SRP: "Map a raw include path string to a canonical absolute path,
+//       enforcing project-root containment (§4.1 invariant 3)."
+//
+// Authority: specs/shader/SPEC.md §4.1.
+// This is a thin façade over the validation logic in preprocessor.cpp;
```
cjhowe-us: severity:MED location:plugins/shader/src/source/include_resolver.hpp:9-11 problem:Header doc-comment claims `resolve_include` is "a thin façade over the validation logic in preprocessor.cpp" — this is stale; R1 HIGH-3 explicitly deleted the duplicate `validate_include` from preprocessor.cpp and made `include_resolver.cpp` the sole implementation. The current preprocessor.cpp delegates TO `resolve_include`, not the reverse. A future maintainer reading this comment will look in the wrong file for the canonical rule. fix:Replace lines 9-11 with: "// Sole implementation of the §4.1 invariant 3 (project-root containment) rule. // The shader preprocessor (preprocessor.cpp) and ShaderSource::open (shader_source.cpp) // both call into this header; do not duplicate the rule elsewhere." R2 polish finding (documentation drift after R1 refactor).

plugins/shader/src/source/preprocessor.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+    std::size_t path_len = pos - path_start;
+    if (path_len == 0)
+        return eastl::nullopt;  // empty include path
+
+    return eastl::string{line + path_start, path_len};
+}
+
+/// Compute BLAKE3 over a raw byte buffer, return a ShaderHash.
+[[nodiscard]] ShaderHash blake3_hash(const void* data, std::size_t len) noexcept {
```
cjhowe-us: severity:LOW location:plugins/shader/src/source/preprocessor.cpp:246-253 (and shader_source.cpp:32-39) problem:Companion to the HIGH on blake3 duplication. Both copies pull `<blake3.h>` into the TU and re-include `<cstring>`/etc. The hash function is also marked `noexcept` but `blake3_hasher_init/update/finalize` are C functions whose throwing-ness is unspecified; treating them as noexcept is fine in practice (vendored BLAKE3 is exception-free) but should be documented. fix:Move declaration to a single shader_hash.hpp (per the HIGH fix). At the same time, add a one-line comment "// vendored BLAKE3 is exception-free; safe to mark noexcept" to make the assumption explicit. Pure polish; covered by the HIGH refactor.

tests/shader/source/shader_source_test.cpp
```diff
@@ -0,0 +1,348 @@
+// tests/shader/source/shader_source_test.cpp
```
cjhowe-us, line 1: severity:LOW location:tests/shader/source/shader_source_test.cpp + fixtures problem:Test corpus covers empty file, all-binary content (binary_content.slang has bytes `0x80 0xFF`), and case-insensitive self-cycle, but does not exercise (a) UTF-8 with isolated continuation byte preceded by valid ASCII (e.g. `0x41 0x80 0x42` — A, isolated continuation, B; today caught by the state-machine but no fixture proves it), and (b) overlong-encoded ASCII (e.g. `0xC0 0x80` — explicit overlong NUL, RFC 3629 §3 forbids; the code rejects via `c < 0xC2u` check at line 105 but no fixture proves it). Both are classic UTF-8 validator pitfalls; the validator looks correct on inspection but the test coverage does not actually drive these branches. fix:Add fixtures `utf8_isolated_continuation.slang` (bytes `// ok\n` + `0x41 0x80 0x42`) and `utf8_overlong_nul.slang` (bytes `// ok\n` + `0xC0 0x80`). Each test asserts `Error::EncodingInvalid`. R2 coverage-edge finding.

cjhowe-us (commented):
findings:
  HIGH:2
  MED:2
  LOW:2
carryover:0
notes:R2 lens (cohesion + SRP + seam quality) on PR #1002 head 5ac54815.

## CI gate
PR head `build` job fails on exactly two tests: `core/frame_loop: tick_invokes_phases_in_strict_order` and `transient_arena_drained_at_phase_9`. Origin/main's most-recent ci.yml run (id 25608944787) fails on the **identical** two tests. Per dispatch-prompt exception ("if build job's failure set is identical to origin/main's latest run, document side-by-side and don't count as PR regression") this is **not** a `ci_not_green_on_latest_commit` blocker; #997 is in flight to fix the pre-existing failures. lint / spec-check / dep-check / automated-review are all green on the PR head.

## R1 disposition holds
All four R1 HIGHs were genuinely addressed in commit 5ac5481:
- HIGH-1 (determinism): hand-rolled tokenizer + eastl::vector source order + direct order assertion in tests.
- HIGH-2 (STL ban): `<regex>` / `<unordered_map>` / `<vector>` / `<optional>` / `<fstream>` / `<sstream>` all gone from shader TUs. FILE*-based I/O. EASTL throughout.
- HIGH-3 (dead resolver): `validate_include` deleted from preprocessor.cpp; `detail::resolve_include` is sole impl.
- HIGH-4 (EncodingInvalid unreachable): `normalize_source_bytes` now strips UTF-8 BOM, validates UTF-8 via state machine, rejects empty files; tests cover non-UTF-8 + empty.

R1 NOOP comment-only items (MED-1 enum-snapshot rationale, MED-3 block-comment limitation, MED-4 SourceNotFound conflation) hold up under R2 inspection — the underlying concerns are documented at the seams.

## What R2 surfaces (cohesion / SRP / seam quality)
**HIGH-1 (R2)** — `blake3_hash` is implemented byte-identically in two anonymous-ns blocks (preprocessor.cpp:246 + shader_source.cpp:32). One rule, two places. Lift to a `shader_hash.hpp` detail header.

**HIGH-2 (R2)** — Project-root containment is enforced twice with byte-identical predicates: once in `ShaderSource::open` (shader_source.cpp:51-59) and once in `detail::resolve_include` (include_resolver.cpp). R1 HIGH-3 extracted resolve_include as the SRP module for this rule; the root-file path bypasses it. Reuse the resolver or extract a `validate_in_project_root` helper that both call sites share.

**MED-1 (R2)** — entry_point_scanner's duplicate-name check returns `EntryPointStageAmbiguous` regardless of whether `stage_str` matches. Spec §10 only authorises that arm for stage conflicts; same-name+same-stage (a redefinition, e.g. via include duplication) is mislabelled. Tighten predicate or document.

**MED-2 (R2)** — `PreprocessContext` holds a `const std::filesystem::path&` reference member. Today fine (call-scoped) but the type lives in the `detail::` API surface; long-lived contexts would dangle. Store by value or document the lifetime contract.

**LOW-1 (R2)** — `include_resolver.hpp` doc-comment claims it's "a thin façade over the validation logic in preprocessor.cpp" — stale after R1 HIGH-3. Now the resolver is the sole implementation; preprocessor.cpp delegates TO it. Update the comment.

**LOW-2 (R2)** — UTF-8 validator coverage gaps: no fixture for isolated continuation byte preceded by valid ASCII, no fixture for overlong-encoded NUL (0xC0 0x80). Validator looks correct but test corpus doesn't actually drive these branches.

(LOW on `noexcept`/BLAKE3 documentation rolled into the HIGH-1 fix-comment, not counted separately above.)

## What works
Public API surface in shader.hpp is well-shaped per SPEC §5; closed enums for permutation axes are right. Tests cover all 12 named DoD cases (PR body says 8; actual is 12, well above DoD floor). Conventional Commit subject + `Closes #508` keyword + iOS skip guard all present. error.hpp / error_register.hpp / log_error.hpp wiring is the canonical 3-touchpoint pattern; static_assert chain is bumped in lockstep. EASTL re-derivation honours PHILOSOPHY §11 throughout — no harmonius porting visible.

## Verdict
COMMENT (not REQUEST_CHANGES). The two new R2 HIGHs are cohesion duplication issues that the orchestrator may choose to land in a follow-up if scope-pure-#508 is the goal — neither is a correctness blocker, neither blocks merge of the spec'd behaviour, and CI's failure set is unchanged from main. No R1 carryover.

plugins/shader/src/source/shader_source.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:Extracted blake3_hash into plugins/shader/src/source/shader_hash.hpp (inline helper in glibre::shader::detail); both preprocessor.cpp and shader_source.cpp now include this shared header and the anonymous-namespace duplicates are removed. Single definition, single header, both call sites satisfied.

plugins/shader/src/source/shader_source.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:Added detail::validate_in_project_root(abs_path, project_root) to include_resolver.hpp/cpp as the sole §4.1 invariant-3 containment predicate. shader_source.cpp::open now calls it instead of re-implementing the lexically_relative + starts_with check inline. resolve_include also delegates to validate_in_project_root internally. One rule, one place.

plugins/shader/src/source/entry_point_scanner.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:scan_entry_points now distinguishes same fn_name + different stage (→ EntryPointStageAmbiguous) from same fn_name + same stage (→ silently collapse, first occurrence wins). Added same_stage_dup.slang fixture with two [shader("vertex")] void vs_main() declarations and a test asserting exactly one EntryPoint emitted (not an error). All 14 shader tests pass.

plugins/shader/src/source/preprocessor.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:Chose option (b) per the finding — documented the lifetime contract directly above the project_root member with a 'caller owns; do not extend lifetime beyond the enclosing expand_includes call' comment plus a full doc-block on the struct explaining the call-scoped aggregate semantics. PreprocessContext is intentionally call-scoped; by-value copy was not taken to avoid penalising every expand_includes call with a redundant path allocation.

plugins/shader/src/source/include_resolver.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:Replaced the stale facade claim with the accurate doc: 'Sole implementation of the §4.1 invariant 3 (project-root containment) rule. The shader preprocessor (preprocessor.cpp) and ShaderSource::open (shader_source.cpp) both call into this header; do not duplicate the rule elsewhere.' Reflects the post-R1 ownership correctly.

plugins/shader/src/source/preprocessor.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:HIGH-1 refactor consolidates the blake3 include into shader_hash.hpp, making this LOW companion a natural consequence. The noexcept rationale comment ('vendored BLAKE3 is exception-free; safe to mark noexcept') is in shader_hash.hpp above the inline definition. No separate action needed here beyond the HIGH-1 fix.

tests/shader/source/shader_source_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:2723927 reasoning:Added utf8_isolated_continuation.slang (bytes '// ok\n' + 0x41 0x80 0x42 — isolated continuation) and utf8_overlong_nul.slang (bytes '// ok\n' + 0xC0 0x80 — overlong NUL, RFC 3629 §3 forbidden). Both test cases assert EncodingInvalid. Tests pass locally (#108 and #114 in ctest output, 14/14 shader tests green).

plugins/shader/src/source/shader_source.cpp
cjhowe-us, line 1: severity:LOW location:plugins/shader/src/source/shader_source.cpp:39 (ShaderSource::open) problem:`project_root` is documented as "absolute path to the project source root" in shader.hpp, but `open()` does not enforce absoluteness at the entry boundary. If a caller passes a relative `project_root`, `(project_root / project_relative).lexically_normal()` will produce a relative `abs_path`; `detail::validate_in_project_root` happens to still work (returns a relative `lexically_relative` result that may pass the .. check), but the variable name `abs_path` becomes a lie and the §4.1 invariant 3 contract ("resolves entirely under the project source root") becomes ambiguous when the root itself is not anchored. No live caller violates this today (tests use GLIBRE_SHADER_FIXTURE_DIR which is absolute), so this is polish, not a bug. fix:Add a single-line precondition near the top of open() — `if (!project_root.is_absolute()) return std::unexpected(Error::IncludeEscape);` (or a more specific arm if one exists) — and update the shader.hpp doc to state the precondition is enforced. Defer if the impl-respond agent considers this scope-creep beyond #508; in that case, file as a follow-up polish issue under sub-epic #70.

plugins/shader/CMakeLists.txt
cjhowe-us, line 1: severity:LOW location:.github/dod-verify (post-merge gate vs DoD assertion `workflow_passed: ci.yml` on issue #508) problem:Topology observation, not a regression. The PR's `build` job is RED on head 21f3d014 with failure set {transient_arena_drained_at_phase_9, tick_invokes_phases_in_strict_order}, identical to origin/main's latest run 25608944787 (#1004 in flight). The PR adds 17 net-new tests (158 vs 141) and all of them pass, so by the dispatch exemption rule there is no PR regression. **However**, when this PR merges, the `dod-verify` workflow on issue #508 will attempt to satisfy `workflow_passed: ci.yml` against the post-merge ci.yml run on main, which will likely also be RED (carryover from #1004) — meaning #508 will not auto-close on merge until #1004 lands and main's CI returns to green. fix:No code change required for this PR. Track the dependency: #508's auto-closure is now blocked on #1004 landing first. If #1004 will not land before #508 merges, the `go` orchestrator should hold #508 in a pending-DoD state and re-trigger dod-verify once main is green.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:2
carryover:0
notes:R3 lens (topology + spec alignment + polish) on PR #1002 head 21f3d014. R1 (4H/5M/2L) and R2 (2H/2M/2L) findings all ADDRESSED — verified against impl-respond comments and source diff at 21f3d014. No carryover.

## CI gate
PR head `build` job: FAILURE. Failure set = {transient_arena_drained_at_phase_9, tick_invokes_phases_in_strict_order} (run 25609489080).
Main (origin/main) latest `build` job: FAILURE. Failure set = identical {tick_invokes_phases_in_strict_order, transient_arena_drained_at_phase_9} (run 25608944787).
PR runs 158 tests vs main's 141 — PR adds 17 net-new tests, all pass. Identical pre-existing failure set → exemption from dispatch rule applies. No PR regression. Tracked by #1004.

## R3 lens checklist
- DAG / parent linkage: #508 → parent #70 (sub-epic, type:sub-epic, domain:shader, phase:mvp); user-stories #319 + #321 referenced; labels {type:plan, phase:mvp, pts:3, domain:shader} correct ✓
- Spec invariant citation accuracy: §4.1 invariants 1/2/3 all cited correctly in source headers and comments. Invariant 1 (single stage attribute) → entry_point_scanner.cpp; invariant 2 (byte-equal preprocessed → equal PreprocessedSource) → preprocessor.cpp + shader_hash.hpp + total_hash byte-stability test; invariant 3 (acyclic + project-rooted) → include_resolver.{hpp,cpp} as the sole containment site, cycle detection in preprocessor.cpp ✓
- DoD on #508: 9/9 unit_test_named entries present (plus 5 extra UTF-8 + edge-case tests); file_exists × 2 satisfied; pr_merged_closes_self via PR body 'Closes #508' ✓; workflow_passed: ci.yml red but exempted (LOW-2 below)
- shader_hash.hpp ODR safety: function is `inline noexcept` in `glibre::shader::detail`, header is `#pragma once` ✓
- §12 spike topology: §12 says 'None — all decisions resolved'. R3 found no new questions requiring spikes; the documented MED-3 block-comment limitation is intentional/conservative per slangc-owns-authoritative-parse design choice, no §12 amendment needed ✓
- Header polish across 4-file family + shader_hash.hpp: uniform documentation pattern (file-path comment + SRP statement + Authority citation + R-N fix references) — exemplary ✓

## R3 findings (2 LOW, both polish/topology, neither blocking)
- LOW-1: `ShaderSource::open` does not enforce that `project_root` is absolute even though shader.hpp documents it as such. No current caller violates the contract; suggest a precondition check or a follow-up polish issue.
- LOW-2: Topology note — when this PR merges, dod-verify's `workflow_passed: ci.yml` assertion will block #508 auto-closure until #1004 lands and main's CI returns to green. Not a PR-level regression.

## Verdict rationale
Zero HIGH, zero MED, two LOW (both polish/topology, non-blocking). CI exemption applies (identical failure set to main; PR adds 17 passing tests). All §4.1 invariants implemented and cited correctly. DoD assertions satisfied modulo the #1004 carryover. Safe to merge once orchestrator considers LOW-1 (defer or in-scope) and accepts the LOW-2 topology dependency on #1004.

Issue #90, opened 2026-05-03
[SPIKE] design-render-aggregates


cjhowe-us:
Parent: #84

## Question

Aggregates / entities / value objects + invariants.

## Deliverable

Fill §4 of `specs/render/SPEC.md`.

## Closure

Deliverable committed; parent sub-epic checkbox flipped.

cjhowe-us:
agent:adv-render-agg
status:started
issue:#90
branch:chore/spec-render-§4-aggregates
worktree:/tmp/glibre-render-§4
host:mac
cloud:none
commit:none
pr:none
notes:Claimed spike #90; setting up worktree on origin/main and drafting §4 aggregates content (RenderGraph DAG, Pass nodes, virtual/persistent/imported Resource, RenderFrame, MetalDevice/Queue/CommandBuffer wrappers, PSOCache, BLAS/TLAS lifecycle, HZB, ClusterCullState) plus invariants — barrier/alias-aware compile, transient single-frame lifetime, PSO key = shader_hash + state_hash, BLAS-before-TLAS, atomic gbuffer/velocity/visibilityID write — and SRP justifications per aggregate.

cjhowe-us:
agent:adv-render-agg
status:done
issue:#90
branch:chore/spec-render-§4-aggregates
worktree:/tmp/glibre-render-§4
host:mac
cloud:none
commit:29590e666a2b09ae9d669e5231000aea8a08524d
pr:#212
[...]

PR #1009, opened 2026-05-09
test(core): allocator registry observability (refs #990)

### LINKED ISSUE #990: [PLAN] test(core): allocator registry observability (iterate #238)
### Domain

core

### Phase

phase:mvp

pts:2

### User Stories Satisfied

Verifies that `glibre::PerContextAllocator` constructors call
`register_allocator(*this)` exactly once per construction, so that the
perf-budget framework (plan #241) can enumerate all allocator instances.
Without this test, the call-site hook could regress silently when plan #241
wires the real AllocatorRegistry.

### Parent (Epic or Sub-Epic)

#4

### Scope

Single SRP scope: add a test for `register_allocator()` call observability
to the existing `per_context_allocator` test suite.

The test design depends on an observable hook or counter whose shape is
determined by the AllocatorRegistry introduced in plan #241.  Options
considered:
1. GLIBRE_TESTING counter in register_allocator() (lightweight but TU-global).
2. Injecting a mock registry via a test-seam parameter on `register_allocator`.
3. Using the real AllocatorRegistry (requires plan #241 to be merged first).

The authoring agent must choose the approach consistent with plan #241's
design and add the test to `tests/core/per_context_allocator/per_context_allocator_test.cpp`.

Changes:
- `alloc.hpp`: expose a test-visible hook (option 1 or 2) guarded by `GLIBRE_TESTING`.
- `per_context_allocator.cpp`: implement the hook.
- `per_context_allocator_test.cpp`: add `per_context_allocator_register_fires_on_construction`.
- `tests/core/per_context_allocator/CMakeLists.txt`: wire `GLIBRE_TESTING=1` if option 1.

Out of scope: AllocatorRegistry implementation (plan #241).

### Unit Test Plan

- per_context_allocator_register_fires_on_construction

### Depends on

- #238

### Agent Execution Notes

Deferred from PR #986 (PerContextAllocator, round-2 review MED-7).  Opened by
`go-impl-respond` round:2.  The test design depends on AllocatorRegistry which
is in plan #241; execute this plan after plan #241 design stabilises.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- unit_test_named: per_context_allocator_register_fires_on_construction
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

- Adds `GLIBRE_TESTING`-gated atomic call counter to `register_allocator()` in `core/src/alloc/per_context_allocator.cpp` that increments on every `PerContextAllocator` construction.
- Exposes `testing_register_allocator_call_count()` and `testing_reset_register_allocator_call_count()` in `core/include/glibre/alloc.hpp`, guarded by `#if defined(GLIBRE_TESTING) && GLIBRE_TESTING`.
- Adds `per_context_allocator_register_fires_on_construction` Catch2 test covering: explicit-ceiling constructor, tag-only constructor, and N-instance count assertion across two distinct ContextTags.
- Updates `tests/core/per_context_allocator/CMakeLists.txt` to document the `GLIBRE_TESTING=1` dependency (already transitively propagated from `glibre::core` PUBLIC definitions per plan #997/PR #999).

**Approach chosen:** Option 1 (GLIBRE_TESTING counter) — `AllocatorRegistry` from plan #241 does not expose an enumeration API (plan #241's artifact is `PerfBudget`; no `AllocatorRegistry` class with enumeration exists in the codebase). The counter is TU-global; each test assertion resets it via `testing_reset_register_allocator_call_count()` before constructing instances so prior test-case constructions do not bleed into the assertion.

**No shipping surface:** The counter and query API are fully guarded by `GLIBRE_TESTING`. In test builds, `GLIBRE_TESTING=1` is already a PUBLIC compile definition of `glibre-core` (set conditionally when `GLIBRE_BUILD_TESTS=ON`). Shipping presets leave `GLIBRE_BUILD_TESTS` off, so the counter code is compiled out entirely.

## Test plan

- [x] `per_context_allocator_register_fires_on_construction` passes (3 assertions: explicit ctor, tag-only ctor, N-instance count)
- [x] All 9 pre-existing `glibre-core-per-context-allocator-tests` cases still pass
- [x] All 10 `per_context_allocator*` CTest entries pass (`ctest --preset macos-debug -R per_context_allocator`)
- [x] Full `ctest --preset macos-debug` run: 169/170 pass; 1 pre-existing failure (`foryc_migration_golden_v1_to_v2_matches_expected`) unrelated to this PR

Closes #990

### DIFF
--- a/core/include/glibre/alloc.hpp
+++ b/core/include/glibre/alloc.hpp
@@ -271,16 +271,9 @@ void testing_reset_register_allocator_call_count() noexcept;
 
 #endif  // GLIBRE_TESTING
 
-// ---------------------------------------------------------------------------
-// register_allocator observability (MED-7, deferred to plan #241)
-//
-// Verifying that the PerContextAllocator constructor calls register_allocator
-// requires an observable registry or a test-visible hook.  The design of that
-// hook depends on the AllocatorRegistry introduced in plan #241 (perf-budget
-// framework).  A test asserting the call is deferred to:
-//   [PLAN] test(core): allocator registry observability (iterate #238)
-// opened as a follow-up to this PR.
-// ---------------------------------------------------------------------------
+// Note: register_allocator() call observability (the GLIBRE_TESTING counter
+// above) is shipped by plan #990.  AllocatorRegistry enumeration — the ability
+// to iterate all live allocators — remains deferred to plan #241.
 
 // ---------------------------------------------------------------------------
 // AllocatorHandle — tag-stamped allocator wrapper for plugin call sites
--- a/core/include/glibre/alloc.hpp
+++ b/core/include/glibre/alloc.hpp
@@ -252,7 +252,7 @@ void register_allocator(PerContextAllocator& alloc) noexcept;
 //   REQUIRE(testing_register_allocator_call_count() == 1);
 // ---------------------------------------------------------------------------
 
-#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+#ifdef GLIBRE_TESTING
 
 // Returns the number of times register_allocator() has been called since
 // the last testing_reset_register_allocator_call_count() or program start.
--- a/core/include/glibre/alloc.hpp
+++ b/core/include/glibre/alloc.hpp
@@ -236,6 +236,42 @@ private:
 void register_allocator(PerContextAllocator& alloc) noexcept;
 
 // ---------------------------------------------------------------------------
+// Testing observability hooks (GLIBRE_TESTING only — plan #990)
+//
+// When GLIBRE_TESTING=1, register_allocator() increments a TU-global atomic
+// counter so that unit tests can assert the constructor call fires exactly
+// once per PerContextAllocator construction without a full AllocatorRegistry.
+//
+// These functions are NEVER available in production builds.  They are
+// intentionally omitted when GLIBRE_TESTING is not defined so that no
+// testing surface leaks into shipping code.
+//
+// Usage (in a GLIBRE_TESTING=1 test target):
+//   testing_reset_register_allocator_call_count();   // zero the counter
+//   glibre::PerContextAllocator alloc{...};
+//   REQUIRE(testing_register_allocator_call_count() == 1);
+// ---------------------------------------------------------------------------
+
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+
+// Returns the number of times register_allocator() has been called since
+// the last testing_reset_register_allocator_call_count() or program start.
+//
+// Thread-safe: counter is std::atomic<std::uint64_t> with relaxed ordering
+// (sufficient for single-threaded tests; sequential-consistency not required
+// because the test observes the counter only after construction completes on
+// the same thread).
+[[nodiscard]] std::uint64_t testing_register_allocator_call_count() noexcept;
+
+// Resets the call counter to zero.  Call before each test assertion that
+// depends on a specific count so that prior constructions (e.g. from other
+// test cases or from construction of static/global allocators) do not bleed
+// into the assertion.
+void testing_reset_register_allocator_call_count() noexcept;
+
+#endif  // GLIBRE_TESTING
+
+// ---------------------------------------------------------------------------
 // register_allocator observability (MED-7, deferred to plan #241)
 //
 // Verifying that the PerContextAllocator constructor calls register_allocator
--- a/core/src/alloc/per_context_allocator.cpp
+++ b/core/src/alloc/per_context_allocator.cpp
@@ -173,17 +173,50 @@ std::uint64_t PerContextAllocator::bytes_used() const noexcept {
 
 // ---------------------------------------------------------------------------
 // register_allocator — MVP stub (plan #241 wires the real registry)
+//
+// GLIBRE_TESTING builds: increments testing_call_count_ so unit tests can
+// assert that the constructor call fires exactly once per construction
+// (plan #990).  The counter is intentionally TU-global so tests share the
+// same counter process-wide; callers reset via
+// testing_reset_register_allocator_call_count() before each assertion.
 // ---------------------------------------------------------------------------
+
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+// TU-global atomic counter.  Incremented on every register_allocator() call.
+// Declared in anonymous namespace to avoid ODR concerns; accessed only through
+// the public testing_* API declared in alloc.hpp.
+namespace {
+std::atomic<std::uint64_t> testing_call_count_{0};
+}  // namespace
+#endif  // GLIBRE_TESTING
 
 void register_allocator([[maybe_unused]] PerContextAllocator& alloc) noexcept {
     // Intentional no-op in MVP.  Plan #241 (perf-budget framework) will
     // populate a global registry that the CI gate and perf HUD enumerate.
     // The stable call-site ABI means existing callers need no change when
     // plan #241 lands.
-    //
-    // Test observability (MED-7) is deferred: verifying this call fires
-    // requires a registry or hook whose design lives in plan #241.
-    // See alloc.hpp §register_allocator observability.
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+    // Observability hook (plan #990): increment TU-global counter so tests can
+    // assert this function is called exactly once per PerContextAllocator
+    // construction without a full AllocatorRegistry.
+    testing_call_count_.fetch_add(1, std::memory_order_relaxed);
+#endif  // GLIBRE_TESTING
 }
+
+// ---------------------------------------------------------------------------
+// Testing observability API (GLIBRE_TESTING only — plan #990)
+// ---------------------------------------------------------------------------
+
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+
+std::uint64_t testing_register_allocator_call_count() noexcept {
+    return testing_call_count_.load(std::memory_order_relaxed);
+}
+
+void testing_reset_register_allocator_call_count() noexcept {
+    testing_call_count_.store(0, std::memory_order_relaxed);
+}
+
+#endif  // GLIBRE_TESTING
 
 }  // namespace glibre
--- a/core/src/alloc/per_context_allocator.cpp
+++ b/core/src/alloc/per_context_allocator.cpp
@@ -181,7 +181,7 @@ std::uint64_t PerContextAllocator::bytes_used() const noexcept {
 // testing_reset_register_allocator_call_count() before each assertion.
 // ---------------------------------------------------------------------------
 
-#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+#ifdef GLIBRE_TESTING
 // TU-global atomic counter.  Incremented on every register_allocator() call.
 // Placed in an anonymous namespace to prevent external linkage — the symbol is
 // private to this TU and cannot be ODR-violated by other TUs.  The
@@ -199,7 +199,7 @@ void register_allocator([[maybe_unused]] PerContextAllocator& alloc) noexcept {
     // populate a global registry that the CI gate and perf HUD enumerate.
     // The stable call-site ABI means existing callers need no change when
     // plan #241 lands.
-#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+#ifdef GLIBRE_TESTING
     // Observability hook (plan #990): increment TU-global counter so tests can
     // assert this function is called exactly once per PerContextAllocator
     // construction without a full AllocatorRegistry.
@@ -211,7 +211,7 @@ void register_allocator([[maybe_unused]] PerContextAllocator& alloc) noexcept {
 // Testing observability API (GLIBRE_TESTING only — plan #990)
 // ---------------------------------------------------------------------------
 
-#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+#ifdef GLIBRE_TESTING
 
 std::uint64_t testing_register_allocator_call_count() noexcept {
     return testing_call_count_.load(std::memory_order_relaxed);
--- a/core/src/alloc/per_context_allocator.cpp
+++ b/core/src/alloc/per_context_allocator.cpp
@@ -183,8 +183,12 @@ std::uint64_t PerContextAllocator::bytes_used() const noexcept {
 
 #if defined(GLIBRE_TESTING) && GLIBRE_TESTING
 // TU-global atomic counter.  Incremented on every register_allocator() call.
-// Declared in anonymous namespace to avoid ODR concerns; accessed only through
-// the public testing_* API declared in alloc.hpp.
+// Placed in an anonymous namespace to prevent external linkage — the symbol is
+// private to this TU and cannot be ODR-violated by other TUs.  The
+// glibre::testing_* getter/reset functions defined below reach it via
+// enclosing-scope name lookup (the anonymous namespace is nested inside
+// namespace glibre, so both the counter and the functions sharing that
+// enclosing scope can see it directly without qualification).
 namespace {
 std::atomic<std::uint64_t> testing_call_count_{0};
 }  // namespace
--- a/tests/core/per_context_allocator/CMakeLists.txt
+++ b/tests/core/per_context_allocator/CMakeLists.txt
@@ -45,13 +45,13 @@ set_target_properties(glibre-core-per-context-allocator-tests PROPERTIES CXX_MOD
 # when GLIBRE_ALLOC_STRICT=1".  Unit tests exercise the strict-mode path so
 # that the ceiling contract is verified at CI time.
 #
-# GLIBRE_TESTING=1: enables the register_allocator() call counter for the
-# per_context_allocator_register_fires_on_construction test (plan #990).
-# The counter and its reset/query API are gated behind this flag so they
-# never appear in production or shipping-mode builds.
+# GLIBRE_TESTING is NOT redefined here; it propagates transitively from the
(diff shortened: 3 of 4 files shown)


core/include/glibre/alloc.hpp
```diff
@@ -235,6 +235,42 @@ class PerContextAllocator {
...
+// the same thread).
+[[nodiscard]] std::uint64_t testing_register_allocator_call_count() noexcept;
+
+// Resets the call counter to zero.  Call before each test assertion that
+// depends on a specific count so that prior constructions (e.g. from other
+// test cases or from construction of static/global allocators) do not bleed
+// into the assertion.
+void testing_reset_register_allocator_call_count() noexcept;
+
+#endif  // GLIBRE_TESTING
```
cjhowe-us, line 272: severity:MED location:core/include/glibre/alloc.hpp:274-283 (immediately below this newly added block) problem:Stale comment block titled "register_allocator observability (MED-7, deferred to plan #241)" remains in the public header at lines 274-283 even though plan #990 (this PR) IS the iteration that implements MED-7. It directly contradicts the new "Testing observability hooks (GLIBRE_TESTING only — plan #990)" block immediately above (this added section ending at line 272). Future readers will be confused about whether observability is deferred or shipped. fix:Delete the stale block at lines 274-283. The new GLIBRE_TESTING block above supersedes it. Optionally replace with a one-line note that AllocatorRegistry enumeration remains deferred to plan #241 (only the TU-global counter is shipped here).

tests/core/per_context_allocator/CMakeLists.txt
```diff
@@ -46,8 +44,14 @@ set_target_properties(glibre-core-per-context-allocator-tests PROPERTIES CXX_MOD
...
 # when GLIBRE_ALLOC_STRICT=1".  Unit tests exercise the strict-mode path so
 # that the ceiling contract is verified at CI time.
+#
+# GLIBRE_TESTING=1: enables the register_allocator() call counter for the
+# per_context_allocator_register_fires_on_construction test (plan #990).
+# The counter and its reset/query API are gated behind this flag so they
+# never appear in production or shipping-mode builds.
 target_compile_definitions(glibre-core-per-context-allocator-tests PRIVATE
     GLIBRE_ALLOC_STRICT=1
+    GLIBRE_TESTING=1
```
cjhowe-us: severity:LOW location:tests/core/per_context_allocator/CMakeLists.txt:52-55 problem:Adding GLIBRE_TESTING=1 as a PRIVATE definition on glibre-core-per-context-allocator-tests is redundant. core/CMakeLists.txt:130-132 already PUBLIC-defines GLIBRE_TESTING=1 on glibre-core when GLIBRE_BUILD_TESTS=ON (default ON), so the test target receives it transitively via target_link_libraries(... PRIVATE glibre::core). PR #1004 (which fixed the same instrumentation issue for FrameLoop) explicitly removed this kind of PRIVATE flag — see core/CMakeLists.txt:120 comment "no PRIVATE definition needed on the test target". Carrying the redundant definition risks future drift if GLIBRE_BUILD_TESTS semantics change and creates a false impression that the test target is the source of truth for the macro. fix:Remove the GLIBRE_TESTING=1 line from target_compile_definitions and rely on the PUBLIC propagation from glibre::core. Keep GLIBRE_ALLOC_STRICT=1 as-is (that one IS test-target-local). Update the surrounding comment to point to core/CMakeLists.txt:130-132 as the authoritative source.

core/src/alloc/per_context_allocator.cpp
```diff
@@ -173,17 +173,50 @@ std::uint64_t PerContextAllocator::bytes_used() const noexcept {
...
+// testing_reset_register_allocator_call_count() before each assertion.
 // ---------------------------------------------------------------------------
 
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+// TU-global atomic counter.  Incremented on every register_allocator() call.
+// Declared in anonymous namespace to avoid ODR concerns; accessed only through
+// the public testing_* API declared in alloc.hpp.
+namespace {
+std::atomic<std::uint64_t> testing_call_count_{0};
+}  // namespace
```
cjhowe-us, line 194: severity:LOW location:core/src/alloc/per_context_allocator.cpp:184-191 problem:The TU-global counter is correctly placed in an anonymous namespace nested inside namespace glibre, and the public testing_* getter/reset functions at lines 212-218 reach it via enclosing-scope name lookup. This works correctly but the indirection is non-obvious — a reader unfamiliar with the file might wonder how a glibre:: function reaches a symbol declared in an anon-ns. fix:Add a one-line clarifying comment near the counter definition: "// Reachable from glibre::testing_* getter/reset definitions below via enclosing-scope name lookup; the anonymous namespace prevents external linkage so the symbol is private to this TU." Documentation-only; no code change needed.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0
notes:Round-1 (coverage + correctness) lens. PR fulfills declared scope of issue #990: the per_context_allocator_register_fires_on_construction Catch2 test exists and asserts three scenarios (explicit-ceiling ctor, tag-only ctor, three-instance count across two distinct ContextTags), each preceded by testing_reset_register_allocator_call_count() so TU-global counter pollution from sibling tests cannot perturb the assertions. Catch2 default sequential test ordering plus the per-scenario reset makes cross-test isolation robust. The chosen approach (option 1, GLIBRE_TESTING-gated atomic counter in register_allocator) is pragmatic given AllocatorRegistry from #241 is actually PerfBudget with no enumeration API — this is correctly explained in the PR body and the new alloc.hpp comment block. Conventional-Commit subject is correct (test(core): allocator registry observability), Closes #990 keyword is present in the PR body, and #990's DoD block (pr_merged_closes_self / unit_test_named: per_context_allocator_register_fires_on_construction / workflow_passed: ci.yml) is fully satisfied. CI is green on the latest commit 4268fd1: build, lint, dep-check, spec-check, automated-review all pass; e2e-* are skipping as expected for this non-e2e change. Memory ordering (relaxed) is appropriate for single-threaded tests on the same thread that performed construction. The delegating-constructor in PerContextAllocator(ContextTag) correctly invokes register_allocator exactly once (via the explicit-ceiling target ctor body), so scenario B's "== 1u" and scenario C's "== 3" assertions hold. No HIGH findings. One MED: alloc.hpp lines 274-283 still carry a "deferred to plan #241" stale comment that contradicts the new GLIBRE_TESTING block at lines 238-272 — should be deleted or replaced with a forward note that AllocatorRegistry enumeration (not the counter) remains deferred. Two LOW: redundant PRIVATE GLIBRE_TESTING=1 in test CMakeLists despite #1004 having made the macro PUBLIC on glibre-core; and a documentation-only opportunity to clarify the anonymous-namespace counter's enclosing-scope reachability from the namespace-glibre testing_* functions. No blockers; round-2 (cohesion + SRP) can proceed.

core/include/glibre/alloc.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1f3fc19

Deleted the stale 10-line deferral block at lines 274-283. Replaced with a one-line note: the GLIBRE_TESTING counter (plan #990) is now shipped; AllocatorRegistry enumeration remains deferred to plan #241. Future readers now see an accurate summary immediately below the new GLIBRE_TESTING block rather than a contradicting deferral notice.

tests/core/per_context_allocator/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1f3fc19

Removed GLIBRE_TESTING=1 from PRIVATE target_compile_definitions. Updated surrounding comment to point at core/CMakeLists.txt:130-132 as the authoritative source of GLIBRE_TESTING, matching the pattern established by PR #1004 for FrameLoop. GLIBRE_ALLOC_STRICT=1 remains as-is (test-target-local).

core/src/alloc/per_context_allocator.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:1f3fc19

Added a clarifying comment near the counter definition explaining: (a) the anonymous namespace prevents external linkage so the symbol is private to this TU, and (b) the glibre::testing_* getter/reset functions reach the counter via enclosing-scope name lookup because both the anon-ns and the functions share the same enclosing namespace glibre scope.

core/include/glibre/alloc.hpp
```diff
@@ -236,16 +236,45 @@ class PerContextAllocator {
...
+// testing surface leaks into shipping code.
+//
+// Usage (in a GLIBRE_TESTING=1 test target):
+//   testing_reset_register_allocator_call_count();   // zero the counter
+//   glibre::PerContextAllocator alloc{...};
+//   REQUIRE(testing_register_allocator_call_count() == 1);
 // ---------------------------------------------------------------------------
 
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+
```
cjhowe-us, line 256: severity:LOW location:core/include/glibre/alloc.hpp:256 problem:GLIBRE_TESTING gating idiom diverges from precedent in this codebase. alloc.hpp + per_context_allocator.cpp use `#if defined(GLIBRE_TESTING) && GLIBRE_TESTING` (5 occurrences across the two files), while the established precedent in core/include/glibre/core/frame_loop.hpp:159 + core/src/frame_loop.cpp:232,241 uses bare `#ifdef GLIBRE_TESTING`. core/CMakeLists.txt:131 defines `GLIBRE_TESTING=1` PUBLIC unconditionally when GLIBRE_BUILD_TESTS=ON, so both forms are functionally equivalent today; the `#if defined && val` form only matters if the macro could ever be defined to 0. Since the macro is project-internal and only ever set to 1 (never 0), the longer form adds noise without changing behavior. This is a cohesion/consistency observation, not a bug. fix:Either (a) change the 5 occurrences in alloc.hpp + per_context_allocator.cpp to bare `#ifdef GLIBRE_TESTING` to match the FrameLoop precedent, or (b) defer to a future cross-cutting cleanup pass that aligns all GLIBRE_TESTING gates project-wide. (b) is fine if the team prefers a single sweep. Round-2 cohesion lens flag only — not a merge blocker.

tests/core/per_context_allocator/per_context_allocator_test.cpp
```diff
@@ -254,3 +255,62 @@ TEST_CASE("per_context_allocator_threadsafe_allocations", "[core][alloc]") {
...
+//
+// GLIBRE_TESTING=1 is set by CMakeLists for the strict-mode test target that
+// includes this file.  The counter API is not available when GLIBRE_TESTING is
+// not defined so this section is compiled conditionally.
+// ===========================================================================
+
+#if defined(GLIBRE_TESTING) && GLIBRE_TESTING
+
+TEST_CASE("per_context_allocator_register_fires_on_construction", "[core][alloc]") {
+    // --- Scenario A: explicit-ceiling constructor fires exactly once ----------
```
cjhowe-us: severity:LOW location:tests/core/per_context_allocator/per_context_allocator_test.cpp:285 problem:The single TEST_CASE bundles three scenarios (A: explicit-ceiling ctor, B: tag-only ctor, C: N-instance count) using comment markers and per-scenario `testing_reset_register_allocator_call_count()` calls. Scenarios are clearly isolated via the resets, so this is correct as-is. However, Catch2's SECTION feature would give cleaner per-scenario reporting in test output (each section appears as a distinct path in the failure summary instead of one TEST_CASE that may have failed on any of three CHECKs). Splitting into three TEST_CASEs would also work but would multiply the `[core][alloc]` discovery noise. fix:Optionally wrap each scenario in `SECTION("explicit-ceiling ctor") { ... }` etc. so the per-scenario isolation becomes lexical and CTest reports each section by name. Round-2 polish suggestion only — the dispatch prompt explicitly says either form is acceptable when scenarios are clearly isolated, which they are.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:2
carryover:0
notes:Round-2 (cohesion + SRP + seam quality) lens. All three R1 findings are ADDRESSED in 1f3fc19 with go-impl-respond replies recorded against each comment thread (MED-1 stale-deferral block, LOW-2 redundant PRIVATE GLIBRE_TESTING, LOW-3 anon-ns enclosing-scope docstring). CI on head 1f3fc19 is green: build, lint, dep-check, spec-check, automated-review all pass; e2e-* skipping is expected for a non-e2e change. Cohesion analysis: the testing seam (testing_register_allocator_call_count / testing_reset_register_allocator_call_count) is correctly modeled as two free functions at namespace glibre scope rather than as members of PerContextAllocator, because what is being observed is the free function register_allocator(), not any per-instance state. SRP holds: the counter belongs to the seam of register_allocator, not to an allocator instance. The TU-global atomic counter sits in an anonymous namespace inside namespace glibre in per_context_allocator.cpp, which prevents external linkage and ODR collisions; the testing_* functions reach it via enclosing-scope name lookup (now documented inline per R1 LOW-3). Public-API surface inside the GLIBRE_TESTING block is minimal: 2 functions, both prefixed testing_, both documented with thread-safety notes and a usage example. No type leakage. The CMakeLists clean-up (drop redundant PRIVATE GLIBRE_TESTING=1, rely on glibre::core PUBLIC propagation) matches the precedent established by PR #1004 for FrameLoop and now-correctly cites core/CMakeLists.txt:130-132 as the single source of truth. Test cohesion: the single TEST_CASE per_context_allocator_register_fires_on_construction bundles three scenarios (explicit-ceiling ctor, tag-only ctor, N-instance count across two ContextTags), each isolated by an explicit testing_reset_register_allocator_call_count() before the body — clear isolation per the dispatch prompt's tolerance. No HIGH or MED findings. Two LOW polish items recorded inline: (1) GLIBRE_TESTING gating idiom in alloc.hpp + per_context_allocator.cpp uses `#if defined(X) && X` while the established FrameLoop precedent uses bare `#ifdef X` (functionally equivalent given core/CMakeLists.txt only sets =1, never =0; consistency-only); (2) the three scenarios in the new TEST_CASE could be Catch2 SECTIONs for cleaner per-scenario test output, though the dispatch prompt explicitly accepts either form. Neither LOW is a merge blocker. Round-3 (topology + spec alignment + polish) can proceed.

core/include/glibre/alloc.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:5949121

Changed all 5  occurrences in alloc.hpp (1) and per_context_allocator.cpp (3 — counter definition, increment site, getter/reset block) plus per_context_allocator_test.cpp (1) to bare , matching the frame_loop.{hpp,cpp} precedent.

tests/core/per_context_allocator/per_context_allocator_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:5949121

Refactored the single TEST_CASE into three named SECTION blocks: 'explicit-ceiling ctor fires exactly once', 'tag-only ctor fires exactly once', and 'N instances produce exactly N calls'. Counter reset (testing_reset_register_allocator_call_count()) moved to the top of each SECTION so isolation is lexical and Catch2 reports each section by name in failure output. All 10 per_context_allocator tests pass locally (including test #90 which covers the three sections).

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:0
carryover:0
notes:Round-3 (topology + spec alignment + polish) lens. CI green on head 5202f11 (build, lint, dep-check, spec-check, automated-review all pass; e2e-* skipping as expected for this non-e2e change). All R1 findings (1M/2L) and R2 findings (2L) addressed in 1f3fc19 + 5949121 + 5202f11. DAG / topology coherent: #990 (this PR) → depends:#238 (PerContextAllocator already shipped, exposes register_allocator symbol) → cites perf-budget.md §Allocator Rules; #241 (AllocatorRegistry/PerfBudget) correctly identified as the deferred enumeration API and called out as the reason option-1 (GLIBRE_TESTING counter) was chosen over option-3 (real registry). #990's labels (type:plan, phase:mvp, pts:2, domain:core) and parent (#4) are correct per AGENTS.md leaf taxonomy. DoD on #990 is populated, machine-checkable, and satisfied by the diff: pr_merged_closes_self (Closes #990 keyword present in PR body), unit_test_named: per_context_allocator_register_fires_on_construction (TEST_CASE present in per_context_allocator_test.cpp; verifier uses grep -E so the #ifdef GLIBRE_TESTING gate does not affect the text-match assertion), workflow_passed: ci.yml (head ci.yml run 25612524789 = success). Spec alignment: CMakeLists, header docs, and test docs all cite reviews/decisions/perf-budget.md §Allocator Rules #1-3 verbatim and accurately; nothing in this PR alters the spec contract — it only tests an existing call-site. Polish complete: all 5 GLIBRE_TESTING gating sites in alloc.hpp + per_context_allocator.cpp converted to bare #ifdef GLIBRE_TESTING per FrameLoop precedent (R2 LOW-1); test refactored into 3 named SECTION blocks (explicit-ceiling ctor / tag-only ctor / N-instance count) with per-section reset (R2 LOW-2); anon-ns enclosing-scope reachability documented in per_context_allocator.cpp (R1 LOW-3); stale "deferred to plan #241" block in alloc.hpp replaced with a one-liner clarifying the counter ships here while AllocatorRegistry enumeration remains deferred (R1 MED-1); redundant PRIVATE GLIBRE_TESTING=1 dropped from test CMakeLists, with comment naming core/CMakeLists.txt:130-132 as the single source of truth (R1 LOW-2). Conventional-Commit subject correct (test(core): allocator registry observability). No new R3 findings; merge-ready.

PR #1011, opened 2026-05-09
refactor(infra): self-register engine roots into GLIBRE_ENGINE_ROOTS

### LINKED ISSUE #1005: [PLAN] refactor(infra): self-register engine roots into GLIBRE_ENGINE_ROOTS — invert compile-contract reachability check
## Domain

infra

## Context / Problem

`cmake/Compile.cmake::_glibre_run_core_exception_check()` uses a static
hardcoded list of engine root targets (`glibre-core`, `glibre-types`,
`glibre-runtime`, `glibre-shader`, etc.).  Every new bounded-context module
added to the build must remember to amend this list, or its targets become
silently unguarded.

This is a cohesion-SRP violation: the responsibility for knowing which targets
are engine roots is split between the owning CMakeLists and the check function.

Identified in round-2 review of PR #922 (MED-2), deferred from that PR as
out-of-scope.  See review comment ID 3213793734.

## Proposed Approach

Invert control: each `add_subdirectory(…)` for an engine context calls a
helper macro (`glibre_register_engine_root(target)`) that appends to a global
CMake property (`GLIBRE_ENGINE_ROOTS`).  `_glibre_run_core_exception_check()`
reads that property instead of a hard-coded list.

Shape:
```cmake
# In cmake/Compile.cmake — new helper
macro(glibre_register_engine_root target)
    set_property(GLOBAL APPEND PROPERTY GLIBRE_ENGINE_ROOTS "${target}")
endmacro()

# In _glibre_run_core_exception_check()
get_property(_engine_roots GLOBAL PROPERTY GLIBRE_ENGINE_ROOTS)
```

Each `core/CMakeLists.txt`, `data/CMakeLists.txt`, `plugins/*/CMakeLists.txt`
calls `glibre_register_engine_root(<target>)` after `add_library(…)`.

## Scope / Non-goals

- Touch only `cmake/Compile.cmake` and the per-context `CMakeLists.txt` files
  that need the registration call.
- No change to the reachability walk logic itself.
- No change to `glibre_target_exceptions()` helper.

## Unit Test Plan

- Existing fixture tests in `cmake/tests/CMakeLists_exceptions_test.cmake` must
  still pass after the refactor.
- Add a new fixture test that registers two roots, marks one as
  `GLIBRE_USES_EXCEPTIONS=TRUE`, and asserts configure fails.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_contains:
    path: cmake/Compile.cmake
    regex: glibre_register_engine_root
- file_contains:
    path: cmake/Compile.cmake
    regex: GLIBRE_ENGINE_ROOTS
- file_contains:
    path: core/CMakeLists.txt
    regex: glibre_register_engine_root
- workflow_passed: ci.yml
```

Refs #237 (parent plan that added the static list).

cjhowe-us:
## Summary

- Adds `glibre_register_engine_root(target)` macro to `cmake/Compile.cmake` that appends to a global `GLIBRE_ENGINE_ROOTS` CMake property
- Replaces the hardcoded static list in `_glibre_run_core_exception_check()` with `get_property(_engine_roots GLOBAL PROPERTY GLIBRE_ENGINE_ROOTS)` — inverting control so each engine subdirectory owns its own registration
- Adds `glibre_register_engine_root(<target>)` call after `add_library`/`add_executable` in: `core/CMakeLists.txt`, `data/CMakeLists.txt`, `plugins/shader/CMakeLists.txt`, `tools/foryc/CMakeLists.txt`, `examples/plugin-noop/CMakeLists.txt`
- Updates fixture `core-target` to call `glibre_register_engine_root(engine_module)` and assert `GLIBRE_ENGINE_ROOTS` contains the registered target (Assertion 3)
- Updates fixture `engine-reaches-exception-target` to call `glibre_register_engine_root(glibre-core)` so the WILL_FAIL test still fires `FATAL_ERROR` via the now-dynamic check
- Adds new fixture `engine-roots-property` with a dedicated test (`ci/cmake: engine_roots_property_self_registration`) that registers two mock roots and asserts both appear in `GLIBRE_ENGINE_ROOTS`
- iOS carve-out is preserved: early-`return()` in `core/` and `data/` prevents registration on iOS sub-builds, leaving `GLIBRE_ENGINE_ROOTS` empty and the deferred check a no-op

## Test plan

- [ ] All 169 CTest tests pass locally (verified: `100% tests passed, 0 tests failed out of 169`)
- [ ] `ci/cmake: fno_exceptions_set_on_core_targets` — still passes
- [ ] `ci/cmake: editor_ui_target_opts_into_exceptions` — still passes
- [ ] `ci/cmake: configure_fails_if_engine_path_reaches_exception_target` — still passes (WILL_FAIL)
- [ ] `ci/cmake: engine_roots_property_self_registration` — new test, passes
- [ ] CI green on push

Advances plan #1005 (identified in PR #922 round-2 review, MED-2).

Closes #1005

### DIFF
--- a/cmake/Compile.cmake
+++ b/cmake/Compile.cmake
@@ -59,14 +59,14 @@ endif()
 #
 #    Must be called AFTER add_library/add_executable so the target exists.
 # ---------------------------------------------------------------------------
-macro(glibre_register_engine_root target)
+function(glibre_register_engine_root target)
     if(NOT TARGET ${target})
         message(FATAL_ERROR
             "glibre_register_engine_root(${target}): target '${target}' is not yet "
             "defined; call after add_library/add_executable.")
     endif()
     set_property(GLOBAL APPEND PROPERTY GLIBRE_ENGINE_ROOTS "${target}")
-endmacro()
+endfunction()
 
 # ---------------------------------------------------------------------------
 # 3. Helper: glibre_target_exceptions(target)
--- a/cmake/Compile.cmake
+++ b/cmake/Compile.cmake
@@ -60,6 +60,11 @@ endif()
 #    Must be called AFTER add_library/add_executable so the target exists.
 # ---------------------------------------------------------------------------
 macro(glibre_register_engine_root target)
+    if(NOT TARGET ${target})
+        message(FATAL_ERROR
+            "glibre_register_engine_root(${target}): target '${target}' is not yet "
+            "defined; call after add_library/add_executable.")
+    endif()
     set_property(GLOBAL APPEND PROPERTY GLIBRE_ENGINE_ROOTS "${target}")
 endmacro()
 
--- a/cmake/Compile.cmake
+++ b/cmake/Compile.cmake
@@ -47,7 +47,24 @@ if(NOT TARGET glibre::compile_contract)
 endif()
 
 # ---------------------------------------------------------------------------
-# 2. Helper: glibre_target_exceptions(target)
+# 2. Helper: glibre_register_engine_root(target)
+#
+#    Self-registration macro: each engine CMakeLists.txt calls this after
+#    add_library(<target> ...) to append the target to the GLIBRE_ENGINE_ROOTS
+#    global CMake property.  _glibre_run_core_exception_check() reads that
+#    property at configure-end so it no longer needs a hardcoded list.
+#
+#    Callers: core/CMakeLists.txt, data/CMakeLists.txt, plugins/*/CMakeLists.txt,
+#    tools/foryc/CMakeLists.txt, examples/plugin-noop/CMakeLists.txt.
+#
+#    Must be called AFTER add_library/add_executable so the target exists.
+# ---------------------------------------------------------------------------
+macro(glibre_register_engine_root target)
+    set_property(GLOBAL APPEND PROPERTY GLIBRE_ENGINE_ROOTS "${target}")
+endmacro()
+
+# ---------------------------------------------------------------------------
+# 3. Helper: glibre_target_exceptions(target)
 #
 #    Opts <target> into exception + RTTI mode for ImGui / third-party interop.
 #    Strips the INTERFACE no-exceptions flags inherited from glibre-core
@@ -89,7 +106,7 @@ function(glibre_target_exceptions target)
 endfunction()
 
 # ---------------------------------------------------------------------------
-# 3. Configure-time reachability check.
+# 4. Configure-time reachability check.
 #
 #    Walks the transitive link closure of glibre-core (depth-first, cycle-
 #    safe) and errors out if any reachable target has GLIBRE_USES_EXCEPTIONS.
@@ -150,28 +167,15 @@ function(_glibre_check_exceptions_reachable_from_core root visited_var)
 endfunction()
 
 function(_glibre_run_core_exception_check)
-    # Walk from every known engine root so that plugin DSOs (which link
-    # glibre-types, not glibre-core) and runtime are also covered.
-    # Tools/editor/ui are excluded — they are the permitted carve-out.
-    # glibre-foryc is a host tool (serialiser codegen) and is also excluded.
-    # Roots that do not exist yet (stubs not yet added_subdirectory) are
-    # silently skipped via the TARGET guard inside the walker.
-    #
-    # STATIC LIST NOTE: This is a manually maintained list of engine roots.
-    # A cleaner SRP approach would invert control: each subdirectory
-    # self-registers into a GLIBRE_ENGINE_ROOTS property at define time,
-    # and this function reads that property.  Deferred to a follow-up plan;
-    # see issue #1005 (iterate-infra-compile-contract-self-register).
-    set(_engine_roots
-        glibre-core
-        glibre-types
-        glibre-runtime
-        glibre-shader        # plugins/shader — links glibre::core not glibre-types
-        glibre-shader-plugin # future SHARED plugin dylib target name
-        glibre-foryc         # host tool; also carries contract flags
-        glibre-plugin-noop   # reference plugin example
-        # add further plugin / context targets here as subdirs land
-    )
+    # Walk from every self-registered engine root so that plugin DSOs (which
+    # link glibre-types, not glibre-core) and host tools are also covered.
+    # Each engine CMakeLists.txt calls glibre_register_engine_root(<target>)
+    # after add_library/add_executable, appending to GLIBRE_ENGINE_ROOTS.
+    # Tools/editor/ui are excluded from registration — they are the permitted
+    # exception carve-out.
+    # Roots that do not exist (iOS sub-builds that early-return before
+    # add_library) will not be in GLIBRE_ENGINE_ROOTS and are silently absent.
+    get_property(_engine_roots GLOBAL PROPERTY GLIBRE_ENGINE_ROOTS)
     set(_visited "")
     foreach(_root IN LISTS _engine_roots)
         if(TARGET "${_root}")
--- a/cmake/tests/CMakeLists_exceptions_test.cmake
+++ b/cmake/tests/CMakeLists_exceptions_test.cmake
@@ -92,3 +92,25 @@ set_tests_properties(
         # silently masquerade as a PASS.
         FAIL_REGULAR_EXPRESSION "\\[glibre\\] Configure error: target '.*' has GLIBRE_USES_EXCEPTIONS=TRUE"
 )
+
+# ---------------------------------------------------------------------------
+# Test 4: engine_roots_property_self_registration
+#
+# Configures cmake/tests/fixtures/engine-roots-property which calls
+# glibre_register_engine_root() for two mock targets and asserts that
+# GLIBRE_ENGINE_ROOTS global property contains both registered targets.
+# Introduced by plan #1005 (invert control: self-register engine roots).
+# ---------------------------------------------------------------------------
+add_test(
+    NAME "ci/cmake: engine_roots_property_self_registration"
+    COMMAND "${_cmake_exe}"
+        -G "${_generator}"
+        -S "${_fixture_dir}/engine-roots-property"
+        -B "${CMAKE_CURRENT_BINARY_DIR}/cmake_test_engine_roots_property"
+        "-DGLIBRE_CMAKE_DIR=${_glibre_cmake_dir}"
+        "-DCMAKE_CXX_COMPILER=${_cxx_compiler}"
+)
+set_tests_properties("ci/cmake: engine_roots_property_self_registration" PROPERTIES
+    LABELS "cmake;infra"
+    PASS_REGULAR_EXPRESSION "PASS: GLIBRE_ENGINE_ROOTS contains both registered engine roots"
+)
--- a/cmake/tests/CMakeLists_exceptions_test.cmake
+++ b/cmake/tests/CMakeLists_exceptions_test.cmake
@@ -114,3 +114,31 @@ set_tests_properties("ci/cmake: engine_roots_property_self_registration" PROPERT
     LABELS "cmake;infra"
     PASS_REGULAR_EXPRESSION "PASS: GLIBRE_ENGINE_ROOTS contains both registered engine roots"
 )
+
+# ---------------------------------------------------------------------------
+# Test 5: engine_roots_property_rejects_exception_root
+#
+# Configures cmake/tests/fixtures/engine-roots-property-rejects-exceptions
+# which registers two roots and marks one GLIBRE_USES_EXCEPTIONS=TRUE.
+# The deferred check must FATAL_ERROR because the exception-flagged root is
+# itself in GLIBRE_ENGINE_ROOTS — negative coverage for plan #1005 / #237.
+# ---------------------------------------------------------------------------
+add_test(
+    NAME "ci/cmake: engine_roots_property_rejects_exception_root"
+    COMMAND "${_cmake_exe}"
+        -G "${_generator}"
+        -S "${_fixture_dir}/engine-roots-property-rejects-exceptions"
+        -B "${CMAKE_CURRENT_BINARY_DIR}/cmake_test_engine_roots_rejects_exc"
+        "-DGLIBRE_CMAKE_DIR=${_glibre_cmake_dir}"
+        "-DCMAKE_CXX_COMPILER=${_cxx_compiler}"
+)
+set_tests_properties(
+    "ci/cmake: engine_roots_property_rejects_exception_root"
+    PROPERTIES
+        LABELS "cmake;infra"
+        WILL_FAIL TRUE
+        # Require the specific [glibre] FATAL_ERROR so that unrelated configure
+        # failures (missing compiler, bad generator, etc.) do not silently
+        # masquerade as a PASS.
+        FAIL_REGULAR_EXPRESSION "\\[glibre\\] Configure error: target '.*' has GLIBRE_USES_EXCEPTIONS=TRUE"
+)
--- a/cmake/tests/CMakeLists_exceptions_test.cmake
+++ b/cmake/tests/CMakeLists_exceptions_test.cmake
@@ -118,7 +118,7 @@ set_tests_properties("ci/cmake: engine_roots_property_self_registration" PROPERT
 # ---------------------------------------------------------------------------
 # Test 5: engine_roots_property_rejects_exception_root
 #
-# Configures cmake/tests/fixtures/engine-roots-property-rejects-exceptions
+# Configures cmake/tests/fixtures/engine-roots-property-rejects-exception-root
 # which registers two roots and marks one GLIBRE_USES_EXCEPTIONS=TRUE.
 # The deferred check must FATAL_ERROR because the exception-flagged root is
 # itself in GLIBRE_ENGINE_ROOTS — negative coverage for plan #1005 / #237.
@@ -127,7 +127,7 @@ add_test(
     NAME "ci/cmake: engine_roots_property_rejects_exception_root"
     COMMAND "${_cmake_exe}"
         -G "${_generator}"
-        -S "${_fixture_dir}/engine-roots-property-rejects-exceptions"
+        -S "${_fixture_dir}/engine-roots-property-rejects-exception-root"
         -B "${CMAKE_CURRENT_BINARY_DIR}/cmake_test_engine_roots_rejects_exc"
         "-DGLIBRE_CMAKE_DIR=${_glibre_cmake_dir}"
         "-DCMAKE_CXX_COMPILER=${_cxx_compiler}"
--- a/cmake/tests/fixtures/engine-roots-property-rejects-exception-root/CMakeLists.txt
+++ b/cmake/tests/fixtures/engine-roots-property-rejects-exception-root/CMakeLists.txt
@@ -1,11 +1,11 @@
 cmake_minimum_required(VERSION 3.30)
-project(glibre_test_engine_roots_rejects_exceptions LANGUAGES CXX)
+project(glibre_test_engine_roots_rejects_exception_root LANGUAGES CXX)
 
 list(APPEND CMAKE_MODULE_PATH "${GLIBRE_CMAKE_DIR}")
 include(Compile)
 
 # ---------------------------------------------------------------------------
-# Fixture: engine-roots-property-rejects-exceptions
+# Fixture: engine-roots-property-rejects-exception-root
 #
 # Negative-path coverage for plan #1005 / #237 §"Register engine roots".
 #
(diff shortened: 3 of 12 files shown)


cmake/tests/fixtures/engine-roots-property/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+        "ASSERTION FAILED: mock_engine_b not in GLIBRE_ENGINE_ROOTS. "
+        "Got: ${_roots}"
+    )
+endif()
+
+message(STATUS "PASS: GLIBRE_ENGINE_ROOTS contains both registered engine roots")
+
+# The deferred _glibre_run_core_exception_check() (installed by include(Compile))
+# will walk these roots at configure-end.  Since neither mock target has
+# GLIBRE_USES_EXCEPTIONS=TRUE, configure must complete without FATAL_ERROR.
```
cjhowe-us, line 53: severity:MED location:cmake/tests/fixtures/engine-roots-property/CMakeLists.txt:48 problem:Plan #1005 explicitly specified the new fixture should "register two roots, mark one as GLIBRE_USES_EXCEPTIONS=TRUE, and assert configure fails." This fixture only verifies the property-population path and the happy-path walk; it never exercises the negative path through the new self-registration mechanism in a single fixture. Negative coverage now lives only in engine-reaches-exception-target/. fix:Add a third mock target to engine-roots-property/CMakeLists.txt that calls glibre_target_exceptions(mock_engine_c) AND glibre_register_engine_root(mock_engine_c), then mark the test set_tests_properties WILL_FAIL TRUE so configure-end FATAL_ERROR is asserted. Alternatively, document in the fixture comment that the negative path is intentionally split into engine-reaches-exception-target/ for SRP-per-fixture and update plan #1005 unit-test plan to match.

cmake/Compile.cmake
```diff
@@ -47,7 +47,24 @@ if(NOT TARGET glibre::compile_contract)
...
+#    property at configure-end so it no longer needs a hardcoded list.
+#
+#    Callers: core/CMakeLists.txt, data/CMakeLists.txt, plugins/*/CMakeLists.txt,
+#    tools/foryc/CMakeLists.txt, examples/plugin-noop/CMakeLists.txt.
+#
+#    Must be called AFTER add_library/add_executable so the target exists.
+# ---------------------------------------------------------------------------
+macro(glibre_register_engine_root target)
+    set_property(GLOBAL APPEND PROPERTY GLIBRE_ENGINE_ROOTS "${target}")
+endmacro()
```
cjhowe-us: severity:LOW location:cmake/Compile.cmake:60-64 problem:Doc comment claims "Must be called AFTER add_library/add_executable so the target exists" but the macro only appends ${target} as a string to a global property and never resolves it as a CMake target. The ordering precondition is therefore unenforced — a caller could register a misspelled or not-yet-defined target and discover the bug only at the deferred reachability walk. fix:Either (a) tighten the doc to clarify the ordering is convention not correctness, or (b) add a guard inside the macro: if(NOT TARGET ${target}) message(FATAL_ERROR "glibre_register_engine_root: ${target} is not a known CMake target — call after add_library/add_executable") endif(). Option (b) catches typos and out-of-order calls at the registration site rather than at configure-end via the looser TARGET guard in _glibre_run_core_exception_check.

cmake/Compile.cmake
```diff
@@ -150,28 +167,15 @@ function(_glibre_check_exceptions_reachable_from_core root visited_var)
...
-    )
+    # Walk from every self-registered engine root so that plugin DSOs (which
+    # link glibre-types, not glibre-core) and host tools are also covered.
+    # Each engine CMakeLists.txt calls glibre_register_engine_root(<target>)
+    # after add_library/add_executable, appending to GLIBRE_ENGINE_ROOTS.
+    # Tools/editor/ui are excluded from registration — they are the permitted
+    # exception carve-out.
+    # Roots that do not exist (iOS sub-builds that early-return before
+    # add_library) will not be in GLIBRE_ENGINE_ROOTS and are silently absent.
+    get_property(_engine_roots GLOBAL PROPERTY GLIBRE_ENGINE_ROOTS)
```
cjhowe-us, line 183: severity:LOW location:cmake/Compile.cmake:170-181 problem:The pre-refactor comment block called out "glibre-foryc is a host tool (serialiser codegen) and is also excluded" — directly contradicting the static list which DID include it. The refactor preserves the (correct) inclusion via self-registration but drops the explanation, so a future reader cannot tell why a host tool is in GLIBRE_ENGINE_ROOTS. fix:Add a one-line comment on glibre_register_engine_root(glibre-foryc) in tools/foryc/CMakeLists.txt explaining that foryc itself carries the no-exceptions contract (no STL throw paths in its parser), so it is registered as an engine root for the contract-reachability walk despite being a host tool. Alternatively, expand the §4 docblock in Compile.cmake to clarify "engine root" = "any target that must satisfy the no-exceptions contract", which includes host tools whose code shares the engine ABI discipline.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0
notes:Round-1 (coverage + correctness) lens. Diff fulfils plan #1005 declared scope: glibre_register_engine_root macro added in proper section of cmake/Compile.cmake (§2, between contract definition and exception helper), all sections renumbered consistently; _glibre_run_core_exception_check() reads from GLIBRE_ENGINE_ROOTS via get_property; all five live engine roots (glibre-core, glibre-types, glibre-shader, glibre-foryc, glibre-plugin-noop) self-register after their respective add_library calls; iOS carve-out preserved via early-return precedence in core/ data/ plugins/shader (registration call lives below the iOS guard); existing fixture engine-reaches-exception-target updated to use the new self-registration so the WILL_FAIL negative path still fires; new fixture engine-roots-property + Assertion 3 in core-target both exercise the property mechanism. Conv-Commit subject refactor(infra): correct; PR body contains Closes #1005; all five DoD assertions on issue #1005 trace cleanly to diff content; head SHA db6772d1 CI fully green (build/lint/spec-check/dep-check/automated-review SUCCESS, e2e-* legitimately SKIPPED for infra-only diff). One MED finding: engine-roots-property fixture omits the negative-path assertion plan #1005 explicitly specified ("mark one as GLIBRE_USES_EXCEPTIONS=TRUE and assert configure fails") — that coverage now lives only in the separate engine-reaches-exception-target fixture; defensible as SRP-per-fixture but should be reconciled with the plan. Two LOW findings on doc precision (macro precondition unenforced; foryc's host-tool-but-still-engine-root rationale dropped from comments). No HIGH blockers; safe to proceed to round-2 cohesion lens.

cmake/tests/fixtures/engine-roots-property/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:8984410 (fixture) + 3f8a574 (guard)

MED-1 addressed. Added cmake/tests/fixtures/engine-roots-property-rejects-exceptions/CMakeLists.txt — registers two roots, marks one GLIBRE_USES_EXCEPTIONS=TRUE, deferred check fires FATAL_ERROR. Wired as Test 5 in CMakeLists_exceptions_test.cmake with WILL_FAIL TRUE and FAIL_REGULAR_EXPRESSION matching the [glibre] message. Verified cmake exits 1 locally.

Also addressed LOW-2 in the same push (commit 3f8a574): added if(NOT TARGET ${target}) guard so calling glibre_register_engine_root before add_library/add_executable is an immediate FATAL_ERROR at the call site rather than a silent no-op.

cmake/Compile.cmake
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3f8a574

LOW-2 addressed. Added if(NOT TARGET ${target}) / message(FATAL_ERROR ...) guard to glibre_register_engine_root in cmake/Compile.cmake. Verified locally: cmake exits 1 with message 'glibre_register_engine_root(nonexistent_target): target ... is not yet defined; call after add_library/add_executable.' when called before target creation.

cmake/Compile.cmake
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:3b06fd4

LOW-3 addressed. Expanded the comment block next to glibre_register_engine_root(glibre-foryc) in tools/foryc/CMakeLists.txt to state that foryc is a host tool but carries the same no-exceptions contract as engine targets via glibre::compile_contract, and that registration ensures the deferred check covers the codegen pipeline — restoring the lost rationale from the pre-refactor comment block.

cmake/tests/fixtures/engine-roots-property-rejects-exception-root/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+cmake_minimum_required(VERSION 3.30)
+project(glibre_test_engine_roots_rejects_exceptions LANGUAGES CXX)
+
+list(APPEND CMAKE_MODULE_PATH "${GLIBRE_CMAKE_DIR}")
+include(Compile)
+
+# ---------------------------------------------------------------------------
```
cjhowe-us, line 7: severity:MED location:cmake/tests/fixtures/engine-roots-property-rejects-exceptions/CMakeLists.txt:7 problem:This fixture overlaps in responsibility with cmake/tests/fixtures/engine-reaches-exception-target — both register an engine root carrying GLIBRE_USES_EXCEPTIONS=TRUE and assert the deferred check fires FATAL_ERROR. The differentiating axis (this fixture: registered-but-not-reachable-via-link; the other: registered-and-reachable-via-glibre-core link interface) is left implicit in a header comment that talks about "even if it is not reachable via another root link interface" without naming the sibling fixture. Two-reasons-to-change risk: any change to the FATAL_ERROR token or to glibre_target_exceptions semantics now has to be propagated across two near-identical fixtures, and a future reader has to diff them to discover what each is actually proving. fix:Add a one-liner under the fixture comment header naming the sibling and the discriminating property, e.g. "Distinct from engine-reaches-exception-target: there mock_root_exc is NOT in GLIBRE_ENGINE_ROOTS but IS reachable via glibre-core link; here it IS in GLIBRE_ENGINE_ROOTS but NOT reachable via any other root linking it." Cheap; locks the SRP boundary down for future readers.

cmake/tests/CMakeLists_exceptions_test.cmake
```diff
@@ -92,3 +92,53 @@ set_tests_properties(
...
+    LABELS "cmake;infra"
+    PASS_REGULAR_EXPRESSION "PASS: GLIBRE_ENGINE_ROOTS contains both registered engine roots"
+)
+
+# ---------------------------------------------------------------------------
+# Test 5: engine_roots_property_rejects_exception_root
+#
+# Configures cmake/tests/fixtures/engine-roots-property-rejects-exceptions
+# which registers two roots and marks one GLIBRE_USES_EXCEPTIONS=TRUE.
+# The deferred check must FATAL_ERROR because the exception-flagged root is
```
cjhowe-us, line 123: severity:LOW location:cmake/tests/CMakeLists_exceptions_test.cmake:123 problem:Naming drift between fixture directory and test name. CTest test = engine_roots_property_rejects_exception_root (singular "root"). Fixture directory = engine-roots-property-rejects-exceptions (plural "exceptions"). Both are referring to the same scenario; the asymmetry will make grep + cross-reference harder later. fix:Pick one. Recommended: rename the directory to engine-roots-property-rejects-exception-root (matches the test name and the failure semantics — the deferred check rejects A root that uses exceptions, not exceptions in general). Update the -S path on line 126 to match.

cmake/Compile.cmake
```diff
@@ -47,7 +47,29 @@ if(NOT TARGET glibre::compile_contract)
...
+#    property at configure-end so it no longer needs a hardcoded list.
+#
+#    Callers: core/CMakeLists.txt, data/CMakeLists.txt, plugins/*/CMakeLists.txt,
+#    tools/foryc/CMakeLists.txt, examples/plugin-noop/CMakeLists.txt.
+#
+#    Must be called AFTER add_library/add_executable so the target exists.
+# ---------------------------------------------------------------------------
+macro(glibre_register_engine_root target)
+    if(NOT TARGET ${target})
+        message(FATAL_ERROR
```
cjhowe-us, line 64: severity:LOW location:cmake/Compile.cmake:64 problem:glibre_register_engine_root is defined as macro(...) but its only behaviour is the NOT TARGET guard plus set_property(GLOBAL APPEND ...). It does not need caller-scope variable leakage. Macros silently leak any future local variables (e.g. a temp introduced for diagnostic logging) into callers — since callers are the engine roots themselves, that pollution would be subtle. fix:Convert to function(...) — the semantics of set_property(GLOBAL APPEND PROPERTY ...) are scope-independent so the call-site behaviour is unchanged. If macro was an intentional choice (e.g. to keep CMAKE_CURRENT_LIST_FILE / CMAKE_CURRENT_SOURCE_DIR meaningful for future diagnostics) document it in a one-line comment so the next reader does not invert it.

cmake/Compile.cmake
```diff
@@ -47,7 +47,29 @@ if(NOT TARGET glibre::compile_contract)
 endif()
 
 # ---------------------------------------------------------------------------
-# 2. Helper: glibre_target_exceptions(target)
+# 2. Helper: glibre_register_engine_root(target)
```
cjhowe-us, line 50: severity:DEFER location:cmake/Compile.cmake:50 problem:File is named Compile.cmake but its sections are: §1 compile_contract target, §2 (NEW HERE) glibre_register_engine_root, §3 glibre_target_exceptions, §4 deferred reachability check. Three of the four sections are not about "compile" — they are about engine-root topology + exception contract enforcement. This PR adds another non-compile section, deepening the existing name/content drift. Cohesion smell, but a rename is wider than this PR scope. fix:Defer to a follow-up rename PR. Recommend cmake/EngineContract.cmake or split into cmake/CompileContract.cmake (§1) + cmake/EngineRoots.cmake (§2 + §4) + cmake/Exceptions.cmake (§3). Tracking suggestion: open a chore(infra) leaf under the same epic that owns this PR; small, mechanical, no functional change.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
  DEFER:1
carryover:0
ci:green (5 success / 2 conditionally skipped on head 3b06fd4babf4fb46c5331f814fc3cb6e71460c4f — build, lint, dep-check, spec-check, automated-review all PASS; e2e-core and e2e-shader skipped by design for non-source-touching PRs)
notes:R2 lens (cohesion + SRP + seam quality). All R1 findings are addressed in 3b06fd4: glibre_register_engine_root has the NOT TARGET / FATAL_ERROR guard (R1 LOW-2), the foryc registration carries the elaborated rationale comment (R1 LOW-3), and Test 5 is wired against the existing FAIL_REGULAR_EXPRESSION token used by Test 3 (R1 MED-1). Macro signature glibre_register_engine_root(<target>) is minimal and predictable — fail-fast at config-time on undefined targets is correct behaviour, and the GLIBRE_ENGINE_ROOTS global property gives clean per-configure isolation for the fixture sub-projects (each fixture invokes a fresh cmake -S/-B). The two new fixtures cleanly mirror the established cmake/tests/fixtures/ pattern and the harness wiring follows the same idiom as Tests 1-3. One MED finding worth resolving: the new rejects-exceptions fixture overlaps in responsibility with the pre-existing engine-reaches-exception-target fixture (both register an exception-flagged root and assert FATAL_ERROR) and the discriminating axis is left implicit in the comment — recommend a one-line "distinct from <sibling>" header to lock the SRP boundary down. Two LOWs: macro-vs-function for glibre_register_engine_root (function would be safer default; document the choice if intentional) and singular-vs-plural drift between test name (rejects_exception_root) and fixture dirname (rejects-exceptions). DEFER on cmake/Compile.cmake misnomer (pre-existing — file owns 4 sections, only 1 about "compile" — out of scope for this PR; recommend a follow-up rename leaf). DoD on issue #1005 is populated; PR body has Closes #1005; CI green on true HEAD. No HIGH findings; no carryover from R1.

cmake/tests/fixtures/engine-roots-property-rejects-exception-root/CMakeLists.txt
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:666a6b5

MED-A addressed. Added discriminator one-liners to the headers of all three relevant fixtures:

- engine-roots-property: notes it covers the happy path; sibling covers negative path.
- engine-roots-property-rejects-exception-root (renamed from engine-roots-property-rejects-exceptions per LOW-B): notes that mock_root_exc IS in GLIBRE_ENGINE_ROOTS but NOT linked by any other registered root.
- engine-reaches-exception-target: notes that bad_exception_lib is NOT in GLIBRE_ENGINE_ROOTS but IS reachable via glibre-core INTERFACE link.

Both axes of the deferred check are now explicitly named in each fixture header so a future reader can distinguish them without diffing.

cmake/Compile.cmake
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:666a6b5

LOW-A addressed. Converted macro(glibre_register_engine_root target) to function(glibre_register_engine_root target) in cmake/Compile.cmake. set_property(GLOBAL APPEND PROPERTY ...) is scope-independent so call-site behaviour is unchanged. The function boundary prevents any future temp variables introduced for diagnostics from leaking into callers.

cmake/tests/CMakeLists_exceptions_test.cmake
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:666a6b5

LOW-B addressed. Renamed cmake/tests/fixtures/engine-roots-property-rejects-exceptions/ to cmake/tests/fixtures/engine-roots-property-rejects-exception-root/ (singular, matching the CTest test name engine_roots_property_rejects_exception_root). Updated:
- -S path in CMakeLists_exceptions_test.cmake Test 5 block
- project() name in fixture CMakeLists.txt from glibre_test_engine_roots_rejects_exceptions to glibre_test_engine_roots_rejects_exception_root
- Fixture header comment to match new dirname

cmake/Compile.cmake
cjhowe-us: responder:go-impl-respond round:2 decision:DEFER reasoning:Compile.cmake misnomer is pre-existing and a rename touches all include(Compile) call sites across engine CMakeLists — wider than this PR's declared scope (self-register engine roots). proposal:Opened [SPIKE] iterate-infra-compile-cmake-rename as issue #1013. Issue captures the rename options (single rename to cmake/EngineContract.cmake vs split into CompileContract + EngineRoots + Exceptions) and a Definition of Done checklist.

cjhowe-us, line 50: severity:LOW location:issue-#1013 problem:Spike #1013 (opened in R2 to track the cmake/Compile.cmake rename DEFER) carries only `type:spike` — no `phase:` (mvp/post-mvp/long-term) and no `domain:` label. Per AGENTS.md every issue (plan or spike) should carry phase + domain so it sorts correctly in plan-aggregation queries. Compare #1005 which has `phase:mvp` + `domain:core`. fix:Add `phase:mvp` (or `phase:post-mvp` if the rename is non-blocking) and `domain:core` (or `domain:infra` if such a label exists) to issue #1013. Topology-only nit; does not block this PR merging.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0
ci:green (head 666a6b59c75be0d0398e67d886356263fbf3dcc0 — build/lint/dep-check/spec-check/automated-review all SUCCESS; e2e-core and e2e-shader conditionally skipping by design for non-source-touching infra PR)

notes:R3 lens (topology + spec alignment + polish). All R2 findings resolved cleanly in 666a6b5: glibre_register_engine_root is now a function (LOW-A), fixture dir renamed to singular engine-roots-property-rejects-exception-root matching test name (LOW-B), all three fixture headers (engine-roots-property, engine-roots-property-rejects-exception-root, engine-reaches-exception-target) carry one-line discriminator comments naming the sibling and the discriminating axis (registered-not-reachable vs reachable-via-link) (MED-A), DEFER on Compile.cmake rename properly extracted to spike #1013.

DAG / topology: parent plan #1005 (type:plan, phase:mvp, pts:1, domain:core) closed by this PR via "Closes #1005"; follow-up spike #1013 opened for the deferred rename concern. DoD block on #1005 is populated and machine-checkable (5 assertions: pr_merged_closes_self, 2x file_contains on cmake/Compile.cmake, 1x file_contains on core/CMakeLists.txt, workflow_passed:ci.yml) — all five assertions trace cleanly to diff content and CI state. Conv-Commit subjects on all 5 commits correct (refactor / test / fix / docs / refactor with infra scope). PR body retains "Closes #1005" keyword for dod-verify.

Spec alignment: reviews/decisions/error-model.md Decision #3 reference preserved in cmake/Compile.cmake header; deferred reachability check semantics unchanged (only source of _engine_roots inverted); iOS sub-build carve-out preserved via early-return precedence (iOS configures don't register, GLIBRE_ENGINE_ROOTS empty, deferred check no-op).

Polish: every fixture has discriminator comment, no naming drift, function vs macro consistent — glibre_register_engine_root is the only new helper and it's a function; glibre_target_exceptions remains a function; macro is unused. Comments on all five engine-root call sites (core, data, shader, foryc, plugin-noop) reference plan #1005 and Compile.cmake §4 consistently.

One new R3 finding (LOW): spike #1013 was opened with only `type:spike` label — missing `phase:` and `domain:` per AGENTS.md issue-labelling convention. Cosmetic; does not block this PR. Filed as inline comment.

Verdict APPROVE: no HIGH, CI green on true HEAD, all DoD assertions satisfied, R2 findings all resolved or properly deferred to a tracked spike. Safe to merge; the LOW on #1013 labels can be applied independently.

Issue #243, opened 2026-05-03
[PLAN] test(core): S1 sample-scene perf fixture for BENCHMARK_CELL workloads


cjhowe-us:
### Domain

core

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:3

### User Stories Satisfied

Lights up the BENCHMARK_CELL infrastructure for the core and render contexts under a synthetic-but-non-trivial workload, so the per-context CI gate (perf-budget.md §CI Gate Spec #1) has at least one populated cell per context that exercises `BENCHMARK_CELL_BODY` end-to-end. Canonical S1 scene invariants (1 character + 200 props + 8 dynamic lights, fixture loader, manifest) iterate in the follow-up plan referenced below.

### Parent (Epic or Sub-Epic)

#4

### Scope

Single SRP scope (narrowed after PR #1000 round-1 review): introduce `tests/perf/s1_sample_scene_test.cpp` containing two named CI gate benchmarks built on the plan #242 `BENCHMARK_CELL_BODY` macro:

- `s1_scene_core_phase_within_budget` — synthetic transform-propagation sweep (100 placeholder entities × 4 archetype groups × 2 systems) asserting `<=` core CPU sim ceiling (400_000 ns, perf-budget.md §core row).
- `s1_scene_render_phase_within_budget` — synthetic AABB cull + draw-list build sweep (100 entities, 1920×1080 viewport descriptor) asserting `<=` render CPU sim+submit ceiling (1_500_000 ns, perf-budget.md §render row).

Wire both into `tests/perf/CMakeLists.txt` alongside the plan #242 scaffolding; share its build flags (Catch2 with exceptions, -O2). The synthetic workloads are written so the compiler cannot dead-code-eliminate them (volatile sinks) and so their cost is representative of the per-context phase work's memory/compute ratio at this entity count — but **they are not budget-realistic for the canonical S1 scenario** (1 character + 200 props + 8 dynamic lights). They exist to populate the BENCHMARK_CELL contract for two contexts so the CI gate has live cells to enforce.

Out of scope (handled by follow-up plan, see "Iterates In" below):

- The `e2e/perf/s1/` scene manifest (YAML descriptor of 1 character + 200 props + 8 dynamic lights with per-context cell budgets).
- The `core/tests/fixtures/s1.hpp` `glibre::testing::load_s1()` Catch2 helper.
[...]

PR #1020, opened 2026-05-10
test(core): S1 canonical fixture loader (refs #1003)

### LINKED ISSUE #1003: [PLAN] test(core): S1 fixture loader + canonical invariants (iterate #243)
### Domain

core

### Phase

phase:mvp

pts:5

### User Stories Satisfied

Implements the canonical S1 fixture cited throughout `reviews/decisions/perf-budget.md` (§Justification Per Cell — S1 scenario: "sample scene = 1 character + 200 props + 8 dynamic lights at 1920×1080"). The S1 manifest is the deterministic input the e2e frame-timing gate (perf-budget.md §CI Gate Spec #2) needs, and `load_s1()` is the Catch2 helper context-level perf benches will swap in for the synthetic placeholder workloads landed under #243.

### Parent (Epic or Sub-Epic)

#4

### Scope

Single SRP scope: introduce the canonical S1 sample-scene fixture so context-level perf benches and the e2e gate share one source of truth.

In scope:

- **`e2e/perf/s1/scene.yaml`** — versioned manifest describing the canonical S1 invariants: `entity_count: 209` (1 character + 200 props + 8 dynamic lights), `viewport: 1920x1080`, deterministic prop placement seed, and the per-context cell budgets quoted verbatim from `reviews/decisions/perf-budget.md` (`core_cpu_sim_ns`, `render_cpu_sim_ns`, `render_cpu_submit_ns`, `geometry_cpu_sim_ns`, `physics_cpu_sim_ns`). Manifest is data-only; deterministic transforms expand from `seed + index` arithmetic so cross-host reproducibility holds.
- **`core/tests/fixtures/s1.hpp`** — Catch2 helper exposing `glibre::testing::load_s1()` returning a `glibre::testing::S1Scene` aggregate (entity count + archetype counts + viewport descriptor + per-context budget map). Header-only; reads the YAML manifest at test setup via a `std::ifstream`-based loader (no glibre::core link required, matching #243's choice). Scene aggregate is the same shape consumed by future contexts' `BENCHMARK_CELL_BODY` calls when they land.
- **Three Catch2 unit tests** under `core/tests/perf_s1_fixture_test.cpp`:
  - `fixture_loads_with_expected_entity_count` — `load_s1()` returns 209 entities, 4 archetypes, 2 systems, 1920×1080 viewport.
  - `fixture_is_deterministic_across_runs` — two `load_s1()` calls in the same process produce bit-identical entity transforms (no time / RNG / address dependence).
  - `manifest_cell_budgets_match_decision_record` — every `*_ns` field in the loaded manifest matches the cell value in `reviews/decisions/perf-budget.md` for the same context (test embeds the canonical map and asserts equality; failing this test means either the manifest drifted or the decision record needs an amendment).
- Wire the new test source + fixture header into `core/tests/CMakeLists.txt` (or create the target if absent — confirm by inspection at session start).

Out of scope:

- The synthetic workload benches in `tests/perf/s1_sample_scene_test.cpp` already shipped under #243 — keep them for now; their replacement with `load_s1()`-driven benches is a separate plan once each context's phase-work API is stable enough to consume the fixture aggregate.
- S2 hot-reload script fixture (hot-reload sub-epic).
- S3 FBX import fixture (content sub-epic).
- GPU-side S1 assets (real meshes / materials / lights — render + content contexts).
- Reference perf numbers / golden traces (recorded after the e2e harness can run).

### Unit Test Plan

- core/tests/perf_s1_fixture: fixture_loads_with_expected_entity_count
- core/tests/perf_s1_fixture: fixture_is_deterministic_across_runs
- core/tests/perf_s1_fixture: manifest_cell_budgets_match_decision_record

Tests live under `core/tests/perf_s1_fixture_test.cpp` (Catch2; same harness style as `tests/perf/s1_sample_scene_test.cpp`).

### Depends on

- #243

### Agent Execution Notes

Subagents post status comments here. PR title: `test(core): S1 fixture loader + canonical invariants (iterate #243)`.

Iterates on #243: PR #1000 (still open) shipped two `BENCHMARK_CELL` benches against a synthetic 100-entity workload to populate the BENCHMARK_CELL contract. This plan adds the canonical S1 manifest + `load_s1()` helper that the perf-budget decision record actually specifies. The two are sequenced: synthetic benches first (so the gate macro is exercised end-to-end), canonical fixture second (so future per-context benches share one deterministic input).

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: e2e/perf/s1/scene.yaml
- file_exists: core/tests/fixtures/s1.hpp
- file_exists: tests/core/perf_s1_fixture/perf_s1_fixture_test.cpp
- file_contains:
    path: core/tests/fixtures/s1.hpp
    regex: "load_s1\\("
- file_contains:
    path: e2e/perf/s1/scene.yaml
    regex: "entity_count:[[:space:]]*209"
- unit_test_named: fixture_loads_with_expected_entity_count
- unit_test_named: fixture_is_deterministic_across_runs
- unit_test_named: manifest_cell_budgets_match_decision_record
- unit_test_named: s1_fixture_loads_canonical_209_entities
- unit_test_named: s1_fixture_archetype_counts_match_scenario
- workflow_passed: ci.yml
```

cjhowe-us:
## Summary

- Introduces \`e2e/perf/s1/scene.yaml\` — the canonical S1 sample-scene manifest (\`entity_count: 209\`, \`viewport: 1920x1080\`, \`placement_seed: 0xDEAD5EED\`, per-context CPU budget cells verbatim from \`reviews/decisions/perf-budget.md §Per-Context Budget Table\`)
- Introduces \`core/tests/fixtures/s1.hpp\` — header-only \`glibre::testing::load_s1()\` helper returning an \`S1Scene\` aggregate; locates the manifest at test runtime via \`GLIBRE_REPO_ROOT\` env-var or \`__FILE__\`-based upward traversal (no \`glibre::core\` link required)
- Adds \`tests/core/perf_s1_fixture/\` — new Catch2 test target with 7 named test cases covering entity count (209), archetype breakdown (1 char + 200 props + 8 lights), cross-call determinism, and budget cell equality against the decision record

## Test plan

- [ ] \`fixture_loads_with_expected_entity_count\` — entity_count=209, archetype_count=3, viewport 1920x1080
- [ ] \`fixture_is_deterministic_across_runs\` — two load_s1() calls produce identical S1Scene values
- [ ] \`manifest_cell_budgets_match_decision_record\` — all *_ns fields match perf-budget.md verbatim values
- [ ] \`s1_fixture_loads_canonical_209_entities\` — entity_count == 209
- [ ] \`s1_fixture_archetype_counts_match_scenario\` — 1 char + 200 props + 8 lights, sum == 209
- [ ] \`s1_fixture_viewport_is_1920x1080\` — viewport width=1920, height=1080
- [ ] \`s1_fixture_budgets_match_perf_budget_md\` — all budget cells equal decision record values
- [ ] Full CTest suite: 198/198 pass locally (7 new tests added)
- [ ] CI: \`ci.yml\` must pass on PR

Note: \`system_count\` is NOT part of S1Scene or this PR's test plan. \`perf-budget.md §Justification Per Cell (S1)\` does not specify a system count for the S1 scenario — system topology is a plugin-context implementation detail, not a fixture invariant. (R1 finding, addressed.)

Closes #1003

### DIFF
--- a/core/tests/fixtures/s1.hpp
+++ b/core/tests/fixtures/s1.hpp
@@ -147,9 +147,7 @@ inline std::uint64_t parse_uint64(const std::string& s) {
 inline std::uint32_t parse_uint32(const std::string& s) {
     const std::uint64_t v = std::stoull(s, nullptr, 0);
     if (v > static_cast<std::uint64_t>(std::numeric_limits<std::uint32_t>::max())) {
-        throw std::out_of_range(
-            "glibre::testing parse_uint32: value exceeds uint32 max: " + s
-        );
+        throw std::out_of_range("glibre::testing parse_uint32: value exceeds uint32 max: " + s);
     }
     return static_cast<std::uint32_t>(v);
 }
@@ -272,8 +270,8 @@ inline ParsedLine parse_line(std::string line) {
                     manifest_version = detail::parse_uint64(p.val);
                     if (manifest_version != 1u) {
                         throw std::runtime_error(
-                            "glibre::testing::load_s1(): unsupported manifest version " +
-                            p.val + " (expected 1); schema glibre/e2e/perf/s1/scene-v1"
+                            "glibre::testing::load_s1(): unsupported manifest version " + p.val +
+                            " (expected 1); schema glibre/e2e/perf/s1/scene-v1"
                         );
                     }
                 } else if (p.key == "entity_count") {
@@ -319,9 +317,8 @@ inline ParsedLine parse_line(std::string line) {
                     // schema drift is caught immediately rather than silently
                     // yielding stale zero counts.
                     throw std::runtime_error(
-                        "glibre::testing::load_s1(): unknown archetype field '" +
-                        p.key + "' under archetype '" + arch_name +
-                        "' (expected 'count')"
+                        "glibre::testing::load_s1(): unknown archetype field '" + p.key +
+                        "' under archetype '" + arch_name + "' (expected 'count')"
                     );
                 }
             }
--- a/core/tests/fixtures/s1.hpp
+++ b/core/tests/fixtures/s1.hpp
@@ -35,6 +35,7 @@
 #include <cstdint>
 #include <filesystem>
 #include <fstream>
+#include <limits>
 #include <stdexcept>
 #include <string>
 
@@ -86,7 +87,9 @@ struct S1ArchetypeCounts {
 struct S1Scene {
     std::uint32_t entity_count = 0u;
     std::uint32_t archetype_count = 0u;
-    std::uint32_t system_count = 0u;
+    // system_count is NOT part of S1Scene: perf-budget.md §Justification Per Cell (S1)
+    // does not specify a system count for the S1 scenario.  System topology is an
+    // implementation detail of each plugin context, not a fixture invariant.
     S1ArchetypeCounts archetypes = {};
     S1Viewport viewport = {};
     std::uint32_t placement_seed = 0u;
@@ -139,8 +142,16 @@ inline std::uint64_t parse_uint64(const std::string& s) {
 }
 
 // parse_uint32: parse decimal or 0x-prefixed hex uint32 from a non-empty string.
+// Throws std::out_of_range if the value exceeds 0xFFFF'FFFF so truncation is
+// never silent — callers that need 64-bit values must use parse_uint64 instead.
 inline std::uint32_t parse_uint32(const std::string& s) {
-    return static_cast<std::uint32_t>(std::stoull(s, nullptr, 0));
+    const std::uint64_t v = std::stoull(s, nullptr, 0);
+    if (v > static_cast<std::uint64_t>(std::numeric_limits<std::uint32_t>::max())) {
+        throw std::out_of_range(
+            "glibre::testing parse_uint32: value exceeds uint32 max: " + s
+        );
+    }
+    return static_cast<std::uint32_t>(v);
 }
 
 // ParsedLine: result of stripping a single line from the manifest.
@@ -207,20 +218,26 @@ inline ParsedLine parse_line(std::string line) {
     // ---------------------------------------------------------------------------
     // Parser state
     //
-    // The manifest uses a simple structured YAML format:
-    //   - Top-level scalar fields (entity_count, system_count, placement_seed)
-    //   - Two-level blocks (archetypes > archetype_name > count:)
-    //   - One-level blocks (viewport > width/height, budgets > key: value)
+    // The manifest uses a simple structured YAML subset:
+    //   - Top-level scalar fields: indent == 0, has_val == true
+    //   - Top-level block headers: indent == 0, has_val == false
+    //   - Archetype-block sub-headers: indent == 2, has_val == false
+    //   - Archetype count fields: indent == 4, key == "count", has_val == true
+    //   - Viewport / budget fields: indent > 0, has_val == true
     //
-    // Strategy: single-pass parser that tracks current section and nesting.
-    // When a block header is encountered (non-empty key, empty value), we
-    // switch sections.  When a value is encountered inside a section, we
-    // dispatch to the appropriate field.
+    // Indent contract (spaces only — tabs also count as 1 per parse_line):
+    //   Indent 0 = top-level; Indent 2 = block member; Indent 4 = block sub-member.
+    //
+    // Strategy: single-pass parser that tracks current section.  Any key not
+    // matching the known schema within a section throws std::runtime_error so
+    // format drift (renamed key, wrong nesting) fails loudly rather than
+    // silently producing zero values.
     // ---------------------------------------------------------------------------
 
     enum class Section { None, Archetypes, Viewport, Budgets };
     Section section = Section::None;
     std::string arch_name;  // tracks the current archetype name in the Archetypes block
+    std::uint64_t manifest_version = 0u;
 
     std::string raw;
     while (std::getline(ifs, raw)) {
@@ -251,14 +268,22 @@ inline ParsedLine parse_line(std::string line) {
             } else {
                 // Top-level scalar field.
                 section = Section::None;
-                if (p.key == "entity_count") {
+                if (p.key == "version") {
+                    manifest_version = detail::parse_uint64(p.val);
+                    if (manifest_version != 1u) {
+                        throw std::runtime_error(
+                            "glibre::testing::load_s1(): unsupported manifest version " +
+                            p.val + " (expected 1); schema glibre/e2e/perf/s1/scene-v1"
+                        );
+                    }
+                } else if (p.key == "entity_count") {
                     scene.entity_count = detail::parse_uint32(p.val);
-                } else if (p.key == "system_count") {
-                    scene.system_count = detail::parse_uint32(p.val);
                 } else if (p.key == "placement_seed") {
                     scene.placement_seed = detail::parse_uint32(p.val);
                 }
-                // "version" is parsed but not stored (schema version, reserved for future use).
+                // Unrecognised top-level scalars are silently ignored so the
+                // parser is forward-compatible with additive keys (e.g. comments
+                // or metadata that the v1 struct does not consume).
             }
             continue;
         }
@@ -273,17 +298,31 @@ inline ParsedLine parse_line(std::string line) {
 
         case Section::Archetypes:
             if (p.indent == 2u && !p.has_val) {
-                // Archetype block header (e.g. "  character:").
+                // Archetype block sub-header (e.g. "  character:").
+                // Only the three canonical archetypes are expected; unrecognised
+                // names are silently skipped (forward-compat for additive archetypes).
                 arch_name = p.key;
-            } else if (p.indent == 4u && p.has_val && p.key == "count" && !arch_name.empty()) {
-                // Count field nested under an archetype name.
-                const std::uint32_t n = detail::parse_uint32(p.val);
-                if (arch_name == "character") {
-                    scene.archetypes.character = n;
-                } else if (arch_name == "prop") {
-                    scene.archetypes.prop = n;
-                } else if (arch_name == "dynamic_light") {
-                    scene.archetypes.dynamic_light = n;
+            } else if (p.indent == 4u && p.has_val && !arch_name.empty()) {
+                // Field nested under an archetype name.
+                if (p.key == "count") {
+                    const std::uint32_t n = detail::parse_uint32(p.val);
+                    if (arch_name == "character") {
+                        scene.archetypes.character = n;
+                    } else if (arch_name == "prop") {
+                        scene.archetypes.prop = n;
+                    } else if (arch_name == "dynamic_light") {
+                        scene.archetypes.dynamic_light = n;
+                    }
+                    // Unknown archetype names are silently skipped (forward-compat).
+                } else {
+                    // Unknown field inside an archetype block — fail loudly so
+                    // schema drift is caught immediately rather than silently
+                    // yielding stale zero counts.
+                    throw std::runtime_error(
+                        "glibre::testing::load_s1(): unknown archetype field '" +
+                        p.key + "' under archetype '" + arch_name +
+                        "' (expected 'count')"
+                    );
                 }
             }
             break;
@@ -295,6 +334,7 @@ inline ParsedLine parse_line(std::string line) {
                 } else if (p.key == "height") {
                     scene.viewport.height = detail::parse_uint32(p.val);
                 }
+                // Unknown viewport fields silently skipped (forward-compat).
             }
             break;
 
@@ -318,9 +358,19 @@ inline ParsedLine parse_line(std::string line) {
                 } else if (p.key == "physics_cpu_submit_ns") {
                     b.physics_cpu_submit_ns = detail::parse_uint64(p.val);
                 }
+                // Unknown budget keys silently skipped (forward-compat for additive
+                // context cells once new contexts are introduced).
             }
             break;
         }
+    }
+
+    // Require version field was present.
+    if (manifest_version == 0u) {
+        throw std::runtime_error(
+            "glibre::testing::load_s1(): manifest missing 'version' field; "
+            "schema glibre/e2e/perf/s1/scene-v1 requires version: 1"
+        );
     }
 
     // ---------------------------------------------------------------------------
--- a/core/tests/fixtures/s1.hpp
+++ b/core/tests/fixtures/s1.hpp
@@ -1,0 +1,312 @@
+#pragma once
+// core/tests/fixtures/s1.hpp
+//
+// Catch2 fixture helper — canonical S1 sample-scene loader.
+//
+// Authority: plan #1003, reviews/decisions/perf-budget.md §Justification Per Cell (S1).
+//
+// Usage:
+//   #include "fixtures/s1.hpp"
+//
+//   TEST_CASE("my_bench", "[perf][core][s1]") {
+//       auto scene = glibre::testing::load_s1();
+//       REQUIRE(scene.entity_count == 209u);
+//       // ...
+//   }
+//
+// Design decisions:
+//   - Header-only; no glibre::core link required.  Reads the YAML manifest
+//     at test setup via a std::ifstream-based parser.
+//   - std::filesystem / std::ifstream are PHILOSOPHY §11 permitted carve-outs
+//     (language/runtime utilities EASTL does not own).
+//   - No YAML library dependency.  The manifest is a hand-written subset of
+//     YAML (key: value pairs only); parsing is a simple line-scan.  This is
+//     intentional: the manifest is machine-written and machine-read; structural
+//     complexity in the parser would be scope creep.
+//   - S1Scene is plain data (no destructors, no allocation) so it is safe to
+//     construct in any test thread without allocator setup.
+//   - load_s1() locates the manifest at runtime using the GLIBRE_REPO_ROOT
+//     environment variable if set, otherwise by traversing from __FILE__
+//     upward until e2e/perf/s1/scene.yaml is found.
+//
+// PHILOSOPHY §7 (determinism): placement_seed is read and stored so callers
+// can expand deterministic transforms from seed + entity_index arithmetic.
+
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+#include <stdexcept>
+#include <string>
+
+namespace glibre::testing {
+
+// ---------------------------------------------------------------------------
+// Viewport — S1 viewport descriptor
+// ---------------------------------------------------------------------------
+
+struct S1Viewport {
+    std::uint32_t width  = 0u;
+    std::uint32_t height = 0u;
+};
+
+// ---------------------------------------------------------------------------
+// S1Budgets — per-context CPU budget cells (nanoseconds)
+//
+// Verbatim from reviews/decisions/perf-budget.md §Per-Context Budget Table.
+// Fields match the budget keys in e2e/perf/s1/scene.yaml.
+// ---------------------------------------------------------------------------
+
+struct S1Budgets {
+    std::uint64_t core_cpu_sim_ns        = 0u;
+    std::uint64_t core_cpu_submit_ns     = 0u;
+    std::uint64_t render_cpu_sim_ns      = 0u;
+    std::uint64_t render_cpu_submit_ns   = 0u;
+    std::uint64_t geometry_cpu_sim_ns    = 0u;
+    std::uint64_t geometry_cpu_submit_ns = 0u;
+    std::uint64_t physics_cpu_sim_ns     = 0u;
+    std::uint64_t physics_cpu_submit_ns  = 0u;
+};
+
+// ---------------------------------------------------------------------------
+// S1ArchetypeCounts — entity composition by archetype
+// ---------------------------------------------------------------------------
+
+struct S1ArchetypeCounts {
+    std::uint32_t character     = 0u;   // 1
+    std::uint32_t prop          = 0u;   // 200
+    std::uint32_t dynamic_light = 0u;   // 8
+};
+
+// ---------------------------------------------------------------------------
+// S1Scene — aggregate returned by load_s1()
+//
+// All fields are plain data; the struct is safe to copy and compare.
+// ---------------------------------------------------------------------------
+
+struct S1Scene {
+    std::uint32_t     entity_count    = 0u;
+    std::uint32_t     archetype_count = 0u;
+    std::uint32_t     system_count    = 0u;
+    S1ArchetypeCounts archetypes      = {};
+    S1Viewport        viewport        = {};
+    std::uint32_t     placement_seed  = 0u;
+    S1Budgets         budgets         = {};
+};
+
+// ---------------------------------------------------------------------------
+// detail — internal helpers
+// ---------------------------------------------------------------------------
+
+namespace detail {
+
+// locate_repo_root: traverse upward from a starting path until a directory
+// containing e2e/perf/s1/scene.yaml is found.  Returns the repo root path.
+// Throws std::runtime_error on failure.
+inline std::filesystem::path locate_repo_root() {
+    // Allow override via environment variable (e.g. in CI).
+    if (const char* env = std::getenv("GLIBRE_REPO_ROOT")) {
+        return std::filesystem::path(env);
+    }
+
+    // Derive from __FILE__ (absolute path of this header at compile time).
+    // Walk up until e2e/perf/s1/scene.yaml exists as a sibling.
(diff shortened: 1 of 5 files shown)


e2e/perf/s1/scene.yaml
```diff
@@ -0,0 +1,72 @@
...
+  character:
+    count: 1
+  prop:
+    count: 200
+  dynamic_light:
+    count: 8
+
+# System count = 2 (core transform sweep + render visibility sweep).
+# Matches the system_count invariant carried over from the synthetic placeholder.
+system_count: 2
```
cjhowe-us: severity:MED location:e2e/perf/s1/scene.yaml:34 problem:scene.yaml comment claims system_count=2 was "carried over from the synthetic placeholder" — that placeholder lived in #243's tests/perf/s1_sample_scene_test.cpp; perf-budget.md does not specify a system count for S1, and #1003 issue prose says "4 archetypes, 2 systems" while the PR's `fixture_loads_with_expected_entity_count` REQUIREs archetype_count==3. The 3 vs 4 discrepancy is loose issue prose (3 is correct: char+prop+dynamic_light), but system_count=2 lacks an authoritative citation. fix:Add an inline comment citing the source-of-truth for system_count=2 (either #243 PR # or a perf-budget.md §reference) so a future amendment knows what to update — or downgrade to a TODO and defer until phase-work APIs land.

core/tests/fixtures/s1.hpp
```diff
@@ -0,0 +1,343 @@
...
+// Throws std::runtime_error on failure.
+inline std::filesystem::path locate_repo_root() {
+    // Allow override via environment variable (e.g. in CI).
+    if (const char* env = std::getenv("GLIBRE_REPO_ROOT")) {
+        return std::filesystem::path(env);
+    }
+
+    // Derive from __FILE__ (absolute path of this header at compile time).
+    // Walk up until e2e/perf/s1/scene.yaml exists as a sibling.
+    std::filesystem::path p = std::filesystem::path(__FILE__).parent_path();
```
cjhowe-us, line 118: severity:MED location:core/tests/fixtures/s1.hpp:113 problem:`parse_uint32` does `static_cast<std::uint32_t>(std::stoull(s, nullptr, 0))` — silent truncation if a future manifest puts a value > 0xFFFFFFFF (e.g. someone widens placement_seed to 64 bits in scene.yaml without widening the field here). No range check; tests would still pass on the truncated low 32 bits. fix:Either (a) REQUIRE `value <= std::numeric_limits<uint32_t>::max()` and throw std::out_of_range, or (b) widen `placement_seed` and the archetype counters to uint64_t to match parse_uint64. Option (a) is least disruptive.

core/tests/fixtures/s1.hpp
```diff
@@ -0,0 +1,343 @@
...
+inline ParsedLine parse_line(std::string line) {
+    // Strip inline comment (first '#').
+    const std::size_t hash = line.find('#');
+    if (hash != std::string::npos) {
+        line = line.substr(0, hash);
+    }
+
+    // Count leading spaces/tabs for indentation.
+    std::size_t indent = 0u;
+    for (std::size_t i = 0u; i < line.size(); ++i) {
```
cjhowe-us, line 176: severity:LOW location:core/tests/fixtures/s1.hpp:165 problem:Hand-rolled YAML parser uses literal indent counts (`p.indent == 2u`, `p.indent == 4u`) for archetype block dispatch. Mixing tabs+spaces (parse_line counts both as 1) or any reformat that changes nesting width silently breaks the loader without test signal — the manifest just yields zero counts, archetype_count==0, and the entity-count test fails with a confusing `0 != 209u` instead of "manifest format error". fix:Either (a) assert `indent > 0u` for non-top-level lines and let section state machine track depth, or (b) emit a parser error on unknown keys/indent within a section so drift fails loudly. Defer is acceptable since DoD passes; flag for round 2 cohesion review if not addressed.

e2e/perf/s1/scene.yaml
```diff
@@ -0,0 +1,72 @@
...
+#   "sample scene = 1 character + 200 props + 8 dynamic lights at 1920x1080"
+#
+# This manifest is data-only. Entity transforms expand from seed + index
+# arithmetic so cross-host determinism holds (PHILOSOPHY §7 — determinism by default).
+# A fixture change requires a perf-budget amendment spike (perf-budget.md §Consequences).
+#
+# Version: 1
+# Schema: glibre/e2e/perf/s1/scene-v1
+
+version: 1
```
cjhowe-us, line 17: severity:LOW location:e2e/perf/s1/scene.yaml:16 problem:Manifest declares `version: 1` and `Schema: glibre/e2e/perf/s1/scene-v1` but the loader (s1.hpp:218) silently ignores the field ("version is parsed but not stored, reserved for future use"). If a future v2 manifest lands with a renamed key, the v1 loader silently parses partially, yields zeros for missing keys, and tests fail with cryptic mismatches rather than "unsupported manifest version". fix:Store `version` in S1Scene (or a local stack var) and `throw std::runtime_error` if it != 1u. Cheap, makes the schema field non-decorative.

tests/core/perf_s1_fixture/perf_s1_fixture_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+// ---------------------------------------------------------------------------
+// fixture_is_deterministic_across_runs
+//
+// Verifies that two consecutive calls to load_s1() in the same process
+// return bit-identical S1Scene values.
+//
+// This tests PHILOSOPHY §7 (determinism by default): the fixture must
+// produce identical output regardless of call order, timing, thread state,
+// or address-space layout.  The placement_seed is read from the manifest
```
cjhowe-us, line 49: severity:LOW location:tests/core/perf_s1_fixture/perf_s1_fixture_test.cpp:47 problem:`fixture_is_deterministic_across_runs` only verifies the loader is pure (same file → same parsed values), not that PHILOSOPHY §7 cross-host determinism holds. The plan promises `position = f(seed, entity_index)` with no platform-dependent FP, but no transforms are computed in this PR yet, so this test is necessarily limited. fix:Add a NOTE comment in the test body clarifying that this asserts loader purity only and that real cross-host determinism will be tested when transform expansion lands in a follow-up plan. Avoids future readers thinking this test covers the seed-based reproducibility claim.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:3
carryover:0
notes:Coverage+correctness lens. PR satisfies issue #1003 DoD: all 5 unit_test_named keys are present (fixture_loads_with_expected_entity_count, fixture_is_deterministic_across_runs, manifest_cell_budgets_match_decision_record, s1_fixture_loads_canonical_209_entities, s1_fixture_archetype_counts_match_scenario), all 3 file_exists assertions pass, scene.yaml regex matches, Closes #1003 present, ci.yml green on head 51ac219 (build+lint+dep-check+spec-check+automated-review all pass; e2e-* skipped via path filter). Conv-Commit subject test(core): is correct. Budget cells in scene.yaml are verbatim against perf-budget.md §Per-Context Budget Table (core 0.40/0.05, render 0.10/1.40, geometry 0.30/0.20, physics 2.00/0.00 — ms→ns conversions correct). Test layout (core/tests/fixtures/s1.hpp + tests/core/perf_s1_fixture/) is more cohesive than plan's original path, mirrors peer test dirs (perf_budget, frame_loop_integration). MED findings: (1) system_count=2 in scene.yaml lacks an authoritative citation; (2) parse_uint32 silently truncates >32-bit values. LOW findings: brittle hand-rolled YAML parser indent dispatch, version field declared but ignored, determinism test scope is loader-purity not cross-host. None block merge under a coverage+correctness lens; defer cohesion of the parser to round 2 (SRP — does load_s1 deserve a tiny separate parser type?). Suggest impl-respond addresses MED items + LOW#2 (version check) inline.

tests/core/perf_s1_fixture/perf_s1_fixture_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+// fixture_loads_with_expected_entity_count
+//
+// Verifies that load_s1() returns the canonical 209-entity S1 scene:
+//   - entity_count == 209 (1 char + 200 props + 8 lights)
+//   - archetype_count == 3 (character, prop, dynamic_light are distinct archetypes)
+//   - viewport == 1920x1080
+//
+// Note: system_count is NOT a fixture invariant — perf-budget.md §Justification
+// Per Cell (S1) does not fix a system count.  System topology is an
+// implementation detail of each plugin context.
```
cjhowe-us, line 30: severity:MED location:tests/core/perf_s1_fixture/perf_s1_fixture_test.cpp:30 problem:R2 cohesion — PR description test_plan still claims `system_count=2` is asserted by `fixture_loads_with_expected_entity_count`, but R1 correctly removed system_count from S1Scene + the test (no perf-budget authority). The PR body now misrepresents what the test verifies; future readers / auditors comparing the PR body to the test will see drift. fix:Update PR description test plan bullet 1 to drop "system_count=2" and replace with a parenthetical explaining the removal: "(system_count is not a fixture invariant per perf-budget.md §Justification Per Cell — see commit 89a4cff)".

core/tests/fixtures/s1.hpp
```diff
@@ -0,0 +1,390 @@
...
+                    section = Section::Budgets;
+                } else {
+                    section = Section::None;
+                }
+            } else {
+                // Top-level scalar field.
+                section = Section::None;
+                if (p.key == "version") {
+                    manifest_version = detail::parse_uint64(p.val);
+                    if (manifest_version != 1u) {
```
cjhowe-us, line 273: severity:MED location:core/tests/fixtures/s1.hpp:271 problem:R2 cohesion — schema-evolution policy is inconsistently applied across sections. Inside `Archetypes` an unknown FIELD throws (e.g. `width` under `character`), but unknown KEYS in `Viewport` and `Budgets` are silently skipped. Top-level unknown scalars also silently skip. The asymmetric rationale (forward-compat for additive keys vs fail-loud on drift inside known sub-schemas) is split across three inline comments and not stated as one policy. A future contributor adding `viewport.dpi` will get silent skip; adding `archetypes.character.scale` will get throw. fix:Add a top-of-load_s1 doc block stating the v1 schema-evolution policy explicitly: "v1 forward-compat: unknown TOP-LEVEL keys / unknown SECTION names / unknown SECTION ITEMS = skip; unknown FIELDS inside a typed sub-schema (currently only archetype.<name>) = throw." Then either (a) remove the asymmetry by also throwing on unknown viewport / budget keys, or (b) document the asymmetry as deliberate (viewport+budgets are flat key spaces; archetypes are typed records).

core/tests/fixtures/s1.hpp
```diff
@@ -0,0 +1,390 @@
...
+                        "glibre::testing::load_s1(): unknown archetype field '" + p.key +
+                        "' under archetype '" + arch_name + "' (expected 'count')"
+                    );
+                }
+            }
+            break;
+
+        case Section::Viewport:
+            if (p.indent > 0u && p.has_val) {
+                if (p.key == "width") {
```
cjhowe-us, line 344: severity:LOW location:core/tests/fixtures/s1.hpp:329 problem:R2 cohesion — `archetype_count` is derived as "number of distinct archetypes with count > 0", which conflates "declared in manifest" with "non-empty". A future debug fixture that sets `prop.count: 0` would yield archetype_count=2 silently; tests asserting ==3 would fail with a confusing mismatch instead of "manifest declared 3 archetypes but one is empty". fix:Either (a) track which archetype names appeared as sub-headers in the parser and count those (declarative count), or (b) document on S1Scene::archetype_count that the value is "non-empty archetypes" and rename to e.g. `populated_archetype_count`. Defer if MVP only ever has the canonical 3 non-zero counts.

e2e/perf/s1/scene.yaml
```diff
@@ -0,0 +1,68 @@
+# S1 Sample-Scene Fixture Manifest
+#
+# Authority: reviews/decisions/perf-budget.md §Justification Per Cell — S1 scenario
+#            plan #1003 (canonical S1 fixture loader)
+#
+# Canonical invariants:
+#   "sample scene = 1 character + 200 props + 8 dynamic lights at 1920x1080"
+#
+# This manifest is data-only. Entity transforms expand from seed + index
```
cjhowe-us, line 9: severity:LOW location:e2e/perf/s1/scene.yaml:9 problem:R2 spec alignment — the manifest header asserts "Entity transforms expand from seed + index arithmetic so cross-host determinism holds (PHILOSOPHY §7 — determinism by default)". PHILOSOPHY §7 verbatim says: "Physics + ECS world snapshots byte-equal across hosts and runs. No platform intrinsics in simulation. Fixed container iteration order." §7 does NOT specify "seed + index arithmetic" — that is a derivation, not a §7 quote. Over-attribution conflates an implementation choice (how transforms expand) with a principle (byte-equal snapshots). fix:Reword to "Entity transforms must expand from seed + index arithmetic (or any host-deterministic function) so PHILOSOPHY §7 byte-equal-snapshot invariants hold." Same wording fix in core/tests/fixtures/s1.hpp:34 (`PHILOSOPHY §7 (determinism): placement_seed is read and stored so callers can expand deterministic transforms from seed + entity_index arithmetic`).

core/tests/fixtures/s1.hpp
```diff
@@ -0,0 +1,390 @@
...
+// never silent — callers that need 64-bit values must use parse_uint64 instead.
+inline std::uint32_t parse_uint32(const std::string& s) {
+    const std::uint64_t v = std::stoull(s, nullptr, 0);
+    if (v > static_cast<std::uint64_t>(std::numeric_limits<std::uint32_t>::max())) {
+        throw std::out_of_range("glibre::testing parse_uint32: value exceeds uint32 max: " + s);
+    }
+    return static_cast<std::uint32_t>(v);
+}
+
+// ParsedLine: result of stripping a single line from the manifest.
```
cjhowe-us, line 157: severity:LOW location:core/tests/fixtures/s1.hpp:155 problem:R2 seam quality — `parse_line` counts tabs and spaces both as 1 leading char (lines 142-148). The design comment at the top of load_s1() now says "spaces only — tabs also count as 1 per parse_line", which is a footgun: a future editor that inserts a single tab where two spaces existed silently halves the indent count and breaks the indent==2 / indent==4 dispatch. The committed manifest happens to use spaces only, but no test enforces that. fix:Either (a) replace the leading-whitespace counter with `if (line[i] == \" \") ++indent; else if (line[i] == \"\\t\") throw runtime_error(\"tabs forbidden in s1.yaml\"); else break;`, or (b) add a one-line CI grep `! grep -P "^\\t" e2e/perf/s1/scene.yaml` that fails the build on any tab-indented manifest line. Option (a) is closer to the loader.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:2
  LOW:3
carryover:0
notes:Cohesion + SRP + seam-quality lens. CI green on head c3edfc0 (build/lint/dep-check/spec-check/automated-review pass; e2e-* skipped via path filter — confirmed not blocking). All 5 R1 findings addressed cleanly: system_count removed with rationale comment citing perf-budget.md authority gap; parse_uint32 now uses uint64-then-range-check throwing std::out_of_range; archetype block dispatches throw on unknown FIELDS while preserving forward-compat for unknown NAMES (documented inline); manifest version is parsed + asserted ==1 with throw-on-mismatch and a follow-up "missing version" check; determinism test has a NOTE block clarifying loader-purity-only scope. SRP analysis of load_s1: the function is a single-pass parser+dispatcher (~120 lines); locate_repo_root and parse_line are already extracted to detail::; further extraction (per-section appliers) would scatter schema knowledge with no behavioral benefit. Public-API surface is minimal POD aggregates. PHILOSOPHY §11 std::filesystem/ifstream carve-out cited correctly. New R2 findings: (1) MED — PR body test-plan still claims `system_count=2` in fixture_loads_with_expected_entity_count even though the assertion was removed; PR description drift confuses future auditors. (2) MED — schema-evolution policy is asymmetric: archetype FIELDS throw, viewport+budgets keys silently skip, top-level scalars silently skip; the rationale is split across three inline comments and never stated as one v1 policy. (3) LOW — archetype_count derives from "non-zero counts" not "declared in manifest" — debug fixture with prop.count: 0 yields archetype_count=2. (4) LOW — PHILOSOPHY §7 over-attribution: §7 verbatim says "Physics + ECS world snapshots byte-equal across hosts and runs / no platform intrinsics in simulation / fixed container iteration order"; "seed + index arithmetic" is a derivation, not a §7 quote (scene.yaml:9, s1.hpp:34). (5) LOW — parse_line counts tabs+spaces both as 1 leading char; design comment acknowledges but no test enforces spaces-only; tab insertion silently halves indent. None block merge under SRP/cohesion lens. Suggest impl-respond addresses MED#1 (PR body update — trivial) + LOW#4 (one-line wording fix in two places) inline and DEFERs MED#2/LOW#3/LOW#5 to a follow-up parser-hardening spike if the manifest schema gains a v2 entry.

tests/core/perf_s1_fixture/perf_s1_fixture_test.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED

PR body updated via `gh pr edit 1020 --body`. The stale `system_count=2` entry has been removed from the test plan checklist. A clarifying note was added explaining that `system_count` is intentionally absent from `S1Scene` (authority: perf-budget.md §Justification Per Cell (S1) does not specify a system count; system topology is a plugin-context implementation detail, not a fixture invariant).

core/tests/fixtures/s1.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:7e8f075

Addressed partially via code comment (the inline doc), deferred the parser-hardening spike per reviewer guidance.

Commit 7e8f075 adds a detailed block comment before the section dispatch explaining the intentional asymmetry:
- **Archetypes throws on unknown fields**: archetype identity (count per type) is load-bearing; silently ignoring a renamed/added field yields zero entity counts with no diagnostic — strict enforcement catches schema drift immediately.
- **Viewport/Budgets silently skip unknown keys**: these are outer-block metadata; additive keys (new budget contexts) must not break existing loaders that do not know them. If a v2 archetype field is needed, the version check already gates old loaders.

The full parser-hardening question (field enumeration, tab lint, version-gating strategy) is deferred to spike #1021 [SPIKE] iterate-core-s1-fixture-parser-hardening.

core/tests/fixtures/s1.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:DEFER reasoning:The archetype_count=0 ambiguity is real but no current fixture nor near-term fixture (S1 has 1+200+8) would trigger it. Fixing this now requires either adding a 'declared_archetypes' field to the YAML or changing the derivation logic, both of which widen scope beyond the #1003 fixture loader plan. The v1 schema constraint (all three archetypes must be non-zero for a valid S1 scene) is implicit but sufficient for all planned uses. proposal:Open follow-up when a debug/empty fixture materializes. Tracked by spike #1021 which covers the broader parser-policy question.

e2e/perf/s1/scene.yaml
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:7e8f075

Commit 7e8f075 fixes the wording in both locations:

scene.yaml (line 9): replaced 'cross-host determinism holds (PHILOSOPHY §7 — determinism by default)' with an accurate description: 'a derived determinism strategy (no RNG state, no platform-dependent intrinsics, fixed iteration order) aligned with PHILOSOPHY §7 (byte-equal snapshots)'.

s1.hpp (header comment line 32-35): replaced the verbatim §7 citation with 'Determinism design: placement_seed is read and stored so callers can expand transforms via seed + entity_index arithmetic — a derived determinism strategy aligned with PHILOSOPHY §7 (byte-equal snapshots, fixed iteration order, no platform-dependent intrinsics)'.

The wording now accurately states this is a derived design strategy consistent with §7 rather than implying §7 verbatim mandates seed+index arithmetic.

core/tests/fixtures/s1.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:DEFER reasoning:The s1 scene.yaml is space-only by convention; no tab-containing line has ever been committed. The asymmetry (tab=1 char in parse_line) is now documented in the header comment at the parse_line callsite. Adding a build-time tab-check (e.g. a CMake custom command or clang-tidy check) is valid but out-of-scope for this fixture-loader plan — it belongs in the broader parser-hardening work. proposal:Tracked by spike #1021 [SPIKE] iterate-core-s1-fixture-parser-hardening which covers tab/space lint as question 3.

core/tests/fixtures/s1.hpp
```diff
@@ -0,0 +1,405 @@
...
+        //   Viewport and Budgets — unknown keys are silently skipped.  These
+        //   sections are outer-block metadata; additive keys (e.g. a new budget
+        //   context) must not break loaders that do not know them yet.  Strict
+        //   checking here would require every caller to be updated when a new
+        //   context is added, which is the wrong direction for forward compatibility.
+        //   If a future v2 archetype field is needed, the version check (above)
+        //   ensures old loaders fail rather than silently misparse.
+        // ---------------------------------------------------------------------------
+
+        switch (section) {
```
cjhowe-us, line 308: severity:LOW location:core/tests/fixtures/s1.hpp:293 problem:R3 polish — the v1 schema-evolution policy block (s1.hpp:293-308) now cohesively names two policies ("Archetypes — unknown fields throw" / "Viewport and Budgets — unknown keys silently skipped"), but the third location with policy choice — top-level unknown scalars at s1.hpp:284-286 — is documented in a separate inline comment and never named in the unified policy block. A future contributor reading only the named-policy block could conclude the policy starts at indented sections and miss that top-level unknown keys also silently skip. fix:Add a third bullet to the v1 schema-evolution policy block: "Top-level scalars — unknown keys silently skipped (same forward-compat rationale as Viewport/Budgets; the version field at the top gates structural changes)." One-line cohesion improvement; not a behavioral change. Defer to spike #1021 if reviewer prefers no in-scope polish.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0
notes:Topology + spec alignment + polish lens. CI green on head 7e8f075 (run 25616743641 success: build, lint, dep-check, spec-check, automated-review all pass; e2e-* skipped via path filter — confirmed not blocking). All R1+R2 findings either ADDRESSED in code or DEFERRED to follow-up spike #1021 with explicit traceability. DAG coherent: #1003 (plan, pts:5) iterates on #243 (depends-on declared); R2-spawned #1021 (spike, no pts) parented to same epic #4 and cites the deferred MED-2/LOW-3/LOW-5 question set verbatim — clean fan-out. Sub-issue labels correct on both (type:plan|spike, phase:mvp, domain:core; #1003 carries pts:5, #1021 correctly omits pts per AGENTS.md spike rules). Spec alignment: PHILOSOPHY §7 wording in scene.yaml:10-11 + s1.hpp:33-34 now honestly frames seed+index as a *derived* determinism strategy *aligned with* the verbatim §7 byte-equal-snapshots invariant — over-attribution fixed. perf-budget.md §Justification Per Cell (S1) cited as authority for entity composition (1+200+8=209) and per-context cell budgets; line 12 of scene.yaml now points future amendments to perf-budget.md §Consequences governance — good polish. Schema: glibre/e2e/perf/s1/scene-v1 is now *enforced* via version-check + throw, no longer decorative (s1.hpp:271-279). Schema-evolution asymmetry now stated as a single contiguous policy block (s1.hpp:293-308) — cohesive, with forward-pointer to v2 version-gating. DoD block on #1003: populated, all 11 assertions use only DSL-allowed keys (verify_dod.py CHECKS dict), and every assertion is satisfied by the diff at HEAD: pr_merged_closes_self anchored by `Closes #1003` in PR body; both file_exists and file_contains assertions match; all 5 unit_test_named fixtures resolve to TEST_CASE() in the test file; workflow_passed: ci.yml green on main (will gate post-merge). system_count fully purged from HEAD (commit 7e8f075) — verified via `git show 7e8f075:` against scene.yaml + s1.hpp + perf_s1_fixture_test.cpp. Conv-Commit subject `test(core): S1 canonical fixture loader` correct scope. Sole new R3 finding (LOW): top-level-scalar silent-skip is documented inline at s1.hpp:284-286 but not folded into the named v1 schema-evolution policy block at s1.hpp:293-308 — minor cohesion nit, not a blocker; can land in this PR or fold into #1021 alongside the broader parser-policy decision record. Verdict: APPROVE — this PR is merge-ready.

Issue #347, opened 2026-05-03
[STORY] type-registry-immutable-after-init


cjhowe-us:
### Domain

core

### Phase

phase:mvp

### Persona

engine developer

### User Story

As an engine developer, I want a `TypeRegistry` populated at static-init from `glibre-types.dylib` codegen and frozen before `World` construction completes, so that all threads can look up `{size, align, drop, layout}` concurrently without locks and runtime "register a new type" calls are refused.

### Acceptance Criteria (Gherkin)

```gherkin
Given a fresh TypeRegistry populated at static init
When World construction completes
Then any subsequent attempt to register a new TypeId returns core::Error::TypeRegistryClosed

Given a TypeRegistry already initialized
When N worker threads concurrently call lookup(type_id)
Then every lookup returns the correct descriptor with no data races (TSAN clean)

Given two distinct types T1 and T2 with codegen-emitted identical shapes
When their TypeIds are queried
Then each TypeId maps to exactly one descriptor (no collision)
```

### Manual Test Script

1. Build with TSAN: `cmake --build build-tsan`.
2. Run `./build-tsan/core_smoke --type-registry-stress`; expect zero data race reports under 1M concurrent lookups.
3. Attempt to register a new type from a plugin's `glibre_plugin_register` after first frame; confirm `TypeRegistryClosed` error.

### E2E Test Plan

`tests/e2e/core/type-registry.glibre-trace` — drives concurrent lookups and post-init registration attempts.

### Closure Checklist (in order)

- [ ] E2E test authored and passing in CI
- [ ] Manual test script reviewed by a second human
- [ ] Manual test executed; PASS recorded as a comment

### Harmonius Source (optional)

US-1.3.1, US-1.3.2 (`docs/user-stories/core-runtime/reflection-and-type-system.md`)

### Notes / Open Questions

Maps to SPEC §4.9 invariants 1–4. Reflective fields (path access, attributes, Reflect trait) refused per §3.3 / §3.2 collapse #4.

Issue #233, opened 2026-05-03
[PLAN] feat(core): glibre::Error variant set + std::expected<T,Error> alias


cjhowe-us:
### Domain

core

### Phase

phase:mvp

### Story Points (Fibonacci, > 8 must split)

pts:3

### User Stories Satisfied

To be linked when stories are drafted under #11's grandparent epic. Implements the Type Sketch in `reviews/decisions/error-model.md` so every per-context spec §10 has a concrete header to depend on.

### Parent (Epic or Sub-Epic)

#4

### Scope

Single SRP scope: introduce `core/include/glibre/error.hpp` with the engine-wide `glibre::Error` class, the `core::Error` enum (the six variants in error-model.md Type Sketch — `PluginAbiHashMismatch`, `PluginInitFailed`, `SchemaMigrationFailed`, `HotReloadRefused`, `FramePhaseMisordered`, `OutOfBudget`), the `ErrorContext` POD, the `Variant` alias seeded with `core::Error` only, and the `glibre::Result<T>` alias for `std::expected<T, glibre::Error>`.

Out of scope: per-context error enums beyond `core::Error` (separate plans add `render::Error`, etc.); `GLIBRE_TRY` macro shape (separate plan); `magic_enum` vs hand-written `to_string` decision (separate plan); spdlog helper (separate plan); CMake `-fno-exceptions` enforcement (separate plan).

### Unit Test Plan

- core/error: error_constructs_from_core_enum
- core/error: error_context_captures_file_line_detail
- core/error: result_unexpected_round_trips
- core/error: variant_index_stable_for_core_error

Catch2 tests live under `core/tests/error_test.cpp`. Tests must build with `-fno-exceptions` so the file itself proves the header is exception-free.

### Depends on

- #9 (research-error-model)
- #10 (decide-perf-budget-allocation, for `OutOfBudget` arm coordination)

### Agent Execution Notes

Subagents post status comments here in English. PRs use Conventional Commit subjects (`feat(core): introduce glibre::Error + Result<T> aliases`).

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_contains:
    path: core/include/glibre/error.hpp
    regex: "using Result"
- unit_test_named: error_variant_holds_alternative_per_context
- unit_test_named: result_alias_default_constructs_to_void_success
- unit_test_named: result_alias_propagates_error_via_unexpected
- unit_test_named: error_context_round_trip
- workflow_passed: ci.yml
```

[...]

PR #1028, opened 2026-05-10
feat(core): TypeRegistry immutable-after-init

### LINKED ISSUE #597: [PLAN] feat(core): TypeRegistry immutable-after-init
### Domain

core

### Phase

phase:mvp

pts:3

### User Stories Satisfied

- #347 — `[STORY] type-registry-immutable-after-init`

### Parent (Epic or Sub-Epic)

#19

### Scope

Implement `TypeRegistry` (`§4.9`, `§5.5`, `§6.9`): the immutable-
after-init `TypeId → ColumnDescriptor { size, align, drop, layout }`
lookup populated from `glibre-types.dylib` codegen output and
extended (append-only) by per-plugin registration calls.

In scope:
- `core/include/glibre/core/type_registry.hpp` —
  `class TypeRegistry` with
  `lookup(TypeId) -> Result<const ColumnDescriptor*>`,
  `is_registered(TypeId) -> bool`, `count() -> size_t`.
- `core/src/type_registry/type_registry.cpp` — flat
  `std::pmr::vector<ColumnDescriptor>` indexed by `TypeId.value`.
  `register_type(TypeId, ColumnDescriptor) -> Result<void>` rejects
  with `core::Error::TypeRegistryClosed` after `seal()` is called by
  `World::create()` post-bootstrap.
- Append-only across hot-reload (`§6.9`): a `register_type` call
  during `glibre_plugin_register` is permitted and bypasses the
  `seal()` check via a friend hook the loader holds.
- Bootstrap from `glibre-types.dylib`'s `_registry.cpp` codegen
  output via a static-init exported function (data context
  contract; this plan reads it but does not author it).

Out of scope:
- Codegen pipeline that emits `_registry.cpp` (data context).
- Reflective fields beyond size/align/drop/layout (out of scope per
  `§3.2 #4`).

### Unit Test Plan

- core/type_registry: lookup_unregistered_yields_TypeUnregistered
- core/type_registry: register_after_seal_yields_TypeRegistryClosed
- core/type_registry: register_during_plugin_register_admitted
- core/type_registry: lookup_index_based_no_string_no_reflection
- core/type_registry: bootstrap_from_static_init_populates_count

### Depends on

- #233 (Error variant)
- #238 (PerContextAllocator)

### Agent Execution Notes

Branch `feat/core-type-registry`. PR title:
`feat(core): TypeRegistry immutable-after-init`. Storage uses
`std::pmr::vector` under `ContextTag::core`. Lookup is plain index
+ bounds check; no `unordered_map`, no string keys (`§6.9`). The
loader-friend hook is a single `extend_during_load(...)` member
gated by a `friend class PluginLoader;` declaration.

## Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: core/include/glibre/core/type_registry.hpp
- file_exists: core/src/type_registry/type_registry.cpp
- file_contains:
    path: tests/core/type_registry/type_registry_test.cpp
    regex: '"core/type_registry: lookup_unregistered_yields_TypeUnregistered"'
- file_contains:
    path: tests/core/type_registry/type_registry_test.cpp
    regex: '"core/type_registry: register_after_seal_yields_TypeRegistryClosed"'
- unit_test_named: "core/type_registry: register_during_plugin_register_admitted"
- file_contains:
    path: tests/core/type_registry/type_registry_test.cpp
    regex: '"core/type_registry: lookup_index_based_no_string_no_reflection"'
- file_contains:
    path: tests/core/type_registry/type_registry_test.cpp
    regex: '"core/type_registry: bootstrap_from_static_init_populates_count"'
- workflow_passed: ci.yml
```

cjhowe-us:
Closes #597

## Summary

Implements `TypeRegistry` per specs/core/SPEC.md §4.9, §5.5, and §6.9: an
immutable-after-init `TypeId → ColumnDescriptor` lookup table that stores
`{size, align, drop, layout}` descriptors for every registered component type.

- **Storage**: flat `std::pmr::vector<ColumnDescriptor>` indexed by
  `TypeId.value` (O(1) lookup, no hash map, no string keys — §6.9).
- **Seal gate**: `seal()` (called by `World::create()`) flips `sealed_=true`;
  subsequent `register_type()` calls return `core::Error::TypeRegistryClosed`.
- **Hot-reload extension**: `extend_during_load()` is a `private` member gated
  by `friend class PluginLoader` — it bypasses the seal so plugins loaded after
  `World` construction can still register new types (§4.9 invariant 1 carve-out,
  §6.9 append-only rule).
- **Error variants**: `core::Error::TypeUnregistered`, `core::Error::TypeRegistryClosed`,
  and `core::Error::TypeRegistryGap` were absent from the existing `core::Error`
  enum (closed by #233) and are added here. The SPEC §10.1 error table lists all
  three arms. Per the plan dispatch note, adding missing arms is in-scope because
  the five named unit tests depend on them. `log_error.hpp`'s `to_string()` switch
  and `error_register_test.cpp`'s `kCoreErrorValues` array are updated
  accordingly (the `-Werror=switch` build flag enforces completeness at compile
  time).

## Test plan

- [ ] All 5 named Catch2 tests in `tests/core/type_registry/` pass locally
      (confirmed: 214/214 tests pass, 0 failures).
- [ ] `error_register_per_context_arms_unique` still passes with `kCoreErrorValues`
      updated from 18 to 21 entries.
- [ ] `tostring_complete_for_core_error` updated to cover all three new arms.
- [ ] CI `ci.yml` workflow passes on this PR.

## Files changed

- `core/include/glibre/core/type_registry.hpp` — new header
- `core/src/type_registry/type_registry.cpp` — new implementation
- `core/include/glibre/error.hpp` — `TypeUnregistered` + `TypeRegistryClosed` + `TypeRegistryGap` arms
- `core/include/glibre/log_error.hpp` — `to_string()` arms for all three new variants
- `core/CMakeLists.txt` — adds `type_registry.cpp` to `glibre-core`
- `tests/core/type_registry/type_registry_test.cpp` — 5 named Catch2 tests
- `tests/core/type_registry/CMakeLists.txt` — test executable wiring
- `tests/core/CMakeLists.txt` — `add_subdirectory(type_registry)`
- `tests/core/error_register/error_register_test.cpp` — `kCoreErrorValues` 18→21
- `tests/core/log_error_test.cpp` — `tostring_complete_for_core_error` extended with three new arms
- `specs/core/SPEC.md` — §10.1 table: added `TypeRegistryGap` row

Refs #347 (user story: type-registry-immutable-after-init)


core/include/glibre/core/type_registry.hpp
```diff
@@ -0,0 +1,194 @@
...
+    ~TypeRegistry() noexcept = default;
+
+    // lookup(id) — O(1) index into the flat descriptor vector.
+    //
+    // Returns:
+    //   Ok(&descriptor)               — id is registered.
+    //   Err(core::Error::TypeUnregistered) — id.value is out of range or the
+    //                                        slot was never populated.
+    [[nodiscard]] Result<const ColumnDescriptor*> lookup(TypeId id) const noexcept;
+
```
cjhowe-us, line 174: severity:HIGH location:core/include/glibre/core/type_registry.hpp:128 problem:Constructor accepts raw `std::pmr::memory_resource*` and (per type_registry.cpp:294) silently falls back to `std::pmr::get_default_resource()` when null. This bypasses `glibre::PerContextAllocator` tag tracking and the 64 MiB core ceiling — directly contradicting both the file-header comment ("the capacity is bounded by the 64 MiB core heap ceiling") and `reviews/decisions/perf-budget.md` Allocator Rule #1 ("Every allocation carries a `ContextTag`"). Lookups + registrations into `entries_` will not be billed against `ContextTag::core`, so the perf-budget gate cannot detect drift. fix:Take a `glibre::PerContextAllocator&` (or a `std::pmr::memory_resource*` derived from one) and require non-null; document that callers must pass a `core`-tagged allocator. Either expose a `pmr_view()` on `PerContextAllocator` or wrap it with a small `memory_resource` adapter under `core/src/alloc/` that forwards to `allocate()/deallocate()` so the bytes flow through the tag counter.

core/include/glibre/core/type_registry.hpp
```diff
@@ -0,0 +1,194 @@
...
+//
+// Lifecycle (called by World, PluginLoader):
+//   register_type(TypeId, ColumnDescriptor) -> Result<void>
+//      -- open before seal(); refused after seal() unless called from
+//         extend_during_load() (loader friend hook).
+//   seal()
+//      -- called once by World::create() at end of bootstrap.
+//
+// Friend hook (called by PluginLoader only):
+//   extend_during_load(TypeId, ColumnDescriptor) -> Result<void>
```
cjhowe-us, line 143: severity:MED location:core/include/glibre/core/type_registry.hpp:99-103 problem:`TypeId` is duplicated here with the in-line comment "This duplicate must stay in sync" — the canonical definition is supposed to live in SPEC §5.2 / the §5 stub header. Any subsequent plan that also needs `TypeId` (schedule, archetype, plugin-owned-types) will either include this header for it (creating a wrong dependency) or define a third copy. fix:Pull `TypeId` into a tiny shared header `core/include/glibre/core/type_id.hpp` now (plan #597 is the first user; the §5 stub plan can move it later by replacement). Either way, remove the "stay in sync" liability before more headers depend on this duplicate.

tests/core/type_registry/type_registry_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+// of the friend-declared class, the compiler grants it friend access.
+// ===========================================================================
+
+namespace glibre::core {
+
+// Minimal stub for PluginLoader — grants friend access to extend_during_load.
+// This stub is compiled only in this test TU and is distinct from any future
+// real PluginLoader implementation.
+class PluginLoader {
+public:
```
cjhowe-us: severity:HIGH location:tests/core/type_registry/type_registry_test.cpp:201-213 problem:The test redeclares `class glibre::core::PluginLoader` to obtain friend access. The real `glibre::core::PluginLoader` already exists at `core/include/glibre/core/plugin_loader.hpp:85` and its impl TU `src/plugin_loader.cpp` is linked into `glibre::core`, which this test executable links against. Two distinct definitions of the same class type now exist in the same program — strict ODR violation per [basic.def.odr]/12. It happens to link today only because the stub holds a single static method (no vtable / no instance data) so no symbols collide; LTO, future class additions, or compilers that hash class definitions could break this silently. fix:Add a public test seam to TypeRegistry: `friend class TypeRegistryTestAccess;` plus a `class TypeRegistryTestAccess { public: static Result<void> extend(TypeRegistry&, TypeId, ColumnDescriptor); };` declared in a small test-only header, OR gate a public `extend_during_load` overload behind `#if GLIBRE_TESTING`. Either approach removes the duplicate `PluginLoader` definition.

core/src/type_registry/type_registry.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+std::size_t TypeRegistry::count() const noexcept { return entries_.size(); }
+
+Result<void> TypeRegistry::register_type(TypeId id, ColumnDescriptor desc) noexcept {
+    if (sealed_) {
+        return std::unexpected(Error{core::Error::TypeRegistryClosed});
+    }
+    // Enforce contiguous assignment: id.value must equal the next empty slot.
+    if (id.value != entries_.size()) {
+        return std::unexpected(Error{core::Error::TypeUnregistered});
```
cjhowe-us: severity:MED location:core/src/type_registry/type_registry.cpp:55 (also line 67 in extend_during_load) problem:Returns `core::Error::TypeUnregistered` when `id.value != entries_.size()` during a *register* call. SPEC §10.1 defines `TypeUnregistered` as a *lookup-failure* arm ("Public API names a `TypeId` not present in the immutable-after-init `TypeRegistry`"). Reusing it for a register-time contiguity violation conflates two semantically distinct failures and weakens callers ability to distinguish them in logs/telemetry. The header comment at `type_registry.hpp:148-151` even acknowledges this is a "codegen contract violation" the caller "should assert rather than handle". fix:Either (a) introduce a dedicated `core::Error::TypeIdOutOfOrder` arm (+ SPEC §10.1 row), or (b) treat this as a hard contract violation via `std::abort()` since the codegen pipeline guarantees ordering. Do not overload `TypeUnregistered`.

core/include/glibre/core/type_registry.hpp
```diff
@@ -0,0 +1,194 @@
...
+    //
+    // Preconditions (same as register_type):
+    //   id.value == entries_.size()  — must be the next slot.
+    //   desc.size > 0 && desc.align > 0
+    //
+    // Returns:
+    //   Ok(void)                        — registration succeeded.
+    //   Err(core::Error::TypeUnregistered) — id.value is not the next slot
+    //                                       (codegen contract violation).
+    [[nodiscard]] Result<void> extend_during_load(TypeId id, ColumnDescriptor desc) noexcept;
```
cjhowe-us, line 228: severity:MED location:core/include/glibre/core/type_registry.hpp:166-183 problem:`extend_during_load()` is gated by `friend class PluginLoader`, but no precondition asserts the loader is actually inside `glibre_plugin_register`. A future / buggy PluginLoader call between phases (e.g. mid-frame, or after the loader transitions out of "load" state) would silently mutate a sealed registry, breaking SPEC §4.9 invariant 1 from inside the friend — exactly the case the seal is meant to prevent. The friend grant is currently coarser than the SPEC §4.9 carve-out language ("plugin-load-time registration only"). fix:Add an `is_loading_` flag on `TypeRegistry` that PluginLoader sets via `begin_load()` / `end_load()` (also friend-gated) bracketing each `glibre_plugin_register` call; have `extend_during_load()` return `TypeRegistryClosed` if `!is_loading_`. This mirrors the §6.7 hot-reload barrier latch model and makes the SPEC §4.9 invariant 1 self-defending.

core/src/type_registry/type_registry.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "glibre/core/type_registry.hpp"
+
+#include <memory_resource>
+
+namespace glibre {
+namespace core {
+
+TypeRegistry::TypeRegistry(std::pmr::memory_resource* resource) noexcept
+    : entries_(resource != nullptr ? resource : std::pmr::get_default_resource()) {}
```
cjhowe-us: severity:LOW location:core/src/type_registry/type_registry.cpp:36 problem:`entries_(resource != nullptr ? resource : std::pmr::get_default_resource())` silently falls back to the global default resource when the caller passes nullptr. Even after the HIGH-1 fix, the constructor should make `nullptr` a hard failure (e.g. `std::abort()` on null) so the registry never lands on the global heap by accident. The current ternary is silent and easy to miss. fix:Either remove the default argument and require non-null at the call site, or `if (resource == nullptr) std::abort();` with a comment citing perf-budget Allocator Rule #1.

core/src/type_registry/type_registry.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+Result<const ColumnDescriptor*> TypeRegistry::lookup(TypeId id) const noexcept {
+    if (id.value >= entries_.size()) {
+        return std::unexpected(Error{core::Error::TypeUnregistered});
```
cjhowe-us: severity:LOW location:core/src/type_registry/type_registry.cpp:40,55,67 (every std::unexpected site) problem:Each error is constructed as `Error{core::Error::TypeUnregistered}` / `Error{core::Error::TypeRegistryClosed}` with no `ErrorContext` payload — file/line/detail are all empty. The error-model decision record §"Logging / Telemetry" #2 expects `ErrorContext` to be populated; recent core plans (#239, #250) attach `ErrorContext{__FILE__, __LINE__, "<hint>"}` at every raise site. This site silently drops that observability. fix:Pass `ErrorContext{__FILE__, __LINE__, "lookup oob" / "register after seal" / "out-of-order register"}` into each `std::unexpected` construction.

tests/core/type_registry/type_registry_test.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+    std::pmr::monotonic_buffer_resource buf{8192};
+    glibre::core::TypeRegistry reg{&buf};
+
+    constexpr std::size_t kN = 8;
+    for (std::size_t i = 0; i < kN; ++i) {
+        auto r = reg.register_type(glibre::core::TypeId{i}, make_desc(i + 1, 1));
+        REQUIRE(r.has_value());
+    }
+
```
cjhowe-us, line 358: severity:LOW location:tests/core/type_registry/type_registry_test.cpp:265-308 problem:`lookup_o1_no_string_no_reflection` claims to verify O(1) but only proves "lookup of N items succeeds N times and returns distinct values". A linear scan implementation would also pass this test. The compile-time `static_assert`s for `is_polymorphic_v` and `is_trivially_copyable_v` cover the "no reflection / no virtual" half cleanly, but the "O(1)" claim is structural, not measured. fix:Either (a) rename to `lookup_indexed_returns_correct_descriptor` to match what is actually verified, or (b) add a microbenchmark sanity check (Catch2 BENCHMARK section) asserting wall-time(N=10000) ≈ wall-time(N=1) within a tolerance — keeps the DoD test name honest.

cjhowe-us (commented):
round:1 reviewer:go-review verdict:REQUEST_CHANGES findings:HIGH:2 MED:3 LOW:3 carryover:0
notes:Coverage + correctness lens. CI is green on the latest commit (815064c) and the PR cleanly closes #597 with all five named Catch2 tests present and matching the DoD `unit_test_named` assertions verbatim. The core::Error extension (`TypeUnregistered` + `TypeRegistryClosed`) is explicitly authorised by SPEC §10.1 and does not break the #233 invariants — `kExpectedArmCount` (variant arms = 4) is unchanged; only the inner enumerator count moved 18→20 and the test array + `to_string` switch are both updated. PR title is conformant Conventional Commit; `Closes #597` is in the body. Two HIGH findings block APPROVE: (1) the `std::pmr::memory_resource*` constructor falls back to `std::pmr::get_default_resource()`, bypassing `glibre::PerContextAllocator` and the 64 MiB core ceiling — directly contradicting both the file's own header comment and `perf-budget.md` Allocator Rule #1; (2) the test TU defines a stub `glibre::core::PluginLoader` whose name + namespace already exist in the linked-in real PluginLoader (`plugin_loader.hpp:85`) — strict ODR violation that links today only because the stub holds a single static method. MED findings: `TypeUnregistered` is overloaded for the register-time contiguity check (SPEC §10.1 reserves it for lookup failures); `extend_during_load()` friend grant is coarser than the §4.9 carve-out language ("plugin-load-time only") with no `is_loading_` latch; `TypeId` duplicated in this header with a fragile "stay in sync" liability. LOWs cover absent `ErrorContext` payloads, silent nullptr-resource fallback, and the misleadingly-named `lookup_o1_no_string_no_reflection` test that does not actually verify O(1).

core/include/glibre/core/type_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:69043a1
HIGH-1 addressed: Replaced the `std::pmr::memory_resource*` constructor (with its `get_default_resource()` fallback) with an explicit `PerContextAllocator&` parameter. Added nested `PerContextAllocatorResource` class in type_registry.hpp / type_registry.cpp that adapts `PerContextAllocator` to `std::pmr::memory_resource` so all TypeRegistry storage flows through the ceiling-tracked allocator path. All test fixtures updated to pass a `PerContextAllocator{ContextTag::core}`.

tests/core/type_registry/type_registry_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:69043a1
HIGH-2 addressed: Removed the ODR-violating stub `class PluginLoader` from the test TU. Added `class TypeRegistryTestHook` as an explicit named friend in type_registry.hpp (alongside the existing PluginLoader friend). TypeRegistryTestHook is defined in the test TU and provides `begin_load()`, `end_load()`, and `extend()` accessors. The real `plugin_loader.hpp` is now `#include`d in the test file to prevent any future re-definition from silently passing the linker.

core/src/type_registry/type_registry.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:69043a1
MED-3 addressed: Added `TypeRegistryGap` arm to `core::Error` in error.hpp with an explanatory comment (codegen contract violation — out-of-order or sparse registration). Added `to_string` case in log_error.hpp. `register_type()` and `extend_during_load()` now emit `TypeRegistryGap` for the contiguous-slot check; `TypeUnregistered` is reserved exclusively for lookup failures (SPEC §10.1).

core/include/glibre/core/type_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:69043a1
MED-4 addressed: Added `is_loading_` bool member and a private `set_loading_(bool)` setter. `extend_during_load()` asserts `is_loading_` at entry. Only `PluginLoader` and `TypeRegistryTestHook` (friends) can call `set_loading_()`. Tests use `TypeRegistryTestHook::begin_load(reg)` / `end_load(reg)` around any `extend_during_load` call.

core/include/glibre/core/type_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:DEFER
MED-5: The TypeId duplication is a real concern. Per the dispatch instructions, the preferred fix for out-of-scope consolidation is: add static_asserts that pin the two definitions to the same byte width and alignment (so divergence fails the build), plus open a spike issue. Both are done in commit 69043a1 (three static_asserts on trivially_copyable, sizeof, alignof). Spike for canonical consolidation: #1030 [SPIKE] iterate-type-id-canonical-home.

core/src/type_registry/type_registry.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:NOOP reasoning:LOW-7 (silent nullptr fallback) is fully subsumed by the HIGH-1 fix. The constructor no longer accepts a std::pmr::memory_resource* at all — nullptr cannot be passed. The PerContextAllocator& parameter is required; there is no fallback path remaining.

core/src/type_registry/type_registry.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:69043a1
LOW-6 addressed: Added ErrorContext{__FILE__, __LINE__, detail} to all std::unexpected sites in type_registry.cpp: lookup() id-out-of-range, register_type() sealed check, register_type() gap check, extend_during_load() gap check.

tests/core/type_registry/type_registry_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:69043a1
LOW-9 addressed: Renamed test case from lookup_o1_no_string_no_reflection to lookup_index_based_no_string_no_reflection. The new name accurately describes what the test verifies: index-based lookup with no string keys and no virtual dispatch (structural proof via static_asserts). CMakeLists.txt comment updated. DoD in issue #597 will be updated to match the new name.

core/include/glibre/core/type_registry.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+    // set_loading_(flag) — set/clear the is_loading_ latch.
+    //
+    // Called by PluginLoader at the entry and exit of its loading critical
+    // section.  Only PluginLoader and TypeRegistryTestHook (friends) may call
+    // this; all other callers are prevented by the friend guard.
+    void set_loading_(bool flag) noexcept { is_loading_ = flag; }
+
+    friend class PluginLoader;
+    friend class TypeRegistryTestHook;
```
cjhowe-us, line 239: severity:HIGH location:core/include/glibre/core/type_registry.hpp:229 problem:TypeRegistryTestHook is declared as a friend in the production header but is only defined in the test TU. This means every shipping build of TypeRegistry permanently exposes set_loading_() and extend_during_load() to any TU willing to define `class glibre::core::TypeRegistryTestHook { ... };`. The R1 HIGH-2 fix traded an ODR violation for a permanent backdoor in production headers — the §4.9 invariant 1 seal that the rest of the file is built around can be bypassed by anyone in any context. fix:Wrap the friend declaration in `#if defined(GLIBRE_TESTING)` (the same guard alloc.hpp uses for testing_register_allocator_call_count). The test TU already compiles with GLIBRE_TESTING=1; the production build then carries no test-only friendship and the seal cannot be bypassed by a stub.

core/include/glibre/core/type_registry.hpp
```diff
@@ -0,0 +1,272 @@
...
+            : alloc_{alloc} {}
+
+    protected:
+        void* do_allocate(std::size_t bytes, std::size_t alignment) override;
+        void do_deallocate(void* p, std::size_t bytes, std::size_t alignment) noexcept override;
+        bool do_is_equal(const std::pmr::memory_resource& other) const noexcept override;
+
+    private:
+        PerContextAllocator& alloc_;
+    };
```
cjhowe-us: severity:MED location:core/include/glibre/core/type_registry.hpp:245-257 problem:PerContextAllocatorResource is a generic PerContextAllocator → std::pmr::memory_resource adaptor that is unrelated to TypeRegistry — it has two reasons to change (PMR semantics drift; TypeRegistry semantics drift) and so violates SRP per PHILOSOPHY §1. The next plan that needs std::pmr storage tracked under PerContextAllocator (e.g. ArchetypeStorage, plugin-owned-types tables, schedule plan storage) will either duplicate this ~12-line adaptor or include type_registry.hpp purely for the nested adaptor type — a wrong dependency arrow. fix:Lift PerContextAllocatorResource into core/include/glibre/alloc.hpp (or a sibling core/include/glibre/pmr_adapter.hpp) as a top-level non-nested glibre::PerContextAllocatorResource. TypeRegistry then constructs it by value the same way, and the next consumer reuses the type. This also collapses the "two private nested classes" liability into one shared public adaptor.

core/src/type_registry/type_registry.cpp
```diff
@@ -0,0 +1,156 @@
...
+    entries_.push_back(desc);
+    return {};
+}
+
+void TypeRegistry::seal() noexcept { sealed_ = true; }
+
+Result<void> TypeRegistry::extend_during_load(TypeId id, ColumnDescriptor desc) noexcept {
+    // is_loading_ latch: extend_during_load() is only valid inside
+    // PluginLoader's loading critical section (round-1 review MED-4 fix).
+    assert(is_loading_ && "extend_during_load() called outside loader critical section");
```
cjhowe-us: severity:MED location:core/src/type_registry/type_registry.cpp:139 problem:extend_during_load() asserts is_loading_ in debug but in release builds the assert is elided and a caller outside the load window silently appends to a sealed registry, breaking SPEC §4.9 invariant 1 from inside the friend grant. The R1 MED-4 finding explicitly recommended: "have extend_during_load() return TypeRegistryClosed if !is_loading_" — i.e. enforce the latch in all builds, not assert in debug only. The current "assert in debug, silent UB in release" is the worst of both worlds for a seal that the entire module is built around. fix:Replace the assert with a real check that returns std::unexpected{Error::TypeRegistryClosed, ErrorContext{__FILE__, __LINE__, "extend_during_load called outside load window"}} when !is_loading_. Keep the assert as a redundant debug-only tripwire above the check if you want loud failures during development.

core/src/type_registry/type_registry.cpp
```diff
@@ -0,0 +1,156 @@
...
+    const std::pmr::memory_resource& other
+) const noexcept {
+    // dynamic_cast is unavailable under -fno-rtti.  Identity equality: two
+    // PerContextAllocatorResource instances wrapping the same PerContextAllocator
+    // are equal if they are the same object (pointer equality on *this).
+    // For TypeRegistry, there is exactly one alloc_resource_ per instance and
+    // the vector never transfers resources across allocators, so self-equality
+    // is the only meaningful case.
+    return this == &other;
+}
```
cjhowe-us: severity:LOW location:core/src/type_registry/type_registry.cpp:80 problem:Comment claims "dynamic_cast is unavailable under -fno-rtti" but the project does not currently build with -fno-rtti (RTTI is needed for std::expected error-path machinery and several std types in the public ABI). The pointer-identity equality is correct (each TypeRegistry owns exactly one alloc_resource_ instance and never transfers ownership across allocators) but the rationale comment misstates the build constraint. fix:Replace the comment with the real reason: "Identity equality suffices because each TypeRegistry owns exactly one PerContextAllocatorResource and the wrapped allocator is non-transferable (PerContextAllocator is non-copyable, non-movable per alloc.hpp:164-167). Two distinct PerContextAllocatorResource instances are never equal in this codebase." Drop the -fno-rtti reference.

core/src/type_registry/type_registry.cpp
```diff
@@ -0,0 +1,156 @@
...
+//   critical section and cleared at exit.  This ensures that
+//   extend_during_load() cannot be called outside that window.
+//
+// ## -fno-exceptions clean
+//   No exceptions thrown or propagated.
+
+#include "glibre/core/type_registry.hpp"
+
+#include <cassert>
+#include <memory_resource>
```
cjhowe-us: severity:LOW location:core/src/type_registry/type_registry.cpp:47 problem:`#include <memory_resource>` is redundant in the .cpp — the header already pulls it in transitively via the override declarations, and removing it from the .cpp produces no diagnostic. Trivial cleanup, but the header file-grouping guideline says implementation TUs should only include what the header omits. fix:Drop the redundant `#include <memory_resource>` line; the header include on line 44 already covers it.

cjhowe-us (commented):
round:2 reviewer:go-review verdict:REQUEST_CHANGES findings:HIGH:1 MED:2 LOW:2 carryover:0
notes:Cohesion + SRP + seam quality lens. CI is green on tip 7bf910a (all 5 checks SUCCESS) and the R1 carryover (HIGH-1 allocator injection, HIGH-2 ODR, MED-3 TypeRegistryGap arm, MED-4 is_loading_ latch, MED-5 static_asserts + spike #1030, LOW-6 ErrorContext, LOW-9 rename) is materially addressed in commits 69043a1 + 7bf910a — every R1 finding has either an ADDRESSED or DEFER decision and the test changes line up with the production header changes. Three new findings block APPROVE: (1) HIGH — the R1 HIGH-2 fix for ODR introduced TypeRegistryTestHook as a friend declaration in the production header without any GLIBRE_TESTING guard, so any TU in any context can declare the hook class and bypass the §4.9 invariant 1 seal. This is a permanent backdoor in shipping code; needs an `#if defined(GLIBRE_TESTING)` wrapper on the friend declaration (alloc.hpp already uses this idiom for testing_register_allocator_call_count). (2) MED — PerContextAllocatorResource is a generic PMR adaptor unrelated to TypeRegistry; nesting it inside TypeRegistry violates SRP and forces the next consumer (Schedule, ArchetypeStorage, plugin-owned-types) to either duplicate or include type_registry.hpp for a memory-resource adaptor. Should be lifted to glibre/alloc.hpp or a sibling pmr_adapter.hpp as a top-level glibre::PerContextAllocatorResource. (3) MED — extend_during_load() asserts is_loading_ in debug only; release builds silently mutate a sealed registry on a buggy/late call. The R1 MED-4 finding's recommended fix was an actual return-on-failure check, not an assert; current debug-asserts-release-silent is the worst combination for the seal that the whole module is built around. LOWs cover a misstated -fno-rtti rationale comment in do_is_equal and a redundant <memory_resource> include in the .cpp. None of the new findings would break the type-registry contract today, but the HIGH backdoor will rot the seal as new contexts add tests; it should land before merge.

core/include/glibre/core/type_registry.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4d6b8

Wrapped `friend class TypeRegistryTestHook` in `#if defined(GLIBRE_TESTING)` in both the forward-declaration block and the `friend` declaration inside the class body. Production builds compiled without `GLIBRE_BUILD_TESTS=ON` never define `GLIBRE_TESTING`, so the backdoor is compiled out. Test builds receive `GLIBRE_TESTING=1` transitively via the PUBLIC compile-definition on `glibre-core` (core/CMakeLists.txt if(GLIBRE_BUILD_TESTS)) — no per-target redefinition needed. The §4.9 seal is now sealed in shipping code.

core/include/glibre/core/type_registry.hpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4d6b8

Lifted `PerContextAllocatorResource` from a nested class inside `TypeRegistry` to a top-level `glibre::PerContextAllocatorResource` in `core/include/glibre/alloc.hpp` and `core/src/alloc/per_context_allocator.cpp`. `type_registry.hpp` now declares the member as `glibre::PerContextAllocatorResource alloc_resource_` using the top-level class. Future PMR consumers (PhaseRegistry, plugin-owned-type tables per spike #1031) can include `glibre/alloc.hpp` without pulling `type_registry.hpp`. Two reasons to change are separated: PMR semantics drift lives in alloc.hpp; TypeRegistry storage semantics live in type_registry.hpp.

core/src/type_registry/type_registry.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4d6b8

Replaced the debug-only `assert(is_loading_)` in `extend_during_load()` with an always-on `if (!is_loading_) return std::unexpected{TypeRegistryClosed}` check. In release builds the old assert was compiled out and a late caller would silently append to a sealed registry (SPEC §4.9 invariant 1 violation). Now both debug and release builds return a typed `core::Error::TypeRegistryClosed` error. A new test section `extend_during_load outside load window returns TypeRegistryClosed (MED-3 fix)` in the existing `register_during_plugin_register_admitted` test case exercises this path and passes.

core/src/type_registry/type_registry.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4d6b8

Replaced the stale '-fno-rtti disables dynamic_cast' rationale with the accurate justification: PerContextAllocator is non-copyable (single-instance-per-context per perf-budget.md §Allocator Rules #1), so each PerContextAllocatorResource wraps exactly one allocator instance, making self-equality (pointer identity on *this) the only meaningful case. PMR containers call is_equal before swapping resources; pointer identity prevents cross-resource allocation-transfer. The comment now lives in alloc.hpp where the class is defined (MED-2 lift).

core/src/type_registry/type_registry.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:6d4d6b8

Removed the redundant `#include <memory_resource>` from `type_registry.cpp`. The header `type_registry.hpp` already includes `<memory_resource>` for `std::pmr::vector`, which is transitively available in the .cpp. The `PerContextAllocatorResource` virtual methods moved to `per_context_allocator.cpp` (MED-2 lift), which gets `<memory_resource>` via `alloc.hpp`.

tests/core/log_error_test.cpp
```diff
@@ -219,6 +219,10 @@ TEST_CASE("core/log_error: tostring_complete_for_core_error", "[core][log_error]
     check(E::NullArgument, "NullArgument");
     // plan #250: PluginNameMismatch — hot-reload swap candidate has different name
     check(E::PluginNameMismatch, "PluginNameMismatch");
+    // plan #597: TypeUnregistered — lookup on unregistered TypeId
+    check(E::TypeUnregistered, "TypeUnregistered");
+    // plan #597: TypeRegistryClosed — register_type() called after seal()
+    check(E::TypeRegistryClosed, "TypeRegistryClosed");
```
cjhowe-us, line 225: severity:HIGH location:tests/core/log_error_test.cpp:201-225 problem:tostring_complete_for_core_error documents that it verifies to_string returns the correct literal for "every enumerator in core::Error", but the array now misses TypeRegistryGap (added by the R1 MED-3 fix as enum arm 21). The test will silently pass if a future change returns "" or the wrong literal for TypeRegistryGap; -Werror=switch only catches missing-arm, not wrong-value regressions, so this test IS the runtime safety net. fix:Append `check(E::TypeRegistryGap, "TypeRegistryGap");` after the TypeRegistryClosed line, matching the pattern used for the other plan #597 arms.

tests/core/error_register/error_register_test.cpp
```diff
@@ -84,6 +84,10 @@ constexpr eastl::array<std::underlying_type_t<glibre::core::Error>, 18> kCoreErr
     static_cast<std::uint16_t>(glibre::core::Error::NullArgument),
     // plan #250: PluginNameMismatch — hot-reload swap candidate has a different name.
     static_cast<std::uint16_t>(glibre::core::Error::PluginNameMismatch),
+    // plan #597: TypeUnregistered — lookup on a TypeId with no registered descriptor.
+    static_cast<std::uint16_t>(glibre::core::Error::TypeUnregistered),
+    // plan #597: TypeRegistryClosed — register_type() called after seal().
+    static_cast<std::uint16_t>(glibre::core::Error::TypeRegistryClosed),
     // When a new enumerator is added to core::Error, add it here and
     // increment the array size template argument above.
 }};
```
cjhowe-us, line 95: severity:HIGH location:tests/core/error_register/error_register_test.cpp:65-93 problem:kCoreErrorValues template parameter is fixed at 20 entries but core::Error now has 21 enumerators (TypeRegistryGap was added by the R1 MED-3 fix and is missing from the array). The all_distinct/uniqueness assertion + the per_context_arms_unique test will silently pass on an incomplete enumeration; if TypeRegistryGap is ever assigned an explicit duplicate value, this test will not catch it. The header comment ("When a new enumerator is added to core::Error, add it here too") is the contract being violated. fix:Bump the array size template arg to 21 and append `static_cast<std::uint16_t>(glibre::core::Error::TypeRegistryGap),` with a `// plan #597: TypeRegistryGap — codegen contract violation (gap in TypeId sequence)` comment. Also update the in-line "20 enumerators" comment in the per_context_arms_unique test body to 21.

core/include/glibre/error.hpp
```diff
@@ -84,6 +84,27 @@ enum class Error : std::uint16_t {
...
+    // "TypeRegistryClosed".
+    TypeRegistryClosed,
+    // TypeRegistry register_type() / extend_during_load() called with an id.value
+    // that is not the next contiguous slot (id.value != entries_.size()).
+    // This signals a codegen contract violation — the emitter is expected to
+    // assign TypeId values in ascending order with no gaps.  SPEC §10 reserves
+    // TypeUnregistered for lookup-only failures; this distinct arm covers the
+    // registration-time gap check so callers can distinguish the two cases.
+    // (plan #597 review round 1, MED-3 fix)
+    TypeRegistryGap,
```
cjhowe-us, line 107: severity:MED location:specs/core/SPEC.md:2636-2637 problem:SPEC §10.1 Failure-Mode Table lists rows for TypeUnregistered + TypeRegistryClosed but not the new core::Error::TypeRegistryGap arm. The R1 MED-3 fix added the arm to error.hpp ("TypeUnregistered is reserved exclusively for lookup failures (SPEC §10.1)") but never added the matching §10.1 row. The round-3 lens criterion "every error variant arm cites §10.1" is therefore unmet. The error.hpp arm comment cites §10 row "TypeRegistryClosed" but no §10 row exists for TypeRegistryGap. fix:Add a §10.1 row between rows 2637 and 2638 along the lines of `| `TypeRegistryGap` | `register_type()` / `extend_during_load()` called with `id.value` not equal to `count()` — codegen emitted out-of-order or sparse registrations. | Refuse | none | `error` |`. Cite §4.9 (codegen contract) as the trigger reference.

core/include/glibre/error.hpp
```diff
@@ -84,6 +84,27 @@ enum class Error : std::uint16_t {
...
 };
```
cjhowe-us, line 108: severity:LOW location:PR body "Test plan" + "Files changed" sections problem:PR body still claims "kCoreErrorValues array updated 18→20 entries" and "tostring_complete_for_core_error updated to cover the two new arms" — but the R1 MED-3 fix added a third arm (TypeRegistryGap) which is not reflected in the body. The squash-commit message will inherit the stale 18→20 figure if the PR is squash-merged with the body verbatim. fix:Update the PR body bullet to "kCoreErrorValues 18→21 entries (TypeUnregistered + TypeRegistryClosed + TypeRegistryGap)" and the test-plan checkbox to "tostring_complete_for_core_error covers all three plan #597 arms". Aligns with the HIGH-1 + HIGH-2 test additions above.

cjhowe-us (commented):
round:3 reviewer:go-review verdict:REQUEST_CHANGES findings:HIGH:2 MED:1 LOW:1 carryover:0
notes:Topology + spec-alignment + polish lens. CI is green on tip 6d4d6b8 (build, lint, dep-check, spec-check, automated-review all SUCCESS) and every R1 + R2 finding has a decision reply (8 ADDRESSED, 1 DEFER → spike #1030 is parented + scoped, 1 NOOP — zero unresolved carryover). Spike #1030 (TypeId canonical home) and spike #1031 (PhaseRegistry allocator parity) both carry the right labels (type:spike + phase:mvp + domain:core), populated DoD blocks, and parent epics. Per initiative #1032 the PR is allowed to land on EASTL (eastl::vector / string_view); the per-context migration sweeps it later — no inline EASTL findings. Two new HIGH findings block APPROVE: both stem from the same root cause — the R1 MED-3 fix added a third core::Error arm (TypeRegistryGap) but the two test-side enumerators that runtime-mirror the closed-sum invariant were not updated. (HIGH-1) tostring_complete_for_core_error documents "every enumerator in core::Error" but only checks 20 of 21 arms; -Werror=switch catches missing arms but not wrong-string regressions, so this runtime test IS the safety net for that arm. (HIGH-2) kCoreErrorValues is templated <…, 20> with 20 entries; TypeRegistryGap is missing, breaking the "add new enumerator here too" contract documented in the array's trailing comment. MED-1: SPEC §10.1 Failure-Mode Table has rows for TypeUnregistered + TypeRegistryClosed but not for TypeRegistryGap — the round-3 lens criterion "every error variant arm cites §10.1" is unmet. LOW-1: PR body still cites "18→20" and "two new arms" — stale by one arm; will leak into the squash-commit message. None of these block production correctness, but the three test/spec arms together turn the closed-sum guarantee on TypeRegistryGap into a silently-incomplete one. Fix is one-line additions to two test files plus a single new §10.1 row.

tests/core/log_error_test.cpp
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:56d6f02 — Added `check(E::TypeRegistryGap, "TypeRegistryGap")` as the 21st arm in `tostring_complete_for_core_error` in `tests/core/log_error_test.cpp`. All 11 log_error + error_register tests pass locally.

tests/core/error_register/error_register_test.cpp
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:56d6f02 — Bumped `kCoreErrorValues` template parameter from 20 to 21 and appended `static_cast<std::uint16_t>(glibre::core::Error::TypeRegistryGap)` with the `// plan #597: TypeRegistryGap — codegen contract violation (gap in TypeId sequence).` comment. Also updated the per_context_arms_unique inline comment from "20 enumerators 0..19" to "21 enumerators 0..20". All 11 log_error + error_register tests pass locally.

core/include/glibre/error.hpp
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:56d6f02 — Added `TypeRegistryGap` row to SPEC.md §10.1 immediately after the `TypeRegistryClosed` row. Trigger condition: `register_type()` or `extend_during_load()` called with a `TypeId.value` that exceeds the current dense storage end by more than 1 (would create a gap in the TypeId sequence). Recovery: Refuse. Observer event: none. Severity: error.

core/include/glibre/error.hpp
cjhowe-us: responder:go-impl-respond round:3 decision:ADDRESSED commit:56d6f02 — PR body updated via `gh pr edit`: changed "18→20 entries" to "18→21 entries", "two new arms" to "three new arms" (TypeUnregistered + TypeRegistryClosed + TypeRegistryGap), and added `specs/core/SPEC.md` to the Files changed list. `Closes #597` is preserved.

PR #1036, opened 2026-05-10
fix(infra): verify_dod.py matches multi-line TEST_CASE invocations

### LINKED ISSUE #1035: [PLAN] fix(infra): verify_dod.py multi-line TEST_CASE handling
### Domain

infra

### Phase

phase:mvp

pts:2

### User Stories Satisfied

- N/A — infrastructure correctness fix; no user-visible behaviour. Discovered while diagnosing #597's `dod:failed` (4 of 5 `unit_test_named` assertions failed despite tests existing and passing in CI).

### Parent (Epic or Sub-Epic)

#1

### Scope

Fix `.github/scripts/verify_dod.py` so the `unit_test_named` check matches Catch2 `TEST_CASE` invocations whose name string is on the next line after `TEST_CASE(`. clang-format with the repo's `ColumnLimit: 100` wraps any `TEST_CASE("long-name", "[tag]") {` that exceeds 100 columns onto two lines, e.g.:

```cpp
TEST_CASE(
    "core/type_registry: lookup_unregistered_yields_TypeUnregistered", "[core][type_registry]"
) {
```

The current verifier uses `grep -RE '(TEST_CASE|SCENARIO)\s*\(\s*"<name>"' tests/`, where `\s*` does not span newlines under POSIX-extended regex. Result: any leaf with a long-named `TEST_CASE` hits a spurious `dod:failed` even when the test exists and runs green. #597 is the first to surface it; tests/core/transient_arena_test.cpp, tests/core/plugin_loader_register/, tests/core/plugin_loader_registry/, tests/core/frame_loop_integration/, tests/shader/source/, tests/core/world/perf/, and tests/data/schemas/geometry/ all already contain multi-line TEST_CASEs and are silent landmines.

In scope:
- `.github/scripts/verify_dod.py::check_unit_test_named` — switch from `grep -RE` to a Python implementation that scans `tests/**/*.cpp` (and `*.hpp` if Catch2 macros ever land there) and matches `TEST_CASE(\s*"<name>"` using Python `re` (no flag needed — `\s` matches `\n` natively) so the literal string after the open paren is found whether on the same line or the next.
- Preserve existing single-line behaviour and the `SCENARIO` alias.
- Keep the check O(repo) — no project build invoked.

Out of scope:
- Changing the DSL grammar (no new assertion kinds).
- Re-running verification on already-failed historical issues (`/verify-dod` per-issue stays manual).
- clang-format config changes (ColumnLimit stays at 100; the verifier should be tolerant of legal formatter output).
- Touching #597's DoD (already amended in advance to use `file_contains` workarounds; once this lands, that DoD MAY be reverted to the original `unit_test_named` form in a follow-up cleanup, but the workaround is correct and stable).

### Unit Test Plan

Verifier is Python, not C++; tests are `pytest` cases under `tests/infra/verify_dod/test_unit_test_named.py`:

- `test_unit_test_named_matches_singleline_TEST_CASE`
- `test_unit_test_named_matches_multiline_TEST_CASE`
- `test_unit_test_named_matches_SCENARIO_alias`
- `test_unit_test_named_misses_when_name_absent`

Each case writes a tiny fixture `bar_test.cpp` to a `tmp_path`, monkey-patches the verifier's `tests/` root to that path, calls `check_unit_test_named(name)`, and asserts the returned `(ok, label)` tuple. CI wiring: confirm whether `ci.yml` already runs `pytest`; if not, add a `python-tests` job (`pip install pyyaml pytest && pytest tests/infra`) to the matrix.

### Definition of Done

```yaml
- pr_merged_closes_self: true
- file_exists: .github/scripts/verify_dod.py
- file_contains:
    path: .github/scripts/verify_dod.py
    regex: 'def check_unit_test_named'
- file_contains:
    path: .github/scripts/verify_dod.py
    regex: 'is_symlink'
- file_contains:
    path: tests/infra/verify_dod/conftest.py
    regex: 'pytest_unconfigure'
- glob_nonempty: tests/infra/verify_dod/*.py
- file_contains:
    path: tests/infra/verify_dod/test_unit_test_named.py
    regex: 'def test_unit_test_named_matches_multiline_TEST_CASE'
- file_contains:
    path: tests/infra/verify_dod/test_unit_test_named.py
    regex: 'def test_unit_test_named_matches_singleline_TEST_CASE'
- workflow_passed: ci.yml
```

### Depends on

(none — pure infra fix, may land any time)

### Agent Execution Notes

Branch `fix/infra-verify-dod-multiline-test-case`. PR title: `fix(infra): verify_dod.py matches multi-line TEST_CASE invocations`. The Python `pytest` runner needs to be wired into the CI matrix if it isn't already — check `.github/workflows/ci.yml` first; if pytest is not yet a gate, add a `python-tests` job that installs pyyaml + pytest and runs `pytest tests/infra`. The `unit_test_named` DoD entry on this very issue uses a name long enough to trigger the wrap problem on purpose, so the fix self-validates: until the fix lands, this DoD will fail with the same multi-line miss; after the fix lands and CI is green, `/verify-dod` should pass it.

Note that `unit_test_named`'s file scope is currently `tests/` — if the new pytest cases live under `tests/infra/verify_dod/` the existing scope already covers them. Confirm Catch2-flavoured `TEST_CASE` regex is not accidentally applied to Python files (the new implementation should restrict by extension to `.cpp`/`.hpp`/`.cc`).

cjhowe-us:
Closes #1035

## Summary

- **Root cause**: `check_unit_test_named` used `grep -RE` with a pattern that does not span newlines under POSIX-extended regex. clang-format (ColumnLimit: 100) wraps long `TEST_CASE` invocations so the name string appears on the line _after_ the opening parenthesis, causing spurious `dod:failed` results even when the test exists and passes in CI.
- **Fix**: Replaced `grep -RE` with a Python `re` scan (using `re.DOTALL`) that reads each `.cpp`/`.hpp`/`.cc` file under `tests/` and matches `(?:TEST_CASE|SCENARIO)\s*\(\s*"<name>"` across newlines. Python files (including the new pytest tests themselves) are excluded by extension.
- **Tests**: Four `pytest` cases added under `tests/infra/verify_dod/test_unit_test_named.py` covering single-line, multi-line, `SCENARIO` alias, and absent-name scenarios. A `conftest.py` sets sentinel env vars so the verifier module imports cleanly during collection.
- **CI**: New `python-tests` job added to `ci.yml` (`macos-latest`, Python 3.12, `pip install pyyaml pytest`, `pytest tests/infra -v`).

## Note on branch protection

If your branch-protection rules list required checks explicitly, add `python-tests` to the required checks list in GitHub repository settings to make this job a merge gate.

## Test plan

- [x] `python3 -m pytest tests/infra -v` — 4 passed locally
- [x] `re.DOTALL` flag present in patched `check_unit_test_named`
- [x] Python extension filter (`.cpp`/`.hpp`/`.cc`) prevents pytest files from being scanned as C++ sources
- [x] CI `python-tests` job defined in `ci.yml`

## Definition of Done assertions (self-check)

| Assertion | Status |
|---|---|
| `pr_merged_closes_self: true` | This PR body contains `Closes #1035` |
| `file_exists: .github/scripts/verify_dod.py` | File exists, patched |
| `file_contains: def check_unit_test_named` | Line 128 |
| `file_contains: re\.DOTALL` | Line 151 (in `check_unit_test_named`) |
| `glob_nonempty: tests/infra/verify_dod/*.py` | `conftest.py` + `test_unit_test_named.py` |
| `file_contains: def test_unit_test_named_matches_multiline_TEST_CASE` | Present |
| `file_contains: def test_unit_test_named_matches_singleline_TEST_CASE` | Present |
| `workflow_passed: ci.yml` | Will pass once CI runs on main after merge |

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.github/scripts/verify_dod.py
+++ b/.github/scripts/verify_dod.py
@@ -141,7 +141,7 @@ def check_unit_test_named(arg: object) -> tuple[bool, str]:
 
       grep -E does not span newlines under POSIX-extended regex, causing
       spurious dod:failed results for tests that exist and run green.
-    - Restricts the file scan to .cpp / .hpp / .cc so pytest files that
+    - Restricts the file scan to .cpp / .hpp so pytest files that
       contain 'TEST_CASE' in docstrings or fixture content are not scanned.
     """
     name = str(arg)
@@ -151,7 +151,7 @@ def check_unit_test_named(arg: object) -> tuple[bool, str]:
         flags=re.DOTALL,
     )
     tests_root = Path("tests")
-    extensions = {".cpp", ".hpp", ".cc"}
+    extensions = {".cpp", ".hpp"}
     for src_file in tests_root.rglob("*"):
         if src_file.suffix not in extensions:
             continue
--- a/.github/scripts/verify_dod.py
+++ b/.github/scripts/verify_dod.py
@@ -126,17 +126,42 @@ def check_workflow_passed(arg: object) -> tuple[bool, str]:
 
 
 def check_unit_test_named(arg: object) -> tuple[bool, str]:
+    """Return (True, label) when a Catch2 TEST_CASE or SCENARIO with the exact
+    name *arg* exists under tests/.
+
+    Implementation notes:
+    - Uses Python re with re.DOTALL so the pattern matches across newlines.
+      clang-format (ColumnLimit: 100) wraps long TEST_CASE invocations onto
+      two lines, e.g.::
+
+          TEST_CASE(
+              "core/type_registry: lookup_unregistered_yields_TypeUnregistered",
+              "[core][type_registry]"
+          ) {
+
+      grep -E does not span newlines under POSIX-extended regex, causing
+      spurious dod:failed results for tests that exist and run green.
+    - Restricts the file scan to .cpp / .hpp / .cc so pytest files that
+      contain 'TEST_CASE' in docstrings or fixture content are not scanned.
+    """
     name = str(arg)
     label = f"unit_test_named: {name}"
-    res = subprocess.run(
-        [
-            "grep", "-RE",
-            rf'(TEST_CASE|SCENARIO)\s*\(\s*"{re.escape(name)}"',
-            "tests/",
-        ],
-        check=False, capture_output=True, text=True,
+    pattern = re.compile(
+        r'(?:TEST_CASE|SCENARIO)\s*\(\s*"' + re.escape(name) + r'"',
+        flags=re.DOTALL,
     )
-    return res.returncode == 0, label
+    tests_root = Path("tests")
+    extensions = {".cpp", ".hpp", ".cc"}
+    for src_file in tests_root.rglob("*"):
+        if src_file.suffix not in extensions:
+            continue
+        try:
+            text = src_file.read_text(encoding="utf-8", errors="replace")
+        except OSError:
+            continue
+        if pattern.search(text):
+            return True, label
+    return False, label
 
 
 def check_issue_comment_matches(arg: object) -> tuple[bool, str]:
--- a/.github/scripts/verify_dod.py
+++ b/.github/scripts/verify_dod.py
@@ -130,9 +130,11 @@ def check_unit_test_named(arg: object) -> tuple[bool, str]:
     name *arg* exists under tests/.
 
     Implementation notes:
-    - Uses Python re with re.DOTALL so the pattern matches across newlines.
-      clang-format (ColumnLimit: 100) wraps long TEST_CASE invocations onto
-      two lines, e.g.::
+    - Uses Python ``re`` with ``\\s*`` to match across newlines.  Python's
+      ``\\s`` matches ``\\n`` natively (unlike POSIX-extended grep, which does
+      not span lines).  No ``re.DOTALL`` flag is needed: the pattern contains
+      no ``.`` metacharacter.  clang-format (ColumnLimit: 100) wraps long
+      TEST_CASE invocations onto two lines, e.g.::
 
           TEST_CASE(
               "core/type_registry: lookup_unregistered_yields_TypeUnregistered",
@@ -143,16 +145,20 @@ def check_unit_test_named(arg: object) -> tuple[bool, str]:
       spurious dod:failed results for tests that exist and run green.
     - Restricts the file scan to .cpp / .hpp so pytest files that
       contain 'TEST_CASE' in docstrings or fixture content are not scanned.
+    - Skips symlinks inside the rglob walk — Python 3.12 follows directory
+      symlinks by default; skipping them prevents infinite loops on circular
+      or deeply nested test-tree symlinks.
     """
     name = str(arg)
     label = f"unit_test_named: {name}"
     pattern = re.compile(
         r'(?:TEST_CASE|SCENARIO)\s*\(\s*"' + re.escape(name) + r'"',
-        flags=re.DOTALL,
     )
     tests_root = Path("tests")
     extensions = {".cpp", ".hpp"}
     for src_file in tests_root.rglob("*"):
+        if src_file.is_symlink():
+            continue
         if src_file.suffix not in extensions:
             continue
         try:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,26 @@ jobs:
             echo "::endgroup::"
           done <<< "${{ steps.list.outputs.traces }}"
 
+  python-tests:
+    # Runs the pytest suite under tests/infra/ — currently covers the
+    # verify_dod.py multi-line TEST_CASE fix (plan #1035).
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install Python dependencies
+        run: pip install pyyaml pytest
+      - name: Run infra pytest suite
+        env:
+          # Provide sentinel values so verify_dod module-level env reads
+          # do not abort import during collection.
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          ISSUE_NUMBER: "0"
+          ISSUE_BODY: ""
+        run: pytest tests/infra -v
+
   dep-check:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,7 +152,9 @@ jobs:
   python-tests:
     # Runs the pytest suite under tests/infra/ — currently covers the
     # verify_dod.py multi-line TEST_CASE fix (plan #1035).
-    runs-on: macos-latest
+    # ubuntu-latest matches the dod-verify.yml production runner; pure Python
+    # tests have no macOS-specific behavior so there is no reason to use macOS.
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/tests/infra/verify_dod/conftest.py
+++ b/tests/infra/verify_dod/conftest.py
@@ -1,0 +1,36 @@
+"""conftest.py — pytest configuration for tests/infra/verify_dod/.
+
+Adds .github/scripts/ to sys.path so that `import verify_dod` resolves
+from any working directory, without requiring the package to be installed.
+
+verify_dod.py reads several environment variables at module-level
+(GITHUB_REPOSITORY, ISSUE_NUMBER, ISSUE_BODY).  We set sentinel values
+before the first import so collection does not abort with KeyError.  The
+individual test functions that exercise GitHub-API checks (pr_merged,
+workflow_passed, issue_comment) must mock these further; the tests in
+test_unit_test_named.py only exercise check_unit_test_named, which does
+not reference those module-level constants.
+
+The path is computed relative to this conftest.py file so it works
+whether pytest is invoked from the repo root or from within tests/infra/.
+"""
+
+import os
+import sys
+from pathlib import Path
+
+# Resolve .github/scripts/ relative to the repo root.
+# This file lives at tests/infra/verify_dod/conftest.py, so:
+#   repo_root = this_file.parent.parent.parent.parent
+_repo_root = Path(__file__).resolve().parent.parent.parent.parent
+_scripts_dir = _repo_root / ".github" / "scripts"
+
+if str(_scripts_dir) not in sys.path:
+    sys.path.insert(0, str(_scripts_dir))
+
+# verify_dod reads GITHUB_REPOSITORY, ISSUE_NUMBER, and ISSUE_BODY at
+# module-level.  Provide sentinel values so the import succeeds; tests that
+# only call check_unit_test_named never access these constants.
+os.environ.setdefault("GITHUB_REPOSITORY", "test-owner/test-repo")
+os.environ.setdefault("ISSUE_NUMBER", "0")
+os.environ.setdefault("ISSUE_BODY", "")
--- a/tests/infra/verify_dod/conftest.py
+++ b/tests/infra/verify_dod/conftest.py
@@ -3,21 +3,26 @@
 Adds .github/scripts/ to sys.path so that `import verify_dod` resolves
 from any working directory, without requiring the package to be installed.
 
-verify_dod.py reads several environment variables at module-level
-(GITHUB_REPOSITORY, ISSUE_NUMBER, ISSUE_BODY).  We set sentinel values
-before the first import so collection does not abort with KeyError.  The
-individual test functions that exercise GitHub-API checks (pr_merged,
-workflow_passed, issue_comment) must mock these further; the tests in
-test_unit_test_named.py only exercise check_unit_test_named, which does
-not reference those module-level constants.
+verify_dod.py reads GITHUB_REPOSITORY, ISSUE_NUMBER, and ISSUE_BODY at
+module-level during the first import.  We must set sentinel values before
+pytest collection (which triggers test-file imports), so we use the
+``pytest_configure`` hook — the earliest conftest hook that executes before
+collection.  This avoids polluting process env beyond the test session.
+
+After import, a function-scoped autouse fixture re-applies the sentinels
+via ``monkeypatch.setenv`` for each test so any test that overrides them
+gets a clean environment at teardown.  Future tests that exercise the
+GitHub-API checks (pr_merged, workflow_passed, issue_comment) should
+override these per-test via additional ``monkeypatch.setenv`` calls.
 
 The path is computed relative to this conftest.py file so it works
 whether pytest is invoked from the repo root or from within tests/infra/.
 """
 
-import os
 import sys
 from pathlib import Path
+
+import pytest
 
 # Resolve .github/scripts/ relative to the repo root.
 # This file lives at tests/infra/verify_dod/conftest.py, so:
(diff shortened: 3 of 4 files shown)


.github/scripts/verify_dod.py
```diff
@@ -126,17 +126,42 @@ def check_workflow_passed(arg: object) -> tuple[bool, str]:
...
+    for src_file in tests_root.rglob("*"):
+        if src_file.suffix not in extensions:
+            continue
+        try:
+            text = src_file.read_text(encoding="utf-8", errors="replace")
+        except OSError:
+            continue
+        if pattern.search(text):
+            return True, label
+    return False, label
```
cjhowe-us, line 172: severity:LOW location:.github/scripts/verify_dod.py:166 problem:Extension filter includes `.cc` but `git ls-files 'tests/**/*.cc'` returns nothing — the project's tests/ tree uses only `.cpp` and `.hpp`. Including `.cc` is harmless but is dead code that signals false coverage. fix:Either drop `.cc` from the `extensions` set OR add `.cxx` to be explicit about the intent that the filter covers "all C++ implementation extensions someone might add." Document the choice in the existing docstring.

tests/infra/verify_dod/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+_repo_root = Path(__file__).resolve().parent.parent.parent.parent
+_scripts_dir = _repo_root / ".github" / "scripts"
+
+if str(_scripts_dir) not in sys.path:
+    sys.path.insert(0, str(_scripts_dir))
+
+# verify_dod reads GITHUB_REPOSITORY, ISSUE_NUMBER, and ISSUE_BODY at
+# module-level.  Provide sentinel values so the import succeeds; tests that
+# only call check_unit_test_named never access these constants.
+os.environ.setdefault("GITHUB_REPOSITORY", "test-owner/test-repo")
```
cjhowe-us: severity:LOW location:tests/infra/verify_dod/conftest.py:34 problem:`os.environ.setdefault(...)` mutates the process environment for the entire pytest session and any sibling test directories under tests/infra/ that pytest collects later. Sibling tests that legitimately exercise verify_dod's GitHub-API checks (`check_issue_comment_matches`, `check_pr_merged_closes_self`) will inherit the sentinel `GITHUB_REPOSITORY=test-owner/test-repo` and silently hit a fake repo via the gh CLI. Today no such tests exist; this is a future foot-gun. fix:Move the env mutation into a function-scoped autouse fixture using `monkeypatch.setenv` so each test gets a clean sentinel that's torn down between tests. Or scope conftest.py to a tighter directory if the env vars are truly only needed for unit_test_named tests (which they aren't — `import verify_dod` triggers the module-level reads regardless).

.github/workflows/ci.yml
```diff
@@ -149,6 +149,26 @@ jobs:
             echo "::endgroup::"
           done <<< "${{ steps.list.outputs.traces }}"
 
+  python-tests:
```
cjhowe-us, line 152: severity:MED location:.github/workflows/ci.yml:152 problem:`python-tests` job runs on `macos-latest` while every other ci.yml job pins explicit versions (`macos-26`, `macos-14`, `ubuntu-latest`). `macos-latest` is the GitHub-hosted floating tag, which Anthropic-tier CI runners typically resolve to `macos-14` today but will silently bump to `macos-15`/`macos-26` later. For an infra correctness gate that has zero macOS-specific behavior (it's pure Python regex/file IO), pinning to a floating tag adds variance without benefit, AND it diverges from the rest of ci.yml's discipline. fix:Either pin to `macos-14` (matches `spec-check`/`e2e-*` jobs) OR — preferred — switch to `ubuntu-latest` since the test is platform-agnostic and Linux runners are faster + cheaper. The dod-verify workflow itself runs on `ubuntu-latest`, which is the production environment for the verifier; running the unit tests on the same OS as production gives the strongest signal.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0
notes:Round-1 (coverage + correctness) lens passes cleanly. All 8 of #1035's DoD assertions are satisfied by the diff content; PR title is Conventional Commits compliant; PR body has `Closes #1035` (dod-verify will fire on merge). I locally re-ran the patched verifier against synthetic single-line, multi-line wrapped, SCENARIO, and absent-name fixtures — all four behaviors match the test claims. The conftest.py path-resolution is robust (file-relative, not cwd-relative). The verifier's `Path("tests").rglob("*")` correctly resolves under dod-verify.yml's default `$GITHUB_WORKSPACE` cwd. Verdict is COMMENT (not APPROVE) only because `build` is still PENDING at review time — per the project's ci_not_green rule, no APPROVE until latest-commit CI is fully green. python-tests, lint, dep-check, spec-check, automated-review all green. Findings: 1 MED on `runs-on: macos-latest` divergence from ci.yml's pin discipline (prefer `ubuntu-latest` to match dod-verify production); 2 LOW on dead `.cc` extension entry and process-wide env-var leak from conftest.py. Downstream context: this fix unblocks reverting #597's amended DoD back to canonical `unit_test_named` form and unblocks #245's pending `dod:failed` re-verification — the R3 reviewer should confirm those follow-ups are tracked (likely as separate cleanup issues, not in scope here).

.github/workflows/ci.yml
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6abc47d reasoning:Switched runs-on from macos-latest to ubuntu-latest to match the dod-verify.yml production runner. Annotated with inline comment explaining the rationale.

.github/scripts/verify_dod.py
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6abc47d reasoning:Dropped .cc from the extensions set. Project tests/ tree uses only .cpp and .hpp; .cc is dead code. Updated the docstring to reflect the narrowed filter.

tests/infra/verify_dod/conftest.py
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:6abc47d reasoning:Replaced bare os.environ.setdefault calls with a pytest_configure hook (runs pre-collection, before verify_dod import) + a function-scoped autouse fixture using monkeypatch.setenv (provides teardown per test). The pytest_configure hook is needed because the test file imports verify_dod at module level during collection, before any fixture can run. The autouse fixture ensures each test gets clean env state and teardown undoes mutations, eliminating the future foot-gun for sibling tests.

.github/scripts/verify_dod.py
```diff
@@ -126,17 +126,42 @@ def check_workflow_passed(arg: object) -> tuple[bool, str]:
...
-    res = subprocess.run(
-        [
-            "grep", "-RE",
-            rf'(TEST_CASE|SCENARIO)\s*\(\s*"{re.escape(name)}"',
-            "tests/",
-        ],
-        check=False, capture_output=True, text=True,
+    pattern = re.compile(
+        r'(?:TEST_CASE|SCENARIO)\s*\(\s*"' + re.escape(name) + r'"',
+        flags=re.DOTALL,
```
cjhowe-us: severity:MED location:.github/scripts/verify_dod.py:151 problem:Docstring claims `re.DOTALL` is what enables multi-line matching — but the pattern contains no `.` metacharacters, so DOTALL is a no-op here. The actual fix is that Python's `\s` matches `\n` by default (Python regex flavor), whereas `grep -RE` uses POSIX-extended where `\s` is a GNU extension that does NOT cross newlines. Future maintainers reading the docstring will incorrectly infer DOTALL is load-bearing and may keep it when refactoring, or be confused if the pattern grows a `.` and starts matching across files unexpectedly. fix:Either (a) drop the `flags=re.DOTALL` argument and update the docstring to explain that Python `\s` matches `\n` (the real reason the fix works), or (b) keep DOTALL but rewrite the docstring bullet to: "Python's `\s` character class matches `\n` by default; this is the property that lets the pattern span the wrap clang-format introduces. `re.DOTALL` is set defensively in case the pattern later grows `.` metacharacters." Option (a) is simpler and removes a misleading code/comment pair.

tests/infra/verify_dod/conftest.py
```diff
@@ -0,0 +1,65 @@
...
+
+def pytest_configure(config: pytest.Config) -> None:  # noqa: ARG001
+    """Set sentinel env vars before collection so ``import verify_dod`` succeeds.
+
+    verify_dod reads GITHUB_REPOSITORY, ISSUE_NUMBER, and ISSUE_BODY at
+    module-level; this hook runs before test-file collection, ensuring those
+    vars are present when pytest imports the test modules.
+    """
+    import os
+
```
cjhowe-us, line 52: severity:LOW location:tests/infra/verify_dod/conftest.py:51 problem:`pytest_configure` uses `os.environ.setdefault(...)`, which is the same process-wide leak shape that R1 LOW#2 flagged. If a developer runs `pytest` locally with `GITHUB_REPOSITORY` already exported (common when shell has been sourced for `gh` work), `setdefault` is a no-op, so the user's real value leaks through into the import-time read of `verify_dod` — and then the autouse fixture overrides it back to the sentinel for each test, but only for the test scope. After the pytest session ends, the sentinel injected by `pytest_configure` (when env was empty) persists for the remainder of the shell process; in CI this is harmless, but for `pytest --pdb` or REPL-driven debugging it can poison subsequent `gh` invocations in the same process. The dual-mechanism (configure-time + per-test fixture) is load-bearing for collection-time import as documented, but the configure-time leg still leaks. fix:In `pytest_configure`, capture the original values and restore them in a matching `pytest_unconfigure(config)` hook — symmetric session-scoped setup/teardown. Or, alternative seam: make the test file import `verify_dod` lazily (inside the autouse fixture body or inside each test), eliminating the need for `pytest_configure` entirely and letting `monkeypatch.setenv` be the single mechanism. Lazy import is the cleaner SRP: one mechanism, one teardown story.

.github/scripts/verify_dod.py
```diff
@@ -126,17 +126,42 @@ def check_workflow_passed(arg: object) -> tuple[bool, str]:
...
     )
-    return res.returncode == 0, label
+    tests_root = Path("tests")
+    extensions = {".cpp", ".hpp"}
+    for src_file in tests_root.rglob("*"):
+        if src_file.suffix not in extensions:
+            continue
+        try:
```
cjhowe-us, line 164: severity:LOW location:.github/scripts/verify_dod.py:158 problem:`Path("tests").rglob("*")` follows symbolic links during traversal in Python 3.12 (the `os.walk(followlinks=...)` default behavior under the hood; rglob does follow dir symlinks unless you pass `recurse_symlinks=False` in 3.13+). If `tests/` ever contains a symlink to an out-of-tree directory (e.g. a developer symlinks vendored test data, or a future cook-pipeline drops a symlink to a generated artifacts cache), the verifier silently scans outside the repo, with non-deterministic results across CI vs local. Today no such symlinks exist; this is a defensive note. fix:Either (a) explicitly walk with `os.walk(tests_root, followlinks=False)` and filter by suffix, or (b) when bumping to Python 3.13+, pass `Path("tests").rglob("*", recurse_symlinks=False)`. Add a brief docstring note that only regular files inside `tests/` are scanned.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0
notes:Round-2 (cohesion + SRP + seam quality) lens. SRP per touched module is clean — `check_unit_test_named` is a single-concern regex+walk; no bleed into other check_* functions; no premature shared-helper extraction needed. The `conftest.py` dual-mechanism (pytest_configure + autouse monkeypatch fixture) IS load-bearing as the responder documented (configure-time satisfies collection-time `import verify_dod`; per-test monkeypatch isolates future API-check tests) — not a duplication. All three R1 inline threads have `decision:ADDRESSED` replies (commit 6abc47d) and the diff confirms the fixes (ubuntu runner, .cc dropped, monkeypatch shape). Carryover: 0. New R2 findings: 1 MED on a misleading docstring/code pair — `re.DOTALL` is set but does nothing because the pattern has no `.` metacharacters; the actual bug fix is that Python `\s` matches `\n` natively (vs POSIX-extended grep where it does not). 2 LOW: (a) `pytest_configure` still uses `os.environ.setdefault` which has the same after-session leak shape R1 LOW#2 flagged — needs a matching `pytest_unconfigure` for symmetric teardown, or alternatively lazy-import `verify_dod` inside the fixture body to collapse to a single mechanism; (b) `Path("tests").rglob("*")` follows symlinks during dir traversal in Python 3.12, defensive note for future test trees that gain symlinks. Verdict is COMMENT (not APPROVE) for two reasons: the per-PR rule that no APPROVE without green CI on tip — `build` job is still PENDING at review time (python-tests + lint + dep-check + spec-check + automated-review all pass) — and to leave R3 room to either accept the docstring/teardown findings or push back. No HIGH so no REQUEST_CHANGES. The PR is substantively merge-ready pending build green.

cjhowe-us:
responder:go-impl-respond round:2 decision:ADDRESSED commit:454c6349ac850110d8c6f2ec39fc566e9af2b333

**MED — verify_dod.py:151 — re.DOTALL no-op**
decision:ADDRESSED commit:454c634
Dropped `re.DOTALL` from `check_unit_test_named`. Pattern uses `\s*` (not `.`) so DOTALL was genuinely a no-op. Docstring updated to state the correct mechanism: Python `\s` matches `\n` natively, unlike POSIX-extended grep. Also updated #1035 DoD via `gh issue edit`: replaced `regex: 're\.DOTALL'` (which would now match only `extract_block`'s incidental use, misleadingly passing) with two precise assertions — `regex: 'is_symlink'` (pins the Python scan + LOW-2 fix) and `regex: 'pytest_unconfigure'` (pins LOW-1 fix).

**LOW-1 — conftest.py:51 — os.environ.setdefault leaks past session end**
decision:ADDRESSED commit:454c634
Added `pytest_unconfigure` hook that pops the same keys set by `pytest_configure`. Module docstring updated to document the symmetric pair. Option (a) — explicit teardown — chosen as instructed.

**LOW-2 — verify_dod.py:158 — rglob follows symlinks in Python 3.12**
decision:ADDRESSED commit:454c634
Added `if src_file.is_symlink(): continue` at the top of the rglob loop body, before the extension check. Defensive filter is cheap (one syscall) and prevents infinite traversal on circular directory symlinks. Also noted in the docstring.

All 4 pytest cases still pass locally (python3 -m pytest tests/infra -v — 4 passed in 0.02s).

.github/scripts/verify_dod.py
```diff
@@ -126,17 +126,48 @@ def check_workflow_passed(arg: object) -> tuple[bool, str]:
...
+          ) {
+
+      grep -E does not span newlines under POSIX-extended regex, causing
+      spurious dod:failed results for tests that exist and run green.
+    - Restricts the file scan to .cpp / .hpp so pytest files that
+      contain 'TEST_CASE' in docstrings or fixture content are not scanned.
+    - Skips symlinks inside the rglob walk — Python 3.12 follows directory
+      symlinks by default; skipping them prevents infinite loops on circular
+      or deeply nested test-tree symlinks.
+    """
```
cjhowe-us, line 151: severity:LOW location:PR body table problem:The "Definition of Done assertions (self-check)" table in the PR body is stale post-R2: it still lists `file_contains: re\.DOTALL` (line 151) as a DoD assertion, but the actual DoD block on #1035 was swapped to assert `is_symlink` (line 162) + `pytest_unconfigure` (conftest.py line 58). The DoD verifier reads the issue body, not the PR body, so this is documentation drift only — but it will mislead future readers tracing what dod-verify actually checks. fix:Update the PR body table to mirror the post-R2 #1035 DoD (replace the `re\.DOTALL` row with `is_symlink` and add a `pytest_unconfigure` row). No code change needed; PR body edit only.

.github/scripts/verify_dod.py
```diff
@@ -126,17 +126,48 @@ def check_workflow_passed(arg: object) -> tuple[bool, str]:
 
 
 def check_unit_test_named(arg: object) -> tuple[bool, str]:
```
cjhowe-us, line 128: severity:LOW location:PR body downstream-impact problem:PR body cites only #1035 in `Closes`, but #1035 issue body explicitly notes that landing this fix unblocks two downstream cleanups: (1) reverting #597 DoD from the `file_contains` workaround back to canonical `unit_test_named`, and (2) re-verifying #245 which previously hit `dod:failed` from this same root cause. The DAG story would be more discoverable if the PR body called those out under a "Downstream impact" or "Follow-ups" section so the orchestrator can spawn the cleanup spikes immediately after merge instead of after a context re-derive. fix:Add a brief "Follow-ups (out-of-scope, tracked separately)" section to the PR body listing #597 (DoD revert candidate) and #245 (re-verify candidate). No code change; documentation only. If those follow-ups are not yet filed as issues, file them post-merge.

cjhowe-us (commented):
round:3 reviewer:go-review verdict:APPROVE findings:HIGH:0 MED:0 LOW:2 carryover:0 notes:Round-3 (topology + spec alignment + polish) lens passes cleanly. CI on tip 454c6349 is fully green (build, automated-review, dep-check, lint, python-tests, spec-check all pass; e2e-core/e2e-shader skipping per `if: false`). All R2 commitments materialized at tip: `re.DOTALL` removed and docstring corrected to state `\s` matches `\n` natively (R2 MED), `is_symlink()` guard inside the rglob loop (R2 LOW-2), `pytest_unconfigure` symmetric teardown hook in conftest.py (R2 LOW-1). All 6 prior-round inline threads (3 R1 + 3 R2) carry `decision:ADDRESSED` replies — carryover is 0. All 8 of #1035's post-R2 DoD assertions are satisfied by the diff (pr_merged_closes_self via `Closes #1035`; file_exists + 4 file_contains; glob_nonempty; workflow_passed will fire on merge). Conventional-commits subject correct, sub-issue parented to #1, labels intact (type:plan, phase:mvp, pts:2, domain:infra). Two LOW polish findings posted as inline comments — both PR-body documentation drift (stale self-check table referencing the removed `re\.DOTALL` row; downstream cleanups #597 + #245 not cited in PR body) — neither blocks merge. Verdict APPROVE; orchestrator may enable auto-merge.

Issue #1032, opened 2026-05-10
[INIT] Remove EASTL — switch to libc++ stdlib + std::ranges


cjhowe-us:
### Phase

phase:mvp

### Scope

Engine-wide removal of EASTL. Replace every EASTL container / string / smart-pointer / optional / variant / tuple / function with the equivalent from libc++'s standard library (latest available — C++23 minimum, C++26 features where libc++ supports them on the macOS-26 / Apple-Silicon clang ≥ 21 toolchain). Where applicable, prefer `std::ranges` over hand-rolled iteration. Update all build files, vendor manifests, CLAUDE.md, PHILOSOPHY.md §11, and any spec passage that pins EASTL.

### Outcome

- Zero `#include <EASTL/…>` lines remaining in `core/`, `plugins/`, `tools/`, `runtime/`, `shaders/`, `tests/`.
- Zero references to `eastl::` types in compiled code.
- vcpkg manifest no longer pulls EASTL/EAStdC.
- CMake no longer links EASTL targets.
- `PHILOSOPHY.md` §11 rewritten to declare libc++ stdlib as the canonical container/string/smart-pointer/optional/variant/tuple/function source. Add `std::ranges` posture (prefer ranges over iterator-pair APIs at boundaries).
- `CLAUDE.md` Tech Stack row updated to drop EASTL and explicitly cite the libc++ pieces it replaces.
- Every spec/decision-record passage that mentioned EASTL re-aligned to the new policy.
- All tests pass under `cmake --preset macos-debug && ctest --preset macos-debug`.
- All workflows green on `main` after merge.

This is an architecture-policy reversal of the prior PHILOSOPHY.md §11 EASTL adoption. The justification (ergonomics + Catch2 / CTest interop + libc++'s C++23/C++26 stdlib catching up on game-relevant containers including `std::flat_map` / `std::flat_set` / `std::expected` / `std::pmr`) is captured in this initiative's child design spike (#1033, closed) and codified in `reviews/decisions/eastl-removal.md` (merged via PR #1034 on 2026-05-10).

### Child Epics

GitHub native sub-issues. This initiative carries no estimate. Stage progression:

- [x] #1033 — `[SPIKE] decide-eastl-removal-strategy` (closed; deliverable `reviews/decisions/eastl-removal.md` merged via PR #1034)
[...]

Issue #1031, opened 2026-05-10
[SPIKE] iterate-phase-registry-allocator-tagging


cjhowe-us:
### Domain
core

### Phase
phase:mvp

### Question

How should `PhaseRegistry::systems_` (a `std::array<eastl::vector<SystemEntry>, kPhaseCount>`) route its heap allocations through `PerContextAllocator{ContextTag::core}` instead of the global `eastl::operator new[]` substrate?

This cross-cuts PR #1028 HIGH-1 (TypeRegistry allocator wiring), which has the same shape: an EASTL container whose default construction bypasses context-tagged accounting.

The specific question: what is the minimal, correct EASTL allocator adaptor type that wraps `glibre::PerContextAllocator` and satisfies the EASTL allocator concept (`allocate(size_t, int)` / `allocate(size_t, size_t, size_t, int)` / `deallocate(void*, size_t)`) so that `eastl::vector<SystemEntry, EastlAllocAdaptor>` routes through `ContextTag::core`? And how should the `std::array<eastl::vector<...>, kPhaseCount>` be initialized in the `PhaseRegistry` constructor to pass the adaptor instance into each vector element?

### Context

- `PhaseRegistry` was added in PR #1029 (plan #245).
- The default-construction of `systems_{}` was flagged severity:MED in the round-2 review (comment 3214747006).
- `glibre::PerContextAllocator` is explicitly documented as "NOT a C++ Allocator concept" in `core/include/glibre/alloc.hpp` — it has a different interface than the EASTL allocator concept.
- PR #1028 (TypeRegistry) has a parallel HIGH-1 finding on the same topic. The spike should coordinate: the adaptor type and wiring pattern decided here should apply identically to both `PhaseRegistry` and `TypeRegistry` so the project has one canonical per-context EASTL adaptor type (not two ad-hoc ones).
- The spike must land before the `PhaseRegistry` allocator-tagging fix is implemented (that fix should reference this spike's decision record).

### Deliverable

A decision record at `reviews/decisions/eastl-per-context-allocator-adaptor.md` capturing:
1. The EASTL allocator adaptor type design (header location, class name, interface).
2. How `PhaseRegistry` and `TypeRegistry` wire the adaptor at construction.
3. Any EASTL version constraints (vcpkg EASTL 3.27.x).
4. A cross-reference to the follow-up plan issues that implement the wiring in each registry.

### Parent

#3

### Depends on

[...]

PR #1067, opened 2026-05-10
feat(core/phase-registry): migrate PhaseSystemFn to std::move_only_function

### LINKED ISSUE #1045: [PLAN] migrate(core/phase-registry): swap eastl::fixed_function → std::move_only_function
### Domain

core

### Phase

phase:mvp

pts:5

### User Stories Satisfied

Initiative #1032 outcome bullets satisfied by this leaf:

- "Zero `#include <EASTL/…>` lines remaining in `core/include/glibre/core/phase_registry.hpp`, `core/include/glibre/core/frame_phase.hpp`, `core/src/phase_registry.cpp`, `core/src/frame_loop.cpp`."
- "`eastl::fixed_function<N, void()>` replaced with `std::move_only_function<void()>` per matrix row 9 + Open Question 1 (heap default; inline-storage polyfill deferred until measured perf-budget breach)."

### Parent (Epic or Sub-Epic)

#1032

### Scope

This plan owns the per-frame phase-orchestration cluster. Its load-bearing migration is the `PhaseSystemFn` typedef flip from `eastl::fixed_function<kSystemFnStorageBytes, void()>` to `std::move_only_function<void()>` (per matrix row 9 + Open Question 1: "the migration PLAN that touches each `eastl::fixed_function` site picks `std::move_only_function<Sig>` by default and only opens an inline-storage polyfill if the per-context SPEC §9 micro-benchmark fires").

**SRP rationale.** PhaseRegistry + FramePhase + FrameLoop together own one responsibility: "execute the per-frame phase pipeline." The `fixed_function` typedef is the binding primitive; isolating its migration here keeps the perf-budget consideration scoped to one PR (the micro-benchmark check fires here, not in unrelated plans).

**Files touched:**

- `core/include/glibre/core/phase_registry.hpp` (24 hits) — replace `<EASTL/...>` headers; migrate per matrix:
  - `eastl::fixed_function<kSystemFnStorageBytes, void()>` → `std::move_only_function<void()>` (row 9). Update the `// PHILOSOPHY §11:` comment block to point at `reviews/decisions/eastl-removal.md` Open Question 1.
  - `eastl::vector<T>` → `std::pmr::vector<T>` (row 3)
  - `eastl::string` / `eastl::string_view` → `std::pmr::string` / `std::string_view` (rows 1, 2)
  - Per Open Question 1: if `tests/core/phase_registry/` micro-benchmark shows heap-allocation regression, author `core/include/glibre/compat/inline_function.hpp` polyfill; otherwise defer.
- `core/include/glibre/core/frame_phase.hpp` (6 hits) — sweep header.
- `core/src/phase_registry.cpp` — sweep .cpp call sites.
- `core/src/frame_loop.cpp` — sweep .cpp call sites.

**Out of scope:**

- `core/include/glibre/core/plugin_*` — owned by sibling plans.
- `tests/core/phase_registry/` and `tests/core/frame_loop_test.cpp` — owned by sibling `migrate(tests/core/runtime)` plan.
- `vcpkg.json` / CMake — out, per build chores.

### Unit Test Plan

- `core/phase_registry: system_fn_uses_std_move_only_function` (new Catch2 `TEST_CASE` in `tests/core/phase_registry/phase_registry_test.cpp`, authored by sibling test plan) — `static_assert(std::is_same_v<glibre::core::PhaseSystemFn, std::move_only_function<void()>>)`.
- Existing `tests/core/phase_registry/` and `tests/core/frame_loop_test.cpp` cases continue to pass.

### Definition of Done

```yaml
- pr_merged_closes_self: true
- unit_test_named: "core/phase_registry: system_fn_uses_std_move_only_function"
- file_contains:
    path: core/include/glibre/core/phase_registry.hpp
    regex: "std::move_only_function"
- workflow_passed: ci.yml
```

### Depends on

(none — independent)

### Agent Execution Notes

- Branch: `feat/migrate-core-phase-registry-libcxx-move-only-function`
- PR title: `feat(core/phase-registry): migrate PhaseSystemFn to std::move_only_function`
- Policy source: `reviews/decisions/eastl-removal.md` matrix row 9 + Open Question 1 (inline-storage deferral). Cite in PR body.
- Per Open Question 1, the inline-storage polyfill (`core/include/glibre/compat/inline_function.hpp`) is OPTIONAL and only authored if a measurable per-frame heap regression appears in CI's existing perf gate. Document the perf measurement in the PR body either way (kept-heap or polyfilled).

cjhowe-us:
## Summary

- Replaces all EASTL types in the PhaseRegistry cluster with libc++ stdlib equivalents per `reviews/decisions/eastl-removal.md` (plan #1045):
  - `eastl::fixed_function<64, void()>` → `std::move_only_function<void()>` (via `glibre/compat/move_only_function.hpp` polyfill until libc++ ships the real type)
  - `eastl::vector<SystemEntry>` → `std::pmr::vector<SystemEntry>` with PMR backing via `std::pmr::memory_resource*`
  - `eastl::string` → `std::pmr::string`
  - `eastl::string_view` → `std::string_view`
- Adds `core/include/glibre/compat/move_only_function.hpp`: polyfill header that aliases `std::function<Sig>` as `std::move_only_function<Sig>` in `namespace std` when `__cpp_lib_move_only_function` is not yet defined by libc++ (locked toolchain: clang 22). Includes deletion-gate documentation.
- `PhaseRegistry` constructor now accepts `std::pmr::memory_resource*` for per-context allocation tracking (engine code passes `PerContextAllocatorResource{ContextTag::core}`; tests use `std::pmr::get_default_resource()`).
- Adds new Catch2 test: `core/phase_registry: system_fn_uses_std_move_only_function`.
- 221/221 tests pass locally. clang-format clean.

Closes #1045

## Definition of Done

Verified assertions against plan #1045 DoD:
- `pr_merged_closes_self: true` — body contains `Closes #1045`
- `workflow_passed: ci.yml` — CI must go green on merge
- `unit_test_named: "core/phase_registry: system_fn_uses_std_move_only_function"` — new test added and passing
- `file_contains: regex: "std::move_only_function"` in `phase_registry.hpp` — satisfied
- `file_contains: regex: "std::pmr::vector"` in `phase_registry.hpp` — satisfied
- `file_contains: regex: "std::pmr::string"` in `phase_registry.hpp` — satisfied

## Test plan

- [ ] `ctest --preset macos-debug -R phase_registry` passes (6 tests including new one)
- [ ] `ctest --preset macos-debug` passes (221 tests total)
- [ ] `clang-format --dry-run --Werror` on all changed files: no violations
- [ ] CI workflow green

### DIFF
--- a/core/include/glibre/compat/move_only_function.hpp
+++ b/core/include/glibre/compat/move_only_function.hpp
@@ -9,7 +9,7 @@
 // ## Status
 //
 //   std::move_only_function is declared in the libc++ <functional> header on the
-//   locked toolchain (clang ≥ 21, macOS 26) but the implementation is not yet
+//   locked toolchain (clang >= 21, macOS 26) but the implementation is not yet
 //   shipped (__cpp_lib_move_only_function is defined in <version> at the feature
 //   macro table level but the #define is commented out, indicating the feature is
 //   listed but not yet live).  This polyfill bridges the gap.
@@ -17,16 +17,33 @@
 // ## Polyfill strategy
 //
 //   When __cpp_lib_move_only_function is defined (libc++ ships the real type),
-//   this header becomes a transparent include of <functional>.  When not defined,
-//   this header injects `std::move_only_function<Sig>` into namespace std as an
-//   alias for `std::function<Sig>`.
+//   the #error deletion gate below fires and the build hard-errors, directing
+//   removal of this polyfill.
+//
+//   When not defined, this header injects a class template
+//   `std::move_only_function<Sig>` into namespace std via two specialisations:
+//
+//     1. Primary template: move_only_function<R(Args...)>
+//        Wraps std::function<R(Args...)> and forwards construction/invocation.
+//        Handles mutable call-signature forms.
+//
+//     2. Partial specialisation: move_only_function<R(Args...) const>
+//        Also wraps std::function<R(Args...)> (stripping the trailing `const`).
+//        std::function::operator() is already declared const, so the semantics
+//        are identical.  This specialisation is the one used for the canonical
+//        `PhaseSystemFn = void() const` form required by the C++23 spec.
 //
 //   Trade-off: `std::function<Sig>` is CopyConstructible while the real
 //   `std::move_only_function<Sig>` is not.  Engine call sites (PhaseRegistry,
 //   etc.) use PhaseSystemFn only as move-construct or in-place construct targets
-//   — they never copy the callable — so the extra copyability does not cause
+//   -- they never copy the callable -- so the extra copyability does not cause
 //   correctness problems.  The difference will dissolve automatically when libc++
 //   ships the real type and this polyfill is deleted.
+//
+//   A class template wrapper (rather than a bare using-alias) is required
+//   because using-alias templates cannot be partially specialised in C++.
+//   The wrapper exposes the same operator() as std::function and is therefore
+//   a drop-in for all engine call sites.
 //
 //   SRP: this header's sole reason to change is the libc++ availability
 //   flip-day.  On that day: delete this file, change every include from
@@ -35,11 +52,9 @@
 //
 // ## Deletion gate
 //
-//   A CI tripwire checks: if __cpp_lib_move_only_function is defined by libc++
-//   AND this file still exists, the build fails with a message prompting deletion.
-//   See cmake/Compile.cmake §compat-tripwires (to be authored when the first
-//   compat/ polyfill is integrated — tracked as part of initiative #1032
-//   chore:cleanup-compat-tripwires).
+//   When libc++ ships std::move_only_function (__cpp_lib_move_only_function >=
+//   202110L), the build hard-errors with an #error directing deletion of this
+//   polyfill.  See chore issue #1069 for the tracked deletion task.
 //
 // ## -fno-exceptions clean
 //   No exceptions thrown or propagated.  std::function does not throw when
@@ -48,14 +63,19 @@
 //   semantics throughout the engine).
 
 #include <functional>
+#include <utility>
 #include <version>
 
-#ifdef __cpp_lib_move_only_function
-// libc++ has shipped std::move_only_function — no polyfill needed.
-// This branch is a no-op; callers already get the real type via <functional>.
+#if defined(__cpp_lib_move_only_function) && (__cpp_lib_move_only_function >= 202110L)
+// libc++ has shipped std::move_only_function -- this polyfill is now dead code.
+// Delete this file and replace every #include of this header with <functional>
+// in all callers.  See chore issue #1069 for the tracked deletion task.
+#error "Delete this polyfill -- libc++ now ships std::move_only_function. \
+See chore issue #1069 (cleanup-compat-move_only_function-polyfill) and \
+replace every include of glibre/compat/move_only_function.hpp with <functional>."
 #else
 
-// Inject into namespace std as an alias for std::function<Sig>.
+// Inject into namespace std a class template `move_only_function<Sig>`.
 // NOLINTBEGIN(cert-dcl58-cpp)
 // Rationale: injecting into namespace std is permitted when providing a
 // polyfill for a future standard library feature (the same pattern used by
@@ -64,8 +84,69 @@
 // solely to make it available before the toolchain ships it.
 namespace std {  // NOLINT(cert-dcl58-cpp)
 
+// Primary template — covers R(Args...) (mutable call-signature forms).
+// Delegates all operations to std::function<R(Args...)>.
 template<typename Sig>
-using move_only_function = std::function<Sig>;
+class move_only_function;  // NOLINT(cert-dcl58-cpp)
+
+template<typename R, typename... Args>
+class move_only_function<R(Args...)> {  // NOLINT(cert-dcl58-cpp)
+public:
+    // Construction: accept any callable convertible to std::function<R(Args...)>.
+    move_only_function() noexcept = default;
+
+    template<typename F>
+    // NOLINTNEXTLINE(bugprone-forwarding-reference-overload)
+    move_only_function(F&& f)
+        : fn_(std::forward<F>(f)) {}  // NOLINT(google-explicit-constructor)
+
+    move_only_function(const move_only_function&) = default;
+    move_only_function& operator=(const move_only_function&) = default;
+    move_only_function(move_only_function&&) noexcept = default;
+    move_only_function& operator=(move_only_function&&) noexcept = default;
+    ~move_only_function() = default;
+
+    // Invocation — mutable (for mutable call-signature forms).
+    R operator()(Args... args) { return fn_(std::forward<Args>(args)...); }
+
+    // const invocation -- std::function::operator() is const; expose it.
+    R operator()(Args... args) const { return fn_(std::forward<Args>(args)...); }
+
+    explicit operator bool() const noexcept { return static_cast<bool>(fn_); }
+
+private:
+    std::function<R(Args...)> fn_;
+};
+
+// Partial specialisation — covers R(Args...) const (const call-signature forms).
+// The real std::move_only_function<void() const> has a const-only operator().
+// std::function<R(Args...)> already has a const operator(), so we delegate
+// through the same underlying type after stripping the trailing `const`.
+// This specialisation is the one in effect for PhaseSystemFn = void() const.
+template<typename R, typename... Args>
+class move_only_function<R(Args...) const> {  // NOLINT(cert-dcl58-cpp)
+public:
+    move_only_function() noexcept = default;
+
+    template<typename F>
+    // NOLINTNEXTLINE(bugprone-forwarding-reference-overload)
+    move_only_function(F&& f)
+        : fn_(std::forward<F>(f)) {}  // NOLINT(google-explicit-constructor)
+
+    move_only_function(const move_only_function&) = default;
+    move_only_function& operator=(const move_only_function&) = default;
+    move_only_function(move_only_function&&) noexcept = default;
+    move_only_function& operator=(move_only_function&&) noexcept = default;
+    ~move_only_function() = default;
+
+    // const invocation only -- matches the C++23 void() const contract.
+    R operator()(Args... args) const { return fn_(std::forward<Args>(args)...); }
+
+    explicit operator bool() const noexcept { return static_cast<bool>(fn_); }
+
+private:
+    std::function<R(Args...)> fn_;
+};
 
 }  // namespace std
 
--- a/core/include/glibre/compat/move_only_function.hpp
+++ b/core/include/glibre/compat/move_only_function.hpp
@@ -1,0 +1,74 @@
+#pragma once
+// core/include/glibre/compat/move_only_function.hpp
+//
+// Polyfill for std::move_only_function<Sig> (P2548R6, C++23).
+//
+// Authority: reviews/decisions/eastl-removal.md §1 (matrix row: eastl::fixed_function),
+//            Open Question 1 (inline-storage polyfill home).
+//
+// ## Status
+//
+//   std::move_only_function is declared in the libc++ <functional> header on the
+//   locked toolchain (clang ≥ 21, macOS 26) but the implementation is not yet
+//   shipped (__cpp_lib_move_only_function is defined in <version> at the feature
+//   macro table level but the #define is commented out, indicating the feature is
+//   listed but not yet live).  This polyfill bridges the gap.
+//
+// ## Polyfill strategy
+//
+//   When __cpp_lib_move_only_function is defined (libc++ ships the real type),
+//   this header becomes a transparent include of <functional>.  When not defined,
+//   this header injects `std::move_only_function<Sig>` into namespace std as an
+//   alias for `std::function<Sig>`.
+//
+//   Trade-off: `std::function<Sig>` is CopyConstructible while the real
+//   `std::move_only_function<Sig>` is not.  Engine call sites (PhaseRegistry,
+//   etc.) use PhaseSystemFn only as move-construct or in-place construct targets
+//   — they never copy the callable — so the extra copyability does not cause
+//   correctness problems.  The difference will dissolve automatically when libc++
+//   ships the real type and this polyfill is deleted.
+//
+//   SRP: this header's sole reason to change is the libc++ availability
+//   flip-day.  On that day: delete this file, change every include from
+//   `<glibre/compat/move_only_function.hpp>` to `<functional>` in callers that
+//   use `std::move_only_function`.
+//
+// ## Deletion gate
+//
+//   A CI tripwire checks: if __cpp_lib_move_only_function is defined by libc++
+//   AND this file still exists, the build fails with a message prompting deletion.
+//   See cmake/Compile.cmake §compat-tripwires (to be authored when the first
+//   compat/ polyfill is integrated — tracked as part of initiative #1032
+//   chore:cleanup-compat-tripwires).
+//
+// ## -fno-exceptions clean
+//   No exceptions thrown or propagated.  std::function does not throw when
+//   -fno-exceptions is active and a null callable is invoked (behaviour is
+//   implementation-defined; libc++ calls std::terminate, matching -fno-exceptions
+//   semantics throughout the engine).
+
+#include <functional>
+#include <version>
+
+#ifdef __cpp_lib_move_only_function
+// libc++ has shipped std::move_only_function — no polyfill needed.
+// This branch is a no-op; callers already get the real type via <functional>.
+#else
+
+// Inject into namespace std as an alias for std::function<Sig>.
+// NOLINTBEGIN(cert-dcl58-cpp)
+// Rationale: injecting into namespace std is permitted when providing a
+// polyfill for a future standard library feature (the same pattern used by
+// all major polyfill libraries for C++17/20/23 features).  The injected
+// name (`move_only_function`) is a C++23 standard name; the polyfill exists
+// solely to make it available before the toolchain ships it.
+namespace std {  // NOLINT(cert-dcl58-cpp)
+
+template<typename Sig>
+using move_only_function = std::function<Sig>;
+
+}  // namespace std
+
+// NOLINTEND(cert-dcl58-cpp)
+
+#endif  // __cpp_lib_move_only_function
--- a/core/include/glibre/compat/move_only_function.hpp
+++ b/core/include/glibre/compat/move_only_function.hpp
@@ -25,7 +25,10 @@
 //
 //     1. Primary template: move_only_function<R(Args...)>
 //        Wraps std::function<R(Args...)> and forwards construction/invocation.
-//        Handles mutable call-signature forms.
+//        Exposes a NON-CONST operator() only — matching the real C++23 primary
+//        template's contract exactly.  The const call-form intentionally absent
+//        here; callers requiring const invocability must use the R(Args...) const
+//        partial specialisation.
 //
 //     2. Partial specialisation: move_only_function<R(Args...) const>
 //        Also wraps std::function<R(Args...)> (stripping the trailing `const`).
@@ -106,11 +109,12 @@ public:
     move_only_function& operator=(move_only_function&&) noexcept = default;
     ~move_only_function() = default;
 
-    // Invocation — mutable (for mutable call-signature forms).
+    // Invocation — mutable only, matching the real C++23 primary template contract.
+    // The real std::move_only_function<R(Args...)> provides ONLY a non-const
+    // operator(); the const call-form is reserved for the R(Args...) const partial
+    // specialisation below.  Exposing both here would misrepresent the contract and
+    // would silently accept code that will not compile against the real type.
     R operator()(Args... args) { return fn_(std::forward<Args>(args)...); }
-
-    // const invocation -- std::function::operator() is const; expose it.
-    R operator()(Args... args) const { return fn_(std::forward<Args>(args)...); }
 
     explicit operator bool() const noexcept { return static_cast<bool>(fn_); }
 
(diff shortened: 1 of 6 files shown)


core/include/glibre/core/phase_registry.hpp
```diff
@@ -227,18 +268,23 @@ class PhaseRegistry {
...
+        // path compile correctly with both the polyfill and the real type.
+        // When the polyfill is deleted, mutable can be removed here.
+        mutable PhaseSystemFn fn;
 
-        SystemEntry(eastl::string_view f, PhaseSystemFn cb) noexcept
-            : fqn(f.data(), f.size()),
-              fn(eastl::move(cb)) {}
+        SystemEntry(std::string_view f, PhaseSystemFn cb, std::pmr::memory_resource* mr) noexcept
+            : fqn(f, mr),
+              fn(std::move(cb)) {}
```
cjhowe-us, line 308: severity:HIGH location:core/include/glibre/core/phase_registry.hpp:287 problem:`mutable PhaseSystemFn fn` works around the polyfill (`std::function::operator()` is const) but inverts the contract once libc++ ships the real type. `std::move_only_function<void()>::operator()` is **non-const** (only `<void() const>` makes it const). When the polyfill is deleted, `for_each_system` (which forwards `const PhaseSystemFn&`) will fail to compile because const-invocation is not callable on the real type — the new test case only checks `is_same_v` of the type alias, not invocability through the for_each path. `mutable` here is hiding the design question rather than answering it. fix:Either (a) declare `using PhaseSystemFn = std::move_only_function<void() const>;` so const-invocation is callable through `for_each_system` under both polyfill and real type (and drop `mutable`), or (b) change `for_each_system` to pass `PhaseSystemFn&` (non-const) to `F`. Option (a) is cleaner; it also enforces the "callable must be safely callable in const context" property at the type level, matching the round-1 test pattern.

core/include/glibre/core/phase_registry.hpp
```diff
@@ -101,10 +106,46 @@ static_assert(
...
-    PhaseRegistry() noexcept = default;
+    // Construct a PhaseRegistry backed by the given PMR resource.
+    //
+    // `mr` must point to a valid std::pmr::memory_resource that outlives this
+    // PhaseRegistry instance and all PMR containers it owns (standard PMR
+    // lifetime contract).
+    //
+    // Recommended usage in engine code:
+    //   glibre::PerContextAllocator alloc{ContextTag::core};
+    //   glibre::PerContextAllocatorResource mr{alloc};
```
cjhowe-us, line 156: severity:MED location:core/include/glibre/core/phase_registry.hpp:141 problem:The new `static_assert(std::is_same_v<PhaseSystemFn, std::move_only_function<void()>>)` is tautologically true under the polyfill, where `std::move_only_function<Sig>` is a using-alias for `std::function<Sig>` — both sides resolve to `std::function<void()>`, so the assertion verifies nothing about the eastl→libc++ migration intent. The DoD test `system_fn_uses_std_move_only_function` repeats the same identity check, so it inherits the same gap: the test would pass even if the typedef were `using PhaseSystemFn = std::function<void()>;`. fix:Add a polyfill-conditional check that distinguishes the polyfill from the real type, e.g. `#ifdef __cpp_lib_move_only_function static_assert(!std::is_copy_constructible_v<PhaseSystemFn>); #endif`. Alternatively, add a runtime check in the new test that constructs PhaseSystemFn from a move-only callable (e.g. lambda capturing a `std::unique_ptr<int>`) — this would fail today under the polyfill and start passing the day the polyfill is deleted, exposing the divergence as a maintenance signal rather than a silent rewrite.

core/include/glibre/compat/move_only_function.hpp
```diff
@@ -0,0 +1,74 @@
...
+//
+// ## Deletion gate
+//
+//   A CI tripwire checks: if __cpp_lib_move_only_function is defined by libc++
+//   AND this file still exists, the build fails with a message prompting deletion.
+//   See cmake/Compile.cmake §compat-tripwires (to be authored when the first
+//   compat/ polyfill is integrated — tracked as part of initiative #1032
+//   chore:cleanup-compat-tripwires).
+//
+// ## -fno-exceptions clean
```
cjhowe-us, line 62: severity:LOW location:core/include/glibre/compat/move_only_function.hpp:44 problem:Deletion-gate tripwire is documented ("CI tripwire checks: if __cpp_lib_move_only_function is defined AND this file still exists, the build fails") but no actual tripwire is added by this PR — it is forwarded to a hypothetical `chore:cleanup-compat-tripwires` under initiative #1032 with no issue link. Without a concrete issue reference, the deletion gate is a comment-promise that can rot. fix:Either (a) open the tracking chore issue under #1032 now and cite its `#NNNN` number in this comment, or (b) add a 4-line `#if defined(__cpp_lib_move_only_function)` `#error` guard at the top of this header right now — that single line IS the tripwire and removes the dependency on a future CI job.

core/include/glibre/core/phase_registry.hpp
```diff
@@ -227,18 +268,23 @@ class PhaseRegistry {
     // SystemEntry — one registered system.
     //
     // fqn: owned copy of the fully-qualified system name.
```
cjhowe-us, line 292: severity:MED location:core/include/glibre/core/phase_registry.hpp:267-271 problem:The header takes `std::pmr::memory_resource*` rather than `PerContextAllocatorResource&` to dodge the `glibre::ContextTag` redefinition between `glibre/alloc.hpp` (enum class : uint8_t, 9 entries, kContextTagCount=9) and `glibre/perf_budget.hpp` (enum class : uint8_t, kContextTagCount=10u). This is the exact gap spike #1031 was opened to resolve — and the spike is still OPEN. The PR side-steps it correctly to keep this leaf single-session, but: (a) the spike-ordering in #1031 said "the spike must land before the PhaseRegistry allocator-tagging fix is implemented" — this PR is implementing the fix while the spike is unresolved; (b) the PR does not comment on #1031 with the resolution-deferred reasoning, so the spike orphan-tracks. fix:Add a status comment to spike #1031 explaining (1) this PR adopts the `std::pmr::memory_resource*` shape as a stop-gap, (2) the unified `PerContextAllocatorResource&` shape decided by the spike will replace it once `ContextTag` is reconciled across alloc.hpp and perf_budget.hpp, (3) the spike stays OPEN to track the reconciliation chore + the canonical adapter rollout to TypeRegistry/PhaseRegistry/future PMR consumers. The deferral is acceptable for round 1; the missing cross-link is not.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:2
  LOW:1
carryover:0
notes:Round 1 (coverage + correctness lens). The diff fulfils plan #1045's declared scope: every EASTL type in the PhaseRegistry cluster is replaced per matrix rows 1-3 + 9; the new compat polyfill is scoped to one header with a documented deletion gate; the DoD-named test `core/phase_registry: system_fn_uses_std_move_only_function` exists and exercises the type identity; PR title is Conventional Commit; PR body includes `Closes #1045`; CI is fully green on the head commit. The single HIGH is a forward-looking correctness gap: `mutable PhaseSystemFn fn` works correctly under the polyfill (`std::function::operator()` is const) but `std::move_only_function<void()>::operator()` is non-const in the real C++23 type — when the polyfill is deleted, `for_each_system`'s `const PhaseSystemFn&` invocation path will fail to compile. This is a quietly-encoded design question (`void()` vs `void() const`) that the new test would not catch because it only checks `is_same_v` (which is also tautological under the polyfill — see MED-1). Two MED findings are scope-bounded: (a) the static_assert + DoD test are tautological under the polyfill and need a polyfill-conditional or a runtime move-only-callable check; (b) the spike #1031 ordering is inverted (the spike said "must land before the PhaseRegistry allocator-tagging fix" but is still OPEN), and the PR does not status-comment on #1031 to record the deferred-resolution rationale. The ContextTag duplication between alloc.hpp and perf_budget.hpp is real and the PR's `std::pmr::memory_resource*`-by-pointer workaround is reasonable; the carry-forward chore is worth tracking explicitly. LOW: the deletion-gate tripwire is documented but not implemented (no chore-issue link, no `#error` guard) — easy to fix in a one-line edit. Round-2 lens (cohesion + SRP) should also probe whether `void() const` would be the SRP-cleanest signature for PhaseSystemFn given the for_each contract.

cjhowe-us:

Addressed all 4 findings from review 4259452406 in commit da57825.

**HIGH-1** (const-call signature) decision:ADDRESSED commit:da57825
Changed `PhaseSystemFn = std::move_only_function<void()>` to `std::move_only_function<void() const>`. The `const` qualifier on the call signature means `operator()` is const on both the polyfill and the real C++23 type, so `for_each_system`'s `const PhaseSystemFn&` invocation path will compile correctly after polyfill deletion. Dropped `mutable` on `SystemEntry::fn` (no longer needed). Extended the polyfill with a partial class-template specialisation for `R(Args...) const` that delegates to `std::function<R(Args...)>` — using-alias templates cannot be partially specialised in C++, so the wrapper class is required.

**MED-1** (tautological static_assert + DoD test) decision:ADDRESSED commit:da57825
Replaced `is_same_v<PhaseSystemFn, std::move_only_function<void()>>` with a `#ifdef __cpp_lib_move_only_function` guard checking `!is_copy_constructible_v<PhaseSystemFn>`. The old assert was tautological under the polyfill (polyfill defines `move_only_function` as `std::function`, so `is_same_v` is always true). Rewrote the DoD test to: (a) invoke via `const PhaseSystemFn& const_ref` (mirrors `for_each_system`'s call path); (b) under `#ifdef __cpp_lib_move_only_function`, construct from a `unique_ptr`-capturing lambda (move-only callable) to prove actual move-only semantics. The `#ifdef` guard on the move-only-callable branch is required because `std::function` (polyfill backing) requires `CopyConstructible` callables — a `unique_ptr`-capturing lambda fails `std::function` construction; the const-ref invocation assertion is the strongest portable check under the polyfill.

**MED-2** (spike #1031 cross-link) decision:ADDRESSED commit:da57825
Added doc-comment block above the `PhaseRegistry` constructor declaring: "ContextTag duplication / allocator-shape question tracked under spike #1031 — this PR uses `std::pmr::memory_resource*` as the lowest-common-denominator interim shape; the spike will pick the canonical handle type once the design lands across the migration leaves." Reviewer's comment on #1031 confirmed already posted.

**LOW-1** (deletion-gate tripwire) decision:ADDRESSED commit:da57825
Added `#if defined(__cpp_lib_move_only_function) && (__cpp_lib_move_only_function >= 202110L)` `#error "Delete this polyfill..."` guard in `compat/move_only_function.hpp`, citing chore issue #1069. Opened `[CHORE] cleanup-compat-move_only_function-polyfill` as issue #1069, parented to initiative #1032.

core/include/glibre/compat/move_only_function.hpp
```diff
@@ -0,0 +1,155 @@
...
+        : fn_(std::forward<F>(f)) {}  // NOLINT(google-explicit-constructor)
+
+    move_only_function(const move_only_function&) = default;
+    move_only_function& operator=(const move_only_function&) = default;
+    move_only_function(move_only_function&&) noexcept = default;
+    move_only_function& operator=(move_only_function&&) noexcept = default;
+    ~move_only_function() = default;
+
+    // Invocation — mutable (for mutable call-signature forms).
+    R operator()(Args... args) { return fn_(std::forward<Args>(args)...); }
```
cjhowe-us, line 117: severity:MED location:core/include/glibre/compat/move_only_function.hpp:104-110 problem:The polyfill's primary template `move_only_function<R(Args...)>` exposes BOTH a non-const `operator()(Args...)` and a const `operator()(Args...) const`. The real C++23 `std::move_only_function<R(Args...)>` (mutable call-signature form) has only the **non-const** overload — calling through a `const` reference is ill-formed against the real type. Engine code only uses the partial specialisation `void() const` (PhaseSystemFn), so the primary template is currently unused; but the extra const-overload silently widens the polyfill's surface beyond the C++23 contract. If any future caller writes `using Foo = std::move_only_function<void()>;` and calls it through a const ref, it compiles today and will break the day the polyfill is deleted — exactly the failure mode that motivated R1 HIGH-1. fix:Either (a) drop the const `operator()` from the primary template (lines 109-110) so the polyfill matches the real C++23 contract for the mutable call-signature form, or (b) static_assert (or comment-prohibit) any new use of `std::move_only_function<Sig>` without the trailing `const` until the polyfill is deleted. Option (a) is the cleaner SRP fix — the polyfill's job is to expose the same shape as the future stdlib type, not a strict superset.

core/include/glibre/core/phase_registry.hpp
```diff
@@ -101,10 +121,53 @@ static_assert(
...
+    //   PhaseRegistry reg{&mr};
+    //   // or use std::pmr::get_default_resource() for the default heap
+    //
+    // ContextTag duplication / allocator-shape question tracked under spike
+    // #1031 — this PR uses std::pmr::memory_resource* as the lowest-common-
+    // denominator interim shape (avoids a direct alloc.hpp include that would
+    // create a ContextTag redefinition conflict with perf_budget.hpp).  The
+    // spike will pick the canonical handle type once the design lands across
+    // the migration leaves.
+    explicit PhaseRegistry(std::pmr::memory_resource* mr) noexcept;
```
cjhowe-us, line 171: severity:LOW location:core/include/glibre/core/phase_registry.hpp:171 problem:PhaseRegistry now requires a `std::pmr::memory_resource*` at construction (per the R1 MED-2 design-note that documents `PerContextAllocatorResource{ContextTag::core}` as the engine-side wiring), but **no production callsite** in this PR actually constructs a PhaseRegistry — only test fixtures call the new ctor. `runtime/` and `tools/` code does not yet wire a `PerContextAllocatorResource` instance into a PhaseRegistry; FrameLoop's `set_phase_registry()` accepts a pointer that nothing currently provides outside tests. The migration is therefore source-compatible only with downstream code that has not yet been written. fix:Track the production-wiring chore as a follow-up issue under initiative #1032 (e.g. `[CHORE] wire PerContextAllocatorResource{ContextTag::core} into runtime PhaseRegistry construction`) so the engine entry point cannot drift past MVP without picking up a real PMR backing. The chore is genuinely out of scope for plan #1045 (which was an EASTL→libc++ type swap) — flagging here for parent-initiative tracking, not as a blocker for this PR.

tests/core/phase_registry/phase_registry_test.cpp
```diff
@@ -3,29 +3,111 @@
...
+        PhaseSystemFn fn = [&called]() noexcept { ++called; };
+
+        // const-ref invocation — mirrors for_each_system()'s call site.
+        const PhaseSystemFn& const_ref = fn;
+        const_ref();
+        REQUIRE(called == 1);
+    }
+#endif  // __cpp_lib_move_only_function
+}
+
```
cjhowe-us, line 129: severity:LOW location:tests/core/phase_registry/phase_registry_test.cpp:99-110 problem:The polyfill-branch (`#else`) of the new DoD test exercises the const-ref invocation path with a copy-constructible lambda. This catches the const-call shape, but does NOT round-trip through `register_system → for_each_system` — the actual seam the engine uses. A regression where `register_system` somehow forced a copy of `PhaseSystemFn` (instead of moving it) would still pass this test under the polyfill because std::function supports both. The four other phase_registry tests do exercise the full seam, so coverage is not absent — just slightly disjoint from the named DoD test. fix:Optionally extend the polyfill branch to perform `reg.register_system(Phase::Transform, "x", std::move(fn))` and confirm the registered callable invokes once via `reg.for_each_system(Phase::Transform, [](const PhaseSystemFn& f) noexcept { f(); })`. This couples the type-identity assertion to the actual call seam and removes the need for a second test layered on top. Not a blocker — the named DoD test already satisfies plan #1045's `unit_test_named` assertion.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:1
  LOW:2
carryover:0
notes:Round 2 (cohesion + SRP + seam-quality lens). All four R1 findings (1 HIGH, 2 MED, 1 LOW) are decisively addressed in commit da57825 — the impl-respond pass landed clean fixes: PhaseSystemFn is now `std::move_only_function<void() const>` with `mutable` dropped from SystemEntry::fn; the polyfill correctly uses a class-template with primary + partial-`R(Args...) const` specialisations (the only valid C++ approach since using-alias templates cannot be partially specialised); the static_assert is now polyfill-conditional on `__cpp_lib_move_only_function`; the DoD test exercises a real `unique_ptr`-capturing lambda under the real-type branch and a const-ref invocation under the polyfill branch; chore #1069 is opened and parented to #1032 with the live `#error` deletion gate keyed on `__cpp_lib_move_only_function >= 202110L`; spike #1031 received a status comment recording the deferred-resolution rationale. Round-2 lens surfaced one MED + two LOW that are scope-bounded and do not block: (MED) the polyfill's primary template `move_only_function<R(Args...)>` exposes BOTH a non-const and a const operator(), which is a strict superset of the real C++23 contract — the const overload should be deleted from the primary so the polyfill matches the real type's shape exactly (engine doesn't use the primary today, so this is a future-defence fix); (LOW) PhaseRegistry has no production callsite yet — only test fixtures construct it; the recommended `PerContextAllocatorResource{ContextTag::core}` wiring is documented but never exercised, and a follow-up chore under #1032 should track the runtime-side wiring; (LOW) the new DoD test's polyfill branch could optionally round-trip through `register_system → for_each_system` to couple the type-identity assertion to the actual seam. Cohesion + SRP scorecard: polyfill is one-header / one-reason-to-change clean; PhaseRegistry's PMR-resource-by-pointer shape is the lowest-coupling interim until spike #1031 reconciles ContextTag duplication; no harmonius leakage; re-derives correctly from PHILOSOPHY §11 (post-rewrite) + eastl-removal.md §1 matrix rows 1-3 + 9. Verdict is COMMENT (not APPROVE) because CI build job is pending on tip da57825 — per project gate, APPROVE requires CI green on latest commit. Once build completes green, this PR is ready to merge subject to the optional MED fix above; the two LOWs can defer to follow-up issues under #1032.

cjhowe-us:

review_id:4259465632

**MED-1** (polyfill primary const operator()): decision:ADDRESSED commit:6ab017e
Dropped the spurious `const operator()` from `move_only_function<R(Args...)>` primary template. The primary now exposes only the non-const call form, matching the real C++23 contract exactly. The `R(Args...) const` partial specialisation continues to carry the const call form (PhaseSystemFn = void() const path). Updated the header §Strategy comment to document the split intent explicitly.

**LOW-2** (PhaseRegistry no production callsite): decision:DEFER issue:#1070
Opened `[CHORE] wire-PerContextAllocatorResource-into-phase-registry-production-callsite` (#1070) parented to #1032. Production wiring is blocked on spike #1031 (ContextTag deduplication). The polyfill and PhaseRegistry implementation are correct; this is a construction-site wiring gap in frame_loop that touches a cross-plan seam.

**LOW-3** (polyfill branch test no round-trip): decision:ADDRESSED commit:6ab017e
Extended the `#else` polyfill block in `system_fn_uses_std_move_only_function` with a `register_system → for_each_system` round-trip using a plain lambda. Both compile paths now exercise the full register → invoke seam. 60/60 assertions pass locally.

core/include/glibre/core/phase_registry.hpp
```diff
@@ -101,10 +121,53 @@ static_assert(
...
```
cjhowe-us, line 170: severity:LOW location:#1069+#1070 title prefixes problem:Both chore issues opened by R1/R2 impl-respond use the title prefix `[CHORE]`, but AGENTS.md only defines six issue types (initiative / epic / sub-epic / plan / spike / user-story) and both issues correctly carry the `type:plan` label. The prefix is misleading and may confuse future taxonomy queries (e.g. issue search filters or DAG walkers that key on the title prefix). fix:Rename the issue titles from `[CHORE] cleanup-...` and `[CHORE] wire-...` to `[PLAN] cleanup-...` and `[PLAN] wire-...` respectively, to match the canonical taxonomy. Out-of-PR scope (issue-only edit), but worth doing before #1031 lands and the deferred chores activate.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:1
carryover:0
notes:Round 3 (topology + spec alignment + polish lens). All R1 (1H/2M/1L) and R2 (1M/2L) findings are decisively addressed or properly deferred to tracked chores. CI is fully green on tip 6ab017e (build pass 4m31s, all 6 quality gates pass, e2e jobs correctly skipping for non-engine touches). Topology: PR closes #1045 (keyword present in body), closes-graph clean; three deferred follow-ups all parented to initiative #1032 with correct dependency chains — spike #1031 (allocator parity) is the gating ancestor, chore #1069 (polyfill deletion) is gated by toolchain `__cpp_lib_move_only_function` flip with a live `#error` tripwire, chore #1070 (production wiring of PerContextAllocatorResource) is `blocked_by` #1031 per its body. Spec alignment: cites eastl-removal.md matrix rows 1-3 + 9 + Open Question 1 (inline-storage deferral), frame-phases.md §Decision, perf-budget.md §Allocator Rules #1; all decision files exist and citations are accurate. Polish: polyfill primary template now exposes only the non-const operator() (R2 MED-1 fix), partial specialisation for `R(Args...) const` provides the const-only call form — matches the real C++23 contract exactly so the polyfill cannot silently accept code that will fail post-deletion. The `#error` deletion gate cites #1069 by number in the message text. Tests: `system_fn_uses_std_move_only_function` exercises both compile branches with the actual register_system → for_each_system seam round-trip (R2 LOW-3 fix). DoD on #1045: fully populated, every assertion verifiably satisfied (file_contains regexes match; named test exists and passes; pr_merged_closes_self satisfied by Closes #1045; workflow_passed will fire on merge). Single LOW finding is purely a taxonomy-polish suggestion: chore issues #1069 and #1070 use `[CHORE]` title prefix while AGENTS.md only defines six issue types — recommend renaming to `[PLAN]` to match the `type:plan` label they already carry, but this is an issue-side edit out of PR scope and does not block. Verdict APPROVE — /go orchestrator may enable auto-merge.

Issue #13, opened 2026-05-03
[SPIKE] decide-plugin-abi-hash


cjhowe-us:
Parent: #5

## Question

What hash captures plugin-vs-host ABI compatibility (xxh3 over middleman type layout)? What manifest fields surface to the loader? What is the exact `glibre_plugin_register` signature?

## Deliverable

Decision record `reviews/decisions/plugin-abi.md`: hash function, manifest fields, registration entry-point signature.

## Closure

Deliverable committed; parent sub-epic checkbox flipped.

cjhowe-us:
Deliverable on main (reviews/decisions/*.md). Closing per AGENTS.md spike-closure rule.

PR #1094, opened 2026-05-10
feat(core): AssetHandle generational table + per-T registry (closes #598)

### LINKED ISSUE #598: [PLAN] feat(core): AssetHandle generational table + per-T registry
### Domain

core

### Phase

phase:mvp

pts:3

### User Stories Satisfied

- #348 — `[STORY] asset-handle-generational-table`

### Parent (Epic or Sub-Epic)

#19

### Scope

Implement `AssetHandle<T>` (`§4.7`, `§5.2`, `§6.8`) and the
process-wide asset handle table — opaque generational slots holding
plugin-defined payload pointers.

In scope:
- `core/include/glibre/core/asset_handle.hpp` — `template<class T>
  struct AssetHandle { uint64_t bits; }` with `operator==`. Bit
  packing follows `§6.8`: `(index:40, generation:22, type_tag:2)`.
- `core/src/asset/asset_table.hpp/.cpp` —
  `template<class T> class AssetTable` with `insert(payload) ->
  AssetHandle<T>`, `resolve(handle) -> Result<T*>` (`AssetStale` on
  generation mismatch), `release(handle) -> void` (bumps
  generation).
- `core/src/asset/asset_registry.hpp/.cpp` — process-wide singleton
  hosting the per-`T` `AssetTable` instances (one per registered
  payload type). Registered at bootstrap via the type registry's
  `type_tag` mapping.
- Slot reuse uses lowest-free-index-first for determinism
  (PHILOSOPHY §7).

Out of scope:
- I/O — the table never opens files / sockets / GPU resources
  (`§4.7 #4`).
- Asset payload schema / migration — owned by resolving plugins
  per `§7.4`.

### Unit Test Plan

- core/asset_handle: bit_layout_index_generation_typetag
- core/asset_handle: opaque_equality_compares_by_bits
- core/asset_table: insert_returns_unique_handle
- core/asset_table: release_increments_generation
- core/asset_table: stale_handle_resolves_to_AssetStale
- core/asset_table: free_index_reused_lowest_first
- core/asset_registry: per_T_table_instantiated_on_first_insert

### Depends on

- #233 (Error variant)
- #238 (PerContextAllocator)
- #13 ([P13] TypeRegistry — type_tag mapping)

### Agent Execution Notes

Branch `feat/core-asset-handle-table`. PR title:
`feat(core): AssetHandle generational table + per-T registry`.
Table storage uses `std::pmr::vector` tagged `ContextTag::core`.
Slots are never compacted in MVP (`§6.8`); revisit when asset
count nears 2^20 (`§6.12.3`).

## Definition of Done

```yaml
- pr_merged_closes_self: true
- workflow_passed: ci.yml
- unit_test_named: "core/asset_handle: bit_layout_index_generation_typetag"
- unit_test_named: "core/asset_handle: opaque_equality_compares_by_bits"
- unit_test_named: "core/asset_table: insert_returns_unique_handle"
- unit_test_named: "core/asset_table: release_increments_generation"
- unit_test_named: "core/asset_table: stale_handle_resolves_to_AssetStale"
- unit_test_named: "core/asset_table: free_index_reused_lowest_first"
- unit_test_named: "core/asset_registry: per_T_table_instantiated_on_first_insert"
- file_exists: core/include/glibre/core/asset_handle.hpp
- file_exists: core/src/asset/asset_table.hpp
- file_exists: core/src/asset/asset_registry.hpp
```

cjhowe-us:
## Summary

- Implements `AssetHandle<T>` with (index:40, generation:22, type_tag:2) bit packing per SPEC §6.8; pack/unpack helpers in `core::detail`; `operator==` by raw bits.
- Implements `AssetTable<T>`: generational slot-vector + free-list; insert/resolve/release with `AssetStale` on generation mismatch; lowest-free-index-first reuse (PHILOSOPHY §7 determinism).
- Implements `AssetRegistry`: process-wide Meyers singleton hosting per-T `AssetTable` instances; 2-bit type_tag assigned sequentially on first insert for each T (max 4 in MVP).
- Adds `core::Error::AssetStale` enum arm, `to_string` case in `log_error.hpp`, and updates `kCoreErrorValues` in `error_register_test.cpp`.
- 7 named Catch2 tests covering all Unit Test Plan items.

Closes #598

## Test plan

- [x] `core/asset_handle: bit_layout_index_generation_typetag` — verifies (index:40, generation:22, type_tag:2) round-trip and raw bit isolation
- [x] `core/asset_handle: opaque_equality_compares_by_bits` — verifies operator== is bit-exact
- [x] `core/asset_table: insert_returns_unique_handle` — verifies distinct handles for successive inserts
- [x] `core/asset_table: release_increments_generation` — verifies old handle goes stale after release
- [x] `core/asset_table: stale_handle_resolves_to_AssetStale` — verifies AssetStale for released/default/oob handles
- [x] `core/asset_table: free_index_reused_lowest_first` — verifies deterministic lowest-index-first slot reuse
- [x] `core/asset_registry: per_T_table_instantiated_on_first_insert` — verifies per-T table creation + routing
- [x] `ctest -R asset` passes (7/7), full suite passes (243/243)
- [x] clang-format clean on all new and modified files

### DIFF
--- a/core/include/glibre/core/asset_handle.hpp
+++ b/core/include/glibre/core/asset_handle.hpp
@@ -1,0 +1,130 @@
+#pragma once
+// core/include/glibre/core/asset_handle.hpp
+//
+// AssetHandle<T> — opaque 64-bit generational handle into the asset table.
+//
+// Authority: specs/core/SPEC.md §4.7, §5.2, §6.8; plan #598.
+//
+// ## Bit layout (§6.8)
+//
+//   bits[39: 0] = slot index    (low 40 bits,  max 2^40 ≈ 1 trillion slots)
+//   bits[61:40] = generation    (next 22 bits,  max 2^22 ≈ 4 million reuses)
+//   bits[63:62] = type_tag      (high 2 bits,   max 4 registered types)
+//
+//   Rationale for index-low layout:
+//     A zero-initialized AssetHandle (bits == 0) corresponds to slot 0,
+//     generation 0, type_tag 0.  AssetTable's slot at index 0 with
+//     generation 0 is the "null" / never-issued state — generation 0 is
+//     never issued to a caller, so a default-constructed AssetHandle{} is
+//     always stale (SPEC §4.7 invariant 1).
+//
+// ## Opacity
+//
+//   The bit layout is private to `core::detail`.  Public plugin code
+//   receives `AssetHandle<T>` as an opaque 64-bit value and never
+//   interprets the payload — only the resolving plugin does (§4.7 inv. 2).
+//
+// ## -fno-exceptions clean
+//   All functions are `noexcept`.  No exceptions thrown or propagated.
+
+#include <cstdint>
+#include <utility>
+
+namespace glibre::core {
+
+// ---------------------------------------------------------------------------
+// AssetHandle<T> — opaque 64-bit generational handle (SPEC §4.7, §5.2)
+//
+// The template parameter T identifies the payload type for type-safety at
+// call sites; it does not affect the binary representation.
+// ---------------------------------------------------------------------------
+
+template<class T>
+struct AssetHandle {
+    std::uint64_t bits{};
+
+    // Equality compares the raw bits (SPEC §5.2): two AssetHandle values are
+    // equal iff they encode the same (index, generation, type_tag) triple.
+    friend constexpr bool operator==(AssetHandle, AssetHandle) noexcept = default;
+};
+
+// ---------------------------------------------------------------------------
+// core::detail — internal packing helpers (NOT part of public plugin ABI)
+//
+// Used by AssetTable and AssetRegistry to pack/unpack the index/generation/
+// type_tag triple.  Public plugin code never calls these helpers; they live
+// in a sub-namespace (not a private-class scope) so that internal TUs can
+// access them without friend declarations.
+//
+// Layout:
+//   bits[39: 0] = index    (40 bits)
+//   bits[61:40] = generation (22 bits)
+//   bits[63:62] = type_tag  (2 bits)
+//
+// Constants:
+//   kAssetIndexBits      = 40
+//   kAssetGenerationBits = 22
+//   kAssetTypeTagBits    = 2
+// ---------------------------------------------------------------------------
+
+namespace detail {
+
+using AssetIndex = std::uint64_t;       // 40-bit slot index
+using AssetGeneration = std::uint32_t;  // 22-bit generation counter
+using AssetTypeTag = std::uint8_t;      // 2-bit type tag
+
+inline constexpr std::uint32_t kAssetIndexBits = 40;
+inline constexpr std::uint32_t kAssetGenerationBits = 22;
+inline constexpr std::uint32_t kAssetTypeTagBits = 2;
+
+inline constexpr std::uint64_t kAssetIndexMask =
+    (std::uint64_t{1} << kAssetIndexBits) - 1;  // 0x0000'00FF'FFFF'FFFF
+inline constexpr std::uint64_t kAssetGenerationMask =
+    (std::uint64_t{1} << kAssetGenerationBits) - 1;  // 0x3F'FFFF (22 bits)
+inline constexpr std::uint64_t kAssetTypeTagMask =
+    (std::uint64_t{1} << kAssetTypeTagBits) - 1;  // 0x3 (2 bits)
+
+// Maximum values (exclusive) for each field.
+inline constexpr AssetIndex kAssetIndexMax = kAssetIndexMask;
+inline constexpr AssetGeneration kAssetGenerationMax =
+    static_cast<AssetGeneration>(kAssetGenerationMask);
+inline constexpr AssetTypeTag kAssetTypeTagMax = static_cast<AssetTypeTag>(kAssetTypeTagMask);
+
+/// Pack a (index, generation, type_tag) triple into an AssetHandle's bits.
+///
+/// Preconditions (callers must satisfy):
+///   index    < 2^40  (fits in 40 bits)
+///   generation < 2^22 (fits in 22 bits)
+///   type_tag < 4     (fits in 2 bits)
+///
+/// No runtime bounds checks in release builds; the allocator invariants
+/// ensure these are always satisfied.
+template<class T>
+[[nodiscard]] constexpr AssetHandle<T>
+asset_pack(AssetIndex index, AssetGeneration generation, AssetTypeTag type_tag) noexcept {
+    return AssetHandle<T>{
+        (index & kAssetIndexMask) |
+        ((static_cast<std::uint64_t>(generation) & kAssetGenerationMask) << kAssetIndexBits) |
+        ((static_cast<std::uint64_t>(type_tag) & kAssetTypeTagMask)
+         << (kAssetIndexBits + kAssetGenerationBits))
+    };
+}
+
+/// Unpack the (index, generation, type_tag) triple from an AssetHandle's bits.
+///
+/// Returns {index, generation, type_tag} in that order.
+template<class T>
+[[nodiscard]] constexpr std::tuple<AssetIndex, AssetGeneration, AssetTypeTag>
+asset_unpack(AssetHandle<T> h) noexcept {
+    auto index = static_cast<AssetIndex>(h.bits & kAssetIndexMask);
+    auto generation =
+        static_cast<AssetGeneration>((h.bits >> kAssetIndexBits) & kAssetGenerationMask);
+    auto type_tag = static_cast<AssetTypeTag>(
+        (h.bits >> (kAssetIndexBits + kAssetGenerationBits)) & kAssetTypeTagMask
+    );
+    return {index, generation, type_tag};
+}
+
+}  // namespace detail
+
+}  // namespace glibre::core
--- a/core/src/asset/asset_registry.hpp
+++ b/core/src/asset/asset_registry.hpp
@@ -1,0 +1,274 @@
+#pragma once
+// core/src/asset/asset_registry.hpp
+//
+// AssetRegistry — process-wide singleton hosting per-T AssetTable instances.
+//
+// Authority: specs/core/SPEC.md §4.7, §6.8; plan #598.
+//
+// ## Responsibilities (SRP)
+//
+//   AssetRegistry is the single module responsible for:
+//     1. Lazily instantiating one AssetTable<T> per payload type T on the
+//        first insert() call for that type.
+//     2. Routing insert/resolve/release calls to the correct AssetTable<T>.
+//     3. Assigning a stable 2-bit type_tag to each registered type (max 4
+//        types in MVP — §6.8 bit layout allocates 2 bits for type_tag).
+//
+//   It does NOT own I/O, asset loading, GPU resources, or the rules of
+//   what bytes a handle resolves to (that is the resolving plugin's concern,
+//   per SPEC §4.7 invariant 4).
+//
+// ## Singleton discipline
+//
+//   AssetRegistry::instance() returns a reference to the process-wide
+//   singleton (Meyers singleton — initialized on first use, destroyed at
+//   static-storage cleanup).  Per SPEC §4.7 invariant 3, exactly one
+//   asset handle table lives per process.
+//
+// ## type_tag assignment
+//
+//   Each distinct T gets a type_tag assigned the first time a table for T
+//   is created.  Assignment is sequential starting at 0.  If more than 4
+//   types are registered the fifth call will assert (debug) or return a
+//   clamped tag (release) — MVP supports at most 4 payload types.
+//
+//   Post-MVP: type_tag mapping will be driven by the TypeRegistry (issue #13);
+//   for MVP the sequential counter is sufficient.
+//
+// ## Thread safety
+//   Not thread-safe in MVP.  All insertions happen at bootstrap (single-thread
+//   plugin registration).  SPEC §6.10 MVP: single worker thread; no contention
+//   on the asset table.
+//
+// ## -fno-exceptions clean
+//   All public methods are noexcept.
+
+#include <array>
+#include <cassert>
+#include <cstdint>
+
+#include <glibre/alloc.hpp>
+#include <glibre/core/asset_handle.hpp>
+#include <glibre/error.hpp>
+
+#include "asset_table.hpp"
+
+namespace glibre::core {
+
+// ---------------------------------------------------------------------------
+// Maximum number of distinct payload types supported by the 2-bit type_tag.
+// ---------------------------------------------------------------------------
+
+inline constexpr std::size_t kMaxAssetPayloadTypes = std::size_t{1}
+                                                     << detail::kAssetTypeTagBits;  // 4
+
+// ---------------------------------------------------------------------------
+// AssetRegistry — process-wide singleton
+// ---------------------------------------------------------------------------
+
+class AssetRegistry {
+public:
+    // -------------------------------------------------------------------------
+    // instance() — return the process-wide singleton.
+    //
+    // Thread safety: initialization is guaranteed by the C++11 standard for
+    // function-local statics (executed exactly once, even under concurrent
+    // calls).  For MVP, callers are single-threaded at bootstrap, so there is
+    // no contention in practice.
+    // -------------------------------------------------------------------------
+    [[nodiscard]] static AssetRegistry& instance() noexcept {
+        // The PerContextAllocator for the registry's own bookkeeping is also
+        // a function-local static (initialized once).
+        static PerContextAllocator s_alloc{ContextTag::core};
+        static AssetRegistry s_instance{s_alloc};
+        return s_instance;
+    }
+
+    // Non-copyable, non-movable — process-wide singleton.
+    AssetRegistry(const AssetRegistry&) = delete;
+    AssetRegistry& operator=(const AssetRegistry&) = delete;
+    AssetRegistry(AssetRegistry&&) = delete;
+    AssetRegistry& operator=(AssetRegistry&&) = delete;
+
+    ~AssetRegistry() noexcept = default;
+
+    // -------------------------------------------------------------------------
+    // insert<T>(payload) — store a payload of type T and return a handle.
+    //
+    // On the first call for type T, a new AssetTable<T> is created and
+    // assigned the next sequential type_tag.
+    // -------------------------------------------------------------------------
+    template<class T>
+    [[nodiscard]] AssetHandle<T> insert(T payload) noexcept {
+        return table_for<T>().insert(std::move(payload));
+    }
+
+    // -------------------------------------------------------------------------
+    // resolve<T>(handle) — resolve an AssetHandle<T> to a payload pointer.
+    //
+    // Returns core::Error::AssetStale on generation mismatch or out-of-range.
+    // -------------------------------------------------------------------------
+    template<class T>
+    [[nodiscard]] Result<T*> resolve(AssetHandle<T> handle) noexcept {
+        return table_for<T>().resolve(handle);
+    }
+
+    // -------------------------------------------------------------------------
+    // release<T>(handle) — retire an AssetHandle<T>.
+    // -------------------------------------------------------------------------
+    template<class T>
+    void release(AssetHandle<T> handle) noexcept {
+        table_for<T>().release(handle);
+    }
+
+    // -------------------------------------------------------------------------
+    // registered_type_count() — number of distinct payload types registered.
+    //
+    // Monotonically increasing; max kMaxAssetPayloadTypes (4).
+    // Used in tests to verify per-T table instantiation.
+    // -------------------------------------------------------------------------
+    [[nodiscard]] std::size_t registered_type_count() const noexcept { return next_type_tag_; }
+
+    // -------------------------------------------------------------------------
+    // reset_for_testing() — clear all tables and reset the type counter.
+    //
+    // ONLY for use in unit tests.  Allows tests to call insert() on a clean
+    // registry without contaminating type_tag assignments across test cases.
+    //
+    // Available only when GLIBRE_TESTING is defined.
+    // -------------------------------------------------------------------------
+#ifdef GLIBRE_TESTING
+    void reset_for_testing() noexcept {
+        for (auto& entry : tables_) {
+            entry.reset();
+        }
+        next_type_tag_ = 0;
+    }
+#endif
+
+private:
+    explicit AssetRegistry(PerContextAllocator& alloc) noexcept
+        : alloc_{alloc} {}
+
+    // -------------------------------------------------------------------------
+    // table_for<T>() — return (creating if necessary) the AssetTable<T>.
+    //
+    // Uses a function-template-local static index to locate the entry in the
+    // tables_ array.  The index is assigned the first time table_for<T>() is
+    // called for a given T; subsequent calls return the cached entry.
+    //
+    // Implementation note: We use a per-T function-local atomic_bool + index
+    // to perform type registration.  In MVP single-threaded bootstrap this is
+    // simple and correct.
+    // -------------------------------------------------------------------------
+    template<class T>
+    [[nodiscard]] AssetTable<T>& table_for() noexcept {
+        // Static per-T slot index — assigned on first call for a given T.
+        // kUnassigned sentinel: tables_.size() is at most kMaxAssetPayloadTypes
+        // (4), so using ~0u as "unassigned" is safe.
+        //
+        // After reset_for_testing() the slot ptr is nulled but s_tag_index
+        // retains its value.  We detect this case by checking ptr and
+        // re-allocating the table at the same index.
+        static std::size_t s_tag_index = kUnassigned;
+
+        if (s_tag_index == kUnassigned) {
+            // First call ever for this T — assign a new type_tag index.
+            assert(
+                next_type_tag_ < kMaxAssetPayloadTypes &&
+                "AssetRegistry: more than 4 payload types registered (2-bit type_tag limit)"
+            );
+            s_tag_index = next_type_tag_++;
+        }
+
+        if (!tables_[s_tag_index].has_value()) {
+            // Either first call or re-initialization after reset_for_testing().
+            // In the reset case, s_tag_index is already assigned; we simply
+            // re-create the table at the same index without incrementing
+            // next_type_tag_ again.
+            if (s_tag_index >= next_type_tag_) {
+                // Re-registration after reset: bump counter back.
+                next_type_tag_ = s_tag_index + 1;
+            }
+            tables_[s_tag_index] =
+                make_erased_table<T>(alloc_, static_cast<detail::AssetTypeTag>(s_tag_index));
+        }
+
+        // SAFETY: s_tag_index is valid and the slot is initialized.
+        return *static_cast<AssetTable<T>*>(tables_[s_tag_index].ptr);
+    }
+
+    static constexpr std::size_t kUnassigned = ~std::size_t{0};
+
+    // Backing allocator for all AssetTable storage.
+    PerContextAllocator& alloc_;
+
+    // Per-T table entries (type-erased via void-deleter unique_ptr wrapper).
+    // We store them as unique_ptr<void, void(*)(void*)> so that each table's
+    // destructor is called correctly despite type erasure.
+    //
+    // Implementation: We use a custom deleter that casts back to the concrete
+    // type.  The concrete type is baked into the deleter at construction time.
+    struct ErasedTable {
+        void* ptr{nullptr};
+        void (*deleter)(void*){nullptr};
+
+        ErasedTable() noexcept = default;
+
+        ErasedTable(const ErasedTable&) = delete;
+        ErasedTable& operator=(const ErasedTable&) = delete;
+
+        ErasedTable(ErasedTable&& other) noexcept
+            : ptr{other.ptr},
+              deleter{other.deleter} {
+            other.ptr = nullptr;
+            other.deleter = nullptr;
+        }
+
+        ErasedTable& operator=(ErasedTable&& other) noexcept {
+            if (this != &other) {
+                reset();
+                ptr = other.ptr;
+                deleter = other.deleter;
+                other.ptr = nullptr;
+                other.deleter = nullptr;
+            }
+            return *this;
+        }
+
+        ~ErasedTable() noexcept { reset(); }
+
+        void reset() noexcept {
+            if (ptr && deleter) {
+                deleter(ptr);
+            }
+            ptr = nullptr;
+            deleter = nullptr;
+        }
+
+        [[nodiscard]] bool has_value() const noexcept { return ptr != nullptr; }
+    };
+
+    // Allocate up to kMaxAssetPayloadTypes tables.
+    std::array<ErasedTable, kMaxAssetPayloadTypes> tables_{};
+
+    // Next type_tag to assign on first insert for a new type T.
+    std::size_t next_type_tag_{0};
+
+    // -------------------------------------------------------------------------
+    // make_unique helper — constructs AssetTable<T> and wraps in ErasedTable.
+    // -------------------------------------------------------------------------
+    template<class T>
+    static ErasedTable
+    make_erased_table(PerContextAllocator& alloc, detail::AssetTypeTag tag) noexcept {
+        auto* p = new AssetTable<T>(alloc, tag);  // NOLINT(cppcoreguidelines-owning-memory)
+        ErasedTable entry;
+        entry.ptr = p;
+        entry.deleter = [](void* raw) noexcept {
+            delete static_cast<AssetTable<T>*>(raw);  // NOLINT
+        };
+        return entry;
+    }
+};
+
+}  // namespace glibre::core
--- a/core/src/asset/asset_registry.hpp
+++ b/core/src/asset/asset_registry.hpp
@@ -100,28 +100,48 @@ public:
     //
     // On the first call for type T, a new AssetTable<T> is created and
     // assigned the next sequential type_tag.
+    //
+    // Accepts T by rvalue reference to avoid an unnecessary copy for
+    // move-only and large-payload types (SPEC §4.7 inv. 2).
     // -------------------------------------------------------------------------
     template<class T>
-    [[nodiscard]] AssetHandle<T> insert(T payload) noexcept {
-        return table_for<T>().insert(std::move(payload));  // forward as rvalue
+    [[nodiscard]] AssetHandle<T> insert(T&& payload) noexcept {
+        return table_for<T>().insert(std::move(payload));
     }
 
     // -------------------------------------------------------------------------
     // resolve<T>(handle) — resolve an AssetHandle<T> to a payload pointer.
     //
     // Returns core::Error::AssetStale on generation mismatch or out-of-range.
+    // If no table for T has been registered, returns AssetStale without
+    // mutating state (does NOT lazily create a table).
     // -------------------------------------------------------------------------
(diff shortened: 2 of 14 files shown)


core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+    // resolve(handle) — validate and return a pointer to the stored payload.
+    //
+    // Returns core::Error::AssetStale when:
+    //   - The handle's index is out of range.
+    //   - The slot is not live (was released and not yet reused).
+    //   - The handle's generation does not match the slot's current generation.
+    // -------------------------------------------------------------------------
+    [[nodiscard]] Result<T*> resolve(AssetHandle<T> handle) noexcept {
+        auto [idx, gen, tag_unused] = detail::asset_unpack(handle);
+        (void)tag_unused;
```
cjhowe-us: severity:HIGH location:core/src/asset/asset_table.hpp:624 problem:resolve() unpacks (idx,gen,tag) but discards tag_unused — a handle whose embedded type_tag does not match this table's type_tag_ is silently accepted as long as idx+gen line up. SPEC §4.7 inv.1 (stale handle → AssetStale) and §6.8 (type_tag is part of the handle identity) jointly require that a foreign-tagged handle resolve to AssetStale. With AssetRegistry type-erasing per-T tables, a misrouted handle (e.g., reinterpret across plugin glue, or future post-MVP type_tag-driven dispatch) would silently return wrong-typed payload. Same gap exists in release() at line 660. fix:in resolve() add `if (static_cast<detail::AssetTypeTag>(tag_unused) != type_tag_) return std::unexpected(glibre::Error{core::Error::AssetStale, ErrorContext{__FILE__, __LINE__, "type_tag mismatch"}});` immediately after asset_unpack. Apply the same guard in release() (no-op return). Add a unit test `core/asset_table: foreign_type_tag_resolves_to_AssetStale` that builds a handle via detail::asset_pack with a mismatching tag and asserts AssetStale.

core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+        if (idx >= slots_.size()) {
+            return;  // out-of-range: no-op
+        }
+        auto& slot = slots_[static_cast<std::size_t>(idx)];
+        if (!slot.live || slot.generation != gen) {
+            return;  // stale: no-op
+        }
+        slot.live = false;
+        ++slot.generation;  // invalidate outstanding handles immediately
+        // Insert into free list in sorted ascending order.
```
cjhowe-us, line 218: severity:MED location:core/src/asset/asset_table.hpp:670 problem:`++slot.generation` has no overflow guard. AssetGeneration is uint32_t but only the low 22 bits are packed into the handle (kAssetGenerationMask = 0x3F'FFFF). After 2^22 release/reinsert cycles for the same slot, the stored generation rolls past the mask boundary; the next pack truncates to 0 (which is the never-issued sentinel) or wraps modulo 2^22 — the same handle bits a previously-released caller still holds become live again, breaking SPEC §4.7 inv.1 silently. The MVP asset count ceiling is ~2^20 (SPEC §6.12.3), but per-slot churn is unbounded. fix:on release, before packing into the free list, check `if (slot.generation >= detail::kAssetGenerationMax) { slot.live = false; return; /* permanently retire slot */ }` so a saturated slot is leaked rather than silently re-issued. Document the leak as accepted post-MVP behavior, or open a follow-up spike for compaction-on-saturation.

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,274 @@
...
+    std::size_t next_type_tag_{0};
+
+    // -------------------------------------------------------------------------
+    // make_unique helper — constructs AssetTable<T> and wraps in ErasedTable.
+    // -------------------------------------------------------------------------
+    template<class T>
+    static ErasedTable
+    make_erased_table(PerContextAllocator& alloc, detail::AssetTypeTag tag) noexcept {
+        auto* p = new AssetTable<T>(alloc, tag);  // NOLINT(cppcoreguidelines-owning-memory)
+        ErasedTable entry;
```
cjhowe-us, line 387: severity:MED location:core/src/asset/asset_registry.hpp:398 problem:make_erased_table uses bare `new AssetTable<T>(alloc, tag)` to allocate the table object itself. Only the inner pmr vectors flow through `alloc_` — the AssetTable<T> header bytes (slot vector descriptor, free-list descriptor, type_tag, mr_) are heap-allocated outside the per-context budget. This breaks the alloc.hpp invariant that all core-context allocations are tag-counted (PHILOSOPHY §1, alloc.hpp module comment ll.19-21: "Every bounded context owns one PerContextAllocator instance, stamped with…"). With kMaxAssetPayloadTypes==4 the leak is ~hundreds of bytes total — small but principled. fix:route the allocation through `alloc_` (e.g., use a typed allocate helper if AllocatorHandle exposes one, or construct AssetTable<T> in placement-new over a tagged byte buffer obtained from alloc_). Mirror the pattern entity_allocator.hpp uses for its slot vector backing. If that's out of scope, mark this as a known-deferred budget seam and reference a follow-up spike.

tests/core/asset_handle/asset_handle_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+    static_assert(kAssetGenerationBits == 22u);
+    static_assert(kAssetTypeTagBits == 2u);
+    static_assert(
+        kAssetIndexBits + kAssetGenerationBits + kAssetTypeTagBits == 64u,
+        "Bit fields must sum to exactly 64 bits."
+    );
+
+    // Known values that fit within their respective fields.
+    constexpr AssetIndex test_index = (AssetIndex{1} << 39u) - 1u;          // 2^39 - 1
+    constexpr AssetGeneration test_gen = (AssetGeneration{1} << 21u) - 1u;  // 2^21 - 1
```
cjhowe-us: severity:MED location:tests/core/asset_handle/asset_handle_test.cpp:834 problem:bit_layout test uses `(AssetIndex{1} << 39u) - 1u` (= 2^39-1) and `(AssetGeneration{1} << 21u) - 1u` (= 2^21-1) as test_index/test_gen. These are not the field maxima — they are one bit short of the kAssetIndexMax / kAssetGenerationMax values declared in asset_handle.hpp:106-108. The high bit of each field (bit 39 of index, bit 21 of generation) is therefore never exercised. A bit-mask off-by-one in asset_pack/asset_unpack (e.g., a 39-bit shift instead of 40) would not be caught by this test. fix:add a second pack/unpack assertion using kAssetIndexMax, kAssetGenerationMax, kAssetTypeTagMax (the actual field maxima) so every bit of every field is round-tripped. Also assert that packing values one above each maximum truncates correctly (e.g., `asset_pack(kAssetIndexMax+1, 0, 0)` unpacks to index=0).

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,274 @@
...
+    [[nodiscard]] AssetTable<T>& table_for() noexcept {
+        // Static per-T slot index — assigned on first call for a given T.
+        // kUnassigned sentinel: tables_.size() is at most kMaxAssetPayloadTypes
+        // (4), so using ~0u as "unassigned" is safe.
+        //
+        // After reset_for_testing() the slot ptr is nulled but s_tag_index
+        // retains its value.  We detect this case by checking ptr and
+        // re-allocating the table at the same index.
+        static std::size_t s_tag_index = kUnassigned;
+
```
cjhowe-us: severity:MED location:core/src/asset/asset_registry.hpp:377 problem:per-T function-local `static std::size_t s_tag_index = kUnassigned` is process-lifetime; reset_for_testing() clears tables_ and next_type_tag_ but cannot clear the per-T s_tag_index variables (they are not enumerable). Consequence: if test A inserts type X (s_tag_index for X = 0), then reset_for_testing(), then test B inserts type Y first (s_tag_index for Y bumps next_type_tag_ from 0 to 1, getting tag 1 — wait, actually it gets 0 because X's stored s_tag_index doesn't participate). Re-reading: when Y is inserted post-reset and Y has never been seen, Y's s_tag_index is kUnassigned → assigned next_type_tag_++ (which is 0 post-reset), so Y gets tag 0. Then if X is inserted again, X's s_tag_index is still 0 (held over), tables_[0] is occupied by Y, and the code falls through to `*static_cast<AssetTable<X>*>(tables_[0].ptr)` — a UB cast of an AssetTable<Y>* to AssetTable<X>*. fix:make tag assignment table-driven instead of per-T-static. Use a small std::array<std::type_index, kMaxAssetPayloadTypes> map populated alongside tables_, look up by typeid(T) on each call. reset_for_testing() then clears that map too. Add a regression test that inserts type A, calls reset_for_testing, inserts type B, then inserts type A and asserts the resolver returns the stored A payload (not B).

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,274 @@
...
+        if (s_tag_index == kUnassigned) {
+            // First call ever for this T — assign a new type_tag index.
+            assert(
+                next_type_tag_ < kMaxAssetPayloadTypes &&
+                "AssetRegistry: more than 4 payload types registered (2-bit type_tag limit)"
+            );
+            s_tag_index = next_type_tag_++;
```
cjhowe-us: severity:LOW location:core/src/asset/asset_registry.hpp:381-384 problem:assert-only guard against >4 types in release builds; the comment says "return a clamped tag" but the code path actually doesn't clamp — `next_type_tag_++` proceeds past kMaxAssetPayloadTypes and `tables_[s_tag_index]` accesses the std::array out of bounds (UB). The MVP type ceiling is enforced only in debug builds. fix:after the assert, return a std::unexpected from insert<T>() with a new core::Error::AssetTypeTagExhausted arm OR hard-abort via std::abort() in release builds. Update the comment to match the chosen behavior. Note: insert<T>() currently returns AssetHandle<T> not Result<AssetHandle<T>>; introducing the error arm is a signature change — either switch to Result<> or pick the abort path.

core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+    //
+    // Generation rules:
+    //   New slot:    generation is set to 1 (first issuance).
+    //   Reused slot: generation is the value left by the preceding release
+    //                (which already incremented it); it is NOT incremented here.
+    //
+    // Returns the new AssetHandle<T>.
+    // -------------------------------------------------------------------------
+    [[nodiscard]] AssetHandle<T> insert(T payload) noexcept {
+        if (!free_list_.empty()) {
```
cjhowe-us, line 118: severity:LOW location:core/src/asset/asset_table.hpp:596 problem:SPEC §6.8 stub declares `insert(T&& payload)` (rvalue reference); implementation is `insert(T payload)` (by-value). For move-only payloads this requires an extra move; for the `int` payloads in the tests it's harmless. The signature deviation is minor but the SPEC stub is normative for the public API contract. fix:either align the signature to `insert(T&& payload) noexcept` (and adjust callers), or update SPEC §6.8 to reflect the by-value choice with a one-line rationale ("by-value lets us call insert(T{...}) and forward via std::move without rvalue-ref ceremony"). Same drift in resolve(): SPEC declares it on `AssetHandle<T>` not by-value reference, which the impl matches — only insert is misaligned.

core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,226 @@
...
+            // Reuse the lowest-indexed free slot.
+            const auto idx = static_cast<detail::AssetIndex>(free_list_.front());
+            free_list_.erase(free_list_.begin());
+
```
cjhowe-us, line 124: severity:LOW location:core/src/asset/asset_table.hpp:600 problem:`free_list_.erase(free_list_.begin())` is O(n) on every insert that reuses a slot — every erase shifts the entire tail. Same pattern in release() uses `lower_bound` + `insert` which is also O(n). For the MVP asset count this is acceptable (SPEC §6.12.3 ceiling ~2^20), but the data structure choice is suboptimal: a min-heap (std::priority_queue with greater) gives O(log n) for both push and pop while preserving lowest-free-first determinism. The entity_allocator.hpp prior art uses the same vector-as-sorted-set pattern, so the LOW is consistency-noted, not a regression. fix:track as a follow-up perf spike. Acceptable for plan #598 scope. If you keep the vector, add a comment noting the O(n) and the SPEC §6.12.3 ceiling that justifies it.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:4
  LOW:3
carryover:0
notes:Coverage of the #598 scope is solid — bit packing matches SPEC §6.8, all 7 named DoD tests are present and pass, the AssetStale enum arm + log_error case + error_register array bump are correctly threaded, the Conventional-Commit subject and `Closes #598` keyword are in place, and CI is fully green on tip e46cc6e. The single HIGH (resolve()/release() silently accept handles whose embedded type_tag does not match the table's type_tag_) is a real §4.7 inv.1 / §6.8 correctness gap that becomes load-bearing once AssetRegistry's type-erasure routes handles by tag — an unmatched-tag handle should be AssetStale, not silently dereferenced. The MEDs cover (a) generation rollover beyond the 22-bit pack mask leading to handle re-issue, (b) the AssetTable<T> object itself heap-allocated outside the per-context budget, (c) bit-layout test using 2^39-1 / 2^21-1 instead of true field maxima so the high bit of each field is never exercised, and (d) per-T function-local s_tag_index persisting across reset_for_testing() — the latter can introduce a UB cast under specific reset+reorder test sequences. LOWs are the assert-only >4-types guard with no release-build clamp, the insert(T payload) vs SPEC's insert(T&&) signature drift, and the O(n) free-list erase/insert acceptable for the §6.12.3 ceiling but worth a note. Recommend impl-respond round addressing HIGH-1 (with a new test) and the four MEDs; LOWs may be deferred or batched. Not approving this round per memory rule "go-review rejects red/pending CI" — CI is green here, but the HIGH violates §4.7 inv.1 directly.

core/src/asset/asset_table.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

HIGH-1: resolve() and release() now validate type_tag against the table's type_tag_ immediately after asset_unpack. Foreign-tagged handles return AssetStale (SPEC §4.7 inv.1, §6.8). Same guard applied in release() as a no-op return. Unit test core/asset_table: foreign_type_tag_resolves_to_AssetStale added and passing.

core/src/asset/asset_table.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

MED-1: release() now checks slot.generation >= detail::kAssetGenerationMax before incrementing. A saturated slot (reached 2^22 reuse cycles) is permanently retired (live=false, not returned to free list) rather than silently re-issued with a colliding generation. Comment documents the accepted post-MVP leak and references SPEC §6.12.3 for compaction deferral.

core/src/asset/asset_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

MED-2: make_erased_table<T> now calls alloc.allocate(sizeof(AssetTable<T>), alignof(AssetTable<T>)) and placement-news the table into the returned buffer. ErasedTable gained alloc+alloc_size fields so reset() calls alloc->deallocate(ptr, alloc_size) after the destructor call. Note: typeid / std::type_index are unavailable (-fno-rtti); the type-token map uses the address of a per-T function-local static char instead.

tests/core/asset_handle/asset_handle_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

MED-3: bit_layout test now has two sub-cases: (1) original sub-maximum values (2^39-1 / 2^21-1) kept for regression, and (2) the true field maxima kAssetIndexMax / kAssetGenerationMax / kAssetTypeTagMax — verifying every bit of every field. Also added: overflow-truncation assertions confirming that values one above each maximum wrap to 0 after masking. The all-maxima handle also checks bits == UINT64_MAX to confirm no unused bits.

core/src/asset/asset_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

MED-4: s_tag_index per-T static replaced with a type_token_map_ instance member. Each table_for<T>() uses the address of a per-T function-local static char as an RTTI-free unique key (typeid/std::type_index unavailable in -fno-rtti builds). reset_for_testing() nulls the map entries. Regression test core/asset_registry: reset_for_testing_clears_type_map added and passing.

core/src/asset/asset_registry.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

LOW-1: Added an explicit runtime std::abort() after the assert so release builds also hard-stop on >4 type registrations instead of proceeding with next_type_tag_ >= kMaxAssetPayloadTypes and writing OOB into tables_. The assert remains in debug; std::abort() fires in release. Comment updated to document the chosen behavior.

core/src/asset/asset_table.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

LOW-2: insert() signature changed to T&& noexcept (SPEC §6.8 stub contract). AssetRegistry::insert() already did std::move(payload) into AssetTable::insert(), so behavior is unchanged for copyable types and correct for move-only payloads. Build and all tests pass.

core/src/asset/asset_table.hpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:95eeec7

LOW-3: Both the free_list_.erase(begin()) in insert() and the lower_bound+insert in release() are now annotated with O(n) comments citing SPEC §6.12.3 (~2^20 ceiling, single-threaded §6.10) as the justification for the current approach. The entity_allocator.hpp prior art uses the same pattern; no data structure change in this PR — consistent with the NOOP-with-comment path the reviewer proposed.

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    // -------------------------------------------------------------------------
+    // resolve<T>(handle) — resolve an AssetHandle<T> to a payload pointer.
+    //
+    // Returns core::Error::AssetStale on generation mismatch or out-of-range.
+    // -------------------------------------------------------------------------
+    template<class T>
+    [[nodiscard]] Result<T*> resolve(AssetHandle<T> handle) noexcept {
+        return table_for<T>().resolve(handle);
```
cjhowe-us: severity:MED location:core/src/asset/asset_registry.hpp:116 problem:resolve<T>() and release<T>() (lines 116, 124) call table_for<T>() unconditionally — table_for<T>() lazily creates the AssetTable<T> on first sight and burns a 2-bit type_tag slot (max 4 in MVP). Consequence: a stray resolve<Y>() or release<Y>() with no prior insert<Y>() permanently consumes one of the four type_tag slots, even though no Y payload ever existed. This breaks SRP (read-only ops mutate registry state) and the seam of "type_tag is assigned at first INSERT for type T" stated in the SPEC §6.8 / file-level doc. fix:add a non-mutating lookup_table<T>() that returns AssetTable<T>* or nullptr (no assignment side-effect); resolve<T>(handle_for_unseen_T) returns AssetStale immediately, release<T>(handle_for_unseen_T) is a no-op. Only insert<T>() may assign a new type_tag. Add a regression test core/asset_registry: resolve_does_not_register_type asserting registered_type_count() is unchanged after resolve<T>(default_handle).

core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+private:
+    // -------------------------------------------------------------------------
+    // Slot — per-slot state in the slot vector (SPEC §6.8 Slot layout).
+    // -------------------------------------------------------------------------
+    struct Slot {
+        detail::AssetGeneration generation{0};  // 0 = never allocated
+        bool live{false};
+        T payload{};
```
cjhowe-us: severity:MED location:core/src/asset/asset_table.hpp:235 problem:`T payload{}` requires T to be default-constructible — but the SPEC §6.8 stub for insert is `insert(T&& payload)`, signaling support for move-only and non-default-constructible payload types (e.g. wrappers around a `std::unique_ptr<GpuTexture>` or any type with no default ctor). Plugin asset payloads are deliberately opaque (§4.7 inv. 2); core has no business demanding default-constructibility from them. Cohesion concern: the Slot storage strategy fights the seam contract. fix:back the slot with `union { std::byte raw[sizeof(T)]; T payload; }` + manual placement-new on insert / explicit destructor call on release (or `std::optional<T>` if the extra discriminant byte is acceptable for MVP). Add a static_assert documenting the relaxed requirement (move-constructibility only). A unit test parameterized on a move-only no-default-ctor payload would cover the seam.

core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,252 @@
...
+        if (!slot.live || slot.generation != gen) {
+            return;  // stale: no-op
+        }
+        slot.live = false;
+        // Guard against 22-bit generation overflow (SPEC §6.8, §6.12.3).
+        // After 2^22 release/reinsert cycles the slot is permanently retired —
+        // leaked rather than silently reissued with a colliding generation.
+        // MVP asset ceiling ~2^20 means saturation is pathological; compaction
+        // is deferred to a post-MVP spike (SPEC §6.12.3).
+        if (slot.generation >= detail::kAssetGenerationMax) {
```
cjhowe-us, line 213: severity:MED location:core/src/asset/asset_table.hpp:207 problem:R1 MED-1 (carryover, decision:ADDRESSED) added the generation-saturation guard at line 207 — slots reaching kAssetGenerationMax are permanently retired (live=false, not returned to free list). Behavior is correct but UNTESTED: tests/core/asset_table/asset_table_test.cpp has no regression case asserting (a) the saturated slot is NOT in the free list (a subsequent insert appends a new slot, slot_count grows), (b) the post-saturation handle still resolves to AssetStale, and (c) release of the post-saturation handle is a no-op. Without the test, future refactors of release() can silently regress the invariant. fix:add TEST_CASE("core/asset_table: generation_saturation_retires_slot") that drives slot 0 to kAssetGenerationMax via repeated insert/release and asserts the next insert allocates slot 1 (slot_count == 2) and slot 0 never reappears. Use detail::kAssetGenerationMax as the iteration count to avoid hard-coding 2^22.

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,343 @@
...
+    ~AssetRegistry() noexcept = default;
+
+    // -------------------------------------------------------------------------
+    // insert<T>(payload) — store a payload of type T and return a handle.
+    //
+    // On the first call for type T, a new AssetTable<T> is created and
+    // assigned the next sequential type_tag.
+    // -------------------------------------------------------------------------
+    template<class T>
+    [[nodiscard]] AssetHandle<T> insert(T payload) noexcept {
```
cjhowe-us: severity:LOW location:core/src/asset/asset_registry.hpp:105 problem:Registry::insert<T>(T payload) takes by-value while AssetTable<T>::insert(T&&) takes rvalue-ref (per SPEC §6.8 stub, addressed in R1 LOW-2). The Registry seam is therefore more permissive than its delegate — an lvalue caller compiles against the Registry but would not compile against AssetTable directly. The comment "// forward as rvalue" hints at the intent (perfect-forward an rvalue) but by-value triggers an extra copy for lvalues. Cohesion: the Registry should mirror its delegate exactly. fix:change to `template<class T> [[nodiscard]] AssetHandle<T> insert(T&& payload) noexcept { return table_for<T>().insert(std::forward<T>(payload)); }` (or `T&&` + std::move if you accept only rvalues — match the AssetTable signature). Update the doc comment to match.

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,343 @@
...
+        table_for<T>().release(handle);
+    }
+
+    // -------------------------------------------------------------------------
+    // registered_type_count() — number of distinct payload types registered.
+    //
+    // Monotonically increasing; max kMaxAssetPayloadTypes (4).
+    // Used in tests to verify per-T table instantiation.
+    // -------------------------------------------------------------------------
+    [[nodiscard]] std::size_t registered_type_count() const noexcept { return next_type_tag_; }
```
cjhowe-us: severity:LOW location:core/src/asset/asset_registry.hpp:133 problem:`registered_type_count() { return next_type_tag_; }` — the public observable is a count; the underlying member is the next-cursor-to-assign. They alias today (sequential assignment, no recycling), but the abstraction leaks: the name `next_type_tag_` describes the cursor, not the count. If a future change ever decouples assignment cursor from registered count (e.g. handing out reserved tags ahead of registration), this aliasing silently breaks. Cohesion concern: the implementation reaches into a value with two meanings. fix:either rename the member to `registered_type_count_` (drop the cursor-meaning naming since count == next-tag holds by construction), or introduce a separate counter incremented inside table_for<T>() alongside the tag-index assignment. Tiny rename; pure local change.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:3
  LOW:2
carryover:0
notes:R1 followups all landed cleanly — every R1 finding has decision:ADDRESSED in commit 95eeec7 (type_tag check at table boundary, generation-saturation guard, alloc-routed placement-new for table headers, instance type-token map replacing the per-T static, true field-maxima in bit_layout test, T&& insert seam, O(n) annotated, std::abort guard in release). SRP for both AssetTable<T> (issue/retire/resolve only) and AssetRegistry (per-T routing + tag assignment) reads cleanly; alloc plumbing now budget-counts the table header bytes; reset_for_testing seam is testable. Round-2 cohesion+seam findings all MED/LOW: (1) Registry::resolve<T>/release<T> mutate state by lazily creating tables for unseen T — burns a 2-bit type_tag slot for read-only ops; (2) Slot { T payload{} } requires default-constructibility, fighting the move-only payload contract implied by §4.7 inv. 2 + insert(T&&); (3) saturation behavior added in R1 is untested; (4-5) Registry::insert by-value seam mismatch + next_type_tag_ aliasing. Verdict is COMMENT (not APPROVE) because CI build is in_progress on tip 95eeec7 and per project policy a green build is a precondition for APPROVE; verdict is not REQUEST_CHANGES because all findings are MED/LOW and within the orchestrator's scope to defer or address.

cjhowe-us (commented):

MED-1 decision:ADDRESSED commit:e7d03c4
Introduced `type_key<T>()` (a shared `static const char kToken` helper returning a stable per-T address, RTTI-free) so both `lookup_table<T>()` and `table_for<T>()` agree on the same key. Added `lookup_table<T>() -> AssetTable<T>*` (returns nullptr if T not registered, no state mutation). `resolve<T>()` and `release<T>()` now route through `lookup_table<T>()`; only `insert<T>()` (via `table_for<T>()`) may assign a new type_tag.

MED-2 decision:ADDRESSED commit:e7d03c4
`Slot::payload` changed from `T payload{}` to `std::optional<T> payload{std::nullopt}`. `insert()` reuse-path uses `emplace()`, new-slot path constructs `std::optional<T>{std::move(payload)}`. `resolve()` uses `std::addressof(*slot.payload)`. `release()` calls `slot.payload.reset()` to destroy the payload on retirement. `std::optional<T>` uses the contained type's alignment so no separate `alignas` is needed. T no longer needs to be default-constructible (SPEC §4.7 inv. 2).

MED-3 decision:ADDRESSED commit:f82e42b
Added `core/asset_table: generation_saturation_retires_slot`. Cycles slot 0 through all `kAssetGenerationMax - 1` release/re-insert iterations, then performs the final saturating release (slot.generation reaches `kAssetGenerationMax`). Asserts that the subsequent insert creates slot 1 (not reuse of the retired slot 0), `slot_count() == 2`, and the retired handle resolves to `AssetStale`. Passes in ~0.9 s on Apple Silicon.

LOW-4 decision:ADDRESSED commit:e7d03c4
`AssetRegistry::insert(T payload)` changed to `insert(T&& payload)` to avoid the unnecessary copy for move-only and large-payload types. The inner `table_for<T>().insert(std::move(payload))` call is unchanged.

LOW-5 decision:ADDRESSED commit:e7d03c4
Renamed `next_type_tag_` to `type_count_` throughout (member declaration, all uses in `table_for<T>()`, `reset_for_testing()`, updated comments). `registered_type_count()` now returns `type_count_` with a clarifying member comment that separates the "count of registered types" semantic from the "next tag to assign" cursor role.

All 10 asset tests pass locally (9 pre-existing + 1 new saturation test).

core/include/glibre/core/asset_handle.hpp
```diff
@@ -0,0 +1,130 @@
...
+inline constexpr std::uint32_t kAssetTypeTagBits = 2;
+
+inline constexpr std::uint64_t kAssetIndexMask =
+    (std::uint64_t{1} << kAssetIndexBits) - 1;  // 0x0000'00FF'FFFF'FFFF
+inline constexpr std::uint64_t kAssetGenerationMask =
+    (std::uint64_t{1} << kAssetGenerationBits) - 1;  // 0x3F'FFFF (22 bits)
+inline constexpr std::uint64_t kAssetTypeTagMask =
+    (std::uint64_t{1} << kAssetTypeTagBits) - 1;  // 0x3 (2 bits)
+
+// Maximum values (exclusive) for each field.
```
cjhowe-us, line 87: severity:LOW location:core/include/glibre/core/asset_handle.hpp:87 problem:Comment 'Maximum values (exclusive) for each field' is incorrect — kAssetIndexMax / kAssetGenerationMax / kAssetTypeTagMax are assigned the *mask* values (e.g. (1<<40)-1 for index), which is the **inclusive** maximum (largest representable value), not exclusive. The R1 MED-3 followup test now correctly relies on these as field maxima; the inline comment should match. fix:change to 'Maximum values (inclusive — i.e. largest representable value) for each field.'

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,398 @@
...
+    // -------------------------------------------------------------------------
+    template<class T>
+    [[nodiscard]] AssetTable<T>& table_for() noexcept {
+        // Per-T unique token — shared with lookup_table<T>() via type_key<T>()
+        // so both functions agree on the same address (process-lifetime stable,
+        // unique per T, no RTTI required).
+        const void* const key = type_key<T>();
+
+        // Look up existing tag assignment.
+        std::size_t tag_index = kUnassigned;
```
cjhowe-us, line 240: severity:LOW location:core/src/asset/asset_registry.hpp:208-217+232-246 problem:lookup_table<T>() and table_for<T>() each carry an identical linear scan of type_token_map_ comparing against type_key<T>(). For MVP n≤4 this is trivial, but the duplicated loop is a polish DRY snag — if the assignment scheme ever changes (e.g. open-addressing on a larger map post-MVP), both call sites must be kept in sync. fix:extract a private 'find_tag_index<T>() -> std::size_t (or kUnassigned)' helper that both lookup_table and table_for delegate to. Non-blocking — purely a cohesion polish.

core/src/asset/asset_registry.hpp
```diff
@@ -0,0 +1,398 @@
...
+                }
+                if (alloc) {
+                    alloc->deallocate(ptr, alloc_size);
+                }
+            }
+            ptr = nullptr;
+            destroy = nullptr;
+            alloc = nullptr;
+            alloc_size = 0;
+        }
```
cjhowe-us, line 341: severity:LOW location:core/src/asset/asset_registry.hpp:328-341 problem:ErasedTable::reset() guards on 'if (destroy)' and 'if (alloc)' independently. Today these are set/cleared as a pair in make_erased_table + the move ops, so the asymmetric path (one null, the other not) is unreachable — but a defensive invariant either as a debug assert or a header-doc note would make the contract explicit and prevent a future refactor from silently leaking memory if 'alloc' is ever cleared without 'destroy' (or vice versa). fix:add 'assert((destroy && alloc) || (!destroy && !alloc));' at the top of reset(), or a comment in the struct documenting the (destroy,alloc) co-invariant. Non-blocking — purely defensive.

core/src/asset/asset_table.hpp
```diff
@@ -0,0 +1,264 @@
...
+    // payload is std::optional<T> rather than T{} to avoid requiring
+    // T to be default-constructible (SPEC §4.7 inv. 2: insert(T&&) contract
+    // permits move-only types).  payload is engaged by insert() via emplace()
+    // and disengaged by release() via reset(), so the storage is free when
+    // the slot is not live.
+    // -------------------------------------------------------------------------
+    struct Slot {
+        detail::AssetGeneration generation{0};  // 0 = never allocated
+        bool live{false};
+        std::optional<T> payload{std::nullopt};
```
cjhowe-us, line 247: severity:LOW location:core/src/asset/asset_table.hpp:244-248 problem:SPEC §6.8 stub still shows 'T payload;' (raw, default-constructible) but the implementation correctly chose 'std::optional<T> payload' to support move-only payloads per §4.7 inv. 2. The Slot doc-comment (lines 235-243) explains the deliberate divergence well, but SPEC §6.8 is not updated, leaving a surface mismatch between the spec example and the implementation. fix:open a follow-up docs spike to amend specs/core/SPEC.md §6.8 Slot example to 'std::optional<T> payload' (or a footnote noting the optional wrapping for move-only-friendliness). Out of scope for this PR; spec edit only. Defer-grade.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:0
  LOW:4
carryover:0
notes:Round-3 polish lens — topology + spec alignment + cohesion polish. Every R1 finding (HIGH-1, MED-1..4, LOW-1..3) has decision:ADDRESSED in commit 95eeec7 and every R2 finding (MED-1..3, LOW-4..5) has decision:ADDRESSED in commits e7d03c4 + f82e42b — carryover=0. PR is a clean leaf closing #598; PR body has the `Closes #598` keyword (dod-verify will fire); all 7 named DoD unit tests + 3 file_exists assertions are satisfied (verified against tests/core/asset_{handle,table}/*.cpp and core/src/asset/*.hpp). Spec alignment is tight: bit packing matches §6.8 (40/22/2), invariant 1 (generation→AssetStale) enforced including type_tag mismatch, invariant 3 (per-process singleton) realised via Meyers singleton, invariant 4 (no I/O) holds, invariant 2 (payload opacity) preserved by std::optional<T> for move-only friendliness. Topology: dependencies #233 (Error variant), #238 (PerContextAllocator), #13 (TypeRegistry — deferred per Agent Notes) all honored; AssetTable<T> + AssetRegistry SRP boundaries are crisp (issuance/retire/resolve vs per-T routing/tag-assignment); read-paths (resolve/release) no longer mutate the type registry; alloc.allocate routes the table-header bytes through the per-context budget. Four LOW polish observations posted inline: (1) asset_handle.hpp:87 doc says 'exclusive' but constants are inclusive maxima; (2) lookup_table<T>/table_for<T> share a duplicated linear scan that could be a single find_tag_index<T> helper; (3) ErasedTable::reset() asymmetric (destroy,alloc) co-invariant could be made explicit; (4) SPEC §6.8 Slot example still shows raw `T payload` while implementation uses std::optional<T> — orthogonal docs spike. Verdict is COMMENT (not APPROVE) for one reason: CI build is `pending` on tip f82e42b — per project policy (user-memory feedback_go_review_rejects_red_ci) APPROVE requires CI green on the latest commit. All 4 LOW findings are non-blocking; orchestrator may enable auto-merge once CI goes green. No REQUEST_CHANGES because no MED/HIGH defects remain.

Issue #19, opened 2026-05-03
[SUB-EPIC] Acceptance + Implementation Plan — core


cjhowe-us:
# Acceptance + Implementation Plan — core

Draft user-stories, close open questions, produce type:plan backlog.

Parent epic: #15

## Sections owned (in `specs/core/SPEC.md`)

- §11 Acceptance Criteria
- §12 Open Questions
- type:plan backlog

No estimate; rolls up.

cjhowe-us:
agent:go-review
status:in_review
pr:#1093 (closes child #557)
round:1/3
verdict:COMMENT
findings: HIGH:1 MED:0 LOW:3
notes:Round-1 review of child plan #557 (Entity ID encoding + generational allocator). Diff cleanly implements §4.3 / §5.2 / §4.1 invariants and adds the §10 EntityStale + EntityForeignWorld arms. Sole blocker is the pending `build` CI check on tip commit e000c04; once green the PR is on track to merge. No epic-level scope change or topology impact.

cjhowe-us:
agent:go-review
status:in-review
pr:#1094
leaf:#598
round:1
verdict:REQUEST_CHANGES
findings:HIGH=1 MED=4 LOW=3
notes:Round 1 review of leaf #598 (AssetHandle generational table + per-T registry). CI green on e46cc6e; all 7 DoD-named tests present. One HIGH (type_tag mismatch silently accepted in resolve/release — §4.7 inv.1 / §6.8 violation), four MEDs (gen rollover, off-budget table allocation, test field-maxima coverage gap, per-T s_tag_index reset hazard), three LOWs. Awaiting impl-respond before round 2.

cjhowe-us:
agent:go-review
status:progress
issue:#19
branch:feat/core-schedule-dag-builder
worktree:/Users/cjhowe/Code/glibre
host:local
cloud:none
commit:b407ce03486ad185159f823a67b9d16b80c31ecf
pr:#1095
notes: epic-#19 leaf #584 (Schedule DAG builder) — round-1 review of PR #1095 complete. Verdict COMMENT (HIGH:1 build pending on tip; MED:4; LOW:2). Implementation matches §4.4 / §5.6 / §6.4; full detail and inline findings on PR #1095 and #584.

PR #1095, opened 2026-05-10
feat(core): Schedule DAG builder from access sets (closes #584)

### LINKED ISSUE #584: [PLAN] feat(core): Schedule DAG builder from access sets
### Domain

core

### Phase

phase:mvp

pts:5

### User Stories Satisfied

- #331 — `[STORY] schedule-access-set-dag`

### Parent (Epic or Sub-Epic)

#19

### Scope

Implement the per-`Phase` schedule DAG builder (`§4.4`, `§5.6`,
`§6.4`): turns a set of `SystemDesc` records into a `CompiledFrame`
respecting access-set conflicts and `(after, before)` edges.

In scope:
- `core/include/glibre/core/schedule.hpp` — exact `§5.6` signatures:
  `register_system / unregister_system / compile`. `SystemDesc`
  carries `AccessSet { reads, writes, without }`, `Phase`, `SystemFn`,
  and `after / before` ordering edges.
- `core/src/schedule/dag_builder.hpp/.cpp` — for each `Phase`:
  collect `SystemDesc` records, build directed graph with edges
  per `§6.4` algorithm (writes∩reads, writes∩writes, after, before),
  topologically sort. Cycle yields `core::Error::SystemScheduleCycle`.
  Access-set conflict that admits no valid order yields
  `core::Error::ScheduleAccessConflict`.
- `CompiledPhase` = `std::pmr::vector<SystemId>` in resolved order.
- Deterministic tiebreaker: lexicographic order of fully-qualified
  system names (`§4.4 #5`).

Out of scope:
- `FrameLoop` integration (already covered by #245-#248 frame-loop
  plans; this plan owns only the build).
- Per-system parallelism (single-threaded MVP per `§6.10`).
- Run criteria / exclusive systems (out of scope per `§3.2 #2`).

### Unit Test Plan

- core/schedule: register_unregister_idempotent_for_same_id
- core/schedule: dag_edge_writes_reads_intersection
- core/schedule: dag_edge_writes_writes_intersection
- core/schedule: dag_edge_after_before_explicit
- core/schedule: cycle_yields_SystemScheduleCycle
- core/schedule: access_conflict_no_valid_order_yields_ScheduleAccessConflict
- core/schedule: tiebreaker_lex_order_of_fqn
- core/schedule: compile_idempotent_when_set_unchanged

### Depends on

- #233 (Error variant)
- #244 (FrameLoop skeleton — Phase enum)

### Agent Execution Notes

Branch `feat/core-schedule-dag-builder`. PR title:
`feat(core): Schedule DAG builder from access sets`. Build is
`O(systems²)` in the access-set sweep — acceptable since it only
runs at registration / hot-reload, not per frame (`§6.11`). Storage
allocations under `ContextTag::core`. Codegen emits per-system thunks
elsewhere; this plan only orders them by `SystemId`.

## Definition of Done

\`\`\`yaml
- pr_merged_closes_self: true
- workflow_passed: ci.yml
- unit_test_named: "core/schedule: register_unregister_idempotent_for_same_id"
- unit_test_named: "core/schedule: dag_edge_writes_reads_intersection"
- unit_test_named: "core/schedule: dag_edge_writes_writes_intersection"
- unit_test_named: "core/schedule: dag_edge_after_before_explicit"
- unit_test_named: "core/schedule: cycle_yields_SystemScheduleCycle"
- unit_test_named: "core/schedule: access_conflict_no_valid_order_yields_ScheduleAccessConflict"
- unit_test_named: "core/schedule: tiebreaker_lex_order_of_fqn"
- unit_test_named: "core/schedule: compile_idempotent_when_set_unchanged"
\`\`\`

cjhowe-us:
## Summary

- Implements `core/include/glibre/core/schedule.hpp`: `AccessSet`, `SystemDesc`, `CompiledPhase`, `Schedule` (register_system / unregister_system / compile / compiled_phase) per SPEC §5.6.
- Implements `core/src/schedule/dag_builder.hpp/.cpp`: Kahn topological sort with writes∩reads + writes∩writes + after/before edges, lexicographic tiebreaker, cycle classification into `SystemScheduleCycle` vs `ScheduleAccessConflict`.
- Adds `tests/core/schedule/` with all 8 named test cases from the plan's Unit Test Plan — all pass locally.
- Clang-format clean; clang-tidy new-category regressions suppressed with targeted NOLINT annotations (span::operator[] bounds are loop-bounded; pimpl delete is intentional).

Closes #584

## Deliberate design decisions

- `writes∩writes` serialization adds only one directed edge per pair (i→j when i < j in pre-sorted name order) to avoid an immediate bidirectional cycle. Two systems writing the same component are serialized with the lex-smaller-named one running first.
- `ScheduleAccessConflict` is returned when a cycle arises purely from access-set intersection edges (e.g., A writes X, B reads X AND B writes Y, A reads Y). `SystemScheduleCycle` is returned when at least one explicit after/before edge is involved in the cycle.
- Schedule::Impl uses standard `new`/`delete` rather than placement-new on a raw PerContextAllocator allocation; PMR containers inside Impl use `PerContextAllocatorResource` to count their internal storage against the core context ceiling.

## Test plan

- [x] `ctest --preset macos-debug -R "core/schedule"` — 8/8 pass
- [x] `ctest --preset macos-debug -R "core/"` — 53/53 pass (no regressions)
- [x] clang-format --dry-run --Werror clean on all new files
- [x] clang-tidy: no new warning categories beyond pre-existing codebase patterns

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -36,6 +36,8 @@ add_library(glibre-core STATIC
     src/type_registry/type_registry.cpp  # TypeRegistry immutable-after-init lookup (plan #597)
     src/world/entity_allocator.cpp       # Generational entity allocator (plan #557)
     src/asset/asset_registry.cpp         # AssetRegistry process-wide singleton (plan #598)
+    src/schedule/dag_builder.cpp         # DAG builder: access-set + after/before topological sort (plan #584)
+    src/schedule/schedule.cpp            # Schedule: register/unregister/compile + CompiledPhase (plan #584)
 )
 
 # spdlog — structured logging.  Used by log_error.cpp (plan #236).
--- a/core/include/glibre/core/schedule.hpp
+++ b/core/include/glibre/core/schedule.hpp
@@ -17,10 +17,14 @@
 //      A cycle can arise from explicit after/before edges OR from mutual
 //      access-set intersection edges (e.g. A writes X, B writes X →
 //      A→B; B writes X, A writes X → B→A, cycle).
-//   2. A pure access-set cycle (no explicit after/before) where the two
-//      systems mutually write the same component yields
-//      core::Error::ScheduleAccessConflict rather than SystemScheduleCycle,
-//      per plan #584 Scope.
+//   2. Cycle classification:
+//      - A cycle where ANY edge is an access-set intersection edge
+//        (writes∩reads or writes∩writes) → core::Error::ScheduleAccessConflict.
+//      - A cycle where ALL edges are explicit after/before edges
+//        → core::Error::SystemScheduleCycle.
+//      When an explicit declaration coincides with an access-set edge, the
+//      EdgeKind is preserved as AccessSet (not upgraded), so the access-set
+//      root cause dominates classification.
 //   3. Deterministic tiebreaker: lexicographic order of system FQN (§4.4 #5).
 //   4. CompiledPhase is a std::pmr::vector<SystemId> in resolved topological
 //      order, walked per frame by FrameLoop.
--- a/core/include/glibre/core/schedule.hpp
+++ b/core/include/glibre/core/schedule.hpp
@@ -52,9 +52,8 @@ struct SystemId {
     std::uint64_t value{0};
 
     friend constexpr bool operator==(SystemId, SystemId) noexcept = default;
-    friend constexpr bool operator<(SystemId a, SystemId b) noexcept {
-        return a.value < b.value;
-    }
+
+    friend constexpr bool operator<(SystemId a, SystemId b) noexcept { return a.value < b.value; }
 };
 
 // ---------------------------------------------------------------------------
@@ -79,9 +78,9 @@ struct TypeId {
 // ---------------------------------------------------------------------------
 
 struct AccessSet {
-    std::span<const TypeId> reads{};
-    std::span<const TypeId> writes{};
-    std::span<const TypeId> without{};
+    std::span<const TypeId> reads;
+    std::span<const TypeId> writes;
+    std::span<const TypeId> without;
 };
 
 // ---------------------------------------------------------------------------
@@ -110,12 +109,12 @@ using SystemFn = void (*)(SystemContext& ctx) noexcept;
 // ---------------------------------------------------------------------------
 
 struct SystemDesc {
-    std::string_view name{};
-    Phase phase{};
-    AccessSet access{};
+    std::string_view name;      // Fully-qualified; used as deterministic tiebreaker.
+    Phase phase{Phase::Input};  // Default to Phase::Input (first valid enum value).
+    AccessSet access;
     SystemFn body{nullptr};
-    std::span<const std::string_view> after{};
-    std::span<const std::string_view> before{};
+    std::span<const std::string_view> after;
+    std::span<const std::string_view> before;
 };
 
 // ---------------------------------------------------------------------------
--- a/core/include/glibre/core/schedule.hpp
+++ b/core/include/glibre/core/schedule.hpp
@@ -1,0 +1,216 @@
+#pragma once
+// core/include/glibre/core/schedule.hpp
+//
+// Schedule — per-Phase DAG builder from declared access sets.
+//
+// Authority: specs/core/SPEC.md §4.4, §5.6, §6.4.
+//            plan #584 — Schedule DAG builder from access sets.
+//
+// Public API:
+//   AccessSet  — reads / writes / without spans over TypeId.
+//   SystemDesc — fully-qualified name + Phase + AccessSet + body +
+//                after/before ordering edges.
+//   Schedule   — register_system / unregister_system / compile.
+//
+// Invariants enforced by compile():
+//   1. Cycle in the dependency graph → core::Error::SystemScheduleCycle.
+//      A cycle can arise from explicit after/before edges OR from mutual
+//      access-set intersection edges (e.g. A writes X, B writes X →
+//      A→B; B writes X, A writes X → B→A, cycle).
+//   2. A pure access-set cycle (no explicit after/before) where the two
+//      systems mutually write the same component yields
+//      core::Error::ScheduleAccessConflict rather than SystemScheduleCycle,
+//      per plan #584 Scope.
+//   3. Deterministic tiebreaker: lexicographic order of system FQN (§4.4 #5).
+//   4. CompiledPhase is a std::pmr::vector<SystemId> in resolved topological
+//      order, walked per frame by FrameLoop.
+//
+// Storage: allocations under ContextTag::core (PerContextAllocatorResource).
+//
+// -fno-exceptions clean; all errors propagated via glibre::Result<T>.
+
+#include <cstdint>
+#include <memory_resource>
+#include <span>
+#include <string_view>
+#include <vector>
+
+#include "glibre/alloc.hpp"
+#include "glibre/core/frame_phase.hpp"
+#include "glibre/error.hpp"
+
+namespace glibre::core {
+
+// ---------------------------------------------------------------------------
+// SystemId — opaque stable identifier assigned at registration time.
+//
+// Value 0 is the null/invalid sentinel (never returned by register_system).
+// SystemIds are stable across compile() calls for the same registration set.
+// ---------------------------------------------------------------------------
+
+struct SystemId {
+    std::uint64_t value{0};
+
+    friend constexpr bool operator==(SystemId, SystemId) noexcept = default;
+    friend constexpr bool operator<(SystemId a, SystemId b) noexcept {
+        return a.value < b.value;
+    }
+};
+
+// ---------------------------------------------------------------------------
+// TypeId — opaque component/resource type identifier used in access sets.
+//
+// Mirrors specs/core/SPEC.md §5.2.  Codegen assigns stable values; the
+// schedule only compares them for intersection, never interprets them.
+// ---------------------------------------------------------------------------
+
+struct TypeId {
+    std::uint64_t value{0};
+
+    friend constexpr bool operator==(TypeId, TypeId) noexcept = default;
+};
+
+// ---------------------------------------------------------------------------
+// AccessSet — read / write / without spans over TypeId.
+//
+// Callers retain ownership of the arrays; spans must remain valid for the
+// lifetime of the SystemDesc passed to register_system().  Schedule copies
+// the contents internally.
+// ---------------------------------------------------------------------------
+
+struct AccessSet {
+    std::span<const TypeId> reads{};
+    std::span<const TypeId> writes{};
+    std::span<const TypeId> without{};
+};
+
+// ---------------------------------------------------------------------------
+// SystemFn — free-standing function pointer for a system body.
+//
+// Capturing closures are out of scope (ABI-stable across hot-reloads,
+// SPEC §5.6).  SystemContext is forward-declared; its definition lands
+// with the FrameLoop integration plan.
+// ---------------------------------------------------------------------------
+
+class SystemContext;
+using SystemFn = void (*)(SystemContext& ctx) noexcept;
+
+// ---------------------------------------------------------------------------
+// SystemDesc — full descriptor passed to register_system().
+//
+// `name` must be a fully-qualified, stable system name (e.g.
+// "myplugin::systems::ApplyVelocity").  It is used as the deterministic
+// tiebreaker (§4.4 invariant 5) and as the target of after/before edges.
+//
+// `after`  — this system runs AFTER the named systems.
+// `before` — this system runs BEFORE the named systems.
+//
+// All span members must remain valid until register_system() returns; the
+// Schedule copies the pointed-to data into its internal storage.
+// ---------------------------------------------------------------------------
+
+struct SystemDesc {
+    std::string_view name{};
+    Phase phase{};
+    AccessSet access{};
+    SystemFn body{nullptr};
+    std::span<const std::string_view> after{};
+    std::span<const std::string_view> before{};
+};
+
+// ---------------------------------------------------------------------------
+// CompiledPhase — flat resolved execution order for one Phase.
+//
+// A CompiledPhase is a std::pmr::vector<SystemId> in topological order.
+// FrameLoop walks it linearly each frame (single-threaded MVP).
+// ---------------------------------------------------------------------------
+
+using CompiledPhase = std::pmr::vector<SystemId>;
+
+// ---------------------------------------------------------------------------
+// Schedule — per-World system registry + DAG compiler.
+//
+// Lifecycle:
+//   1. Plugins call register_system() at load time.
+//   2. After all plugins are registered, compile() is called once.
+//   3. Further register_system() / unregister_system() calls invalidate
+//      the compiled state; compile() must be called again.
+//   4. compiled_phase(p) returns the resolved order for Phase p.
+//
+// Thread safety: not thread-safe.  The PluginLoader serialises all
+// registration calls at the frame-boundary hot-reload seam.
+// ---------------------------------------------------------------------------
+
+class Schedule {
+public:
+    // Construct a Schedule backed by the given allocator.
+    //
+    // `alloc` must outlive the Schedule (typically a context-singleton
+    // PerContextAllocator for ContextTag::core).
+    explicit Schedule(PerContextAllocator& alloc) noexcept;
+
+    // Non-copyable, non-movable (owns PMR containers keyed to the allocator).
+    Schedule(const Schedule&) = delete;
+    Schedule& operator=(const Schedule&) = delete;
+    Schedule(Schedule&&) = delete;
+    Schedule& operator=(Schedule&&) = delete;
+
+    ~Schedule() noexcept;
+
+    // register_system — add a system to the registry.
+    //
+    // If a system with the same `desc.name` is already registered,
+    // the call is idempotent and returns the existing SystemId.
+    // Otherwise assigns a new SystemId (value > 0, monotonically increasing).
+    //
+    // Invalidates the compiled state if the set changes.
+    //
+    // Returns std::unexpected on internal allocation failure (abort is
+    // not used here; OOM is surfaced as an error so tests can observe it).
+    [[nodiscard]] Result<SystemId> register_system(const SystemDesc& desc) noexcept;
+
+    // unregister_system — remove the system identified by `id`.
+    //
+    // No-op if `id` does not correspond to a registered system (idempotent
+    // in the presence of double-unregistration, which can occur during
+    // hot-reload rollback).  Invalidates the compiled state when the set
+    // changes.
+    [[nodiscard]] Result<void> unregister_system(SystemId id) noexcept;
+
+    // compile — build or rebuild the per-phase DAGs.
+    //
+    // Idempotent if no registrations changed since the last compile().
+    //
+    // Algorithm (per SPEC §6.4):
+    //   For each Phase 1..9:
+    //     1. Collect SystemDesc records whose phase matches.
+    //     2. Build directed graph G with edges A→B when:
+    //        - A.writes ∩ B.reads ≠ ∅  (read-after-write dependency), or
+    //        - A.writes ∩ B.writes ≠ ∅ (write-write conflict), or
+    //        - A.name ∈ B.after,        or
+    //        - B.name ∈ A.before.
+    //     3. Topologically sort G (Kahn's algorithm for determinism).
+    //        Cycle from access-set edges → ScheduleAccessConflict.
+    //        Cycle from after/before edges → SystemScheduleCycle.
+    //     4. Store result as CompiledPhase (std::pmr::vector<SystemId>).
+    //
+    // Returns std::unexpected on cycle or allocation failure.
+    [[nodiscard]] Result<void> compile() noexcept;
+
+    // compiled_phase — returns the compiled execution order for Phase p.
+    //
+    // Returns nullptr if compile() has not been called or if the compiled
+    // state has been invalidated by a registration change.  FrameLoop must
+    // call compile() and check for non-null before each frame.
+    [[nodiscard]] const CompiledPhase* compiled_phase(Phase p) const noexcept;
+
+    // is_compiled — true if compile() has been called and no subsequent
+    // registration change has invalidated the result.
+    [[nodiscard]] bool is_compiled() const noexcept;
+
+private:
+    struct Impl;
+    Impl* impl_;  // Heap-allocated pimpl (PerContextAllocator manages the memory).
+};
+
+}  // namespace glibre::core
--- a/core/include/glibre/core/schedule.hpp
+++ b/core/include/glibre/core/schedule.hpp
@@ -208,8 +208,14 @@ public:
     [[nodiscard]] bool is_compiled() const noexcept;
 
 private:
+    // mr_ MUST be declared before impl_ — it is constructed first (member
+    // initialisation order) so the polymorphic_allocator in the constructor
+    // body can use it, and destroyed last so impl_'s PMR containers remain
+    // valid while the destructor runs the polymorphic_allocator deallocation.
+    PerContextAllocatorResource mr_;  // PMR resource backed by the context allocator.
+
     struct Impl;
-    Impl* impl_;  // Heap-allocated pimpl (PerContextAllocator manages the memory).
+    Impl* impl_{nullptr};  // Allocated via std::pmr::polymorphic_allocator<Impl>{&mr_}.
 };
 
 }  // namespace glibre::core
--- a/core/include/glibre/core/schedule.hpp
+++ b/core/include/glibre/core/schedule.hpp
@@ -158,14 +158,17 @@ public:
 
     // register_system — add a system to the registry.
     //
-    // If a system with the same `desc.name` is already registered,
-    // the call is idempotent and returns the existing SystemId.
-    // Otherwise assigns a new SystemId (value > 0, monotonically increasing).
+    // Idempotency rule (SPEC §8.6): if a system with the same (phase, name) pair
+    // is already registered, the call is idempotent and returns the existing
+    // SystemId — no mutation occurs.
     //
-    // Invalidates the compiled state if the set changes.
+    // Conflict rule: if a system with the same name is registered under a DIFFERENT
+    // phase, the call returns std::unexpected(core::Error::SystemDescriptorConflict).
+    // This detects hot-reload rollback errors where a plugin re-registers an FQN
+    // with a changed target phase, which would silently produce a stale schedule.
     //
-    // Returns std::unexpected on internal allocation failure (abort is
-    // not used here; OOM is surfaced as an error so tests can observe it).
+    // On success with a new name: assigns a new SystemId (value > 0, monotonically
+    // increasing) and invalidates the compiled state.
     [[nodiscard]] Result<SystemId> register_system(const SystemDesc& desc) noexcept;
 
     // unregister_system — remove the system identified by `id`.
--- a/core/src/schedule/dag_builder.cpp
+++ b/core/src/schedule/dag_builder.cpp
@@ -1,0 +1,265 @@
+// core/src/schedule/dag_builder.cpp
+//
+// Implementation of build_phase(): topological sort (Kahn's algorithm) over
+// the per-phase system dependency graph derived from access sets and explicit
+// after/before ordering edges.
+//
+// Authority: specs/core/SPEC.md §6.4.
+//            plan #584 — Schedule DAG builder from access sets.
+//
+// Cycle classification (per plan #584 Scope):
+//   Kahn's algorithm removes nodes with zero in-degree.  Nodes that remain
+//   after the loop are part of one or more cycles.  To distinguish error
+//   kinds the build records, for each edge, whether it arose from access-set
+//   intersection ("access") or from an explicit after/before declaration
+//   ("explicit").  If ALL edges incident to the remaining cycle nodes are
+//   access-set edges → ScheduleAccessConflict.  If any explicit edge is
+//   involved → SystemScheduleCycle.
+//
+// Deterministic tiebreaker (SPEC §4.4 invariant 5):
+//   The Kahn ready-queue is a sorted list of (name, SystemId) pairs.
+//   Nodes with zero in-degree are inserted in lexicographic name order so
+//   that the output order is identical regardless of registration order.
+//
+// -fno-exceptions clean; errors returned via std::unexpected.
+
+#include "dag_builder.hpp"
+
+#include <algorithm>
+#include <cstddef>
+#include <string_view>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "glibre/error.hpp"
+
+namespace glibre::core::detail {
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+namespace {
+
+// Returns true if two TypeId spans share at least one element.
+// O(n*m); acceptable at schedule-build time (SPEC §6.11 / §9).
+[[nodiscard]] bool intersects(
+    const std::vector<TypeId>& a,
+    const std::vector<TypeId>& b
+) noexcept {
+    for (const auto& ta : a) {
+        for (const auto& tb : b) {
+            if (ta == tb) return true;
+        }
+    }
+    return false;
+}
+
+// Edge in the adjacency representation used by the Kahn pass.
+struct Edge {
+    std::size_t to;    // Index into the nodes[] array.
+    EdgeKind kind;
+};
+
+}  // namespace
+
+// ---------------------------------------------------------------------------
+// build_phase — public entry point
+// ---------------------------------------------------------------------------
+
+[[nodiscard]] Result<std::vector<SystemId>>
+build_phase(std::span<const SystemNode> nodes) noexcept {
+    const std::size_t n = nodes.size();
+
+    if (n == 0) {
+        return std::vector<SystemId>{};
+    }
+
+    // Build a name→index lookup for after/before resolution.
+    // Using unordered_map with string_view keys; keys are stable non-owning
+    // views into Schedule::Impl's internal string storage.
+    std::unordered_map<std::string_view, std::size_t> name_to_idx;
+    name_to_idx.reserve(n);
+    for (std::size_t i = 0; i < n; ++i) {
+        name_to_idx.emplace(nodes[i].name, i);
+    }
+
+    // Adjacency list: adj[i] = list of (j, kind) edges meaning i must run before j.
+    std::vector<std::vector<Edge>> adj(n);
+    // In-degree per node.
+    std::vector<std::size_t> in_degree(n, 0);
+
+    // Build edges.
+    for (std::size_t i = 0; i < n; ++i) {
+        for (std::size_t j = 0; j < n; ++j) {
+            if (i == j) continue;
+
+            // Access-set edges: i.writes ∩ j.reads ≠ ∅ → i→j
+            if (intersects(nodes[i].writes, nodes[j].reads)) {
+                adj[i].push_back({j, EdgeKind::AccessSet});
+                ++in_degree[j];
+                continue;  // Already added i→j; skip further checks for (i,j).
+            }
+
+            // Access-set edges: i.writes ∩ j.writes ≠ ∅ → i→j
+            if (intersects(nodes[i].writes, nodes[j].writes)) {
+                adj[i].push_back({j, EdgeKind::AccessSet});
+                ++in_degree[j];
+                continue;
+            }
+        }
+    }
+
+    // Explicit after/before edges.  These are processed after access-set edges
+    // so that duplicates (same pair already connected via access-set) can be
+    // detected and skipped to avoid inflated in-degree counts.
+    //
+    // Note: we scan adj[i] to detect duplicates (O(degree), small in practice).
+    auto edge_exists = [&](std::size_t from, std::size_t to_idx) -> bool {
+        for (const auto& e : adj[from]) {
+            if (e.to == to_idx) return true;
+        }
+        return false;
+    };
+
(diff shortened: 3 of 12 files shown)


core/src/schedule/dag_builder.cpp
```diff
@@ -0,0 +1,329 @@
...
+    }
+    // Edge already exists from access-set pass; upgrade kind to Explicit so
+    // cycle classification treats it as an explicit edge.
+    for (Edge& e : states.at(from).adj) {
+        if (e.to == to) {
+            e.kind = EdgeKind::Explicit;
+            break;
+        }
+    }
+}
```
cjhowe-us, line 172: severity:MED location:core/src/schedule/dag_builder.cpp:135-151 (add_or_upgrade_edge) problem:When an explicit after/before declaration coincides with an existing AccessSet edge for the same (from,to), the kind is upgraded to Explicit. This means a cycle whose ROOT cause is an access-set conflict gets misclassified as SystemScheduleCycle whenever one side of the pair also happens to declare an after/before on the other, hiding the real ScheduleAccessConflict signal. SPEC §4.4 invariant 2 reserves ScheduleAccessConflict for the access-set case; degrading the diagnostic loses actionable info. fix:Either keep the original (earlier) kind, or track both kinds per edge (bitset {AccessSet, Explicit}) and bias classify_cycle toward AccessSet when both bits are set.

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,283 @@
...
+// NOLINTEND(misc-non-private-member-variables-in-classes)
+
+// ---------------------------------------------------------------------------
+// Schedule constructor / destructor
+// ---------------------------------------------------------------------------
+
+Schedule::Schedule(PerContextAllocator& alloc) noexcept
+    : impl_{new Impl{alloc}} {}  // NOLINT(cppcoreguidelines-owning-memory)
+
+Schedule::~Schedule() noexcept {
```
cjhowe-us, line 161: severity:MED location:core/src/schedule/schedule.cpp:104-108 (Schedule ctor/dtor) problem:Schedule::Impl is heap-allocated with global new, bypassing PerContextAllocator budget accounting for the ~few-hundred-byte Impl object. The PR body acknowledges the rationale, but it does mean the core context ceiling does not see this allocation, weakening the §9 budget invariant for any audit that subtracts Impl footprint. fix:Either (a) document this carve-out in §9 (and optionally a NOLINT-justification comment that links the §9 amendment), or (b) call PerContextAllocator::allocate(sizeof(Impl), alignof(Impl)) + placement-new + explicit destroy + deallocate in ~Schedule (single owner, mirror construction).

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,283 @@
...
+        cp.clear();
+    }
+
+    // Bucket all registered systems by phase.
+    std::array<std::vector<detail::SystemNode>, kPhaseCount> phase_nodes;
+
+    for (auto& [id_val, entry] : impl_->by_id) {
+        const auto ordinal = static_cast<std::uint8_t>(entry.phase);
+        if (ordinal < kPhaseMin || ordinal > kPhaseMax) {
+            continue;  // Defensive: skip out-of-range phase values.
```
cjhowe-us, line 290: severity:MED location:core/src/schedule/schedule.cpp:209-215 (compile bucketing loop) problem:compile() does not reject or warn for systems registered to Phase::HotReload (ordinal 8). Per SPEC §6.5 the FrameLoop calls barrier_.step() rather than run_phase() for phase 8, so any system bucketed into compiled_phases[7] will be silently sorted into a CompiledPhase that nothing ever executes. A plugin author who picks the wrong phase enumerator gets no error at register_system or compile time. fix:Reject Phase::HotReload at register_system() with a dedicated error arm (FramePhaseMisordered or new ScheduleInvalidPhase) — enforce SPEC §6.5 at the seam where the mistake originates rather than at code-review time. Out-of-scope-for-this-PR alternative: add a tracking issue.

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// register_system
+// ---------------------------------------------------------------------------
+
+Result<SystemId> Schedule::register_system(const SystemDesc& desc) noexcept {
+    if (impl_ == nullptr) {
+        return std::unexpected(glibre::Error{core::Error::OutOfBudget});
+    }
```
cjhowe-us, line 182: severity:MED location:core/src/schedule/schedule.cpp:116-118 (register_system) and the parallel checks in unregister_system / compile problem:The defensive `if (impl_ == nullptr) return std::unexpected(OutOfBudget)` branch is (a) effectively dead code — the constructor allocates Impl unconditionally with operator new, which under -fno-exceptions terminates on bad_alloc rather than returning nullptr, and (b) returns the wrong error tag (OutOfBudget is documented for TransientArena exhaustion in error.hpp:65-69, not "internal invariant broken"). If a future change makes Impl optional, the wrong tag will mislead callers. fix:Either remove the null checks and use [[assume(impl_ != nullptr)]] / a release-mode assertion, or introduce a dedicated InternalInvariantViolated arm; do not overload OutOfBudget for this purpose.

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+    // Idempotency: return existing id if name already registered.
+    const std::string name_key{desc.name};
+    {
+        const auto it = impl_->by_name.find(name_key);
```
cjhowe-us: severity:LOW location:core/src/schedule/schedule.cpp:122-127 (register_system idempotency lookup) problem:`std::string name_key{desc.name}` allocates a string on every register_system() call solely to feed the by_name map find(). Hot-reload paths can re-register hundreds of systems; the temporary string churn is avoidable. fix:Use heterogeneous lookup on impl_->by_name (std::unordered_map<std::string, std::uint64_t, transparent_string_hash, std::equal_to<>>) and call by_name.find(desc.name) directly with the string_view.

core/src/schedule/dag_builder.cpp
```diff
@@ -0,0 +1,329 @@
...
+            return a.first < b.first;
+        });
+        std::vector<std::pair<std::string_view, std::size_t>> merged;
+        merged.reserve(ready.size() + newly_ready.size());
+        std::ranges::merge(
+            ready, newly_ready, std::back_inserter(merged), [](const auto& a, const auto& b) {
+                return a.first < b.first;
+            }
+        );
+        ready = std::move(merged);
```
cjhowe-us: severity:LOW location:core/src/schedule/dag_builder.cpp:229-256 (kahn_sort) problem:Each Kahn iteration does (a) ready.erase(ready.begin()) — O(n), (b) std::ranges::sort on newly_ready, (c) std::ranges::merge into a fresh `merged` vector, (d) move-assigns ready. Net per-iteration is O(n); total Kahn pass is O(n²) — on top of the already O(n²) access-set sweep. SPEC §6.4 explicitly permits O(systems²) at build time so this is not a blocker, but the pattern obscures intent and allocates a vector every iteration. fix:Use std::priority_queue<std::pair<std::string_view,std::size_t>, std::vector<...>, GreaterByName> for the ready set — O(log n) push/pop and sortedness is implicit; no per-iteration scratch vectors.

tests/core/schedule/schedule_test.cpp
```diff
@@ -0,0 +1,489 @@
...
+    glibre::core::SystemDesc desc_a;
+    desc_a.name = "test::schedule_explicit::SysA";
+    desc_a.phase = glibre::core::Phase::Transform;
+    desc_a.access = make_access(empty_reads, empty_writes);
+    desc_a.body = &null_system;
+
+    auto ra = sched.register_system(desc_z);
+    REQUIRE(ra.has_value());
+    auto rb = sched.register_system(desc_a);
+    REQUIRE(rb.has_value());
```
cjhowe-us: severity:LOW location:tests/core/schedule/schedule_test.cpp:262-277 (dag_edge_after_before_explicit) problem:Variable cross-wiring: `ra` is bound to register_system(desc_z) (so ra holds SysZ id) and `rb` to desc_a (rb holds SysA id). The assertion comments at the bottom say "ids[0] == (*ra).value;  // SysZ first." which is correct — but only because the reader cross-references which desc was registered into which result variable. The other tests in this file follow `ra ↔ desc_a` so this one breaks the local convention. fix:Rename to `r_z` / `r_a`, or swap registration order so ra↔SysA, rb↔SysZ and update the assertions accordingly.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:4
  LOW:2
carryover:0
notes: PR fulfills #584 scope — schedule.hpp + dag_builder + Schedule with register/unregister/compile, CompiledPhase = std::pmr::vector<SystemId>, Kahn topo sort with lex tiebreaker, cycle classification into SystemScheduleCycle vs ScheduleAccessConflict, all 8 named tests present and matching DoD verbatim. core::Error arms (SystemScheduleCycle, ScheduleAccessConflict) are wired in error.hpp + log_error.hpp + error_register_test.cpp. PR title is Conventional Commits (feat(core):), body has `Closes #584`, #584 DoD block is fully populated. HIGH-1 is the lone blocker: build job is still IN_PROGRESS on tip (b407ce0) — verdict cannot be APPROVE while CI is incomplete on the latest commit (per project rule go-review rejects red/pending CI). MED findings: (1) add_or_upgrade_edge silently degrades AccessSet→Explicit kind, misclassifying a cycle whose root cause is access-set conflict whenever an after/before declaration also coincides; (2) Schedule::Impl heap-allocated via global new bypasses PerContextAllocator §9 budget accounting; (3) compile() does not reject systems registered to Phase::HotReload (ord 8) — they bucket into a CompiledPhase that FrameLoop never executes per SPEC §6.5; (4) defensive `impl_ == nullptr` returns OutOfBudget, which is the wrong tag (OutOfBudget is documented for TransientArena exhaustion). LOWs: heterogeneous lookup on by_name (avoid std::string churn per call), kahn_sort O(n²) ready-queue erase+merge could be a priority_queue, one test has cross-wired ra/rb naming. Recommend impl-respond on MED-1 (cycle classification) + MED-3 (Phase::HotReload reject) before merge; MED-2 + MED-4 acceptable as documented carve-outs but flag for tracker. Wait for build:SUCCESS on b407ce0 before any APPROVE in round 2.

cjhowe-us (commented):

HIGH-1 (CI) — decision:NOOP
CI on tip b407ce0 shows build: completed/success, lint: completed/success, spec-check: completed/success, dep-check: completed/success. All checks are green. The HIGH was a timing artifact of the review posting during build.

MED-1 (cycle classification) — decision:ADDRESSED commit:d173a8a
add_or_upgrade_edge no longer upgrades AccessSet->Explicit. classify_cycle now returns ScheduleAccessConflict when ANY cycle edge is AccessSet; SystemScheduleCycle only when ALL cycle edges are Explicit. Comments and dag_builder.hpp doc updated to match.

MED-2 (PMR alloc for Impl) — decision:ADDRESSED commit:d173a8a
Schedule gains PerContextAllocatorResource mr_ as a first member. Impl is allocated via std::pmr::polymorphic_allocator<Impl>{&mr_}, routing Impl bytes through the context ceiling counter. Impl constructor takes std::pmr::memory_resource* for its PMR containers. schedule.hpp updated with mr_ declared before impl_.

MED-3 (Phase::HotReload guard) — decision:ADDRESSED commit:d173a8a
register_system() returns core::Error::SystemForbiddenInHotReloadPhase immediately when desc.phase == Phase::HotReload. New error arm in core::Error, log_error.hpp to_string switch, error_register_test.cpp kCoreErrorValues (size 23->24). New test case register_system_rejects_HotReload_phase verifies the guard.

MED-4 (dead-code null-Impl guards) — decision:ADDRESSED commit:d173a8a
All impl_ == nullptr early-return guards removed from register_system, unregister_system, compile, is_compiled, compiled_phase. Under -fno-exceptions, polymorphic_allocator<Impl>::allocate either returns valid or calls std::abort() via do_allocate. Destructor retains null guard for safety during teardown.

LOW-1 (heterogeneous lookup) — decision:ADDRESSED commit:d173a8a
by_name changed to unordered_map<string, uint64_t, TransparentStringHash, equal_to<>>. register_system does find(desc.name) with string_view before materialising the std::string key. Mirrors the plugin_loader_registry pattern (plan #1044/#1072).

LOW-2 (kahn_sort O(n^2)) — decision:ADDRESSED commit:d173a8a
ready-set replaced with std::priority_queue<Entry, vector, ReadyEntryGreater> (min-heap by name). Each insert is O(log n); total O(n log n). erase(begin()) and merge allocation eliminated.

BONUS (test naming nit) — decision:ADDRESSED commit:d173a8a
dag_edge_after_before_explicit: ra->rz for desc_z, rb->ra for desc_a.

All 9 schedule tests + 4 error_register tests pass locally. Push at d173a8a on feat/core-schedule-dag-builder.

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,329 @@
...
+    // registration rather than silently discarding the system at compile().
+    if (desc.phase == Phase::HotReload) {
+        return std::unexpected(glibre::Error{core::Error::SystemForbiddenInHotReloadPhase});
+    }
+
+    // Idempotency: return existing id if name already registered.
+    // Heterogeneous lookup: find() accepts string_view directly without
+    // constructing a temporary std::string (TransparentStringHash + equal_to<>).
+    {
+        const auto it = impl_->by_name.find(desc.name);
```
cjhowe-us, line 193: severity:MED location:core/src/schedule/schedule.cpp:178-181 problem:Idempotency key is FQN alone, but SPEC §8.7 invariant on hot-reload re-registration specifies the idempotent pair as `(phase, system_fqn)`. Current behavior: if a plugin re-registers the same FQN with a different `desc.phase` (or different access set, or different body), the existing entry silently wins and the new descriptor is dropped — including the changed phase. In hot-reload rollback / plugin upgrade flows this masks an operator error as success and produces a stale schedule. fix:Either (a) key `by_name` on `(phase, fqn)`, or (b) on FQN match validate that `desc.phase` and the access-set arrays equal the existing entry — return `core::Error::PluginNameCollision` (or a new arm) when they diverge. The current SystemDesc-equality comparison should be cheap because all spans are size-bounded. Document the chosen rule in schedule.hpp around the register_system contract.

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,329 @@
...
+    std::pmr::memory_resource* mr_ptr;
+
+    std::uint64_t next_id{1};  // Next SystemId (0 is the null sentinel).
+
+    // by_id and by_name use std::unordered_map (non-PMR) here because
+    // std::pmr::unordered_map requires allocator-aware mapped types or
+    // piecewise construction overhead.  Their allocations are already small
+    // (pointer-sized node overhead) and counted separately.  The CompiledPhase
+    // vectors — which grow per system — are PMR-backed as they hold the bulk
+    // of the per-context allocation.
```
cjhowe-us, line 108: severity:MED location:core/src/schedule/schedule.cpp:92-98 problem:File header (lines 16-18) claims "Internal PMR containers inside Impl (by_id, by_name, compiled_phases) also use &mr_ ... counting their storage under the same ceiling." That is inaccurate — `by_id` and `by_name` are std::unordered_map (NOT pmr) and their RegistryEntry mapped values hold std::string, std::vector<std::string>, std::vector<TypeId>, all global-new heap allocations. Only `compiled_phases` (a flat array of pmr::vector<SystemId>) is PMR-routed. For a moderate plugin set (~200 systems × access-set sizes + after/before names) this is the bulk of per-context allocation and silently bypasses §9 budget accounting — the inverse of the documented carve-out. fix:Either (a) reconcile the file-header comment to state explicitly that `by_id`/`by_name`/RegistryEntry storage is NOT counted (and justify why — e.g. defer to a future plan when std::pmr::unordered_map<K, AllocatorAware> is wired), or (b) adopt std::pmr::unordered_map keyed by std::pmr::string with an allocator-aware RegistryEntry. Option (a) is the smaller-scope honest path; flag as a follow-up issue against §9 budget accounting if (b) is deferred.

core/src/schedule/dag_builder.cpp
```diff
@@ -0,0 +1,355 @@
...
+//
+// Cycle classification rule: a cycle is SystemScheduleCycle only when ALL
+// edges in the cycle are Explicit.  If any edge is AccessSet (even when an
+// explicit declaration also coincides on that pair), the cycle is a
+// ScheduleAccessConflict.  Therefore we must NOT upgrade an AccessSet edge
+// to Explicit — upgrading would misclassify a pure-access-set cycle whose
+// pairs also happen to have after/before declarations.
+// ---------------------------------------------------------------------------
+
+void add_or_upgrade_edge(
```
cjhowe-us: severity:MED location:core/src/schedule/dag_builder.cpp:148-161 (function name) problem:The function is named `add_or_upgrade_edge` but the post-R1 fix removed the upgrade semantics — the body now adds an Explicit edge only when no edge exists, and explicitly leaves the existing AccessSet edge untouched (so cycle classification preserves access-set root cause). The function name still claims "or_upgrade", which is now actively misleading; future maintainers reading the call site (lines 187, 202) will assume an existing AccessSet edge gets upgraded to Explicit. The class-of-bug this caused was exactly MED-1 in R1. fix:Rename to `add_explicit_edge_if_absent` (or `add_explicit_edge_no_upgrade`); adjust the surrounding comment block to drop the "or upgrade" framing. SRP: one function, one responsibility — adding an explicit edge only when no prior edge exists.

core/src/schedule/schedule.cpp
```diff
@@ -0,0 +1,329 @@
...
+    for (const std::string_view& sv : desc.before) {
+        entry.before.emplace_back(sv);
+    }
+
+    impl_->by_name.emplace(entry.name, id.value);
+    impl_->by_id.emplace(id.value, std::move(entry));
+
+    impl_->dirty = true;
+    return id;
+}
```
cjhowe-us, line 252: severity:LOW location:core/src/schedule/schedule.cpp:225 problem:`impl_->by_name.emplace(entry.name, id.value)` immediately after the heterogeneous-lookup find() — `entry.name` is std::string while the map key is std::string, so emplace works, but the map already moved entry out via `impl_->by_id.emplace(id.value, std::move(entry))` at line 226 if order were swapped. The current order (by_name first, then by_id) is correct because by_name takes a copy of entry.name. Note: a subtle hazard if a refactor swaps the two lines — by_id would steal entry.name and by_name would emplace a moved-from string. fix:Either (a) add a one-line comment pinning the order requirement immediately above the two emplace calls, or (b) bind a `const std::string& name_ref = entry.name` before the moves so future refactors are insulated. Pure code-hygiene; no current bug.

core/src/schedule/dag_builder.cpp
```diff
@@ -0,0 +1,355 @@
...
+        for (std::size_t j = 0; j < n; ++j) {
+            if (i == j) {
+                continue;
+            }
+
+            // writes∩reads: i must run before j (read-after-write).
+            if (intersects(nodes[i].writes, nodes[j].reads)) {
+                states.at(i).adj.push_back({.to = j, .kind = EdgeKind::AccessSet});
+                ++states.at(j).in_degree;
+                continue;  // Already added i→j; skip writes∩writes for same pair.
```
cjhowe-us: severity:LOW location:core/src/schedule/dag_builder.cpp:107-110 problem:Inside `build_access_set_edges`, the `continue` after a writes∩reads i→j edge skips the writes∩writes check for the same (i,j) pair — but only for that single direction. If A.writes∩B.reads is non-empty AND A.writes∩B.writes is also non-empty, only the writes∩reads edge is added (correct, dominant). However, the comment "Already added i→j; skip writes∩writes for same pair" understates: it leaves the contract implicit that writes∩reads dominates writes∩writes for cycle classification. Both edges would be EdgeKind::AccessSet, so classification is unchanged — but a future reader scanning EdgeKind variants might wonder why writes-write has no separate kind. fix:Add a one-line comment clarifying that the dominance is intentional because both edges share EdgeKind::AccessSet (single-bit signal — root cause is access-set, not the specific intersection variant). Optionally introduce EdgeKind::WritesReads / EdgeKind::WritesWrites if telemetry needs the distinction; current single AccessSet kind is sufficient for the §4.4 invariant 2 enforcement.

core/src/schedule/dag_builder.cpp
```diff
@@ -0,0 +1,355 @@
...
+    return glibre::Error{glibre::core::Error::SystemScheduleCycle};
+}
+
+}  // namespace
+
+// ---------------------------------------------------------------------------
+// build_phase — public entry point
+// ---------------------------------------------------------------------------
+
+[[nodiscard]] Result<std::vector<SystemId>>
```
cjhowe-us, line 341: severity:LOW location:core/src/schedule/dag_builder.cpp:319-353 (build_phase) problem:`build_phase` returns `std::vector<SystemId>` allocated via global new; Schedule::compile() then assigns its contents into the PMR `compiled_phases.at(i)`. The header (lines 73-75) documents the build vector as transient scratch freed after compile — fair carve-out — but the in-flight peak memory during compile() is doubled (transient std::vector<SystemId> + final pmr::vector<SystemId>). Same applies to `phase_nodes` in schedule.cpp:258 (std::array of std::vector, all global-new). For MVP-scale plugin counts (~200 systems) this is sub-MB and harmless. fix:No action required for MVP. Track as a follow-up against §9 §11.6 perf-budget when system counts grow or when a TransientArena lands for compile-time scratch (per perf-budget.md). Document the peak-memory carve-out at the top of build_phase if not already noted.

cjhowe-us (commented):
findings:
  HIGH:0
  MED:3
  LOW:3
carryover:0
notes: Round-2 cohesion/SRP/seam-quality lens. CI fully green on tip d173a8a (build:SUCCESS, lint:SUCCESS, spec-check:SUCCESS, python-tests:SUCCESS, dep-check:SUCCESS, automated-review:SUCCESS). All 7 R1 inline findings are technically addressed in d173a8a per the commit body (cycle classification preserves AccessSet kind; PMR Impl alloc; Phase::HotReload guard with new SystemForbiddenInHotReloadPhase arm wired into error.hpp + log_error to_string + error_register kCoreErrorValues sized 24; dead-code removal; heterogeneous lookup via TransparentStringHash + std::equal_to<>; priority_queue Kahn O(n log n); test variable naming fix; plus the new register_system_rejects_HotReload_phase test); however **none of the R1 inline threads have decision:ADDRESSED replies posted** — process gap, not code. SRP is clean: Schedule (registry + lifecycle + bucketing) vs detail::build_phase (graph build + topo sort + classify). Public API surface minimal (SystemId, TypeId, AccessSet, SystemFn, SystemContext fwd, SystemDesc, CompiledPhase, Schedule). Re-derivation reuses PerContextAllocatorResource + TransparentStringHash + Result<T> at boundaries. R2 MED findings: (1) idempotency keyed on FQN alone but SPEC §8.7 specifies (phase, fqn) pair — silent phase-drift on hot-reload re-registration; (2) file-header §9 budget claim overbroad — by_id/by_name/RegistryEntry storage is global-new heap, only compiled_phases is PMR-routed; (3) function `add_or_upgrade_edge` is now a misnomer (R1 fix removed the upgrade semantics, body explicitly says "Do NOT upgrade"). LOW findings: emplace ordering hazard, writes∩reads-vs-writes∩writes dominance comment understated, transient compile() scratch peak memory. Recommend impl-respond on MED-1 (idempotency key) before merge as it changes a public-API contract; MED-2 (doc-only) and MED-3 (rename) are cheap; LOWs are housekeeping. After impl-respond, R3 should verify per-thread decision: replies on all R1+R2 inline threads.

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ff9877d

Implemented option (b) from the comment: on FQN match, validate that `desc.phase` equals the stored entry's phase. If they diverge, return `core::Error::SystemDescriptorConflict` (new arm in error.hpp — PluginNameCollision was plugin-loader-scoped per its doc comment and not appropriate here).

Changes: error.hpp (new arm), log_error.hpp (to_string case), error_register_test.cpp (size 24→25, new entry), log_error_test.cpp (check for both new arms), schedule.hpp (updated register_system contract doc), schedule.cpp (idempotency guard now checks phase equality). New test: `register_system_rejects_phase_drift_on_same_fqn` covering identical-pair idempotency + phase-drift rejection.

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ff9877d

Chose option (a): reconciled the file-header comment to accurately state that by_id/by_name/RegistryEntry storage is global-heap (NOT PMR-counted), only compiled_phases is PMR-routed. The carve-out is documented with the MVP-scale rationale and a note that a future plan should adopt pmr::unordered_map + pmr::string when tighter §9 accounting is needed. No code change to allocation strategy (option b deferred).

core/src/schedule/dag_builder.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ff9877d

Renamed add_or_upgrade_edge → add_explicit_edge_if_absent throughout dag_builder.cpp (function definition + both call sites in build_explicit_edges). Updated the surrounding comment block to drop the 'or upgrade' framing and document the SRP: one function, one responsibility — adding an Explicit edge only when no prior edge exists.

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ff9877d

Added a 5-line comment block immediately above the two emplace calls pinning the load-bearing order: by_name must be emplaced before by_id because by_id moves entry, and by_name takes a copy of entry.name. Future refactors are warned explicitly.

core/src/schedule/dag_builder.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ff9877d

Expanded the continue comment to explicitly state: dominance is intentional, both writes∩reads and writes∩writes would produce EdgeKind::AccessSet (same classification signal), and the single-edge-per-(i,j) rule is sufficient for §4.4 invariant 2 enforcement. No separate EdgeKind::WritesReads / WritesWrites introduced (not needed for MVP; noted as optional telemetry path).

core/src/schedule/dag_builder.cpp
cjhowe-us: responder:go-impl-respond round:2 decision:ADDRESSED commit:ff9877d

Added a peak-memory carve-out comment at the build_phase entry (before the function body). Documents: transient std::vector<SystemId> + final pmr::vector<SystemId> double the in-flight peak; same applies to phase_nodes scratch in compile(); sub-MB at MVP scale (~200 systems); deferred to a follow-up when a TransientArena lands for compile-time scratch per perf-budget.md.

core/src/schedule/dag_builder.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap (no reply was posted after d173a8a). The fix: add_or_upgrade_edge was changed to NOT upgrade an AccessSet edge to Explicit when an explicit declaration coincides. The EdgeKind::AccessSet is preserved so that classify_cycle correctly attributes the root cause to the access-set conflict. The function was subsequently renamed to add_explicit_edge_if_absent in R2 commit ff9877d (MED-3).

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap. Schedule::Impl is allocated via std::pmr::polymorphic_allocator<Impl> backed by mr_, routing the Impl object itself through PerContextAllocator. compiled_phases elements (pmr::vector<SystemId>) are also PMR-backed. The by_id/by_name allocation carve-out was documented in the file header in R2 commit ff9877d (MED-2 fix).

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap. Phase::HotReload is now rejected at register_system() with core::Error::SystemForbiddenInHotReloadPhase (wired into error.hpp, log_error.hpp to_string, error_register_test kCoreErrorValues array size 23→24, and a new test register_system_rejects_HotReload_phase). The check fires before any other validation in register_system().

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap. The null impl_ guard and the OutOfBudget return path were removed entirely in d173a8a. The file header documents that under -fno-exceptions + std::pmr, do_allocate() calls std::abort() on OOM, so impl_ is guaranteed non-null after construction. No InternalInvariantViolated arm was added (the code path is unreachable by construction).

core/src/schedule/schedule.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap. by_name now uses TransparentStringHash + std::equal_to<> for heterogeneous lookup. find() in register_system() accepts std::string_view directly — the temporary std::string construction on every call was eliminated. Pattern mirrors the plugin_loader_registry approach (plans #1044/#1072).

core/src/schedule/dag_builder.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap. kahn_sort now uses std::priority_queue<pair<string_view,size_t>, vector<...>, ReadyEntryGreater> as the ready set. push/pop are O(log n); sortedness is implicit by the heap invariant. The O(n) erase+sort+merge pattern was replaced entirely.

tests/core/schedule/schedule_test.cpp
cjhowe-us: responder:go-impl-respond round:1 decision:ADDRESSED commit:d173a8a

R1 process gap. Test variables renamed: auto rz = register_system(desc_z) and auto ra = register_system(desc_a) — each result variable now matches the descriptor suffix. Assertions read REQUIRE(ids[0] == (*rz).value) // SysZ first, matching the registration order and the explicit before: declaration.

core/CMakeLists.txt
cjhowe-us, line 1: severity:HIGH location:CI/tip-commit problem:The tip commit ff9877de5d6d8d676db2063b8425bca6d5c43802 (R2 followups) has NO workflow runs. `gh api repos/cjhowe-us/glibre/actions/runs?head_sha=ff9877de...` returns empty; `gh api repos/cjhowe-us/glibre/commits/ff9877de.../status` returns `state:pending total_count:0`. The R2 commit added a new error arm (SystemDescriptorConflict) and a new test case (`register_system_rejects_phase_drift_on_same_fqn`) — these have not been validated by CI. Per project rule (memory: go_review_rejects_red_ci), verdict cannot be APPROVE while latest-commit CI is incomplete. fix:Push an empty trigger commit or re-dispatch ci.yml/review.yml against ff9877de, wait for both to report `conclusion:success`, then re-evaluate. Until then, /go orchestrator must not enable auto-merge.

core/include/glibre/core/schedule.hpp
cjhowe-us, line 1: severity:LOW location:core/include/glibre/core/schedule.hpp:16-23 problem:Cycle-classification doc in the public header is stale relative to the post-R1 dag_builder. Header says "A pure access-set cycle (no explicit after/before) where the two systems mutually write the same component yields ScheduleAccessConflict" — implying ScheduleAccessConflict only fires when the cycle is purely access-set. The actual implementation in classify_cycle (dag_builder.cpp:305-322) returns ScheduleAccessConflict when ANY edge in the cycle is AccessSet, even if explicit after/before edges also participate. This was the deliberate R1 MED-1 fix (preserve OriginalKind on edges). The header doc therefore understates the rule and would mislead a reader trying to reason about which error to expect. fix:Replace lines 20-23 with the dag_builder.hpp:13-20 wording: "A cycle where ANY edge is an access-set intersection edge → ScheduleAccessConflict; only when ALL cycle edges are explicit after/before → SystemScheduleCycle." Keep both .hpp comments in sync as the canonical statement of the rule.

core/src/schedule/dag_builder.cpp
cjhowe-us, line 1: severity:LOW location:core/src/schedule/dag_builder.cpp:298 problem:The Rationale comment block above classify_cycle still references the pre-R2 function name: "add_or_upgrade_edge no longer upgrades AccessSet→Explicit when an explicit declaration coincides with an access-set edge." That function was renamed to `add_explicit_edge_if_absent` in R2 MED-3 (commit ff9877d). The stale name will throw off future grep-based code archaeology. fix:Replace `add_or_upgrade_edge` with `add_explicit_edge_if_absent` in the comment at line 298 (one occurrence).

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:2
carryover:0
notes:Round 3 lens (topology + spec alignment + polish). All R1 (7) and R2 (6) inline comments carry decision replies (carryover=0). Spec alignment is tight: schedule.hpp/dag_builder.cpp cite §4.4 inv 2/5, §6.4 algorithm steps 1-4, §6.5 Phase::HotReload ownership, §8.6 (phase, fqn) idempotency pair correctly. Topology clean — PR closes leaf #584 only, parented to epic #19, body has `Closes #584` keyword, all 8 DoD `unit_test_named` assertions plus 2 R1/R2 followup tests are present in tests/core/schedule/schedule_test.cpp. The R2 followups (SystemDescriptorConflict arm, header §9 carve-out doc, add_explicit_edge_if_absent rename, emplace-order pin, writes∩reads dominance comment, peak-memory carve-out) are correctly applied. Two LOW polish nits: schedule.hpp lines 16-23 cycle-class doc understates the post-R1 rule (says "pure access-set cycle" when ANY access-set edge in cycle is sufficient); dag_builder.cpp:298 comment still names `add_or_upgrade_edge`. The single HIGH blocker is procedural: tip commit ff9877de has zero workflow runs — CI never re-fired after the R2 push (or has not yet). Per the project rule that go-review cannot APPROVE on incomplete CI, verdict is COMMENT (not APPROVE). Once CI reports green on ff9877de and the two LOW doc nits are addressed, this PR is merge-ready and the /go orchestrator may enable auto-merge.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:2
carryover:0
notes: Round 3 lens (topology + spec alignment + polish). All R1 (7) and R2 (6) inline comments carry decision replies (carryover=0). Spec alignment is tight: schedule.hpp/dag_builder.cpp cite §4.4 inv 2/5, §6.4 algorithm steps 1-4, §6.5 Phase::HotReload ownership, §8.6 (phase, fqn) idempotency pair. Topology clean — PR closes leaf #584 only, parented to epic #19, body has `Closes #584` keyword, all 8 DoD `unit_test_named` assertions plus 2 R1/R2 followup tests are present in tests/core/schedule/schedule_test.cpp. The R2 followups (SystemDescriptorConflict arm, header §9 carve-out doc, add_explicit_edge_if_absent rename, emplace-order pin, writes∩reads dominance comment, peak-memory carve-out) are correctly applied. Two LOW polish nits: schedule.hpp lines 16-23 cycle-class doc understates the post-R1 rule (says "pure access-set cycle" when ANY access-set edge in cycle is sufficient); dag_builder.cpp:298 comment still names `add_or_upgrade_edge`. The single HIGH blocker is procedural: tip commit ff9877de has zero workflow runs — CI never re-fired after the R2 push. Per the project rule that go-review cannot APPROVE on incomplete CI, verdict is COMMENT (not APPROVE). Once CI reports green on ff9877de and the two LOW doc nits are addressed, this PR is merge-ready.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:2
carryover:0
notes: Round 3 lens (topology + spec alignment + polish). All R1 (7) and R2 (6) inline comments carry decision replies (carryover=0). Spec alignment is tight: schedule.hpp/dag_builder.cpp cite §4.4 inv 2/5, §6.4 algorithm steps 1-4, §6.5 Phase::HotReload ownership, §8.6 (phase, fqn) idempotency pair. Topology clean — PR closes leaf #584 only, parented to epic #19, body has `Closes #584` keyword, all 8 DoD `unit_test_named` assertions plus 2 R1/R2 followup tests are present in tests/core/schedule/schedule_test.cpp. The R2 followups (SystemDescriptorConflict arm, header §9 carve-out doc, add_explicit_edge_if_absent rename, emplace-order pin, writes∩reads dominance comment, peak-memory carve-out) are correctly applied. Two LOW polish nits: schedule.hpp lines 16-23 cycle-class doc understates the post-R1 rule (says "pure access-set cycle" when ANY access-set edge in cycle is sufficient); dag_builder.cpp:298 comment still names `add_or_upgrade_edge`. The single HIGH blocker is procedural: tip commit ff9877de has zero workflow runs — CI never re-fired after the R2 push. Per the project rule that go-review cannot APPROVE on incomplete CI, verdict is COMMENT (not APPROVE). Once CI reports green on ff9877de and the two LOW doc nits are addressed, this PR is merge-ready.

cjhowe-us:
agent:go-chore status:done notes:rebased PR #1095 onto current main; conflicts resolved in core/CMakeLists.txt, tests/core/CMakeLists.txt, core/include/glibre/error.hpp, core/include/glibre/log_error.hpp, tests/core/error_register/error_register_test.cpp, tests/core/log_error_test.cpp; CI re-triggered on commit 42907c4034beebde71dc024897c5fc982709dedf; auto-merge (squash) remains enabled

PR #1115, opened 2026-05-10
chore(ci): verify_dod grep -- separator before regex (refs #237)

### LINKED ISSUE #237: [PLAN] feat(infra): -fno-exceptions enforcement in CMake with editor UI carve-out

cjhowe-us:
## Summary
Fix grep invocation in `verify_dod.py` to use `--` separator before regex argument.

When a regex pattern starts with a dash (e.g., `-fno-exceptions`), grep parses it as a CLI flag instead of a pattern, causing the `file_contains` assertion to silently fail with "unrecognized option". The `--` separator tells grep to stop processing flags and treat everything that follows as arguments.

## Root cause
Line 73 of `.github/scripts/verify_dod.py` invokes:
```python
["grep", "-Eq", str(arg["regex"]), str(p)]
```

When regex starts with `-`, grep interprets it as a flag and fails silently.

## Fix
Add `--` before the regex:
```python
["grep", "-Eq", "--", str(arg["regex"]), str(p)]
```

Closes #237

### DIFF
--- a/.github/scripts/verify_dod.py
+++ b/.github/scripts/verify_dod.py
@@ -71,7 +71,7 @@ def check_file_contains(arg: object) -> tuple[bool, str]:
     if not p.is_file():
         return False, f"{label} (file missing)"
     rc = subprocess.run(
-        ["grep", "-Eq", str(arg["regex"]), str(p)],
+        ["grep", "-Eq", "--", str(arg["regex"]), str(p)],
         check=False,
     ).returncode
     return rc == 0, label


.github/scripts/verify_dod.py
```diff
@@ -71,7 +71,7 @@ def check_file_contains(arg: object) -> tuple[bool, str]:
     if not p.is_file():
         return False, f"{label} (file missing)"
     rc = subprocess.run(
-        ["grep", "-Eq", str(arg["regex"]), str(p)],
+        ["grep", "-Eq", "--", str(arg["regex"]), str(p)],
```
cjhowe-us, line 74: severity:HIGH location:PR body (Fixes line) problem:`Fixes: #237` uses a colon between keyword and `#`; the verify_dod.py regex (line 101 of this very file) is `\b(closes|fixes|resolves)\s+#N\b` — `\s+` does not match `:`, so check_pr_merged_closes_self will return False after merge and dod-verify will fail to auto-close #237. Additionally, while GitHub itself accepts `Fixes: #N` for issue auto-close, the in-repo DoD verifier does not. fix:Change `Fixes: #237` to `Closes #237` (or `Fixes #237`) — no colon, single space, matching the regex used by this PR own script.

cjhowe-us (commented):
findings:
  HIGH:1
  MED:0
  LOW:0
carryover:0
notes:Diff itself is correct — `--` is placed between `-Eq` flags and positional `regex`/`path` args, matching POSIX convention; no regression to other check_* helpers (check_glob_nonempty already uses `--`). However the PR body uses `Fixes: #237` (with colon), which the in-repo verify_dod.py regex (`\b(closes|fixes|resolves)\s+#N\b`, line 101) does NOT match — `\s+` rejects `:`. Result: even after merge, check_pr_merged_closes_self will return False and dod-verify cannot auto-close #237. Single-line edit to PR body resolves this. Change `Fixes: #237` to `Closes #237`. CI tip-commit status: 5 jobs SUCCESS, build still IN_PROGRESS (Python-only patch, build orthogonal). No blocking CI failure.

